PR #26, opened 2014-12-08
Apply `the_title` filter to event title

Morhaus:
mqTranslate saves all translations of a post title inside its `post_title` attribute. As such, when `get_event_link` retrieves it, it will display all translations concatenated instead of only the correct one.

Applying the `the_title` filter will only display the correct one.


ckpicker:
@Morhaus - we've had a LOT of refactoring in develop, so could you please re-submit this pull request based off the develop branch?  Just comment back here once you've done so.  We just want to make sure you get credit for the commit.  Thanks!

PR #27, opened 2014-12-08
added method to turn off sslverify

misterbisson:
Issue: when the `PUE_UPDATE_URL` constant is set to a site with a self-signed certificate for development purposes, the PUE client can't access the PUE server. 

This works around that by relaxing the SSL verification done when the `PUE_UPDATE_URL` constant is set.


jazbek:
Closing, per internal discussion, the issue is no longer relevant. If it becomes relevant again, we will reopen.

PR #32, opened 2015-01-21
Hide Event Add-Ons menu if user cannot edit

neverarriving:
Switch permission from edit_posts to edit_tribe_events


jazbek:
Hey @neverarriving thanks for the PR! Great idea to change the capability on the Add-Ons page, but I'm thinking that 'install_plugins' might be even better. If you are so inclined, wrapping the value in a filter called `tribe_events_addon_page_capability` would be even more awesome.

Would you mind updating and resubmitting the PR? If so, we can get it integrated in our upcoming release. Thanks!

neverarriving:
Replaced with #37

PR #35, opened 2015-02-04
Replaced admin events view column conditions with switch

kittsville:
Replaced unnecessary conditionals with a switch; resulting code is
easier to read and more efficient.


kittsville:
Making changes to new class to re-suggest pull request

PR #38, opened 2015-02-10
Make the single event Google map object available externally via tribeEventsSingleMap

tddewey:
This allows plugin or theme code to modify the Google map. The tribeEventsSingleMap var will have all the data needed: initial zoom level, addresses, and the google map object itself.

For example, we may want to disable the scrollwheel. With this commit, the following can be called from outside The Event Calendar:
`tribeEventsSingleMap.map.setOptions({scrollwheel:false});`

### DIFF
--- a/resources/embedded-map.js
+++ b/resources/embedded-map.js
@@ -60,6 +60,9 @@ if ( "function" === typeof jQuery ) jQuery( document ).ready( function( $ ) {
 			title   : venueTitle,
 			position: position
 		} );
+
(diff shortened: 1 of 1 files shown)


borkweb:
@tddewey - sorry for the delay in addressing this PR! We'll plan to have this in the next release! Thanks :)  I've Pull Requested it against our new working branch: https://github.com/moderntribe/the-events-calendar/pull/177

PR #40, opened 2015-02-24
Rename tribe_get_prev_event_link filter

jancbeck:
Currently the filter is called **tribe_get_next_event_link** however, the function is called **tribe_get_prev_event_link** so I think the filter should be named according to the function.


barryhughes:
Hi jancbeck - thanks for submitting this PR - you're absolutely right and we'll look at merging this in :+1:

roblagatta:
Logged as Central ticket #34475 (Current Release).

PR #41, opened 2015-02-24
Let updateEvent set the 'post_type' arg like createEvent does

jancbeck:
When writing an import script that would check if the event already exists before creating a new one I would have a look at **Tribe__Events__API:: createEvent** to see what arguments were required. I saw that it sets the `post_type` argument to **Tribe__Events__Events::POSTTYPE** which is nice, so I don't have to.

Later I stumbled upon that when I found that my import script was not properly updating the existing events and I discovered that **Tribe__Events__API:: updateEvent** requires the `post_type` to be set properly in the passed arguments.

The behavior of updateEvent should be consistent with createEvent so this PR removes the if-clause and instead sets the post_type to the arg.

### DIFF
--- a/lib/API.php
+++ b/lib/API.php
@@ -58,10 +58,7 @@ if ( ! class_exists( 'Tribe__Events__API' ) ) {
 		 */
 		public static function updateEvent( $eventId, $args ) {
 			$args['ID'] = $eventId;
-
-			if ( ! in_array( Tribe__Events__Events::POSTTYPE, (array) $args['post_type'] ) ) {
-				return false;
-			}
+			$args['post_type'] = Tribe__Events__Events::POSTTYPE;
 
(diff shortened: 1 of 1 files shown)


barryhughes:
Hi jancbeck,

Thanks for the request :-)

One thing I'd like to highlight, though, is that our recommended approach for creating and updating events is to use `tribe_create_event()` and `tribe_update_event()` respectively, which should shield you from any class-level changes that might occur in the future.

That definitely does not take away from the validity of your proposed change, though.

> The behavior of updateEvent should be consistent with createEvent so this PR removes the if-clause and instead sets the post_type to the arg. 

I agree - I think this is a sensible change and we'll look at merging your request :-)

roblagatta:
Logged as Central ticket #34474 (Current Release).

PR #42, opened 2015-04-19
Add sortable events table by tax + Refactoring

kittsville:
**New:**  
- Added sorting by event category and tags. Code was based off similar work from [my own plugin](https://github.com/kittsville/WP-Librarian/blob/e2bb0e176633dbb38a83912aeccbb2d4cdae03c8/lib/admin-tables.class.php#L453-L493), which was based on code by Mike Schinkel.

**Refactored:**  

- Removed duplicate is_admin check
- Replaced obfuscated event date sorting logic with single function. Also got rid of unnecessary use of two SQL joins rather than just one.
- Aligned and commented hook registration for an easier to read/understand init(). Now looks more like hook registration in main plugin class
- Replaced 3 if statements with a switch for very slightly faster but mainly easier to read function logic

[Here's a quick look at the sorting in action](http://imgur.com/a/GRpWC). Note that sorting by a taxonomy hides events that have no terms for that tax.

While I was working on this I noticed a query is leaking from `Query.php` and potentially causing a minor performance loss for all events queries. If you check out the generated SQL with any debugging plugin you'll see the offending join being added to the admin tables queries.

If you don't want to add the sortable feature then please still consider accepting the refactoring done in the other commits.

### DIFF
--- a/lib/Admin_List.php
+++ b/lib/Admin_List.php
@@ -20,105 +20,67 @@ if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 		 * @return void
 		 */
 		public static function init() {
-			if ( is_admin() && ! ( defined( 'DOING_AJAX' ) && DOING_AJAX ) ) {
-				add_filter( 'posts_join', array( __CLASS__, 'events_search_join' ), 10, 2 );
-				add_filter( 'posts_orderby', array( __CLASS__, 'events_search_orderby' ), 10, 2 );
-				add_filter( 'posts_fields', array( __CLASS__, 'events_search_fields' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


barryhughes:
Hi kittsville,

This looks great! It's been a busy week however so if you can bear with us a little longer we'll aim to review in a little more depth early next week and will get back to you from there.

Thanks again!

kittsville:
Sure thing. I'm just happy there's a means to contribute back to a plugin I've found so useful.

roblagatta:
Logged as Central ticket #36187 (Next Release).

borkweb:
@kittsville - this is an awesome change. I've re-pulled this against our `release/119` branch and made a few tweaks (mentioned at the bottom of the PR description here: https://github.com/moderntribe/the-events-calendar/pull/148).  Thank you!

PR #44, opened 2015-04-20
Escaping URLs

borkweb:
Escaping URLs as late as possible.

### DIFF
--- a/lib/Main.php
+++ b/lib/Main.php
@@ -2466,7 +2466,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 					break;
 			}
 
-			return apply_filters( 'tribe_events_ugly_link', esc_url( $eventUrl ), $type, $secondary );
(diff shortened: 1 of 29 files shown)


lib/Main.php
```diff
@@ -2464,7 +2466,7 @@ public function uglyLink( $type = 'home', $secondary = false ) {
 					break;
 			}
 
-			return apply_filters( 'tribe_events_ugly_link', $eventUrl, $type, $secondary );
+			return apply_filters( 'tribe_events_ugly_link', esc_url( $eventUrl ), $type, $secondary );
```
borkweb: This function is called from https://github.com/moderntribe/the-events-calendar/blob/develop/lib/Main.php#L2345, which already escapes this value in a less destructive way.  Which means this probably is a bad line to escape the URL on.

borkweb:
Closing this out so we can tackle this against `release/117`.

PR #45, opened 2015-04-20
Escape URLs

borkweb:
This adds more extensive URL escaping for the core Events Calendar plugin.  In addition to addressing URLs that were manipulated with `add_query_arg()` / `remove_query_arg()`, we are adding escaping to other programmatically constructed URLs.


barryhughes:
Looks good, let's merge!

PR #46, opened 2015-04-21
Localize the events url base and use that to generate event URLs

borkweb:
On sites where there aren't any events, pagination links aren't displayed which in turn breaks the `base_url` generation. To prevent that from happening, let's localize the base event URL and start with _that_ rather than `/`.  This ensures that the event URL is properly formed even in the absence of events.


resources/tribe-events-ajax-calendar.js
```diff
@@ -20,12 +20,15 @@
...
 			initial_date = tf.get_url_param( 'tribe-bar-date' ),
 			$tribedate = $( '#tribe-bar-date' ),
 			date_mod = false;
 
-		if ( $nav_link.length ) {
+		var base_url = '/';
+
+		if ( 'undefined' !== typeof TribeCalendar.events_base ) {
+			base_url = TribeCalendar.events_base;
+		} else if ( $nav_link.length ) {
```
borkweb: Only use the link failover if we don't have a `TribeCalendar.events_base` for some strange reason.

resources/tribe-events-ajax-day.js
```diff
@@ -18,10 +18,12 @@
...
 
-		var $nav_link = $( '[class^="tribe-events-nav-"] a' ),
-			base_url = '/';
+		var $nav_link = $( '[class^="tribe-events-nav-"] a' );
+		var base_url = '/';
 
-		if ( $nav_link.length ) {
+		if ( 'undefined' !== typeof TribeCalendar.events_base ) {
+			base_url = TribeCalendar.events_base;
+		} else if ( $nav_link.length ) {
```
borkweb: Only use the link failover if we don't have a `TribeCalendar.events_base` for some strange reason.

bordoni:
If the person doesn't have permalinks active what happens?

borkweb:
A good question. When permalinks aren't active, the URL still functions but looks ugly (as expected):

screen shot 2015-04-22 at 1 57 56 pm

bordoni:
Only asked because I worked on a few tickets CR on the navigation JS bugs.

faction23:
Looks good js side. Ya this whole base_url setup was never optimal. In regards to how we pass it from php, maybe it would be better placed here? https://github.com/moderntribe/the-events-calendar/blob/bugfix/35425-calendar-navigation/lib/tribe-template-factory.class.php#L555

This localize outputs for all of the events views, so you would only need it once, instead of in 3 places in php (ignore the terrible handle of 'calendar-script' for that js file, this is the core events js). We have been slowly moving towards passing php config data to js here. It's shared by all views, its early in our scripts, right before tribe-events.js (which is "core" to all other events scripts) so we know any of the js and inlines (god forbid, looking at you filterbar) can access its values. Down the road the plan was to refactor this whole enqueue stack and then move even more config data to that localize and away from data atts sprayed around the plugin. 

If you disagree and want to keep the base var passing just for the three views then that's fine too, this is good to go, just an option.

borkweb:
Excellent suggestion, @faction23. I've made the move :)

faction23:
Great! Ready for qa. :)

PR #48, opened 2015-04-22
Remove custom styles for the notice box.

borkweb:
The margins and background color are the defaults. The border color gets overridden by the WP dashboard styles. Removing this style declaration removes unneeded duplicate styles and eliminates the border that causes our notices to differ from default notices in WP.


barryhughes:
Looking good, let's do it!

PR #50, opened 2015-05-12
Resolves #30311

bordoni:
As @barryhughes commented, the previous solution wasn't fixing all the issues with Google Calendar.

_Ref:_ [#30311](https://central.tri.be/issues/30311)

### DIFF
--- a/lib/iCal.php
+++ b/lib/iCal.php
@@ -176,7 +176,7 @@ class Tribe__Events__iCal {
 	public static function generate_ical_feed( $post = null ) {
 
 		$tec         = Tribe__Events__Main::instance();
-		$wp_timezone = get_option( 'timezone_string' );
+		$timezone = get_option( 'timezone_string' );
 		$events      = '';
 		$blogHome    = get_bloginfo( 'url' );
 		$blogName    = get_bloginfo( 'name' );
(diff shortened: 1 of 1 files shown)


lib/iCal.php
```diff
@@ -195,56 +194,38 @@ public static function generate_ical_feed( $post = null ) {
...
+			}
 
-				if ( ! empty( $wp_timezone ) ){
-					$item[] = "DTSTART;TZID=\"{$wp_timezone}\":{$tz_startDate}";
-				}
-				$item[] = 'DTSTART:' . $tz_startDate;
-				$item[] = 'DTSTART:' . $startDate;
+			$tzoned = (object) array(
+				'start' => date( $format, $time->start ),
+				'end' => date( $format, $time->end ),
```
borkweb: This doesn't appear to be converting the local timezone into UTC before adding it to the iCal output.

borkweb: It looks like it is using local time and claiming it is UTC.  So January 24, 2015 03:00 EST would be translated to January 24, 2015 03:00 UTC.

barryhughes:
One sidenote: is `Tribe__Events__iCal::wp_date()` now unused (can it be removed altogether)?

lib/iCal.php
bordoni: I thought that the `self::wp_strtotime()` was converting to UTC.
```php
$time = (object) array(
	'start' => self::wp_strtotime( $event_post->EventStartDate ),
	'end' => self::wp_strtotime( $event_post->EventEndDate ),
	'modified' => self::wp_strtotime( $event_post->post_modified ),
	'created' => self::wp_strtotime( $event_post->post_date ),
);
```

borkweb: Looking at the `wp_strtotime` function, it appears to be converting to local time given a UTC or local timestamp.

lib/iCal.php
```diff
@@ -193,58 +192,49 @@ public static function generate_ical_feed( $post = null ) {
...
-						'T',
-						'',
-					), $event_post->post_modified
-				);
-			$item[] = 'UID:' . $event_post->ID . '-' . strtotime( $startDate ) . '-' . strtotime( $endDate ) . '@' . $blogHome;
-			$item[] = 'SUMMARY:' . $event_post->post_title;
-			$item[] = 'DESCRIPTION:' . str_replace( array( ',', "\n", "\r", "\t" ), array( '\,', '\n', '', '\t' ), strip_tags( $event_post->post_content ) );
+			$item[] = 'DTSTAMP:' . date( $full_format, time() );
+			$item[] = 'CREATED:' . date( $full_format, $tzoned->created );
+			$item[] = 'LAST-MODIFIED:' . date( $full_format, $tzoned->modified );
```
zbtirrell: These should just use `$tzoned->created` and `$tzoned->modified` directly as those have already been formatted to `$full_format`.  As it is, passing them through `date()` again corrupts their values.

lib/iCal.php
zbtirrell: This was resolved with the mod to the `wp_strtotime` function in https://github.com/moderntribe/the-events-calendar/commit/c6dc0c3456b19bd7117aa5d8b9fd6797ee2b2ad2

PR #51, opened 2015-05-15
Feature/33500 rename files

lucatume:
The work done so far for ticket https://central.tri.be/issues/33500, also includes work done for ticket https://central.tri.be/issues/36164


borkweb:
While I wasn't able to dig into this PR as a single contained unit because it is too large, I _did_ look at each individual commit, though and it looks really solid - though I found _one_ bug:

1. [This image reference](https://github.com/moderntribe/the-events-calendar/blob/feature/33500-rename-files/src/resources/css/events-admin.css#L276) needs an update

lucatume:
I've fixed the image path in the minified and not minified version of the CSS file. The referenced image is not there though (`src/resources/images/tribe-loading@2x.gif`). That's pre-refactoring and could be addressed by someone that knows his/her way around TEC CSS

borkweb:
Cool, thanks for the update on the image path. I've created a ticket (https://central.tri.be/issues/36222) for the actual file name issue so it can be fixed separate from this PR by someone.

I'd say this PR is looking good!

PR #53, opened 2015-05-18
Make event attendees page title filterable

aaemnnosttv:
Pretty straightforward here, just adding a new filter for an Event's Attendees admin page title.

PR #54, opened 2015-05-18
Add hooks to Event Attendees meta box

aaemnnosttv:
Pretty straightforward here, just adding a few action hooks to the Event Attendees "meta box" (`#tribe-filters`).

One related fix here on the Tickets_Pro class, the path property was 1 directory too shallow in its definition, so all of the templates it was attempting to load were failing as they all had a `src/src/` in the path instead of `src/`.  (This should probably be its own PR, but I included it here as it was necessary to load the attendees page.)


aaemnnosttv:
Added dedicated PR for Tickets_Pro fix, see #55

barryhughes:
Hi @aaemnnosttv - looks like a nice change and it should make things a little more flexible for customization purposes. One thing though - as in https://github.com/moderntribe/the-events-calendar/pull/53 - would be to make the hook names consistent with our existing ones :-)

Would you be down to revise and re-submit?

aaemnnosttv:
@barryhughes, how about now?

PR #55, opened 2015-05-18
fix Tickets_Pro path property

aaemnnosttv:
Just a simple change here. The path property was 1 directory too shallow in its definition, so all of the templates it was attempting to load were failing as they all had a `src/src/` in the path instead of `src/`.


borkweb:
Hey, thanks @aaemnnosttv! Looks great (well, as great as a pile of parentheses can look :smile: )

__MT note__: this PR addresses a bugfix for https://central.tri.be/issues/33500

PR #56, opened 2015-05-18
Enhance venue details on Event Attendees screen

aaemnnosttv:
This PR does a few things:

- Cleaner markup, and more readable php
  - Uses `div` instead of `br` for vertical separation
  - Add classes to venue divs to easier allow for potential alternate styling / custom JS
- Adds full venue address (it can be very difficult to discern between venues with the same name without it.  e.g. "Holiday Inn")
- Displays the venue's website in a much cleaner way, especially for long URLs:
  - yuck: http://cl.ly/image/3E3v2y3r282m
  - much better: http://cl.ly/image/3a3I1w2B1y2B

### DIFF
--- a/src/admin-views/tickets/attendees.php
+++ b/src/admin-views/tickets/attendees.php
@@ -29,41 +29,36 @@ $tickets = Tribe__Events__Tickets__Tickets::get_event_tickets( $event_id );
 						<strong><?php _e( 'End Date / Time:', 'tribe-events-calendar' ) ?></strong>
 						<?php echo tribe_get_end_date( $event_id, false, tribe_get_datetime_format( true ) ) ?>
 
-						<?php
-						// venue
-						$venue_id = tribe_get_venue_id( $event_id );
-						if ( ! empty( $venue_id ) ) {
-							$venue = get_post( $venue_id );
-						}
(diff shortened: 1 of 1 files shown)


barryhughes:
Nice job cleaning things up there, @aaemnnosttv!

I also love the idea of restricting things to the host name as opposed to a straight bit of truncation that keeps the scheme etc. Couple of questions:

* As it stands, even if there is nothing after the hostname, we append `/...` - might we avoid doing so?
* I can readily imagine some users not wishing to lose the path - it's important therefore that we can give them an easy 'escape route' - could you consider making this field filterable?

Thanks!

aaemnnosttv:
Thanks @barryhughes,

With my last commit, I made it so that the `/...` will only be appended if there is a path.
The output there is also now passed through a new filter `tribe_venue_display_url` which passes a few parameters -- the second of which is the unaltered url, so reverting to the current behavior is as simple as:
```php
add_filter('tribe_venue_display_url', function($display, $url) {

	return $url;

}, 10, 2 );
```

borkweb:
This is awesome! Repointed at `release/120`! https://github.com/moderntribe/the-events-calendar/pull/196

PR #57, opened 2015-05-18
Allow Attendees_Table constructor to accept arguments

aaemnnosttv:
This is a simple change which does just what it says, it allows the constructor to accept extra arguments or override any of the defaults.

This has no affect on the normal functionality of the plugin/attendees table.  It is necessary though however for the list table to be used on any other admin page (e.g. WooCommerce order - which has been my use-case).  To do this, we need to be able to pass in the `screen` key in the arguments array.

### DIFF
--- a/src/Tribe/Tickets/Attendees_Table.php
+++ b/src/Tribe/Tickets/Attendees_Table.php
@@ -14,9 +14,18 @@ class Tribe__Events__Tickets__Attendees_Table extends WP_List_Table {
 
 	/**
 	 * Class constructor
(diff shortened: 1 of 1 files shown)


barryhughes:
Thanks @aaemnnosttv - we'll look at merging this in :-)

zbtirrell:
Logged as Central ticket [#36286](https://central.tri.be/issues/36286) (Next Release).

aaemnnosttv:
Merged in #67

PR #58, opened 2015-05-18
Display proper order_id for woocommerce orders

aaemnnosttv:
This PR addresses an issue where the displayed order ID for an attendee does not match the ID elsewhere in the system.  This is because at current, it is just showing the bare ID, rather than calling the `get_order_number` method on the woocommerce order.  _This will only change the current behavior if the item's order_id is a woocommerce order._

Why would the order number be different than the order ID?  See http://www.woothemes.com/products/sequential-order-numbers-pro/

### DIFF
--- a/src/Tribe/Tickets/Attendees_Table.php
+++ b/src/Tribe/Tickets/Attendees_Table.php
@@ -112,16 +112,10 @@ class Tribe__Events__Tickets__Attendees_Table extends WP_List_Table {
 
 		//back compat
 		if ( empty( $item['order_id_link'] ) ) {
-			$display_id = $item['order_id'];
-			if ( function_exists('wc_get_order') && wc_get_order( $item['order_id'] ) ) {
-				$display_id = wc_get_order( $item['order_id'] )->get_order_number();
-			}
-			$id = sprintf( '<a class="row-title" href="%s">%s</a>', esc_url( get_edit_post_link( $item['order_id'], true ) ), esc_html( $display_id ) );
-		} else {
-			$id = $item['order_id_link'];
+			$item['order_id_link'] = sprintf( '<a class="row-title" href="%s">%s</a>', esc_url( get_edit_post_link( $item['order_id'], true ) ), esc_html( $item['order_id'] ) );
 		}
 
-		return $id;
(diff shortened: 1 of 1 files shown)


barryhughes:
Thanks for the PR, @aaemnnosttv!

My only reservation here is that we try to keep module-specific code out of the core plugin: in this case, for instance, any interaction with WooCommerce itself should be handled from within WooCommerce Tickets.

Some options that might be worth thinking about:

* We could set `order_id_link` from within WooCommerce Tickets -- and retain our check for a non empty `$item['order_id_link']` in the core ticketing framework (which until now has only been kept for backward compatibility)
* Or, perhaps, we allow modules like WooCommerce Tickets to optionally set an `order_display_id` and use it where available within our core ticketing code

Are you open to reconsidering things along these lines? (Realizing our WooCommerce Tickets repo is private, you could submit a .diff for any changes need there if you did want to follow something like the second route.)

aaemnnosttv:
Hey @barryhughes!

I agree that WooCommerce-specific code should reside in the WooCommerce Tickets extension.  I forget how versatile this plugin is!  Regarding your ideas for how to implement that, I have slightly mixed feelings.

After thinking it over, I think the best choice here is to go with your first suggestion, and simply pass the complete link from WooTickets.  I really like the idea of adding an `order_display_id` field though, that may be useable elsewhere.

In closing, I still have a tweak or two I'd like to propose here, just for the sake of cleaner code than what's currently there, but we'll keep it mostly the same :)

On a related note, I will submit a diff for WooTickets, I had a few changes there anyways.  What's the best way to do this, and where?

aaemnnosttv:
@barryhughes - See what you think here, it's a minor change but does the same thing with less lines and is more readable I think.

Let me know if you'd like me to rebase/squash this or rename the pr or something.

aaemnnosttv:
@barryhughes see moderntribe/events-tickets-woo#8

borkweb:
Since this has been twiddled and implemented in a simpler manner by @aaemnnosttv in https://github.com/moderntribe/events-tickets-woo/issues/22  - this issue is closable! Thanks!

PR #59, opened 2015-05-18
Further refactor the Template_Factory so that events-pro can override it

borkweb:
Related: https://github.com/moderntribe/events-pro/pull/11

See: https://central.tri.be/issues/36164 and https://central.tri.be/issues/33500


zbtirrell:
:+1:

PR #60, opened 2015-05-18
PHP 5.2 doesn't do `static`. Use `self`

borkweb:
Since PHP 5.2 doesn't like `static`, I needed to restructure my recent restructure a bit so other plugins can have their own Assets.

Related: https://github.com/moderntribe/events-pro/pull/12

See: https://central.tri.be/issues/36164 and https://central.tri.be/issues/33500


barryhughes:
:+1: let's merge in and get to QA

PR #61, opened 2015-05-18
Add Ticket events in tickets.js

aaemnnosttv:
This PR adds new javascript events to the tickets.js script which handles ticket creation, modification, and deletion from an Event edit screen.

No behavioral changes are introduced here, this is all for making the code more extendable.

The events that are triggered are all done on the `#tribetickets` container, and follow a particular convention where `tribe` is the jQuery event namespace:

- `eventSaveTicket.tribe`
- `eventSavedTicket.tribe`
- `eventEditTicket.tribe`
- `eventDeleteTicket.tribe`
- `eventDeletedTicket.tribe`

Note the missing `eventEditedTicket.tribe` - as the event for updating an event is just another Save.  Edit is triggered just after the new edit form is loaded and reset.

The pre-event hooks are passed the jQuery event object that triggered the action, and the completed events are passed the ajax response object.

These hooks make managing additional product fields possible on the ticket, or working with your own custom fields, without resorting to just using the same click bindings. 

In addition to this, there are a few other supporting small changes, like caching the jQuery object for the `#tribetickets` container which is used for emitting the events (as well as a few other places), and updating `$` instead of `jQuery` in a few places for consistency.

These events can be viewed in the console by adding this jQuery for testing
```js
jQuery(function($){

	function log(){
		console.log(arguments[0].type);
	}

	$('#tribetickets')
		.on('eventSaveTicket.tribe', log)
		.on('eventSavedTicket.tribe', log)
		.on('eventEditTicket.tribe', log)
		.on('eventDeleteTicket.tribe', log)
		.on('eventDeletedTicket.tribe', log)
});
```

### DIFF
--- a/src/resources/js/tickets.js
+++ b/src/resources/js/tickets.js
@@ -355,13 +355,13 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 		}
 
 		function tickets_start_spin() {
-			jQuery( '#event_tickets' ).css( 'opacity', '0.5' );
-			jQuery( "#tribe-loading" ).show();
+			$( '#event_tickets' ).css( 'opacity', '0.5' );
+			$( "#tribe-loading" ).show();
 		}
 
 		function tickets_stop_spin() {
-			jQuery( '#event_tickets' ).css( 'opacity', '1' );
-			jQuery( "#tribe-loading" ).hide();
+			$( '#event_tickets' ).css( 'opacity', '1' );
+			$( "#tribe-loading" ).hide();
 		}
 
 		function tribe_fix_image_width() {
--- a/src/resources/js/tickets.js
+++ b/src/resources/js/tickets.js
@@ -126,7 +126,9 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 		var $tribetickets = $('#tribetickets');
 
 		/* "Save Ticket" button action */
-		$( '#ticket_form_save' ).click( function() {
+		$( '#ticket_form_save' ).click( function( e ) {
+
+			$tribetickets.trigger('eventSaveTicket.tribe', e);
 
 			tickets_start_spin();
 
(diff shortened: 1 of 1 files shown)


zbtirrell:
This is a great idea!  I appreciate the thought that went into your event names.  We want to keep the event names in line with the other events we are triggering in JS throughout the plugin.  That being said though, we aren't real happy with the current standards around our JS events.

Here is an example from elsewhere: https://github.com/moderntribe/the-events-calendar/blob/develop/src/resources/js/tribe-events-ajax-list.js#L201

The good news is that this is now a front and center discussion that we will be considering for a near-future release!  The downside is that we may ask you to adjust these event names *and* we can't merge this right away.

Logged as Central ticket [#36302](https://central.tri.be/issues/36302).

aaemnnosttv:
Thanks @zbtirrell,

Yes, I came across those triggers in the list ajax scripts recently too.  Perhaps this isn't the right ticket to ask, but why is the trigger applied to the `tribe_ev.events` object, rather than a DOM node, which would allow for event propagation through the DOM?

I realize this PR is a bit more substantial than some of the others so I understand if there isn't enough time for it to be considered for the upcoming release.

faction23:
Hi. I added those events on that object long ago. The case for using an isolated object that i knew i could rely on is that the front end of events gets heavily customized, and in fact no nodes are truly reliable in those views due to that, though we do gamble here and there. I could have sent on document or body ofcourse, but at that time event isolation on empty objects instead of cluttering those nodes was a pattern you could see around.

aaemnnosttv:
Thanks for the insight @faction23!  Much appreciated :smile:

borkweb:
@aaemnnosttv, this is awesome work and we're going to get this merged [with a few tweaks](https://github.com/moderntribe/the-events-calendar/pull/155). Note that the event names will be changing, but they'll be fired the same way (and in the same location).

I'll close this PR out in favor of the other one - thanks again :) :+1:

PR #68, opened 2015-05-19
Make attendee table nav filterable

aaemnnosttv:
This PR makes the attendee's list table top/bottom nav extendable by first restructuring the output into an array, and then making that filterable via a new filter `tribe_events_tickets_attendees_table_nav ` before output.

The structure of the array, in combination with the `$which` position passed by the filter, makes extending any combination of the `top`, `bottom`, `left`, `right` nav very easy.

No real change in the current behavior here, but there is one minor change which may not be obvious that _could_, which is the use of `esc_attr__()` for the _Print_ and _Email_ action buttons (instead of `__()`, since those translations are passed into the `value` attribute of their respective buttons.


borkweb:
I really like this change. Thanks! [Repointed](https://github.com/moderntribe/the-events-calendar/pull/197) at `release/120`!

PR #72, opened 2015-05-25
Preventing Notices from Current Screen variables

bordoni:
_Resolves:_ [#30696](https://central.tri.be/issues/30696)

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -1594,7 +1594,13 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		 */
 		public function admin_body_class( $classes ) {
 			global $current_screen;
-			if ( isset( $current_screen->post_type ) && ( $current_screen->post_type == self::POSTTYPE || $current_screen->id == 'settings_page_tribe-settings' ) ) {
+			if (
+				isset( $current_screen->post_type )
+				&& (
+					$current_screen->post_type === self::POSTTYPE
+					|| 'settings_page_tribe-settings' === $current_screen->id
+				)
+			) {
 				$classes .= ' events-cal ';
 			}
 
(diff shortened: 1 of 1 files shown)


src/Tribe/Main.php
```diff
@@ -1594,9 +1594,7 @@ public function saved_organizers_dropdown( $current = null, $name = 'organizer[O
 		 */
 		public function admin_body_class( $classes ) {
 			global $current_screen;
-			if ( isset( $current_screen->post_type ) &&
-				 ( $current_screen->post_type == self::POSTTYPE || $current_screen->id == 'settings_page_tribe-settings' )
-			) {
+			if ( isset( $current_screen->post_type ) && ( $current_screen->post_type == self::POSTTYPE || $current_screen->id == 'settings_page_tribe-settings' ) ) {
```
borkweb: Collapsing this to a single line makes it harder to read, so I'm not a huge fan of this particular change.  If anything, _more_ spacing may help:

```php
if (
  isset( $current_screen->post_type )
  && (
    self::POSTTYPE === $current_screen->post_type
    || 'settings_page_tribe-settings' === $current_screen->id
  )
) {
```

borkweb:
This looks like it is for the next release. You can PR this against `release/119` sans the `dont-merge` label if you'd like.

src/Tribe/Main.php
borkweb: Also note the reversal of the scalar comparisons. Since we standardize on WP PHP standards, we might as well fix scalar conditionals as we encounter them to match [Yoda conditions](https://make.wordpress.org/core/handbook/coding-standards/php/#yoda-conditions).

bordoni: So @borkweb the thing is that my Code Sniffer was complaning about multiple lines in the conditionals.

When I change something I will work with Yoda Conditions then, just didn't occur to me that we would do it as of right now.

jbrinley: Seems like a debate over the best wrong way to do something. Why not break the conditional out into a separate function? `if ( $this->is_events_admin_screen() )`

borkweb: A function is a good idea. Yoda matters regardless of where the logic lies.

PR #75, opened 2015-05-26
Adding missing link to the Month Cache blurb on display settings page

borkweb:
Note: The link currently points to an unpublished page (see Leah's note [here](https://central.tri.be/issues/33364#note-5))

See: https://central.tri.be/issues/36369


zbtirrell:
:+1:

PR #77, opened 2015-05-27
Escape $ and \ when builing anchor tags via preg_replace

borkweb:
See: https://central.tri.be/issues/31284


bordoni:
Seem to be good for merge.

PR #79, opened 2015-05-28
get_term_by() returns a false on failure, not a wp_error

barryhughes:
Avoids a notice level issue where bool false is returned by get_term_by() and we continue to try and access the property of a non-existent object.


borkweb:
This looks good to me. :+1:

barryhughes:
Note to self: resubmit against `release/119`

PR #81, opened 2015-05-30
Remove doubled <!--Navigation --> comment

gregrickaby:
Stumbled across this and wasn't sure if this was intentional or not? :grin:


borkweb:
Oh! Good catch :) Yeah, that wasn't intentional. Thank you for fixing it!

PR #82, opened 2015-06-01
Find min/max prices when price ranges are entered in the "Cost" field

borkweb:
See: https://central.tri.be/issues/36518

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -835,35 +835,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	 * @return int the minimum cost.
 	 */
 	function tribe_get_minimum_cost() {
-		global $wpdb;
-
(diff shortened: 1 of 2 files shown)


src/functions/template-tags/general.php
```diff
@@ -841,6 +841,17 @@ function tribe_get_minimum_cost() {
 
 		$costs = array_map( 'tribe_map_cost_array_callback', $costs );
 		foreach ( $costs as $index => $value ) {
+			// try to find the lowest numerical value in a possible range
+			if ( preg_match( '/^(-?[\d]+)[^\d\.]+([\d\.]+)/', $value, $matches ) ) {
```
bordoni: It would be cool to have that Regular expression on a method that returns an `Array`.

bordoni:
I think we will need QA here. Just to make sure we are good for a merge later on.

borkweb:
Closing and re-pull requesting against the `develop` branch as per https://central.tri.be/issues/36620

PR #83, opened 2015-06-01
Abstract out date meta logic. If missing dates, don't update dates.

borkweb:
See: https://central.tri.be/issues/34636


bordoni:
From the CR perspective it's good. Kick it to QA

borkweb:
This was QA'd and approved for merge! Merging.

PR #84, opened 2015-06-01
Allow tickets with a cost of 0 to affect the event price

borkweb:
`! empty( $ticket->price )` evaluates to false if price is `0`. Using `isset` instead allows 0 to be interpreted as a number.

See: https://central.tri.be/issues/36107

---

This fix allows events with `$0` tickets to display `Free` next to it!

screen shot 2015-06-01 at 3 57 38 pm


bordoni:
I don't see the need to QA this since it's a small thing. So `Merge` tag right?

borkweb:
Sounds good to me!

PR #86, opened 2015-06-02
Only inject date fields into queries when appropriate

borkweb:
This resolves a long-standing conflict with the Posts 2 Posts plugin which has historically required [a workaround](https://github.com/scribu/wp-posts-to-posts/issues/249).

**Note:** ~~This probably needs a QA pass before merging and smoketesting.~~ This has been QA'd and can be merged.

See: https://central.tri.be/issues/33288

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -710,7 +710,7 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 				}
 
 				// trim trailing characters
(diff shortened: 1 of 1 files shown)


bordoni:
From the Code Review perspective it seems to be good, but this will need a bunch of QA for sure.

bordoni:
Sorry! Wrong Label but Zach was faster.

PR #90, opened 2015-06-04
Find min/max prices when price ranges are entered in the "Cost" field

borkweb:
See: https://central.tri.be/issues/36518 and https://central.tri.be/issues/36620

Note: this was already code reviewed in https://github.com/moderntribe/the-events-calendar/pull/82 and this is just pointing those changes at `develop` rather than `release/119`.

### DIFF
--- a/src/Tribe/Cost_Utils.php
+++ b/src/Tribe/Cost_Utils.php
@@ -39,13 +39,14 @@ class Tribe__Events__Cost_Utils {
 	}//end get_all_costs
 
 	/**
-	 * Returns a maximum cost in a list of costs. If an array of costs is not passed in, the array of costs is fetched via query.
+	 * Returns a particular cost within an array of costs
 	 *
 	 * @param $costs mixed Cost(s) to review for max value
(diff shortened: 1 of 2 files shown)


src/Tribe/Cost_Utils.php
```diff
@@ -0,0 +1,127 @@
...
+
+	/**
+	 * fetches all event costs from the database
+	 *
+	 * @return array
+	 */
+	public function get_all_costs() {
+		global $wpdb;
+
+		$costs = $wpdb->get_col( "SELECT meta_value FROM {$wpdb->postmeta} WHERE meta_key = '_EventCost'" );
```
zbtirrell: Why not use a native WordPress function here to fetch this?

borkweb: This isn't fetching the value for a single post, instead it is fetching it for _all_ posts.

bordoni: That is for sure missing on WordPress.

borkweb:
This passed QA. Merging!

PR #92, opened 2015-06-04
Don't remove the filter that allows for custom sorting in the event list

borkweb:
See: https://central.tri.be/issues/28668


bordoni:
Looks good.

PR #94, opened 2015-06-04
Bugfix where EventStartDate was excluded from some queries, busting them

borkweb:
See: https://central.tri.be/issues/36770


zbtirrell:
I find that if logic gross, but, minimal viable change is where we are at in the cycle... :+1:

PR #95, opened 2015-06-04
Identify the 0th page when paging as page 1 for the purposes of nav

borkweb:
Ideally, the logic in these two functions would be merged in some way to keep it DRY, however, this late in the release cycle means that I shouldn't refactor at this time. Instead, a duplicated change can be seen in both of the modified functions.

See: https://central.tri.be/issues/36766


zbtirrell:
:+1:

PR #97, opened 2015-06-05
Restore startTime and endTime month tooltip template vars for back co…

barryhughes:
Custom `month/tooltip.php` templates set up with TEC 3.9.3 or earlier may expect the `startTime` and `endTime` vars to be available.

If they are not (they were removed in commit 959251c5f8d - internal Central ref #30805) the JS templating system gets upset and tooltips stop working altogether.

This change restores those vars so that, as a minimum, month view tooltips continue to display for those users who have customized them.

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -1242,6 +1242,16 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 			'categoryClasses' => $category_classes,
 		);
 
+		/**
+		 * Template overrides (of month/tooltip.php) set up in 3.9.3 or earlier may still expect
+		 * these vars and will break without them, so they are being kept temporarily for
+		 * backwards compatibility purposes.
+		 *
+		 * @todo consider removing in 4.0
+		 */
+		$json['startTime'] = tribe_get_start_date( $event );
+		$json['endTime']   = tribe_get_end_date( $event );
+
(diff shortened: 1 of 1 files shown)


src/functions/template-tags/general.php
```diff
@@ -1242,6 +1242,16 @@ function tribe_events_template_data( $event, array $additional = null ) {
...
 
+		/**
+		 * Template overrides (of month/tooltip.php) set up in 3.9.3 or earlier may still expect
+		 * these vars and will break without them, so they are being kept temporarily for
+		 * backwards compatibility purposes.
+		 *
+		 * @todo consider removing in 4.0
+		 */
+		$json['startTime'] = tribe_get_start_date( $event );
+		$json['endTime']   = tribe_get_end_date( $event );
```
borkweb: From Slack chat:

>  zach [1:13 PM]: I wonder if we should hook to our own filter `tribe_events_template_data_array` and add those in via something in `lib/deprecated/`

> matt [1:24 PM]: That doesn’t seem like a bad plan, Zach. That way folks will get the deprecation notices and may be able to proactively react to the inevitable removal of those vars.

borkweb: Barry points out that this would trigger deprecation notices for everyone. That'd be stupid. Ignore me.

zbtirrell:
:+1:

PR #98, opened 2015-06-08
Small interim EventAllDay fix to allow proper imports of CSV files

borkweb:
See: https://central.tri.be/issues/36815


zbtirrell:
:100:

PR #101, opened 2015-06-08
grunt task to pull translations from GlotPress

zbtirrell:
See: https://central.tri.be/issues/36860

### DIFF
--- a/dev/grunt_options/glotpress_download.js
+++ b/dev/grunt_options/glotpress_download.js
@@ -6,7 +6,7 @@
 module.exports = {
   core: {
     options: {
-      url: 'http://translations.tri.be',
+      url: 'http://translations.theeventscalendar.com',
       domainPath: 'lang',
       slug: 'the-events-calendar',
       textdomain: 'tribe-events-calendar',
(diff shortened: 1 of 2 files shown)


dev/grunt_options/glotpress_download.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = {
+  core: {
+    options: {
+      url: 'http://translations.tri.be',
+      domainPath: 'lang',
+      slug: 'the-events-calendar',
+      textdomain: 'tribe-events-calendar',
+    },
+  },
```
borkweb: In JS, remove trailing commas on the last item in a list.

zbtirrell: Aren't trailing commas fine in the context of grunt since we are guaranteed to have modern ecmascript?

borkweb: Yeah, that's right. Even though trailing commas aren't in the JSON spec, doesn't mean we can't use them in config files. Keep 'em!

borkweb:
:+1: Merge it

borkweb:
Needs a URL update now

borkweb:
:fallen_leaf:

PR #103, opened 2015-06-09
EventAllDay should be saved using yes/no rather than true/false

borkweb:
See: https://central.tri.be/issues/36927


zbtirrell:
:whale2: :dash:

PR #104, opened 2015-06-09
Centralize and make consistent the logic for checking for AllDay

borkweb:
The checkbox for determining if something was AllDay was different than the logic in tribe_event_is_all_day where any truthy value (including `'no'` and `'false'`) would be considered true. The new `Tribe__Events__Date_Utils::is_all_day()` method now looks for explicitly true values (`true`, `'true'`, `'TRUE'`, and `'yes'`) and considers all other values to be false.

See: https://central.tri.be/issues/36928


zbtirrell:
:clap:

PR #106, opened 2015-06-10
Codesniffer fixes: pass 1

borkweb:
This is a partial codesniffer cleanup of core.  I'll be doing this in batches to make the review manageable.

Pull Requests to handled the deprecation of the `Tribe__Events__Date_Utils` methods:
* https://github.com/moderntribe/events-facebook/pull/11
* https://github.com/moderntribe/events-pro/pull/28
* https://github.com/moderntribe/events-community/pull/20
* https://github.com/moderntribe/events-filterbar/pull/15

See: https://central.tri.be/issues/36908


src/Tribe/Date_Utils.php
```diff
@@ -351,5 +306,223 @@ public static function week_ends_on( $week_starts_on ) {
...
+		 *
+		 * @deprecated since 3.10
+		 * @todo       remove in 4.1
+		 *
+		 * @param int|string $date        The date (timestamp or string).
+		 * @param bool       $isTimestamp Is $date in timestamp format?
+		 *
+		 * @return string The date and time of the beginning of a given day.
+		 */
+		public static function beginningOfDay( $date, $isTimestamp = false ) {
```
zbtirrell: I know you just relocated these deprecated functions (`endOfDay` as well), but why do they have repeated functionality rather than calling the new functions?

zbtirrell:
other than my non-essential question above, this is good to go

src/Tribe/Date_Utils.php
borkweb: That's a good question. It is _probably_ swappable with the functions, but they _do_ take different parameters.  I could change this one to the following:

```php
public static function beginningOfDay( $date, $isTimestamp = false ) {
  _deprecated_function( __FILE__, '3.10', 'tribe_event_beginning_of_day' );
  if ( $isTimestamp ) {
    $date = date( self::DBDATEFORMAT, $date );
  }
  return tribe_event_beginning_of_day( $date, self::DBDATETIMEFORMAT );
}
```

Perhaps it'd be worth it to do it now.

PR #110, opened 2015-06-11
Codesniffer updates to Tribe__Events__Main

borkweb:
See: https://central.tri.be/issues/36908

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -4247,7 +4247,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		 * @param array $post
 		 *
 		 * @return array
-		 * @see Tribe__Events__Mainfilter_wp_import_data_after()
+		 * @see Tribe__Events__Main::filter_wp_import_data_after()
 		 */
 		public function filter_wp_import_data_before( $post ) {
 			if ( $post['post_type'] == self::POSTTYPE && ! empty( $post['post_parent'] ) ) {
@@ -4275,7 +4275,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		 * @param array $post
 		 *
 		 * @return array
-		 * @see Tribe__Events__Mainfilter_wp_import_data_before()
+		 * @see Tribe__Events__Main::filter_wp_import_data_before()
 		 */
 		public function filter_wp_import_data_after( $post ) {
 			if ( $post['post_type'] == self::POSTTYPE ) {
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -728,7 +728,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 				array(
 					'utm_campaign' => 'in-app',
 					'utm_medium'   => 'plugin-tec',
-					'utm_source'   => 'notice'
+					'utm_source'   => 'notice',
 				), self::$tecUrl . 'license-keys/'
 			);
 
@@ -737,7 +737,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 			$tribe_licences_tab_fields = array(
 				'info-start'               => array(
 					'type' => 'html',
-					'html' => '<div id="modern-tribe-info">'
+					'html' => '<div id="modern-tribe-info">',
 				),
 				'info-box-title'           => array(
 					'type' => 'html',
@@ -757,17 +757,17 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 				),
 				'info-end'                 => array(
 					'type' => 'html',
-					'html' => '</div>'
+					'html' => '</div>',
 				),
 				'tribe-form-content-start' => array(
 					'type' => 'html',
-					'html' => '<div class="tribe-settings-form-wrap">'
+					'html' => '<div class="tribe-settings-form-wrap">',
 				),
 				// TODO: Figure out how properly close this wrapper after the license content
 				'tribe-form-content-end'   => array(
 					'type' => 'html',
-					'html' => '</div>'
-				)
+					'html' => '</div>',
+				),
 			);
 			new Tribe__Events__Settings_Tab( 'general', __( 'General', 'tribe-events-calendar' ), $generalTab );
 			new Tribe__Events__Settings_Tab( 'display', __( 'Display', 'tribe-events-calendar' ), $displayTab );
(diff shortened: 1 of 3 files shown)


src/Tribe/Main.php
```diff
@@ -4249,10 +4247,10 @@ public function register_30min_interval( $schedules ) {
 		 * @param array $post
 		 *
 		 * @return array
-		 * @see Tribe__Events__Main::filter_wp_import_data_after()
+		 * @see Tribe__Events__Mainfilter_wp_import_data_after()
```
bordoni: There is a few instances where the Docblock was replaced like this. ;)

src/functions/template-tags/general.php
```diff
@@ -1135,7 +1135,7 @@ function tribe_get_days_between( $start_date, $end_date, $day_cutoff = '00:00' )
 			$end_date->modify( '-1 day' );
 		}
 
-		return Tribe__Events__Date_Utils::dateDiff( $start_date->format( 'Y-m-d ' . $day_cutoff ), $end_date->format( 'Y-m-d ' . $day_cutoff ) );
+		return Tribe__Events__Date_Utils::date_diff( $start_date->format( 'Y-m-d ' . $day_cutoff ), $end_date->format( 'Y-m-d ' . $day_cutoff ) );
```
bordoni: I see that you've changed this method's name, but you didn't change on `Tribe__Events__Date_Utils`;

My question is, will you do each file in a Pull Request or do everything here then merge?

If you are going to do one class by PR, this change is not meant to be here right?

borkweb: It _is_ changed here: https://github.com/moderntribe/the-events-calendar/blob/release/119/src/Tribe/Date_Utils.php#L102

The change in `Query.php` is uncommitted on my local environment (I missed pushing it with these changes).

borkweb: I added the additional files because I found `dateDiff` to `date_diff` changes in the previous PR.

bordoni: So we are good, I was looking at the wrong page, I used the GH search and it defaults to the `develop` branch instead of `release/119`.

src/Tribe/Main.php
```diff
@@ -1361,8 +1377,8 @@ public function displayEventVenueDropdown( $postId ) {
 
 			?>
 			<tr class="">
-				<td style="width:170px"><?php printf( __('Use Saved %s:', 'tribe-events-calendar'), $this->singular_venue_label ); ?></td>
-				<td><?php $this->saved_venues_dropdown( $VenueID ); ?> <div class="edit-venue-link" <?php if( empty( $VenueID ) ) { ?>style="display:none;"<?php } ?>><a data-admin-url="<?php echo admin_url('post.php?action=edit&post=' ); ?>" href="<?php echo admin_url( sprintf( 'post.php?action=edit&post=%s', $VenueID ) ); ?>" target="_blank"><?php printf( __( 'Edit %s', 'tribe-events-calendar' ), $this->singular_venue_label ); ?></a></div></td>
+				<td style="width:170px"><?php echo esc_html( sprintf( __( 'Use Saved %s:', 'tribe-events-calendar' ), $this->singular_venue_label ) ); ?></td>
```
zbtirrell: this escaping could cause a bug.  (I know you are confirming now based on an offline conversation)

src/Tribe/Main.php
```diff
@@ -1412,9 +1428,9 @@ public function displayEventOrganizerDropdown( $postId ) {
 			?>
 			<tr class="">
 				<td style="width:170px">
-					<label for="saved_organizer"><?php printf( __( 'Use Saved %s:', 'tribe-events-calendar' ), $this->singular_organizer_label ); ?></label>
+					<label for="saved_organizer"><?php echo esc_html( sprintf( __( 'Use Saved %s:', 'tribe-events-calendar' ), $this->singular_organizer_label ) ); ?></label>
```
zbtirrell: this escaping could cause a bug.  (I know you are confirming now based on an offline conversation)

src/Tribe/Main.php
borkweb: It does not cause a bug in the way that you were thinking..._however_ it _may_ cause a bug if that value has been filtered by an end user and has had HTML added via the `tribe_venue_label_singular` filter.

src/Tribe/Main.php
```diff
@@ -1718,30 +1738,7 @@ public function localizeAdmin() {
...
-			}
-			$data .= "\n};\n";
-
-			echo "<script type='text/javascript'>\n";
-			echo "/* <![CDATA[ */\n";
-			echo $data;
-			echo "/* ]]> */\n";
-			echo "</script>\n";
-
+			wp_localize_script( 'tribe-events-admin', 'TEC', $this->localizeAdmin() );
```
zbtirrell: This is a significant change (a good one from the looks of it).  Does this cover all the use cases?  ex: is the function being used anywhere weird, as in, when  `tribe-events-admin` is not yet registered or enqueued?

src/Tribe/Main.php
```diff
@@ -2642,6 +2642,7 @@ public function dateToTimeStamp( $date, $hour, $minute, $meridian ) {
 		 * @todo remove - unused
 		 */
 		protected function dateHelper( $date ) {
+			_deprecated_function( __FILE__, '3.11', 'date' );
```
zbtirrell: interesting...

src/Tribe/Main.php
borkweb: A good question, though luckily one that I checked :D It _is_ being used by `events-community` after `tribe-events-admin` has already been enqueued.

src/Tribe/Main.php
jazbek: note my @todo -- this function is not used anywhere in our code and can actually just be removed. i don't think we need to deprecate class methods.

borkweb: Yeah, though we hadn't thrown deprecation notices for it before, so it is _possible_ that someone is using one of those functions.  With the notice in place, we can trash the functions in a future release.

zbtirrell:
:+1:

src/Tribe/Main.php
jazbek: it's a protected method so only we would be using it.. i searched through the codebase and confirmed we are not when i made that note. all good if you want to keep it and throw the notice though.

borkweb: The `dateToTimeStamp` function is `public`, though - so we'll keep it around for a bit.

PR #111, opened 2015-06-12
Resolving Translations for JS templates

bordoni:
I've changed quite a bit on this Pull Request, the reason been that our previous Template structure
didn't support non-event related template data and for the Header on Mobile I saw this need.

Also I changed the way we deal with the `tribe_mobile_setup_day` now we have little bit better JS.

The `es_ES` l10n files were changed so the testing can be fully accomplish in Spanish.

_Ref:_ [#35475](https://central.tri.be/issues/35475)

--

Friendly reminder, I only touched the `src/resources/js/tribe-events-ajax-calendar.min.js` source file.
As I branched out of `release/119` someone might have changed stuff without doing the grunt task :wink:


src/views/month/loop-grid.php
```diff
@@ -24,37 +24,25 @@
...
-						$day_name = tribe_event_format_date( $daydata['date'], false );
-						?>
-						data-date-name="<?php echo $day_name ?>"
-					<?php
-					}
-					?>
-
-				<?php } ?>
+				data-day="<?php echo esc_attr( isset( $daydata['daynum'] ) ? $daydata['date'] : '' ); ?>"
+				data-tribejson='<?php echo tribe_events_template_data( null, array( 'date_name' => tribe_event_format_date( $daydata['date'], false ) ) ); ?>'
```
borkweb: `esc_attr` this value?

src/resources/js/tribe-events-ajax-calendar.js
```diff
@@ -142,37 +142,37 @@
...
-				day_data = { "date": date, "date_name": full_date_name };
-
-			$mobile_trigger
-				.removeClass( 'mobile-active' );
-
-			// If full_date_name is empty then default to highlighting the first day of the current month
-			var filter = full_date_name.length
-				? '[data-date-name="' + full_date_name + '"]'
-				: '.tribe-events-thismonth[data-day="' + date + '"]';
+		function tribe_mobile_setup_day( $date ) {
```
borkweb: I like the parameters in this updated function. I was originally curious about backward compatibility here, but having read through the code and seeing how/when this is called, I'm doubtful that we'd be impacting customizations by folks. Thoughts?

src/resources/js/tribe-events-ajax-calendar.js
```diff
@@ -142,37 +142,37 @@
...
+				$days = $container.find( '.tribe-mobile-day' ),
+				$triggers = $calendar.find( '.mobile-trigger' ),
+				_active = '[data-day="' + data.date + '"]',
+				$day = $days.filter( _active );
+
+			data.has_events = $date.hasClass( 'tribe-events-has-events' );
+
+			// Backwards compatibility
+			// @todo "Check if we can remove this check"
+			if ( data.has_events ){
```
borkweb: Space between `){`

borkweb:
Nice work in this PR, @bordoni!

src/views/month/loop-grid.php
bordoni: No we shouldn't, since the Functions already does it.

barryhughes:
Yep, looks good to me (from a back-compat angle): let's just be sure to have the tester confirm the same by using template overrides based on those that ship with the last tagged release.

:+1:

PR #113, opened 2015-06-12
Now the Datepickers respect the Option selected

bordoni:
_Ref:_ [#36646](https://central.tri.be/issues/36646)


src/Tribe/Date_Utils.php
```diff
@@ -21,18 +21,50 @@ class Tribe__Events__Date_Utils {
...
 		const DBYEARMONTHTIMEFORMAT = 'Y-m';
 
+
+		/**
+		 * Get the datepicker format, that is used to translate the option from the DB to a string
+		 *
+		 * @param  int $translate The db Option from datepickerFormat
+		 * @return string|array            If $translate is not set returns the full array, if not returns the `Y-m-d`
+		 */
+		public static function datepicker_formats( $translate = null ) {
```
borkweb: This function would probably make more sense with the name `datepicker_format` as it only returns a single value.

borkweb: Nevermind, I didn't see the `return $formats;` below.

src/Tribe/Main.php
```diff
@@ -291,7 +291,7 @@ public function before_html_data_wrapper( $html ) {
 
 			$data_attributes = array(
 				'live_ajax'         => tribe_get_option( 'liveFiltersUpdate', true ) ? 1 : 0,
-				'datepicker_format' => tribe_get_option( 'datepickerFormat' ),
+				'datepickerFormat'  => tribe_get_option( 'datepickerFormat' ),
```
borkweb: We should probably keep this key the same for compatibility reasons.

src/resources/js/events-admin.js
```diff
@@ -217,41 +217,41 @@ jQuery( document ).ready( function( $ ) {
...
-				if ( this.id === "EventStartDate" && $recurrence_type.val() !== 'None' ) {
-
-					var startDate = $( '#EventStartDate' ).data( 'prevDate' ),
-						dateDif = null == startDate ? 0 : date_diff_in_days( startDate, $end_date.datepicker( 'getDate' ) ),
-						endDate = new Date( date.setDate( date.getDate() + dateDif ) );
-
-					$end_date
-						.datepicker( "option", option, endDate )
-						.datepicker( "setDate", endDate );
+				dateFormat     : date_format,
```
borkweb: You tabbed this whole block in 1 tab too far.

src/Tribe/Main.php
bordoni: The problem is that we cannot, since HTML might bug with `data-datepicker_format`, I changed it all the places we used it, on Core.

borkweb: I must be missing something obvious - how would `date-datepicker_format` cause a bug?

I can say with confidence that uppercase letters are not desirable in attribute names [according to W3C](http://www.w3.org/html/wg/drafts/html/master/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes).

> A custom data attribute is an attribute in no namespace whose name starts with the string "data-", has at least one character after the hyphen, is XML-compatible, and contains __no uppercase ASCII letters__.

bordoni: So I think it was my Sublime Text getting crazy. It marks the HTML as invalid on the "colors" for some reason.

I will change to the other one.

src/resources/js/events-admin.js
bordoni: Shouldn't the end of a `object` end on the same structure as the variables below?

Correct?
```
var $recurrence_type = $( '[name="recurrence[type]"]' ),
      $end_date = $( '#EventEndDate' );
      datepickerOpts = {
            variables: true
      };
```

If I did like you suggested the behavior would be like this:
```
var $recurrence_type = $( '[name="recurrence[type]"]' ),
      $end_date = $( '#EventEndDate' );
      datepickerOpts = {
            variables: true
};
```

borkweb: The end of an object should end where the beginning starts (if that makes sense).

How it is currently:
```js
var datepickerOpts = {
		dateFormat     : date_format,
		showAnim       : 'fadeIn',
		changeMonth    : true,
		changeYear     : true,
		numberOfMonths : get_datepicker_num_months(),
		firstDay       : startofweek,
		showButtonPanel: true,
		beforeShow     : function( element, object ) {
			object.input.datepicker( 'option', 'numberOfMonths', get_datepicker_num_months() );
			object.input.data( 'prevDate', object.input.datepicker( "getDate" ) );
		},
		onSelect       : function( selectedDate ) {
			var option = this.id == "EventStartDate" ? "minDate" : "maxDate",
				instance = $( this ).data( "datepicker" ),
				date = $.datepicker.parseDate( instance.settings.dateFormat || $.datepicker._defaults.dateFormat, selectedDate, instance.settings );

			if ( this.id === "EventStartDate" && $recurrence_type.val() !== 'None' ) {

				var startDate = $( '#EventStartDate' ).data( 'prevDate' ),
					dateDif = null == startDate ? 0 : date_diff_in_days( startDate, $end_date.datepicker( 'getDate' ) ),
					endDate = new Date( date.setDate( date.getDate() + dateDif ) );

				$end_date
					.datepicker( "option", option, endDate )
					.datepicker( "setDate", endDate );

			}
			else {
				dates
					.not( this )
					.not( '#recurrence_end' )
					.datepicker( "option", option, date );
			}
		}
	};
```

Tabbed in 1 deep:

```js
var datepickerOpts = {
	dateFormat     : date_format,
	showAnim       : 'fadeIn',
	changeMonth    : true,
	changeYear     : true,
	numberOfMonths : get_datepicker_num_months(),
	firstDay       : startofweek,
	showButtonPanel: true,
	beforeShow     : function( element, object ) {
		object.input.datepicker( 'option', 'numberOfMonths', get_datepicker_num_months() );
		object.input.data( 'prevDate', object.input.datepicker( "getDate" ) );
	},
	onSelect       : function( selectedDate ) {
		var option = this.id == "EventStartDate" ? "minDate" : "maxDate",
			instance = $( this ).data( "datepicker" ),
			date = $.datepicker.parseDate( instance.settings.dateFormat || $.datepicker._defaults.dateFormat, selectedDate, instance.settings );

		if ( this.id === "EventStartDate" && $recurrence_type.val() !== 'None' ) {

			var startDate = $( '#EventStartDate' ).data( 'prevDate' ),
				dateDif = null == startDate ? 0 : date_diff_in_days( startDate, $end_date.datepicker( 'getDate' ) ),
				endDate = new Date( date.setDate( date.getDate() + dateDif ) );

			$end_date
				.datepicker( "option", option, endDate )
				.datepicker( "setDate", endDate );

		}
		else {
			dates
				.not( this )
				.not( '#recurrence_end' )
				.datepicker( "option", option, date );
		}
	}
};
```

The example you gave is interesting because it uses comma-separated vars (which I'm no fan of) and aligns the closing brace with the beginning of a variable where the `var` is absent. Doing so is possible if you start using tabs after the `var` - which is kind of odd.

I hold that a clean way to do your example is as follows:

```js
var $recurrence_type = $( '[name="recurrence[type]"]' );
var $end_date = $( '#EventEndDate' );
var datepickerOpts = {
    variables: true
};
```

The benefits are:

1. Object ends are even with their beginning
1. The same amount of characters are used when writing that block of code (instead of spaces and a tab, you use `var` and a space - instead of a comma you use a semicolon)
1. The block of code is easier to manipulate when you want to add more (or remove) declarations. You don't need to convert semicolons to commas (and vice versa).
1. Diffs in version control will result in fewer lines changed when adding/removing variables.
1. There's [less typing](http://benalman.com/news/2012/05/multiple-var-statements-javascript/) when making an addition!

bordoni: I know what you meant; It's a matter of choosing one and sticking with it. I like the Tabbed variables :stuck_out_tongue_closed_eyes: 

_Note:_ It's a little bit faster, not that it matters; http://jsperf.com/var-type-comparison/5

borkweb: The speed doesn't _really_ matter because we run minified versions in production and the vars declarations are combined at that point.

bordoni: So we will stick with multiple `var declarations = true;`?

borkweb: Not yet. We need to discuss it in a larger group. That being said, _that_ conversation doesn't matter with this changeset/block of code in question. Whichever way we choose, this block is still tabbed in too deep.

bordoni: :wink:

src/resources/js/events-admin.js
```diff
@@ -146,7 +146,7 @@ Date.prototype.format = function( mask, utc ) {
  */
 jQuery( document ).ready( function( $ ) {
 
-	var $date_format      = $( '[data-datepicker_format]' ),
+	var $date_format      = $( '[data-datepickerFormat]' ),
```
borkweb: Looks like this file still has reference to `datepickerFormat` in 3 places rather than `datepicker_format`.

src/Tribe/Main.php
```diff
@@ -291,7 +291,7 @@ public function before_html_data_wrapper( $html ) {
 
 			$data_attributes = array(
 				'live_ajax'         => tribe_get_option( 'liveFiltersUpdate', true ) ? 1 : 0,
-				'datepicker_format' => tribe_get_option( 'datepickerFormat' ),
+				'datepicker_format'  => tribe_get_option( 'datepickerFormat' ),
```
borkweb: Spacing is misaligned here with the change of the variable name :)

borkweb:
Glorious!

PR #114, opened 2015-06-12
Default view is month (#29374)

barryhughes:
Month view is our default calendar view (and shows as such in *Events > Settings > Display* in a clean install).

This fixes a problem where:

* Within a brand new install, month view shows as default in Events → Settings → Display
* On following the view events link we are presented with list view instead
* Behind the scenes, the default view option was also then changed to month view creating a janky experience


borkweb:
Looks good! :100:

PR #115, opened 2015-06-12
Event descriptions (of password protected events) should not be exposed in tooltips (#29379)

barryhughes:
Instead of exposing protected content in tooltips, this change supplies a message telling the user to visit the event and enter the password.

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -1232,7 +1232,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 		} else {
 			$excerpt = $event->post_content;
 		}
-		$excerpt = isset( $do_not_truncate ) ? $excerpt : Tribe__Events__Main::instance()->truncate( $excerpt, 30 );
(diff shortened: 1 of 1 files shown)


borkweb:
That's a good thought! That's probably worth a tweak.

barryhughes:
Updated, can someone take a look at the tweak?

src/functions/template-tags/general.php
```diff
@@ -1222,12 +1222,17 @@ function tribe_events_template_data( $event, array $additional = null ) {
...
+			$excerpt = apply_filters( 'tribe_events_template_data_password_required', $password_required_msg );
+			$do_not_truncate = true;
+		}
+		elseif ( has_excerpt( $event->ID ) ) {
 			$excerpt = $event->post_excerpt;
 		} else {
 			$excerpt = $event->post_content;
 		}
-		$excerpt = Tribe__Events__Main::instance()->truncate( $excerpt, 30 );
+		$excerpt = isset( $do_not_truncate ) ? $excerpt : Tribe__Events__Main::instance()->truncate( $excerpt, 30 );
```
borkweb: Maybe check the truthiness of `$do_not_truncate` at the same time with `empty()` rather than `isset()`?

borkweb:
:dog2: :dash:

PR #116, opened 2015-06-12
Bugfix/29386 maintain default view setting

barryhughes:
Stops the default view setting from automatically switching to the first enabled view even if this is undesirable. For context please see (internal ref) [#29386](https://central.tri.be/issues/29386).


borkweb:
:+1:

PR #117, opened 2015-06-12
Day view should respect the hide-from-listings setting (#29508)

barryhughes:
Stops events where hide-from-listings is enabled from showing in day view.

See: https://central.tri.be/issues/29508


borkweb:
Glorious and complex! :icecream:

PR #121, opened 2015-06-12
Tickets: datepicker opts for one ticket shouldn't be inherited by other tickets (#30688)

barryhughes:
See (internal ref) #30688 for context: the same form is reused for each ticket that is edited in the event editor. This change prevents the min/max date options set for the first ticket being inherited by the next.


borkweb:
:+1:

PR #122, opened 2015-06-12
gcal link always strip html and add newline spacing

jesseeproductions:
Remove HTML from Google Calendar Import Link (Bug #30699)


src/Tribe/iCal.php
```diff
@@ -74,7 +74,7 @@ public static function single_event_links() {
 		}
 
 		echo '<div class="tribe-events-cal-links">';
-		echo '<a class="tribe-events-gcal tribe-events-button" href="' . esc_url( tribe_get_gcal_link() ) . '" title="' . __( 'Add to Google Calendar', 'tribe-events-calendar' ) . '">+ ' . __( 'Google Calendar', 'tribe-events-calendar' ) . '</a>';
+		echo '<a class="tribe-events-gcal tribe-events-button" href="' . tribe_get_gcal_link() . '" title="' . __( 'Add to Google Calendar', 'tribe-events-calendar' ) . '">+ ' . __( 'Google Calendar', 'tribe-events-calendar' ) . '</a>';
```
borkweb: We could make a special escape function like we discussed on the call that could look somewhat like this:

```php
public function esc_gcal_url( $url ) {
  $url = str_replace( '%0a', 'TRIBE-GCAL-LINEBREAK', $url );
  $url = esc_url( $url );
  $url = str_replace( 'TRIBE-GCAL-LINEBREAK', '%0a', $url );
  return $url;
}
```

That should fully escape everything and then re-inject the desired `%0a`. Then we can escape on this line like so:

```php
href="' . $something->esc_gcal_url( tribe_get_gcal_link() ) . '"
```

Where does the escaping function belong? The googleCalendarLink method is in `Tribe__Events__Main` so it feels like it should be in the same location.  Really, those functions belong elsewhere but that's probably a larger conversation.

rochow:
Expanding on the "always strip HTML" point, I've just fixed this on a site. It's a core logic problem in googleCalendarLink();

At the moment:
- Get content
- Strip tags
- Add permalink
- Apply the_content filters

It should be:
- Get content
- Apply the_content filters
- Strip tags
- Add permalink

Visual Composor, Gravity Forms etc (anything shortcode based) at present will have their HTML pulled in as the description because they are part of the_content filters (do_shortcodes), which is run AFTER strip_tags. This should all happen BEFORE strip_tags.

Alternatively a better solution is to use the_excerpt (given that's what it's designed for - general overviews) with the_content as a fallback. And still change the logic order as per above.

I haven't submitting a patch as it's simple to change it around.

jesseeproductions:
Hi Rochow,

Could be something we do is use the excerpt instead of the content. 

I am not seeing us apply the content filters last though, I just see us getting the content and applying the filters then checking for string length.  Either way the order you propose is going to be the order we do it in.

Thanks!

rochow:
```php
function googleCalendarLink()
the-events-calendar.class.php

$event_details = get_the_content();

			//Truncate Event Description and add permalink if greater than 996 characters
			if ( strlen( $event_details ) > 996 ) {
				//Strip tags
				$event_details = strip_tags( $event_details );

				$event_url     = get_permalink();
				$event_details = substr( $event_details, 0, 996 );

				//Only add the permalink if it's shorter than 900 characters, so we don't exceed the browser's URL limits
				if ( strlen( $event_url ) < 900 ) {
					$event_details .= ' (View Full Event Description Here: ' . $event_url . ')';
				}
			}

			$params = array(
				'action'   => 'TEMPLATE',
				'text'     => urlencode( strip_tags( $post->post_title ) ),
				'dates'    => $dates,
				'details'  => urlencode( apply_filters( 'the_content', $event_details ) ),
				'location' => urlencode( $location ),
				'sprop'    => get_option( 'blogname' ),
				'trp'      => 'false',
				'sprop'    => 'website:' . home_url(),
			);
```

I can't figure out how to emphasise the 3 lines inside the code block, you can see though:
- Content is fetched
- Strlen determined, tags stripped etc
- Right at the death when it's being added into the array, filters are being applied. This now runs all shortcodes, adds further formatting.

If keeping apply_filters it should be on the first line of the code so it doesn't bypass the strip_tags etc.

jesseeproductions:
Looks like it might have changed, this is what I am working with:

			```$event_details = apply_filters( 'the_content', get_the_content() );

 			// Hack: Add space after paragraph
			// Normally Google Cal understands the newline character %0a
			// And that character will automatically replace newlines on urlencode()
			$event_details = str_replace ('</p>', '</p> ', $event_details);
			
			$event_details = strip_tags( $event_details );
	
			//Truncate Event Description and add permalink if greater than 996 characters
			if ( strlen( $event_details ) > 996 ) {

				$event_url     = get_permalink();
				$event_details = substr( $event_details, 0, 996 );

				//Only add the permalink if it's shorter than 900 characters, so we don't exceed the browser's URL limits
				if ( strlen( $event_url ) < 900 ) {
					$event_details .= sprintf( ' (View Full %1$s Description Here: %2$s)', $this->singular_event_label, $event_url );
				}
			}

			$params = array(
				'action'   => 'TEMPLATE',
				'text'     => urlencode( strip_tags( $post->post_title ) ),
				'dates'    => $dates,
				'details'  => urlencode( $event_details ),
				'location' => urlencode( $location ),
				'sprop'    => get_option( 'blogname' ),
				'trp'      => 'false',
				'sprop'    => 'website:' . home_url(),
			);
			$params = apply_filters( 'tribe_google_calendar_parameters', $params, $postId );
			$url    = add_query_arg( $params, $base_url );

			return $url;
```

jesseeproductions:
agree though that was backwards way of doing things!

rochow:
Odd, I haven't logged into admin on this one as the timestamps are all 8th June, assumed this site was running the latest version! Perhaps:

$event_details = get_the_excerpt() ?  get_the_excerpt() : apply_filters( 'the_content', get_the_content() );

Or similar might be better. Could always add an option to the admin panel for people to select, depends if you want yet another option though!

jesseeproductions:
Thanks for the feedback, I could see us making it an option in the admin, but that is beyond the scope of my ticket here. 

If you think that is something you would like as a feature I encourage you to visit our User Voice Page and either upvote an existing request or make a new request there. 

http://tribe.uservoice.com/

borkweb:
Looks good! :+1: Let's merge it!

PR #123, opened 2015-06-12
Register meta if Register Meta class exists (#37024)

barryhughes:
Changed the check so we *do* register meta groups etc when the relevant class exists.


borkweb:
:+1:

barryhughes:
Thanks to Leah for catching this!

PR #125, opened 2015-06-15
Fixing Timezone on #30999

bordoni:
Brian is a Genius, really quick fix -- Did a fair bit of testing it works like a charm :wink: 

_Ref:_ [#30999 @ N30](https://central.tri.be/issues/30999#note-30)


zbtirrell:
:+1:

PR #127, opened 2015-06-16
Observe the start of week for ticket datepickers

borkweb:
Use the `#tribe-event-datepickers` div to determine the `firstDay` for the jQuery UI datepicker.

See: https://central.tri.be/issues/33849

_TODO: move that setting to a JS object that gets localized, but that'll come when we start to refactor the JS._


bordoni:
:wine_glass: :wink: 

_When will we do the JS refactor?_

borkweb:
4.0 is the hope!

PR #130, opened 2015-06-17
Empty Defaults Improvements

bordoni:
_Ref:_ [#27096](https://central.tri.be/issues/27096)


src/Tribe/Field.php
```diff
@@ -161,11 +161,10 @@ public function __construct( $id, $field, $value = null ) {
 			$value            = is_array( $value ) ? array_map( 'esc_attr', $value ) : esc_attr( $value );
 			$conditional      = $args['conditional'];
 			$display_callback = $args['display_callback'];
-			$if_empty         = (bool) $args['if_empty'];
+			$if_empty         = ( is_numeric( $args['if_empty'] ) ? (bool) $args['if_empty'] : (string) $args['if_empty'] );
```
borkweb: Why cast it to a boolean at all? Why not simply do:

```php
// if it is a string, trim it so blanks are evaluated as false.
$if_empty = is_string( $args['if_empty'] ) ? trim( $args['if_empty'] ) : $args['if_empty'];
```

borkweb:
:rocket:

PR #131, opened 2015-06-17
i can't get around this error when validating a translation without u…

ggwicz:
…sing simpler string replacement here – https://cloudup.com/cOgaxtGnNmk

### DIFF
--- a/src/views/month/single-day.php
+++ b/src/views/month/single-day.php
@@ -40,7 +40,7 @@ $day = tribe_events_get_current_month_day();
 	<div class="tribe-events-viewmore">
 		<?php
 
-			$view_all_label = sprintf( _n( 'View 1 %1$s', 'View All %2$s %3$s', $day['total_events'], 'tribe-events-calendar' ), $events_label_singular, $day['total_events'], $events_label_plural );
+			$view_all_label = sprintf( _n( 'View 1 %s', 'View All %s %s', $day['total_events'], 'tribe-events-calendar' ), $events_label_singular, $day['total_events'], $events_label_plural );
 
 		?>
 		<a href="<?php echo esc_url( $day['view_more'] ); ?>"><?php echo $view_all_label ?> &raquo;</a>


src/views/month/single-day.php
```diff
@@ -40,7 +40,7 @@
 	<div class="tribe-events-viewmore">
 		<?php
 
-			$view_all_label = sprintf( _n( 'View 1 %1$s', 'View All %2$s %3$s', $day['total_events'], 'tribe-events-calendar' ), $events_label_singular, $day['total_events'], $events_label_plural );
+			$view_all_label = sprintf( _n( 'View 1 %s', 'View All %s %s', $day['total_events'], 'tribe-events-calendar' ), $events_label_singular, $day['total_events'], $events_label_plural );
```
borkweb: Switching to that format results in the following `$view_all_label` when there are more than 1 events in viewmore:

```
View All Event 4
```

borkweb: The `%s` just grabs the params in order starting with the first one (`$events_label_singular`) regardless of which string `_n()` chooses.

afragen:
Perhaps it might be better to change the string from `View 1 %1$s` to `View single %1$s`. The previous code is much better for multiple strings substitution.

src/views/month/single-day.php
ggwicz: Damn, sorry I missed that here. Do you know what could be causing the sort of error in the screenshot? Things both here *and* in the translation file all seem accurate to me ¯\_(ツ)_/¯

borkweb: This [Stackoverflow answer](http://stackoverflow.com/a/27036053) gave a clue, I think. Not all languages handle plurals the same so the same variables need to be in both strings. What if we do the following:

Up above (on lines 16 - 19):
```php
// Change:
$events_label_singular = tribe_get_event_label_singular();
$events_label_plural = tribe_get_event_label_plural();

$day = tribe_events_get_current_month_day();

// To:
$day = tribe_events_get_current_month_day();
$events_label = ( 1 === $day['total_events'] ) ? tribe_get_event_label_singular() : tribe_get_event_label_plural();
```

Then update the line in question with:
```php
$view_all_label = sprintf(
  _n(
    'View %s %s',
    'View All %s %s',
    $day['total_events']
  ),
  $day['total_events'],
  $events_label
);
```

ggwicz:
*Made a ticket for this* → https://central.tri.be/issues/38870

PR #135, opened 2015-06-18
Support the overriding of branch checkouts with command line args

borkweb:
This changeset is aimed to help our plugin release process.

See: https://central.tri.be/issues/35525


bordoni:
:dragon: :dash:

PR #137, opened 2015-06-18
Display cost ranges when multiple tickets/costs exist on an event

borkweb:
Relocated some cost logic into the new `Cost_Utils` class!

See: https://central.tri.be/issues/33715

### DIFF
--- a/src/Tribe/Cost_Utils.php
+++ b/src/Tribe/Cost_Utils.php
@@ -41,12 +41,22 @@ class Tribe__Events__Cost_Utils {
 	/**
 	 * fetches an event's cost values
 	 *
-	 * @param int $event_id The ID of the event
+	 * @param int|WP_Post $event The Event post object or event ID
 	 *
 	 * @return array
 	 */
-	public function get_event_costs( $event_id ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Cost_Utils.php
```diff
@@ -39,6 +39,97 @@ public function get_all_costs() {
...
 
 	/**
+	 * fetches an event's cost values
+	 *
+	 * @param int $event_id The ID of the event
+	 *
+	 * @return array
+	 */
+	public function get_event_costs( $event_id ) {
+		$costs = tribe_get_event_meta( $event_id, '_EventCost', false );
```
bordoni: We could allow the some `WP_Post` to be the param:
```
$event = get_post( $event );
if ( ! is_object( $event ) || ! $event instanceof WP_Post ){
	return; // I dunno what would be the correct return here
}

if ( ! tribe_is_event( $event->ID ) ){
	return; // I dunno what would be the correct return here
}
```

borkweb: Yeah, that's a good call. Done.

bordoni:
:wine_glass: :wink:

PR #138, opened 2015-06-18
Fix and update some German translations

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>


zbtirrell:
Thanks for this submission!  However, we recently moved our translations to be managed here: http://translations.theeventscalendar.com/projects/tribe-events-calendar/

We really appreciate the submissions, but the best place to get these contributions accepted is through that site.

PR #139, opened 2015-06-18
Make note of any rewrite rule flushing that occurs when on the Help tab

borkweb:
See: https://central.tri.be/issues/34885


bordoni:
:wine_glass: :turtle:

PR #112, opened 2015-06-12
Prevent Welcome page on WP Actions

bordoni:
_Ref:_ [#33897](https://central.tri.be/issues/33897)

PR #140, opened 2015-06-18
Suppress splash screen sometimes. Don't be as nitpicky about showing it

borkweb:
When activating a plugin, there is a `nonce`, `action`, and `s` parameter on the URL which prevents the splash screen from displaying. Don't worry about those vars.

See: https://central.tri.be/issues/34265


bordoni:
Doesn't this breaks what I did on: https://github.com/moderntribe/the-events-calendar/pull/112

borkweb:
Some of it. That nonce/s/action line broke the welcome page completely.

bordoni:
:+1:

PR #141, opened 2015-06-19
Allow view selection to function with JS disabled

borkweb:
See: https://central.tri.be/issues/27982

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -707,10 +707,36 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		 */
 		public function template_redirect() {
 			global $wp_query;
+
+			// if JS is disabled, then we need to handle tribe bar submissions manually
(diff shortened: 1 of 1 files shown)


src/Tribe/Main.php
```diff
@@ -707,12 +707,38 @@ public function initOptions() {
...
 		/**
+		 * handles tribe bar post submissions
+		 *
+		 * @param array $postdata Data from $_POST
+		 */
+		public function handle_submit_bar_redirect( $postdata ) {
+			$url = $postdata['tribe-bar-view'];
+
+			foreach ( $postdata as $key => $value ) {
+				if ( 'submit-bar' === $key || 'tribe-bar-view' === $key ) {
```
bordoni: Why not `in_array`?

borkweb: It takes more typing if there are only two values :)

bordoni: Right... Ok...

bordoni:
:wine_glass: :snowman:

PR #142, opened 2015-06-19
Override default args for event category checkboxes to retain hierarchy

borkweb:
See: https://central.tri.be/issues/33700


bordoni:
:dragon: :dash: -- Just needs a Conflict resolution

PR #145, opened 2015-06-19
BugFix #29769 - Convert RSS Event Date to UTC000

jesseeproductions:
WordPress always outputs a pubDate set to 00:00 (UTC) so this takes that
into account with the event date.

Resolves this ticket:
https://central.tri.be/issues/29769

### DIFF
--- a/src/Tribe/Templates.php
+++ b/src/Tribe/Templates.php
@@ -667,7 +667,7 @@ if ( ! class_exists( 'Tribe__Events__Templates' ) ) {
 		}
 
 		/**
-		 * convert the post_date_gmt to the event date for feeds
+		 * Convert the post_date_gmt to the event date for feeds
 		 *
 		 * @param $time the post_date
 		 * @param $d    the date format to return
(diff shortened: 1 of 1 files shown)


src/Tribe/Templates.php
```diff
@@ -679,7 +679,25 @@ public static function event_date_to_pubDate( $time, $d, $gmt ) {
 			global $post;
 
 			if ( is_object( $post ) && $post->post_type == Tribe__Events__Main::POSTTYPE && is_feed() && $gmt ) {
-				$time = tribe_get_start_date( $post->ID, false, $d );
+
+				// Don't interfere if the timezone hasn't been set
+				$zone = get_option( 'timezone_string', false );
+				if ( false === $zone ) {
+					return $time;
```
borkweb: Are we 100% sure that `$time` is in the desired format (`$d`) at this point in time? It may be a good plan to format the date (as UTC) rather than just returning here.

src/Tribe/Templates.php
```diff
@@ -679,7 +679,25 @@ public static function event_date_to_pubDate( $time, $d, $gmt ) {
...
+				$zone = new DateTimeZone( $zone );
+
+				// Apply timezone and calculate offset
+				$time->setTimezone( $zone );
+				$offset = $zone->getOffset( $time );
+				$offset *= - 2;
+				
+				// Apply correction as WordPress always outputs a pubDate set to 00:00 (UTC)
+				$time->modify( "$offset seconds" );
+				$time = $time->format( TribeDateUtils::DBDATETIMEFORMAT );
```
borkweb: I think this whole block could be simplified a bit.

We should be able to check if there is a timezone or not (like you are doing), then initialize a DateTimeZone object to be either `$zone` or `UTC`. Then set up the time object with a second parameter of `$zone`. Convert to UTC and set `$time` to the format we want.

Pseudo code:

```
if zone set
  $zone = DTZ( zone )
else
  $zone = DTZ( UTC )

$time = DT( $time, $zone )
$time->setTimezone( DTZ( UTC ) )
$time = $time->format( $d )
// and mysql2date would be ditchable
```

borkweb:
Looks good! :+1:

PR #146, opened 2015-06-19
Show "View all" in month view and allow Day tpl to render when disabled

borkweb:
See: https://central.tri.be/issues/27348


bordoni:
:man_with_gua_pi_mao: :dragon:

PR #147, opened 2015-06-19
After retrieving date format from db, strip slashes (#35633)

barryhughes:
This change lets users successfully escape chars in the date format settings (internal ref #35633 for context).


borkweb:
:poodle:

PR #148, opened 2015-06-19
Add sortable events table by tax + Refactoring (redux against 119)

borkweb:
This takes the excellent work done by @kittsville and applies it to our `release/119` branch with some tweaks.  Here are the original notes from kittsville from #42:

>
**New:**  
- Added sorting by event category and tags. Code was based off similar work from [my own plugin](https://github.com/kittsville/WP-Librarian/blob/e2bb0e176633dbb38a83912aeccbb2d4cdae03c8/lib/admin-tables.class.php#L453-L493), which was based on code by Mike Schinkel.
>
**Refactored:**  
>
- Removed duplicate is_admin check
- Replaced obfuscated event date sorting logic with single function. Also got rid of unnecessary use of two SQL joins rather than just one.
- Aligned and commented hook registration for an easier to read/understand init(). Now looks more like hook registration in main plugin class
- Replaced 3 if statements with a switch for very slightly faster but mainly easier to read function logic

> [Here's a quick look at the sorting in action](http://imgur.com/a/GRpWC). Note that sorting by a taxonomy hides events that have no terms for that tax.

> While I was working on this I noticed a query is leaking from `Query.php` and potentially causing a minor performance loss for all events queries. If you check out the generated SQL with any debugging plugin you'll see the offending join being added to the admin tables queries.

> If you don't want to add the sortable feature then please still consider accepting the refactoring done in the other commits. 

I've made the following adjustments:

1. Always include date sorting to some degree. Since we're dealing with events (which are date driven), make sure to sort by start & end date - though defer the adding of start/end date sorting until _after_ the new taxonomy sorting.
1. Retain the start/end date fields as part of the fields returned in the query.
1. When events were missing terms, the `GROUP_CONCAT` was excluding those events. Changed how the query sorts that data.
1. Some minor code standards tweaks.

### DIFF
--- a/src/Tribe/Admin_List.php
+++ b/src/Tribe/Admin_List.php
@@ -146,16 +146,26 @@ if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 
 			global $wpdb;
 
-			$clauses['join'] .= "
-				LEFT OUTER JOIN {$wpdb->term_relationships} ON {$wpdb->posts}.ID = {$wpdb->term_relationships}.object_id
-				LEFT OUTER JOIN {$wpdb->term_taxonomy} USING (term_taxonomy_id)
-				LEFT OUTER JOIN {$wpdb->terms} USING (term_id)
+			// collect the terms in the desired taxonomy for the given post into a single string
+			$smashed_terms_sql = "
+				SELECT
+					GROUP_CONCAT( wp_terms.name ORDER BY name ASC ) smashed_terms
+				FROM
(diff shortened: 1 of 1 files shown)


barryhughes:
Following commit d73353d61f9c1c02403b55f7aa83a9e12578b179 we get a db error if PRO is also activated - the problem isn't necessarily in this piece of work but we should take a second look and resolve here or in PRO as needed.

` [Unknown column 'tribe_event_end_date.meta_value' in 'order clause']`

PR #149, opened 2015-06-19
Prep alternate meta query when matching existing *all day* events (#35913)

barryhughes:
Improves ability to detect existing all day events when importing from CSV (internal ref #35913 for context).

### DIFF
--- a/src/Tribe/Importer/File_Importer_Events.php
+++ b/src/Tribe/Importer/File_Importer_Events.php
@@ -10,27 +10,55 @@ class Tribe__Events__Importer__File_Importer_Events extends Tribe__Events__Impor
 	protected function match_existing_post( array $record ) {
 		$start_date = $this->get_event_start_date( $record );
 		$end_date   = $this->get_event_end_date( $record );
+		$all_day    = $this->get_boolean_value_by_key( $record, 'event_all_day' );
+
+		// Base query - only the meta query will be different
 		$query_args = array(
 			'post_type'      => Tribe__Events__Main::POSTTYPE,
 			'post_title'     => $this->get_value_by_key( $record, 'event_name' ),
-			'meta_query'     => array(
+			'fields'         => 'ids',
+			'posts_per_page' => 1,
+		);
+
+		// When trying to find matches for all day events, the comparison should only be against the date
+		// component only since a) the time is irrelevant and b) the time may have been adjusted to match
+		// the eod cutoff setting
+		if ( in_array( $all_day, array( 'yes', 'true', '1' ) ) ) {
+			$meta_query = array(
+				array(
+					'key'     => '_EventStartDate',
+					'value'   => $this->get_event_start_date( $record, true ),
+					'compare' => 'LIKE'
+				),
+				array(
+					'key'     => '_EventAllDay',
+					'value'   => 'yes'
+				)
+			);
+		// For regular, non-all day events, use the full date *and* time in the start date comparison
+		} else {
+			$meta_query = array(
 				array(
 					'key'   => '_EventStartDate',
 					'value' => $start_date,
 				),
-			),
-			'fields'         => 'ids',
-			'posts_per_page' => 1,
-		);
-		if ( ! empty( $end_date ) ) {
-			$query_args['meta_query'][] = array(
+			);
+		}
+
+		// Optionally use the end date/time for matching, where available
+		if ( ! empty( $end_date ) && ! $all_day ) {
+			$meta_query[] = array(
 				'key'   => '_EventEndDate',
 				'value' => $end_date,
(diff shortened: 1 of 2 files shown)


src/Tribe/Importer/File_Importer_Events.php
```diff
@@ -51,13 +79,17 @@ protected function create_post( array $record ) {
...
 		$start_time = $this->get_value_by_key( $record, 'event_start_time' );
-		if ( ! empty( $start_time ) ) {
+
+		if ( ! $date_only && ! empty( $start_time ) ) {
 			$start_date .= ' ' . $start_time;
 		}
-		$start_date = date( 'Y-m-d H:i:s', strtotime( $start_date ) );
+
+		$start_date = $date_only
+			? date( 'Y-m-d', strtotime( $start_date ) )
```
bordoni: We should use the `Date_Utils::DBDATEFORMAT` and `Date_Utils::DBDATETIMEFORMAT` also I think we could make this ternary a real if conditional.

src/Tribe/Importer/File_Importer_Events.php
```diff
@@ -10,27 +10,55 @@ class Tribe__Events__Importer__File_Importer_Events extends Tribe__Events__Impor
...
 			'post_title'     => $this->get_value_by_key( $record, 'event_name' ),
-			'meta_query'     => array(
+			'fields'         => 'ids',
+			'posts_per_page' => 1,
+		);
+
+		// When trying to find matches for all day events, the comparison should only be against the date
+		// component only since a) the time is irrelevant and b) the time may have been adjusted to match
+		// the eod cutoff setting
+		if ( in_array( $all_day, array( 'yes', 'true', '1' ) ) ) {
```
bordoni: We should use: `Date_Utils:: is_all_day()`.

barryhughes: Hmm, this could definitely be revised - but I'm not sure I agree a switch to `*::is_all_day()` makes sense here.

We're not inspecting a post meta field in that line - we're looking at a record from the CSV data - and `*::is_all_day()` does not account for a user using `1` to indicate an event is all day within the source CSV (which is supported by the importer).

bordoni: Question, why not implement the `1` on the `::is_all_day()` do you see any problems around that?

barryhughes: Well, we could do that - I just think it's taking us in the wrong direction. 

I feel what we really need is a separate `is_true()` type helper function that tests for truthiness and which can be localized (if we have one, I've missed it) which is important in this scenario, where information is coming from a CSV source we don't control and where users might reasonably want to use keywords in a language of their choice.

This would be more appropriate than `::is_all_day()` which is very specific, tests for values we control and cannot be localized (and since it is testing against values we set, why should it be?).

Right now I need to match the array defined in the signature of `Tribe__Events__Importer__File_Importer_Events::get_boolean_value_by_key()` which I can't access by other means. I do feel that needs to be addressed, but that it's a separate problem.

borkweb: I like the idea of a localized `is_true()`, @barryhughes. That's a neat idea!

I _do_ think that if we expect a `1` as a possible truthy value in CSV files, then adding support in our current non-localized method for checking `::is_all_day()` seems like a nice way to bridge the gap between the non-existence and existence of `is_true()`.

barryhughes: OK, done, and after a further look, no meed to modify _is_all_day()_ in this case as it turns out we don't support `1` as a possible truthy value for that field (maybe that's what you were hinting at?).

src/Tribe/Importer/File_Importer_Events.php
```diff
@@ -10,27 +10,55 @@ class Tribe__Events__Importer__File_Importer_Events extends Tribe__Events__Impor
 	protected function match_existing_post( array $record ) {
 		$start_date = $this->get_event_start_date( $record );
 		$end_date   = $this->get_event_end_date( $record );
+		$all_day    = $this->get_boolean_value_by_key( $record, 'event_all_day' );
```
borkweb: If we aren't adding `1` to `is_all_day()` (because you noted that isn't a value we are expecting), should this have a third parameter passed that allows it to return a value we are expecting in `is_all_day()`? Perhaps `true`?

barryhughes: No, I don't think so - my note/reply on an earlier commit was wrong (Gustavo was bang on earlier). We're looking for existing records in the db here.

borkweb:
:+1:

PR #151, opened 2015-06-20
Support random ordering of events (#33626)

barryhughes:
Sometimes `orderby` will be set to `rand` for event queries (example: `tribe_get_related_posts()`) - this change ensures that is respected and not overridden - internal ref #33626.


bordoni:
:wine_glass: :wink: - My only concern is that we should have some sort of warning on the Knowledge base about the performance Hit that is the usage of `RAND()`

barryhughes:
True, though, we don't necessarily have to document this in the KB - but we do need it in place for pro's related posts functionality to work as intended.

PR #154, opened 2015-06-21
Resolving DatePicker saving bug

bordoni:
I was trying a bit more the DatePicker with a few different dates, and I found a bug related to another Pull Request I've created #113.

With formats like `dd/mm/yyyy` or `d/m/yy` it would break the date on saving because our plugin was only passing the dates via `strtotime` which breaks on these formats and return a 1970 date.

Since PHP 5.2 doesn't have a `date_parse_from_format` for us to deal with formatted date parsing I've created a Reversed engineer method on our `Date_Utils` that will get these formats and return a Database formatted String that we can pass to `strtotime`.

 **E.g.:**
- `15/01/2015` to `2015-01-15`
- `2/9/2015` to `2015-9-2`
- `12/09/2012 21h 02m 00s` to `2012-09-12 21:02:00`

### DIFF
--- a/src/Tribe/Date_Utils.php
+++ b/src/Tribe/Date_Utils.php
@@ -56,7 +56,7 @@ if ( ! class_exists( 'Tribe__Events__Date_Utils' ) ) {
 		 *
 		 * @return string         A DB formated Date, includes time if possible
 		 */
-		public static function time_from_format( $format, $date ) {
+		public static function datetime_from_format( $format, $date ) {
 			// Reverse engineer the relevant date formats
 			$keys = array(
 				// Year with 4 Digits
@@ -72,7 +72,7 @@ if ( ! class_exists( 'Tribe__Events__Date_Utils' ) ) {
 				'n' => array( 'month', '\d{1,2}' ),
 
 				// Month ABBR 3 letters
-				'M' => array( 'month', '[A-Z][a-z]{3}' ),
+				'M' => array( 'month', '[A-Z][a-z]{2}' ),
 
 				// Month Name
 				'F' => array( 'month', '[A-Z][a-z]{2,8}' ),
@@ -87,7 +87,7 @@ if ( ! class_exists( 'Tribe__Events__Date_Utils' ) ) {
 				'D' => array( 'day', '[A-Z][a-z]{2}' ),
 
 				// Day Name
-				'l' => array( 'day', '[A-Z][a-z]{6,9}' ),
+				'l' => array( 'day', '[A-Z][a-z]{5,8}' ),
 
 				// Hour 12h formatted, with leading 0
 				'h' => array( 'hour', '\d{2}' ),
--- a/src/Tribe/Date_Utils.php
+++ b/src/Tribe/Date_Utils.php
@@ -10,7 +10,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 if ( ! class_exists( 'Tribe__Events__Date_Utils' ) ) {
 	class Tribe__Events__Date_Utils {
-		// default formats, they are overridden by WP options or by arguments to date methods
+		// Default formats, they are overridden by WP options or by arguments to date methods
 		const DATEONLYFORMAT        = 'F j, Y';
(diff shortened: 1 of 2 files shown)


src/Tribe/Date_Utils.php
```diff
@@ -49,6 +48,109 @@ public static function datepicker_formats( $translate = null ) {
...
+				'y' => array( 'year', '\d{2}' ),
+
+				// Month with leading 0
+				'm' => array( 'month', '\d{2}' ),
+
+				// Month without the leading 0
+				'n' => array( 'month', '\d{1,2}' ),
+
+				// Month ABBR 3 letters
+				'M' => array( 'month', '[A-Z][a-z]{3}' ),
```
borkweb: This would allow 4 letters.

src/Tribe/Date_Utils.php
```diff
@@ -49,6 +48,109 @@ public static function datepicker_formats( $translate = null ) {
...
+				'd' => array( 'day', '\d{2}' ),
+
+				// Day without leading 0
+				'j' => array( 'day', '\d{1,2}' ),
+
+				// Day ABBR 3 Letters
+				'D' => array( 'day', '[A-Z][a-z]{2}' ),
+
+				// Day Name
+				'l' => array( 'day', '[A-Z][a-z]{6,9}' ),
```
borkweb: 1 character too long, I think: `{5,8}`

src/Tribe/Date_Utils.php
```diff
@@ -49,6 +48,109 @@ public static function datepicker_formats( $translate = null ) {
...
+				's' => array( 'second', '\d{2}' ),
+			);
+
+			// Convert format string to regex
+			$regex = '';
+			$chars = str_split( $format );
+			foreach ( $chars as $n => $char ) {
+				$last_char = isset( $chars[ $n - 1 ] ) ? $chars[ $n - 1 ] : '';
+				$skip_current = '\\' == $last_char;
+				if ( ! $skip_current && isset( $keys[ $char ] ) ) {
```
borkweb: Isn't `$n` the numeric value within this loop rather than `$char`?

src/Tribe/Date_Utils.php
```diff
@@ -49,6 +48,109 @@ public static function datepicker_formats( $translate = null ) {
...
+					$regex .= '(?P<' . $keys[ $char ][0] . '>' . $keys[ $char ][1] . ')';
```
borkweb: `$n` rather than `$char`?

src/Tribe/Date_Utils.php
```diff
@@ -49,6 +48,109 @@ public static function datepicker_formats( $translate = null ) {
...
 		/**
+		 * As PHP 5.2 doesn't have a good version of `date_parse_from_format`, this is how we deal with
+		 * possible weird datepicker formats not working
+		 *
+		 * @param  string $format The weird format you are using
+		 * @param  string $date   The date string to parse
+		 *
+		 * @return string         A DB formated Date, includes time if possible
+		 */
+		public static function time_from_format( $format, $date ) {
```
borkweb: This method is more of a `datetime_from_format` rather than just `time_from_format`, right?

src/Tribe/Date_Utils.php
bordoni: Here we are looking for the `$char` key within the `$keys` that is defined on the beginning of the function, this prevents non supported date Chars to be converted to a _regex_.

src/Tribe/Date_Utils.php
bordoni: Here we are looking for the `$char` key within the `$keys` that is defined on the beginning of the function, this prevents non supported date Chars to be converted to a _regex_.

borkweb: Ohhhh...right. derp.

src/Tribe/Date_Utils.php
borkweb: Yup, my bad. Thanks.

borkweb:
Looks good! Let's merge it :+1:

PR #158, opened 2015-06-22
Minimum changes to prevent month view args from being overwritten (#36593)

barryhughes:
I'd have liked to take this a little further and eliminate more of the static members, but this smaller change is safer and should require less QA - while still resolving the problem in the [linked bug report](https://central.tri.be/issues/36593).


borkweb:
:cake:

PR #159, opened 2015-06-23
Move single event nav logic to WP_Query and skip hidden events (#35041)

barryhughes:
Use a WP_Query to build the prev/next event links, and avoid linking to events that have been hidden from listings.

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -3675,57 +3675,57 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		 */
 		public function get_event_link( $post, $mode = 'next', $anchor = false ) {
 			global $wpdb;
+			$link = '';
 
-			if ( $mode == 'previous' ) {
-				$order = 'DESC';
-				$sign  = '<';
+			if ( 'previous' === $mode ) {
+				$order      = 'DESC';
+				$direction  = '<';
 			} else {
-				$order = 'ASC';
-				$sign  = '>';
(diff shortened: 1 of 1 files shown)


src/Tribe/Main.php
```diff
@@ -3743,8 +3743,10 @@ public function get_event_link( $post, $mode = 'next', $anchor = false ) {
 					$anchor = preg_replace( '|%title%|', $title, $anchor );
 				}
 
-				return apply_filters( 'tribe_events_get_event_link', '<a href="' . esc_url( tribe_get_event_link( $results ) ) . '">' . $anchor . '</a>' );
+				$link = '<a href="' . esc_url( tribe_get_event_link( $results ) ) . '">' . $anchor . '</a>';
 			}
+
+			return apply_filters( 'tribe_events_get_event_link', $link );
```
bordoni: Here we could pass the `$args` and `$post` variables to allow a better filtering.

barryhughes: Yes, you're right - I was actually toying with providing the function arguments - `$post`, `$mode` and `$anchor` - not sure `$args` is needed or the right fit for this hook (since it can be modified in an earlier filter) - what do you think?

borkweb: `$post`, `$mode`, and `$anchor` sound good to me :+1:

barryhughes:
Added and commented :-)

bordoni:
Looking good!

PR #164, opened 2015-06-23
Allow version_compare to compare real values rather than (float) vals

borkweb:
Casting as a float butchers the version numbers. Examples:

* `(float) '3.10'` becomes `3.1`
* `(float) '3.9.3'` becomes `3.9`
* `(float) '3.9a'` becomes `3.9`

Right now, if someone has upgraded TEC to 3.10 and they have Pro at 3.9, the version check sees that 3.9 is greater than 3.10 (float value comparison: `3.9 > 3.1`). PHP's `version_compare` function supports the version numbers that we tend to use, so let's not cast. This fix will allow the outdated-plugin notification message to display once again.

~~Add-ons do a good job of making sure TEC is up to snuff. The area of code that is being changed is the reverse check - making sure that add-ons are up to date after core has been updated.  We only care about the first to parts of a version number when checking the add-ons. (e.g. We don't care about the `.4` in `3.10.4`).~~

__UPDATE:__ @barryhughes suggested we add a `MIN_ADDON_VERSION` constant. That simplifies things and will allow us to allow older addon versions if they don't make the world burn.

See: https://central.tri.be/issues/37189

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -24,10 +24,11 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		const VENUE_POST_TYPE     = 'tribe_venue';
 		const ORGANIZER_POST_TYPE = 'tribe_organizer';
 
-		const VERSION       = '3.10';
-		const FEED_URL      = 'https://theeventscalendar.com/feed/';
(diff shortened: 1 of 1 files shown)


borkweb:
We probably want to trim of the final decimal, though.  Hrm.

zbtirrell:
:+1:

PR #165, opened 2015-06-23
Update to bootstrap-datepicker 1.4

borkweb:
This resolves a bug that caused a date field to be blanked out when the enter key is pressed while the field is focused. This bug was reported in bootstrap-datepicker and [resolved a while back](https://github.com/eternicode/bootstrap-datepicker/issues/1104).

Additionally: I'm enqueuing the minified JS file.

See: https://central.tri.be/issues/27486


bordoni:
:+1:

PR #166, opened 2015-06-23
Remove phone as a mappable field (it hasn't gone anywhere for a while)

borkweb:
Phone numbers should be imported via venue or organizer data. They have been disappearing into the ether for a while now (deprecated in 2.0).

See: https://central.tri.be/issues/28448


bordoni:
:wine_glass: :dash:

PR #169, opened 2015-06-23
Support the suppression of "Results not found" notices

borkweb:
This is a sister-commit to one in Pro where the Mini Calendar widget - when no results are found - causes a global notice to be applied to a page despite the presence of results in the main event loop.

Related: https://github.com/moderntribe/events-pro/pull/48

See: https://central.tri.be/issues/37194


bordoni:
:dragon: :fire_engine:

PR #171, opened 2015-06-26
move multi-organizer view template to the src/admin-views directory

jbrinley:
When the multi-organizer feature was originally merged, a template ended up at the wrong path, due to code reorganization conflicts. This fixes it.


borkweb:
:+1:

PR #172, opened 2015-06-26
 #33566 - Only super admins should be able to turn on debug in multisite environments

elimn:
Wrapped debugMode fields in is_super_admin(). Chose to do it inline instead of hooking in and removing, that way the logic would be consolidated and easy to understand. Formatted code a bit better to remove the multiple levels of indentation.

See: https://central.tri.be/issues/33566


borkweb:
This looks good (and was much easier to review with `?w=1` on the URL). Nice work! Merge it.

PR #174, opened 2015-06-27
Fix/34202 status header handling

barryhughes:
Please see [#34202](https://central.tri.be/issues/34202) for context. 

The overall objective of this change is to avoid 404s on empty event views, the way I'm hoping to accomplish that is by taking advantage of WP's normal handling of post type archive requests (instead of filtering on `status_header` as done in the linked report).


bordoni:
:wine_glass: :wink:

zbtirrell:
@barryhughes your solution also relates to this ticket: https://central.tri.be/issues/37213  Assuming we feel good about this change, I'll slide that ticket into the current release.

PR #175, opened 2015-06-29
Don't show undefined organizers on the front end

borkweb:
If multiple organizers have been partially set up on the back end (i.e.  organizer data partially filled out), only show the organizers that have been set up in their entirety on the front end

See: [#33808](https://central.tri.be/issues/33808#note-7)


bordoni:
:wine_glass: :+1:

PR #152, opened 2015-06-20
Expose embedded map objects for easier customization potential (#33910)

barryhughes:
[PR/38](https://github.com/moderntribe/the-events-calendar/pull/38) by @tddewey was a great idea and is encapsulated by this new PR, which targets the `release/119` branch and incorporates some further tweaks to ensure we don't overwrite the map object where multiple maps are delivered in the same page.

Props: [@tddewey](https://github.com/tddewey)

PR #177, opened 2015-06-29
Assigns the google map object back to tribeEventsSingleMap global JS var

borkweb:
This is [a contribution](https://github.com/moderntribe/the-events-calendar/pull/38) by @tddewey (thanks!) that I've twiddled to be PR'd against `release/119`. Here's the original PR description:

> This allows plugin or theme code to modify the map. The tribeEventsSingleMap var will have all the data needed: initial zoom level, addresses, and the google map object itself.

> For example, we may want to disable the scrollwheel. With this commit, the following can be called from outside The Event Calendar:
`tribeEventsSingleMap.map.setOptions({scrollwheel:false});`

### DIFF
--- a/src/resources/js/embedded-map.js
+++ b/src/resources/js/embedded-map.js
@@ -71,6 +71,9 @@ if ( "function" === typeof jQuery ) jQuery( document ).ready( function( $ ) {
 			title   : venueTitle,
 			position: position
 		} );
+
+		// Assign the Google Map to a global variable so themes or plugins can interact with it
+		tribeEventsSingleMap.map = map;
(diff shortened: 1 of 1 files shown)


barryhughes:
Hey Matt:

* Wasn't this already [handled here](https://github.com/moderntribe/the-events-calendar/pull/152)
* If multiple maps are embedded, won't only the final map object be accessible with the approach in this PR?

borkweb:
Oh geez. Yes. I'll close this one in shame :). Good news is, the original PR by the contributor is now closed as well.

PR #179, opened 2015-06-30
Use wp_kses as an additional safeguard 

barryhughes:
See [#36065](https://central.tri.be/issues/36065) for context. Small enhancement which applies wp_kses() as an extra safeguard before displaying service messages.

### DIFF
--- a/src/Tribe/PUE/Checker.php
+++ b/src/Tribe/PUE/Checker.php
@@ -428,7 +428,7 @@ if ( ! class_exists( 'Tribe__Events__PUE__Checker' ) ) {
 				} else {
 					$default_success_msg    = sprintf( __( 'Valid Key! Expires on %s', 'tribe-events-calendar' ), $expiration );
 					$response['status']     = isset( $pluginInfo->api_message ) ? 2 : 1;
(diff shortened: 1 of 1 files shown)


src/Tribe/PUE/Checker.php
```diff
@@ -428,7 +428,7 @@ public function ajax_validate_key() {
 				} else {
 					$default_success_msg    = sprintf( __( 'Valid Key! Expires on %s', 'tribe-events-calendar' ), $expiration );
 					$response['status']     = isset( $pluginInfo->api_message ) ? 2 : 1;
-					$response['message']    = isset( $pluginInfo->api_message ) ? $pluginInfo->api_message : $default_success_msg;
+					$response['message']    = isset( $pluginInfo->api_message ) ? wp_kses( $pluginInfo->api_message ) : $default_success_msg;
```
bordoni: `wp_kses` second param is required, you need to pass an `Array` of the Elements that are allowed. To remove all HTML `wp_kses( $data , 'strip' );`

borkweb:
:tada:

PR #181, opened 2015-06-30
Fix/36060 mu domain reporting

barryhughes:
Reliably report the network level *domain* for multisite networks. Relates to [#36060](https://central.tri.be/issues/36060).


ggwicz:
Looks good!

PR #187, opened 2015-07-06
When using classic header, use a smaller font size for "View As" items

borkweb:
See: https://central.tri.be/issues/33823


bordoni:
:+1:

PR #188, opened 2015-07-06
We need sprintf not printf when supplying esc_attr() with a string (spotfix, no related issue)

barryhughes:
Found this while estimating on an unrelated issue: we were printf()ing when we should have been sprintf()ing.


borkweb:
:+1:

bordoni:
Good Catch!

PR #189, opened 2015-07-06
pulling all the langugae files from GlotPress

zbtirrell:
See: https://central.tri.be/issues/37346


borkweb:
:+1: to QA this!

afragen:
Just a quick question, isn't the base name for language files supposed to be the same as the plugin slug?

I know it doesn't have to be but I thought that was the convention.

zbtirrell:
I would tend to agree and prefer that convention, but our language name has been this way forever.

PR #190, opened 2015-07-06
Set wp-util as a dependency when enqueuing events-admin.js

borkweb:
`events-admin[.min].js` uses `wp.template` - a function found in the `wp-util` js file. To ensure that the object/function is available when `events-admin` executes, set it as a dependency.

See: https://central.tri.be/issues/23902


zbtirrell:
:+1:

PR #191, opened 2015-07-07
Only include published and optionally private events in month view

borkweb:
See: https://central.tri.be/issues/37458


zbtirrell:
:+1:

PR #192, opened 2015-07-08
Make month view faster

jazbek:
This PR addresses the fact that month view wasn't as fast in 3.10.1 as it was in 3.9.

Items to note:
- The month view template class now queries the postmeta table directly to get all the posts in the month
- The events are then looped through for each day of the month to find the events on that day, and the IDs found are passed to post__in on each day's query
- If there are no events on the day, no query runs for that day
- The code in `tribe_event_is_on_date()` that calculates whether an event's date falls between the start and end day of a date (accounting for EOD cutoff) was moved to a new method,  `Tribe__Events__Date_Utils::range_coincides()`. This method can now be called directly to bypass the EOD cutoff and event start/end time lookups that happened on every iteration of `tribe_event_is_on_date()`

I have just under 8000 events in my local db, and about 300 events per month. I cut down the page load time on month view by about .5 - .75s from 3.10.1

See: https://central.tri.be/issues/32067


src/Tribe/Template/Month.php
```diff
@@ -288,6 +279,40 @@ private static function view_more_link( $date ) {
...
+					WHERE tribe_event_start.meta_key = '_EventStartDate'
+					AND ( (tribe_event_start.meta_value >= '%1\$s' AND  tribe_event_start.meta_value <= '%2\$s')
+						OR (tribe_event_start.meta_value <= '%1\$s' AND tribe_event_end_date.meta_value >= '%1\$s')
+						OR ( tribe_event_start.meta_value >= '%1\$s' AND  tribe_event_start.meta_value <= '%2\$s')
+					)
+					",
+					tribe_event_beginning_of_day($this->first_grid_date),
+					tribe_event_end_of_day($this->final_grid_date)
+			);
+			$this->events_in_month = $wpdb->get_results( $events_request );
```
borkweb: Since we are opting away from using `WP_Query` (a reasonable decision), it feels like the result of this query should be cached. Maybe something like:

```php
wp_cache_set( $month, $this->events_in_month, 'tribe-events-in-month', MINUTE_IN_SECONDS * 30 );
```

Then we can manually purge the cache for the month (or for two months if the event's date is moved drastically) when an event is saved.

src/Tribe/Template/Month.php
```diff
@@ -288,6 +279,40 @@ private static function view_more_link( $date ) {
...
+					FROM $wpdb->postmeta AS tribe_event_start
+					LEFT JOIN $wpdb->postmeta as tribe_event_end_date ON ( tribe_event_start.post_id = tribe_event_end_date.post_id AND tribe_event_end_date.meta_key = '_EventEndDate' )
+					WHERE tribe_event_start.meta_key = '_EventStartDate'
+					AND ( (tribe_event_start.meta_value >= '%1\$s' AND  tribe_event_start.meta_value <= '%2\$s')
+						OR (tribe_event_start.meta_value <= '%1\$s' AND tribe_event_end_date.meta_value >= '%1\$s')
+						OR ( tribe_event_start.meta_value >= '%1\$s' AND  tribe_event_start.meta_value <= '%2\$s')
+					)
+					",
+					tribe_event_beginning_of_day($this->first_grid_date),
+					tribe_event_end_of_day($this->final_grid_date)
```
borkweb: Need spacing inside the parentheses:

```php
tribe_event_beginning_of_day( $this->first_grid_date ),
tribe_event_end_of_day( $this->final_grid_date )
```

src/Tribe/Template/Month.php
```diff
@@ -156,22 +156,13 @@ public function __construct( $args = null ) {
...
-				'fields'         => 'ids',
-				'start_date'     => $this->first_grid_date,
-				'end_date'       => $this->final_grid_date,
-				'post_status'    => array( 'publish' ),
-				'posts_per_page' => - 1,
-			) );
 
 			if ( is_user_logged_in() ) {
 				$args['post_status'][] = 'private';
 			}
```
borkweb: Looks like you missed incorporating this into the new `set_events_in_month()` query.

src/Tribe/Template/Month.php
```diff
@@ -288,6 +279,40 @@ private static function view_more_link( $date ) {
...
+		protected function set_events_in_month() {
+			global $wpdb;
+
+			// note: this query returns the event dates as timestamps, makes date comparison faster later
+			$events_request =
+				$wpdb->prepare(
+					"SELECT tribe_event_start.post_id as ID,
+							UNIX_TIMESTAMP(tribe_event_start.meta_value) as EventStartDate,
+							UNIX_TIMESTAMP( tribe_event_end_date.meta_value) as EventEndDate
+					FROM $wpdb->postmeta AS tribe_event_start
```
borkweb: We need to make sure we only fetch posts that have a post status of `publish` (and also include `private` if the user is logged in).

src/Tribe/Date_Utils.php
```diff
@@ -459,6 +459,30 @@ public static function is_all_day( $all_day_value ) {
 			);
 		}
 
+		public static function range_coincides( $range_1_start, $range_1_end, $range_2_start, $range_2_end ) {
```
borkweb: Can you add docblock comment to this method?

src/Tribe/Template/Month.php
```diff
@@ -624,7 +665,9 @@ public function event_classes( $classes = '' ) {
 
 			// Get our wrapper classes (for event categories, organizer, venue, and defaults)
 			$classes         = array( 'hentry', 'vevent' );
+//			$GLOBALS['check_post'] = true;
```
borkweb: Remove this line?

src/functions/template-tags/date.php
```diff
@@ -301,23 +300,12 @@ function tribe_event_is_on_date( $date = null, $event = null ) {
...
-		/**
-		 * conditions:
-		 * event starts on this day (event start time is between start and end of day)
-		 * event ends on this day (event end time is between start and end of day)
-		 * event starts before start of day and ends after end of day (spans across this day)
-		 * note:
+		/* note:
 		 * events that start exactly on the EOD cutoff will count on the following day
-		 * events that end exactly on the EOD cutoff will count on the previous day
+         * events that end exactly on the EOD cutoff will count on the previous day
```
borkweb: Tabbing is off on this line.

src/Tribe/Template/Month.php
jazbek: Good call. We actually already have a class that handles object caching in TEC, I used that.

src/Tribe/Date_Utils.php
bordoni: I was looking at it, couldn't we create something that allows more than 2 ranges? just Throwing the idea out in the :wind_chime:

src/Tribe/Template/Month.php
jazbek: Cool, fixed some formatting issues throughout the file too.

src/Tribe/Template/Month.php
jazbek: Don't think it's needed..? Querying the postmeta table now..

src/Tribe/Template/Month.php
jazbek: I was trying not to do an extra join on the posts table for this Query. The wp_query that actually gets the posts to be displayed is run on each day, later, during get_daily_events(). I can set the post status there, was assuming it'd "just work" and select the correct post stati, but you're right, on ajax requests it'll get everything.

src/Tribe/Date_Utils.php
jazbek: Yes, good call.

src/Tribe/Template/Month.php
jazbek: Done.

src/Tribe/Date_Utils.php
jazbek: Good call to @borkweb I meant, I think we were commenting at the same time. :) IDK about the more than 2 ranges thing. Do we have a need for that?

src/Tribe/Date_Utils.php
```diff
@@ -459,6 +459,40 @@ public static function is_all_day( $all_day_value ) {
 			);
 		}
 
+		/**
+		 * Given 2 datetime ranges, return whether they overlap at all
+		 *
+		 * @param $range_1_start
+		 * @param $range_1_end
+		 * @param $range_2_start
+		 * @param $range_2_end
```
borkweb: How about:

```php
* @param int $range_1_start Starting timestamp of the first datetime range
* @param int $range_1_end Ending timestamp of the first datetime range
* @param int $range_2_start Starting timestamp of the second datetime range
* @param int $range_2_end Ending timestamp of the second datetime range
```

jazbek: Have already made that change and haven't pushed yet :grin:

src/Tribe/Template/Month.php
```diff
@@ -288,6 +275,56 @@ private static function view_more_link( $date ) {
...
+					)
+					",
+					$grid_start_datetime,
+					$grid_end_datetime
+				);
+			$this->events_in_month = $wpdb->get_results( $events_request );
+
+			// cache the postmeta and terms for all these posts in one go
+			$event_ids_in_month = wp_list_pluck( $this->events_in_month, 'ID' );
+			update_object_term_cache( $event_ids_in_month, TribeEvents::POSTTYPE );
```
borkweb: We'll need to switch `TribeEvents` to `Tribe__Events__Main`

src/Tribe/Template/Month.php
borkweb: Nice. I just checked out your branch and tested it and your solution here works swimmingly! :+1:

jazbek:
@borkweb you available to re-review this?

src/Tribe/Template/Month.php
```diff
@@ -289,6 +286,99 @@ private static function view_more_link( $date ) {
...
+			}
+
+			$post_stati = array( 'publish' );
+			if ( is_user_logged_in() ) {
+				$post_stati[] = 'private';
+			}
+
+			$post_stati = implode( "','", $post_stati );
+
+			// note: this query returns the event dates as timestamps, makes date comparison faster later
```
borkweb: This comment is no longer applicable

src/Tribe/Template/Month.php
```diff
@@ -289,6 +286,99 @@ private static function view_more_link( $date ) {
...
+			$events_request        =
```
borkweb: Spacing and tabbing is weird here. Maybe:

```php
$events_request = $wpdb->prepare(
  ...
);
```

borkweb:
Just to super minor things remain. This looks pretty solid, @jazbek!

PR #193, opened 2015-07-08
removing escaping as this element is pre-escaped and contains HTML

zbtirrell:
See: https://central.tri.be/issues/37508


bordoni:
:wine_glass: :dragon_face:

PR #65, opened 2015-05-19
Make events-meta-box venue and organizer templates filterable [RC]

aaemnnosttv:
This is a resubmit of #52 for possible inclusion in `release/119`, per @zbtirrell.

PR #198, opened 2015-07-09
Add events-meta-box venue template filter

borkweb:
This incorporates half of @aaemnnosttv's [PR](https://github.com/moderntribe/the-events-calendar/pull/65) that adds metabox template filters. While I have adopted the filter for the venue template, I have opted to not include the organizer one. We now support _multiple_ organizers, so making _that_ template filterable doesn't make as much sense (as originally pull requested).

There's some additional code cleanup work in this PR as well.

See: https://central.tri.be/issues/37571


bordoni:
:wink:

PR #200, opened 2015-07-09
Generate the proper event link for next/previous event pagination

borkweb:
See: https://central.tri.be/issues/37509


bordoni:
:wink:

PR #202, opened 2015-07-13
Adding missing pieces to support paging on "See All" recurring events

borkweb:
This is a follow up to https://github.com/moderntribe/events-pro/pull/55 (already code reviewed and merged).

The sister commit in Pro is: https://github.com/moderntribe/events-pro/pull/55

See: https://central.tri.be/issues/33181


src/functions/template-tags/loop.php
```diff
@@ -223,6 +223,8 @@ function tribe_has_previous_event() {
 			$args['no_found_rows']  = true;
 			$args['posts_per_page'] = 1;
 
+			$args = apply_filters( 'tribe_events_has_previous_event_args', $args );
```
zbtirrell: matt say: oops, I added these but I don't need them

src/functions/template-tags/loop.php
```diff
@@ -263,6 +265,8 @@ function tribe_has_next_event() {
 			$args['no_found_rows'] = true;
 			$args['posts_per_page'] = 1;
 
+			$args = apply_filters( 'tribe_events_has_next_event_args', $args );
```
zbtirrell: matt say: this one too

src/functions/template-tags/link.php
```diff
@@ -184,13 +184,13 @@ function tribe_get_listview_dir_link( $direction = 'next', $term = null, $curren
...
 		if ( ! $page ) {
-			$page = empty( $_GET['tribe_paged'] ) ? 1 : absint( $_GET['tribe_paged'] );
+			$page = empty( $_REQUEST['tribe_paged'] ) ? 1 : absint( $_REQUEST['tribe_paged'] );
 		}
 
 		// if what we are currently displaying is not passed in, let's set a default and check $_GET
 		if ( ! $currently_displaying ) {
 			$currently_displaying = 'list';
-			if ( ! empty( $_GET['tribe_event_display'] ) && 'past' === $_GET['tribe_event_display'] ) {
+			if ( ! empty( $_REQUEST['tribe_event_display'] ) && 'past' === $_REQUEST['tribe_event_display'] ) {
```
zbtirrell: we may be seeking VIP certification at which point, this would be better explicitly looking for GET or POST.

zbtirrell:
after those filters are removed, this look :peach::dash:

zbtirrell:
that is even better. :shipit:

PR #206, opened 2015-07-14
Add jQuery UI sortable as a dependency for events-admin.js

borkweb:
The new multiple-organizer functionality requires it.

See: https://central.tri.be/issues/37673


zbtirrell:
:+1:

PR #208, opened 2015-07-14
Invert the logic for $show_tab so it holds the value we expect

borkweb:
`$show_tab` should hold `true` if we want to show the tab rather than false.

See: https://central.tri.be/issues/37677


zbtirrell:
other than the unnecessary parens, this looks good.

PR #209, opened 2015-07-15
Simplify 'show tab' logic - was unnecessarily stupid (#37362)

barryhughes:
No idea why, but I originally was testing for one of two values to be negative instead of both conditions being positive...


borkweb:
This was already addressed here: https://github.com/moderntribe/the-events-calendar/pull/208

barryhughes:
You're a legend, I somehow missed that :-)

PR #211, opened 2015-07-17
Ensure that the start/end grid dates include the min/max times as well

borkweb:
See: https://central.tri.be/issues/37867


zbtirrell:
:+1:

PR #212, opened 2015-07-17
Datepicker bug Fixed

bordoni:
_Ref:_ #154 [C#36646](https://central.tri.be/issues/36646) [C#37856](https://central.tri.be/issues/37856)


borkweb:
:+1:

PR #213, opened 2015-07-17
Change month view query so it uses EOD cutoff

jazbek:
Change query so it uses EOD cutoff instead of midnight
Change dates on the query only because the first_grid_date is expected in Y-m-d format

Updates #37867


borkweb:
:+1:

PR #214, opened 2015-07-20
Only convert dates to YYYY-MM-DD if they aren't already in that format

borkweb:
See: https://central.tri.be/issues/37929


bordoni:
:wine_glass: :sake:

PR #219, opened 2015-07-23
[PR] Prepend space when building JOIN query to prevent conflicts.

jeremyfelt:
If another plugin has not properly appended a space in their filtering of the JOIN query, the lack of space here results in an invalid SQL query. As is elsewhere in The Events Calendar, there should be a
space before and after the query string.

Note that the same thing is in Events Calendar Pro as well.


borkweb:
Thanks for the PR! That's a good plan (and it looks good for merging). This will be in our next release!

PR #220, opened 2015-07-23
Type Check fix for saving Event Meta

dalethedeveloper:
Adding check for optional Object that is permitted to be passed as an array in Tribe__Events__Api::saveEventMeta()

### DIFF
--- a/src/Tribe/API.php
+++ b/src/Tribe/API.php
@@ -158,13 +158,13 @@ if ( ! class_exists( 'Tribe__Events__API' ) ) {
 			}
 
 			// Set sticky state for calendar view.
-			if ( isset( $data['EventShowInCalendar'] ) && $data['EventShowInCalendar'] == 'yes' && $event->menu_order != '-1' ) {
+			if ( isset( $data['EventShowInCalendar'] ) && $data['EventShowInCalendar'] == 'yes' && is_object($event) && $event->menu_order != '-1' ) {
 				$update_event = array(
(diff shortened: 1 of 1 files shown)


bordoni:
Hi @dalethedeveloper, love your contribution to our codebase, but let me do a few comments on that so that we can improve and then merge it.

1. You could move this check all the way up in the `saveEventMeta` method, and do something like;

```
if ( ! is_scalar( $event ) ){
    $event = (object) $event;
}
```

2. Keep in mind that our code follows the [WP Code Std](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage), so the space usage is important on the Pull Requests.

dalethedeveloper:
Moving this up will still cause those two conditionals to throw a PHP Notice for Undefined Property instead of "Call to a member function on a non-object". Updating the weak type checking at the core of this issue is a larger task, though.

I did not catch the spacing before sending the pull, my apologies.  I will rectify on future requests.

bordoni:
Hi @dalethedeveloper indeed, what I meant was for us to have a the type of the object _"fixed"_ right at the start and then later on we can still do the checks.

If you can just fix the spacing on this one and push the commits I will review it and we can merge it :smile:

dalethedeveloper:
Spacing fixed up, apologies for the delay!

zbtirrell:
I reopened this PR against the latest release code here: https://github.com/moderntribe/the-events-calendar/pull/342

PR #222, opened 2015-07-24
Observe the tribe_events_cat taxonomy when fetching month data

borkweb:
Additionally relocating the HTML cache so the serialized args as a cache key takes into account _ALL_ of the arguments.

See: https://central.tri.be/issues/37604


bordoni:
:wink:

PR #223, opened 2015-07-24
Only bootstrap the Admin_List class on event list pages in dashboard

borkweb:
Note: @bordoni asked if I checked if all of the actions/filters in Tribe__Events__Admin_List still fire. The answer is: **Yes, they all still fire (on the Events Admin List) with this tweak in place**.

See: https://central.tri.be/issues/38105


bordoni:
:8ball:

PR #224, opened 2015-07-24
Label all of the queries in month view as such with eventDisplay

borkweb:
Queries executed from the confines of the Month View template class should be flagged as such so that other plugins can identify and react to the queries as needed.

The foremost example of this is Pro's ability to suppress recurring events. Without labeling the query as a month view query, those event queries are instead identified as `custom`, causing the recurring events to be hidden.

Related: https://github.com/moderntribe/events-pro/pull/64

See: https://central.tri.be/issues/37629


bordoni:
:+1:

jazbek:
I think instead of this pull request we can take advantage of the hide_upcoming query arg. See:
https://github.com/moderntribe/the-events-calendar/blob/release/120/src/Tribe/Query.php#L282
https://github.com/moderntribe/the-events-calendar/blob/release/120/src/Tribe/Query.php#L334

PR #228, opened 2015-07-28
Fix/36817 venue show map logic

barryhughes:
* Show map/map link should be enabled by default for new venues
* Updating a venue via CSV import should not alter these settings

### DIFF
--- a/src/admin-views/venue-meta-box.php
+++ b/src/admin-views/venue-meta-box.php
@@ -163,7 +163,7 @@ if ( $post->post_type != Tribe__Events__Main::VENUE_POST_TYPE ) {
 		</tr>
 	<?php
 	}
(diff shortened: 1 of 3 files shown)


src/admin-views/venue-meta-box.php
```diff
@@ -163,7 +163,7 @@
 		</tr>
 	<?php
 	}
-	$google_map_link_toggle = ( get_post_status( $event->ID ) != 'auto-draft' || get_post_meta( $event->ID, '_VenueShowMapLink', true ) !== 'false' ) ? true : false;
+	$google_map_link_toggle = ( get_post_meta( $event->ID, '_VenueShowMapLink', true ) !== 'false' || get_post_status( $event->ID ) == 'auto-draft' ) ? true : false;
```
bordoni: Couldn't we pull this _logic_ into a template tag?

barryhughes: We probably should and could for this and a few other things in the same template, don't know if we should _right now_ as part of this lift, though.

bordoni: Yeah that's also true.

PR #230, opened 2015-07-29
Use list pagination link args with AJAX fetching prev/next content

borkweb:
Attempt to use the tribe_paged GET argument from the URL in the rendered link. If the argument is absent, generate a page number that is appropriate based on the current page location.

See: https://central.tri.be/issues/38179


src/resources/js/tribe-events-ajax-list.js
```diff
@@ -82,10 +82,28 @@
 
 			td.cur_url = tf.url_path( href );
 
-			if ( ! ts.paged ) {
-				ts.paged = 2;
+			reg = /tribe_paged=([^&]+)/ig;
+			result = reg.exec( href );
+
+			// use what is on the URL if possible
+			if ( 'undefined' !== result[1] ) {
```
bordoni: does it return `undefined` or you need the `typeof` here?

src/resources/js/tribe-events-ajax-list.js
```diff
@@ -111,11 +129,24 @@
...
 			}
 
+			reg = /tribe_paged=([^&]+)/ig;
+			result = reg.exec( href );
+
 			td.cur_url = tf.url_path( $( this ).attr( 'href' ) );
 
-			if ( ts.paged > 1 ) {
-				ts.paged--;
+			if ( 'undefined' !== result[1] ) {
```
bordoni: Same as the above.

borkweb: ermagerd. Yeah, it needs `typeof`

bordoni:
looking good :boom:

PR #231, opened 2015-07-29
Make sure we export ALL events when exporting from month view

borkweb:
And by "All", I mean all events that aren't hidden. Without doing `posts_per_page = -1`, the query was only retrieving events up to the "Number of events to show per page" setting.

See: https://central.tri.be/issues/38182


bordoni:
:+1:

PR #232, opened 2015-07-29
Feature/24221 import more fields

jazbek:
Adds functionality to import more fields using the csv importer:

- Event Tags
- Event Currency Symbol
- Event Currency Position

Additionally, adds 2 filters so that pro (or other customizations) can hook in to import more fields:

`tribe_events_importer_event_column_names`
`tribe_events_csv_import_event_additional_fields`

Internal ref: https://central.tri.be/issues/24221


jazbek:
Please add in-qa label and send ticket to Pending QA once code review passes. :)

PR #234, opened 2015-07-30
Spotfix for untranslated string in import general options (#38234)

barryhughes:
Minor fix, highlighted in the wp.org forums by @tititou36


bordoni:
:+1:

PR #233, opened 2015-07-30
Google Maps venue Links fixed

bordoni:
_Ref:_ [C#37660](https://central.tri.be/issues/37660)
- Deprecating `tribe_event_meta_gmap_link` filter
- Deprecating `Tribe__Events__Advanced_Functions__Register_Meta::gmap_link`

PR #237, opened 2015-08-03
Follow-up to PR #233 that fixes fatal for gmap link

borkweb:
We removed a method but there was one reference still hanging around.

See: https://github.com/moderntribe/the-events-calendar/pull/233 and https://central.tri.be/issues/37660


bordoni:
Looks good to me.

PR #238, opened 2015-08-03
Constrain datepicker Skeleton styles when in TwentFifteen theme

borkweb:
This doesn't appear to be a problem for TwentyFourteen or TwentyThirteen

See: https://central.tri.be/issues/37731


bordoni:
:dragon: :boat:

PR #240, opened 2015-08-04
Add support for a new query var called: tribe_render_context

borkweb:
Also adding a helper function to fetch render context from queries.

See: https://central.tri.be/issues/35421

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -1665,4 +1665,24 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 		$value = call_user_func( array( $defaults, $field ) );
 		return $value;
 }
+
+	/**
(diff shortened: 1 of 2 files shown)


src/functions/template-tags/general.php
```diff
@@ -1665,4 +1665,24 @@ function tribe_get_default_value( $field ) {
...
+	/**
+	 * Gets the render context of the given query
+	 *
+	 * @param WP_Query $query Query object
+	 * @return string
+	 */
+	function tribe_get_render_context( $query = null ) {
+		global $wp_query;
+
+		if ( ! $query ) {
```
bordoni: Can we also check if the variable is `instaceof WP_Query`?

bordoni:
:8ball:

PR #241, opened 2015-08-04
Only unhook addEventMeta if the function is actually executing

borkweb:
The location that this `remove_action` was placed was causing incompatibilities. When `wp_insert_post` was called during hooks like `save_post`/`transition_post_status`/etc (that execute before addEventMeta) those post insertions were triggering the `remove_action` call on `addEventMeta`. When it came time for the actual event to be inserted, the addEventMeta hook was non existent and the meta data wouldn't get inserted. This fixes that.

See: https://central.tri.be/issues/35491


bordoni:
Sounds good.

PR #245, opened 2015-08-05
Document 39 hooks for WP-Parser

zbtirrell:
See: https://central.tri.be/issues/30184


admin-views/tribe-options-display.php
```diff
@@ -10,6 +10,18 @@
 	$template_options[$templates[$template]] = $template;
 }
 
+/**
+ * Filter the array of views that are registered for the tribe bar
+ * @param array array() {
+ *     Associative array of views
+ *
```
barryhughes: Wondering, should it be made clearer that this is an array of arrays (only the inner arrays are associative)?

zbtirrell: "Array of Associative array views"?

barryhughes: "Array of views, where each view is itself represented by an associative array consisting of these keys:"

...too wordy?

borkweb:
:+1:

PR #246, opened 2015-08-05
Fix #36153 issue where _EventCost would not update when tickets updated

elimn:
This function is only called when via an ajax request when a user edits a ticket while editing an event. A separate function updates these fields when the page is POSTed (The user clicks the Update button). That function worked flawlessly, but this called via ajax did not.

I am not sure what the code was supposed to do before, why there was an array shift nor why we had a $base_cost variable. However, the filter 'tribe_events_event_costs' is only applied in one other place in all of our plugins, and in that place it passes an empty array exactly like this now does. Now that there are no values passed, it behaves as expected and never has a duplicate value.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -312,6 +312,11 @@ At no point during the 3.0 lifecycle will the major version change. But you can 
 
 == Changelog ==
 
+
+= [Unreleased] unreleased =
+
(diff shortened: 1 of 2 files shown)


bordoni:
@elimn don't forget to label it as _"Code Review"_

bordoni:
@elimn don't forget to delete the Branch after you Merge it. :wink:

PR #247, opened 2015-08-06
WPML Compatibility

bordoni:
### Working in Progress

_Ref:_ [Q#16797](https://central.tri.be/issues/16797)

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -2506,6 +2506,20 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		}
 
 		/**
+		 * Deprecated alias to `filter_rewrite_rules`
+		 *
+		 * @param object $wp_rewrite
+		 *
+		 * @deprecated
+		 * @todo remove 4.2
+		 *
+		 * @return void
+		 */
+		public function filterRewriteRules( $wp_rewrite ) {
+			$this->filter_rewrite_rules( $wp_rewrite );
+		}
+
+		/**
 		 * Redirect the legacy past/upcoming view URLs to list
 		 */
 		public function redirect_past_upcoming_view_urls() {
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -109,13 +109,13 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		public $rewriteSlugSingular = 'event';
 		public $taxRewriteSlug = 'event/category';
 		public $tagRewriteSlug = 'event/tag';
-		protected $monthSlug = 'month';
+		public $monthSlug = 'month';
 
 		// @todo remove in 4.0
-		protected $upcomingSlug = 'upcoming';
-		protected $pastSlug = 'past';
+		public $upcomingSlug = 'upcoming';
+		public $pastSlug = 'past';
 
-		protected $listSlug = 'list';
+		public $listSlug = 'list';
 		public $daySlug = 'day';
 		public $todaySlug = 'today';
 		protected $postExceptionThrown = false;
@@ -187,14 +187,14 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		public $plural_event_label;
 
 		/** @var Tribe__Events__Default_Values */
-		private $default_values = NULL;
+		private $default_values = null;
 
 		public static $tribeEventsMuDefaults;
 
 		/**
 		 * Static Singleton Factory Method
 		 *
-		 *@return Tribe__Events__Main
+		 * @return Tribe__Events__Main
 		 */
 		public static function instance() {
 			if ( ! isset( self::$instance ) ) {
@@ -337,6 +337,9 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		 * Add filters and actions
 		 */
 		protected function addHooks() {
+			// Load Rewrite
+			add_action( 'plugins_loaded', array( Tribe__Events__Rewrite::instance(), 'hooks' ) );
+
 			add_action( 'init', array( $this, 'init' ), 10 );
 
 			// Frontend Javascript
@@ -358,7 +361,6 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 			add_filter( 'nav_menu_items_' . self::POSTTYPE, array( $this, 'add_events_checkbox_to_menu' ), null, 3 );
 			add_filter( 'wp_nav_menu_objects', array( $this, 'add_current_menu_item_class_to_events' ), null, 2 );
 
-			add_filter( 'generate_rewrite_rules', array( $this, 'filter_rewrite_rules' ) );
 			add_filter( 'template_redirect', array( $this, 'redirect_past_upcoming_view_urls' ), 11 );
 
 			/* Setup Tribe Events Bar */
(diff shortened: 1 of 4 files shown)


src/Tribe/Main.php
```diff
@@ -358,7 +358,7 @@ protected function addHooks() {
 			add_filter( 'nav_menu_items_' . self::POSTTYPE, array( $this, 'add_events_checkbox_to_menu' ), null, 3 );
 			add_filter( 'wp_nav_menu_objects', array( $this, 'add_current_menu_item_class_to_events' ), null, 2 );
 
-			add_filter( 'generate_rewrite_rules', array( $this, 'filterRewriteRules' ) );
+			add_filter( 'generate_rewrite_rules', array( $this, 'filter_rewrite_rules' ) );
```
borkweb: If you are going to change the method name, we probably should create a deprecated (scheduled for removal in 4.2) method with the same name (`filterRewriteRules`) that just calls `filter_rewrite_rules`. This filter should then have a comment above it stating that it should be swapped out for the new function in 4.2.

src/Tribe/Main.php
```diff
@@ -2272,6 +2272,157 @@ public function eventQueryVars( $qvars ) {
...
+		 * @return array                    A multi level array with the possible translations for the given strings
+		 */
+		public function get_i18n_strings( $strings, $languages, $domains = array( 'tribe-events-calendar' ), $default_language = 'en_US' ) {
+			foreach ( $languages as $language ) {
+				// Prevent re-loading the default language
+				if ( $language === $default_language ){
+					continue;
+				}
+
+				foreach ( (array) $domains as $domain => $file ) {
```
borkweb: I'm curious about `$domain => $file` and how it relates to the default `$domains` variable in the method declaration. The default declaration just has a value and is lacking an index of note that would relate to `$domain`.

src/Tribe/Main.php
```diff
@@ -2272,6 +2272,157 @@ public function eventQueryVars( $qvars ) {
...
+							continue;
+						}
+
+						// Make sure we have an Array
+						$strings[ $key ] = (array) $strings[ $key ];
+
+						// Grab the possible strings for Default and Any other domain
+						if ( 'default' === $domain ) {
+							$strings[ $key ][] = __( $value );
+							$strings[ $key ][] = __( strtolower( $value ) );
```
borkweb: Don't you [need string literals](http://codex.wordpress.org/Function_Reference/_2) for `__()`?

bordoni: On this case the Literals will be _"defined"_ elsewhere. I'm not sure where I will do it.

src/Tribe/Main.php
bordoni: We can but I don't think anyone using it. I will do it.

src/Tribe/Rewrite.php
```diff
@@ -0,0 +1,311 @@
...
+		 */
+		public $bases = array();
+
+		private function __construct() {}
+
+		public function hooks() {
+			add_filter( 'generate_rewrite_rules', array( $this, '_filter_generate' ) );
+		}
+
+		public function _filter_generate( WP_Rewrite $wp_rewrite ) {
```
barryhughes: Are you using the leading underscore as an alternative to hard enforcement through the `protected` or `private` modifiers? I actually quite like that but isn't it going against the current coding guidelines for the project?

src/Tribe/Rewrite.php
```diff
@@ -0,0 +1,311 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	die( '-1' );
+}
+
+if ( ! class_exists( 'Tribe__Events__Rewrite' ) ) {
+
+	/**
+	 * Permalinks magic Happens over here!
+	 */
+	class Tribe__Events__Rewrite {
```
barryhughes: It's great to see this being broken out of `Tribe__Events__Main` :+1:

src/Tribe/Rewrite.php
```diff
@@ -0,0 +1,311 @@
...
+				->tag( array( '(\d{4}-\d{2})' ), array( 'eventDisplay' => 'month', 'eventDate' => '%2' ) )
+				->tag( array( '(\d{4}-\d{2}-\d{2})' ), array( 'eventDisplay' => 'day', 'eventDate' => '%2' ) )
+				->tag( array( 'feed' ), array( 'eventDisplay' => 'list', 'feed' => 'rss2' ) )
+				->tag( array( 'ical' ), array( 'ical' => 1 ) )
+				->tag( array( 'feed', '(feed|rdf|rss|rss2|atom)' ), array( 'feed' => '%2' ) )
+				->tag( array(), array( 'eventDisplay' => $options['default_view'] ) );
+
+			do_action( 'tribe_events_pre_rewrite', $this );
+
+			$wp_rewrite->rules = apply_filters( 'tribe_events_rewrite_rules', $this->rules + $wp_rewrite->rules, $this );
```
barryhughes: Add a docblock for this hook?

src/Tribe/Rewrite.php
```diff
@@ -0,0 +1,311 @@
...
+				$bases = Tribe__Events__Main::instance()->get_i18n_strings( $bases, $languages, $domains, $current_locale );
+			}
+
+			if ( 'regex' === $method ){
+				foreach ( $bases as $type => $base ) {
+					$bases[ $type ] = '(?:' . implode( '|', $base ) . ')';
+				}
+			}
+
+			return (object) apply_filters( 'tribe_events_rewrite_i18n_slugs', $bases, $method );
```
barryhughes: Docblock?

src/Tribe/Rewrite.php
```diff
@@ -0,0 +1,311 @@
...
+				->tag( array( '{{ today }}' ), array( 'eventDisplay' => 'day' ) )
+				->tag( array( '{{ day }}', '(\d{4}-\d{2}-\d{2})' ), array( 'eventDisplay' => 'day', 'eventDate' => '%2' ) )
+				->tag( array( '(\d{4}-\d{2})' ), array( 'eventDisplay' => 'month', 'eventDate' => '%2' ) )
+				->tag( array( '(\d{4}-\d{2}-\d{2})' ), array( 'eventDisplay' => 'day', 'eventDate' => '%2' ) )
+				->tag( array( 'feed' ), array( 'eventDisplay' => 'list', 'feed' => 'rss2' ) )
+				->tag( array( 'ical' ), array( 'ical' => 1 ) )
+				->tag( array( 'feed', '(feed|rdf|rss|rss2|atom)' ), array( 'feed' => '%2' ) )
+				->tag( array(), array( 'eventDisplay' => $options['default_view'] ) );
+
+			do_action( 'tribe_events_pre_rewrite', $this );
```
barryhughes: Docblock? ;-)

src/Tribe/Rewrite.php
borkweb: Yeah, let's not prefix functions with `_` and instead use `protected` and `private` where necessary.

src/Tribe/Main.php
```diff
@@ -337,6 +337,9 @@ public function after_html_data_wrapper( $html ) {
 		 * Add filters and actions
 		 */
 		protected function addHooks() {
```
afragen: Shouldn't this be snake_case not camelCase?

bordoni: Agreed, we could change it, but we might do some changes around the _"camelCase"_ on the next release, because of "breaking" changes. On this case it shouldn't break because of `private` but...

PR #248, opened 2015-08-06
adding numbered placeholders to translated strings

zbtirrell:
This is a big help for translators to be able to get the grammar correct
when translating.

See: https://central.tri.be/issues/37552


bordoni:
Seems to be good.

PR #249, opened 2015-08-06
Added check for get_current_screen() before trying to use it

Enchiridion:
Hi,
This is a fix for an issue that I (and several others) have been having where TEC is trying to use `get_current_screen()`, but for some reason it's unavailable, which is breaking the entire admin side when trying to add a new event.

The bug can be recreated by creating a WordPress plugin that use ```WP_Query``` when ```wp_loaded``` is fired.  The ```pre_get_posts``` hook in TEC then fires and tries to use ```get_current_screen()```, which is undefined at this point.  Which is weird because WP_ADMIN *is* defined, so that function should exist.  Oh well.

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -299,7 +299,11 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 				// Because this method filters out drafts without EventStartDate.
 				// For this screen we're doing the JOIN manually in Tribe__Events__Admin_List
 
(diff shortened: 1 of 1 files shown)


borkweb:
Thank you for the Pull Request! Sure enough, `get_current_screen()` isn't available until the `_admin_menu` action. We'll get this merged in and the fix will be in our next release!

_Internal reference ticket: [#38452](https://central.tri.be/issues/38452)_

PR #250, opened 2015-08-06
When parsing dates w/ single digit months/days, convert them to 2-digits

borkweb:
When a date format contains a single-digit month or single-digit day, the `Tribe__Events__Date_Utils::datetime_from_format()` returns those months/days in the datetime as single digit rather than the desired `Y-m-d` format.

Examples:

Format: `j/n/Y`
Input: `15/1/2015`
Returns: `2015-1-15`
Should be: `2015-01-15`

Format: `j/n/Y`
Input: `1/1/2015`
Returns: `2015-1-1`
Should be: `2015-01-01`

See: https://central.tri.be/issues/38451

### DIFF
--- a/src/Tribe/Date_Utils.php
+++ b/src/Tribe/Date_Utils.php
@@ -150,6 +150,9 @@ if ( ! class_exists( 'Tribe__Events__Date_Utils' ) ) {
 				return false;
 			}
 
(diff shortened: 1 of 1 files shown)


bordoni:
:8ball:

PR #252, opened 2015-08-06
Avoid a notice when _GET['plugins'] isn't set on the update page

borkweb:
See: https://central.tri.be/issues/38061


afragen:
Thanks, I was seeing this too.

bordoni:
:8ball:

PR #254, opened 2015-08-09
swapped the_title() for the_title_attribute()

elimn:
Swapped the_title() for the_title_attribute() in list and day view <a> links.


bordoni:
Looks good.

PR #256, opened 2015-08-10
Standalone Tickets Framework

MZAWeb:
This branch removes the Ticket framework from the core plugin, and adds the new standalone Tickets plugin as a submodule in the vendor folder. It also adds a thin loader to load the framework if the standalone plugin is not installed.

Related:

* https://github.com/moderntribe/events-pro/pull/157
* https://github.com/moderntribe/events-community/pull/65
* https://github.com/moderntribe/events-eventbrite/pull/83
* https://github.com/moderntribe/events-facebook/pull/35
* https://github.com/moderntribe/events-filterbar/pull/38
* https://github.com/moderntribe/events-importer-ical/pull/38

See: [C#38979](https://central.tri.be/issues/38979) and [C#38488](https://central.tri.be/issues/38488)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$update' | [Timezone_Updater.php:60](https://github.com/moderntribe/the-events-calendar/blob/6afd5b7e72f50169802ba262cd86f049312463d2/src/Tribe/Admin/Timezone_Updater.php#L60)

src/Tribe/Main.php
```diff
@@ -4542,15 +4490,8 @@ public static function array_insert_after_key( $key, $source_array, $insert_arra
...
-				$position     = array_search( $key, array_keys( $source_array ) );
-				$source_array = array_slice( $source_array, 0, $position, true ) + $insert_array + array_slice( $source_array, $position, null, true );
-			} else {
-				// If no key is found, then add it to the end of the array.
-				$source_array += $insert_array;
-			}
-
-			return $source_array;
+			_deprecated_function( __METHOD__, '4.0', 'Tribe__Main::array_insert_before_key' );
+			return $this->common()->array_insert_before_key( $key, $source_array, $insert_array );
```
barryhughes: `$this` doesn't make sense here - we're inside an instance method at this point.

barryhughes: (Updated^! I meant, we're in a static method so can't use `$this`)

src/Tribe/Main.php
```diff
@@ -4610,5 +4557,225 @@ public function set_meta_factory_global() {
...
+		}
+
+		public function tribe_help_sidebar_top() {
+			$tec_info = wp_remote_get(
+			/**
+			 * Filter the tribe info API url
+			 *
+			 * @param string $url
+			 */
+			apply_filters( 'tribe_help_tab_api_info_url', Tribe__Events__Main::INFO_API_URL ), array(
```
barryhughes: Feels like this line (and the above docblock) should be tabbed in a level)

src/Tribe/Main.php
```diff
@@ -4610,5 +4557,225 @@ public function set_meta_factory_global() {
...
+					'timeout' => 15, //seconds
+					'headers' => array( 'Accept' => 'application/json' ),
+				)
+			);
+			if ( ! is_wp_error( $tec_info ) ) {
```
barryhughes: Should we also handle a bad response (we go on to render `$version` for example, but it's not guaranteed it will be set).

src/Tribe/Main.php
```diff
@@ -4610,5 +4557,225 @@ public function set_meta_factory_global() {
...
+				if ( isset( $tec_info['version'] ) ) {
+					$version = $tec_info['version'];
+				}
+				$total_downloads = ( isset( $tec_info['total_downloads'] ) ) ? number_format( $tec_info['total_downloads'] ) : _x( 'n/a', 'not available', 'tribe-common' );
+				$up_to_date      = ( isset( $tec_info['version'] ) && version_compare( Tribe__Events__Main::VERSION, $tec_info['version'], '<' ) ) ? __( 'You need to upgrade!', 'tribe-common' ) : __( 'You are up to date!', 'tribe-common' );
+			}
+
+			?>
+			<div id="tribe-help-plugin-info">
+				<h3><?php esc_html_e( 'The Events Calendar', 'tribe-common' ); ?></h3>
```
barryhughes: Here and a few other places we're using the `tribe-common` textdomain - is that the right choice here, where we're not inside the common libs module?

src/Tribe/Main.php
```diff
@@ -4610,5 +4557,225 @@ public function set_meta_factory_global() {
...
+				<?php
+			}
+		}
+
+		/**
+		 * Adds post types to the post_types array used to determine if on a post type screen
+		 *
+		 * @param array $post_types Collection of post types
+		 */
+		public function is_post_type_screen_post_types( $post_types ) {
```
barryhughes: I realize the method name is just echoing the name of the matching hook - but might something like `add_event_post_types` be clearer?

src/Tribe/Main.php
```diff
@@ -4610,5 +4557,225 @@ public function set_meta_factory_global() {
...
+			foreach ( self::getPostTypes() as $post_type ) {
```
barryhughes: Might it be clearer to collapse this into an array_merge(), if we can?

src/Tribe/Template_Factory.php
```diff
@@ -10,639 +10,520 @@
...
+	 *
+	 * @static
+	 * @var array
+	 */
+	protected $comments_off = false;
+
+	/**
+	 * Run include packages, set up hooks
+	 *
+	 * @return void
```
barryhughes: Don't really need a @return annotation for the constructor - but if we do it probably shouldn't be void

src/Tribe/Template_Factory.php
```diff
@@ -10,639 +10,520 @@
...
-			$asset->set_prefix( $prefix );
-			$asset->set_tec( $tec );
+	/**
+	 * Handles an asset package request.
+	 *
+	 * @param string              $name          The asset name in the `hyphen-separated-format`
+	 * @param array               $deps          An array of dependency handles
+	 * @param string              $vendor_url    URL to vendor scripts and styles dir
+	 * @param string              $prefix        MT script and style prefix
+	 * @param Tribe__Events__Main $tec           An instance of the main plugin class
```
barryhughes: Should the type be `Tribe__Main` now?

src/Tribe/Template_Factory.php
```diff
@@ -10,639 +10,520 @@
...
-		 * @return string[] An array of registered vendor script handles.
-		 */
-		public static function get_vendor_scripts() {
-			return self::$vendor_scripts;
+	/**
+	 * Filter the body class
+	 *
+	 * @param array $classes
+	 *
+	 * @return void
```
barryhughes: Not void, we return an array from this method

src/Tribe/Template_Factory.php
```diff
@@ -10,639 +10,520 @@
...
+	 */
+	public function excerpt_more( $more ) {
+		return $this->excerpt_more;
+	}
 
-			self::handle_asset_package_request( $name, $deps, $vendor_url, $prefix, $tec );
+	/**
+	 * Check if comments are disabled on this view
+	 *
+	 * @param int $more
```
barryhughes: Docblock is out of sync with the actual method signature

barryhughes:
A lot of the above is probably just picking up on old code that's been relocated.

src/Tribe/Main.php
```diff
@@ -2010,9 +2002,19 @@ public function addNetworkOptionsPage() {
 		 * @return void
 		 */
 		public function doNetworkSettingTab() {
-			include_once( $this->pluginPath . 'src/admin-views/tribe-options-network.php' );
+			_deprecated_function( __METHOD__, '4.0', 'Tribe__Settings_Manager::do_network_settings_tab' );
+			Tribe__Settings_Manager::do_network_settings_tab();
+		}
+
+		/**
+		 * Filters the post types across all of the Tribe plugins
```
bordoni: The docblock missing `@return` and `@param`

src/Tribe/Main.php
```diff
@@ -216,12 +236,17 @@ public static function instance() {
 		 * Initializes plugin variables and sets up WordPress hooks/actions.
 		 */
 		protected function __construct() {
-			$this->pluginPath = trailingslashit( dirname( dirname( dirname( __FILE__ ) ) ) );
-			$this->pluginDir  = trailingslashit( basename( $this->pluginPath ) );
-			$this->pluginUrl  = plugins_url( $this->pluginDir );
+			$this->pluginPath = $this->plugin_path = trailingslashit( dirname( dirname( dirname( __FILE__ ) ) ) );
```
bordoni: Instead of setting the `$this-> pluginPath` + `$this->pluginDir` + `$this-> pluginUrl` here, we could have a Magic Method `__get` set for this class to notify users that are using this class variable. What do you think @borkweb?

MZAWeb: Not sure what you mean by "to notify users", but why would you want to have a magic `__get` instead of explicit vars? Explicit is almost always a better choice, IMHO.

bordoni: It's because we are going to deprecate those right?

src/Tribe/Main.php
```diff
@@ -26,17 +34,10 @@ class Tribe__Events__Main {
 
 		const VERSION           = '3.12.3';
 		const MIN_ADDON_VERSION = '3.12';
-		const FEED_URL          = 'https://theeventscalendar.com/feed/';
```
barryhughes: Sure we don't need this? We still refer to it on [line 3901](https://github.com/moderntribe/the-events-calendar/blob/feature/standalone-tickets-framework/src/Tribe/Main.php#L3901) (the `outputDashboardWidget()` method) - that will cause a fatal so we should change that method or else add it back as appropriate.

barryhughes:
GH doesn't seem able to let me add line comments for these (due to the volume of changes in this PR) but we've deprecated a few functions including `tribe_events_resource_url()` - but are still calling them from our own codebase (`Tribe__Events__Asset__Admin_Menu`).

src/Tribe/Main.php
MZAWeb: Good catch, fixed.

MZAWeb:
All comments are addressed, and I fixed a few other assorted issues and phpDocs.

barryhughes:
:8ball:

PR #259, opened 2015-08-12
Performance enhancements that target month view

borkweb:
Note: I'm Pull Requesting this into a feature branch (that is [this PR](https://github.com/moderntribe/the-events-calendar/pull/192)) so the code within this changeset can be reviewed in isolation.

**The Cost_Utils class**

`get_cost_by_func` was pretty slow for two reasons: `array_merge` is kind of slow _and_ when there are a large number of events, fetching every non-unique cost means we have to loop over and parse each one.  That's expensive. Let's select distinct costs and avoid doing `array_merge`.

**Month view class**

When fetching the daily events, we were looping over every event in the month and determining if it fell on the given day. Instead, we could reduce the number of times we loop by _instead_ looping over all the events in the month and breaking them out into the days that they fall on. We get the same data, it is just built in reverse and requires fewer loops.

See: https://github.com/moderntribe/the-events-calendar/pull/192 and https://central.tri.be/issues/32067

### DIFF
--- a/src/Tribe/Template/Month.php
+++ b/src/Tribe/Template/Month.php
@@ -571,7 +571,6 @@ if ( ! class_exists( 'Tribe__Events__Template__Month' ) ) {
 
 			$date = $this->first_grid_date; // Start with the first grid date
 
-
 			// Populate complete date range including leading/trailing days from adjacent months
 			while ( $date <= $this->final_grid_date ) {
 
--- a/src/Tribe/Template/Month.php
+++ b/src/Tribe/Template/Month.php
@@ -36,6 +36,12 @@ if ( ! class_exists( 'Tribe__Events__Template__Month' ) ) {
 		private $events_per_day;
 
 		/**
+		 * Grid day events
+		 * @var array
+		 */
+		private $event_ids_by_day;
+
+		/**
 		 * Array of days of the month
 		 * @var array
 		 */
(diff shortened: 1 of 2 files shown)


src/Tribe/Template/Month.php
```diff
@@ -421,42 +426,105 @@ protected function set_events_in_month() {
...
+						$diff_in_days = $diff / DAY_IN_SECONDS;
+
+						// add the event to each day until the event end
+						$new_start = $start;
+						for ( $i = 0; $i <= $diff_in_days; $i++ ) {
+							if ( ! isset( $this->event_ids_by_day[ $new_start ] ) ) {
+								$this->event_ids_by_day[ $new_start ] = array();
+							}
+
+							$this->event_ids_by_day[ $new_start ][] = $event->ID;
```
barryhughes: Hey Matt: got a slight concern about this re EOD cutoffs.

Looking back to line 471, `$start` and `$end` are just dates (rather than complete datetimes) - so, if execution has entered the block beginning line 473, `$diff_in_days` will be at least `1` and this `for` loop is going to run for at least 2 iterations.

Given an event with a late finish (ie, the next day but _within_ the cutoff period) the ID will be added incorrectly to that final day - throwing the count of events on that day (ie, we might see *"View all X events"* where X is one unit higher than it should be).

borkweb: If there's a difference of 1 day, we'd want it to run for two iterations. The first iteration would be day 1 and the second iteration would be the end date (the diff).

borkweb: Ohhhh...I had to read through your message a few times. I get what you're saying!

barryhughes: Yar, sorry - tricky one to summarize :-/

barryhughes:
:+1:

Nice job on the commentary, too - will definitely be useful in the future :-)

src/Tribe/Template/Month.php
```diff
@@ -421,42 +426,159 @@ protected function set_events_in_month() {
...
+					// if we're querying by category and the event doesn't have it, skip the event
 					if ( ! empty ( $this->queried_event_cats ) ) {
 						if ( ! has_term( $this->queried_event_cats, Tribe__Events__Main::TAXONOMY, $event ) ) {
 							continue;
 						}
 					}
-					$event_ids_on_date[] = $event->ID;
+
+					$event_start = strtotime( $event->EventStartDate );
+					$event_end   = strtotime( $event->EventEndDate );
```
barryhughes: Can we switch to using `tribe_get_start_date()` and `tribe_get_end_date()` when setting these two vars? The benefit is they automatically account for the current timezone setting which is important if the current timezone mode is "sitewide" and some of the events are in a different timezone.

borkweb: Ooo, good catch! Done.

PR #260, opened 2015-08-12
Alter query to respect events that are set to be 'sticky in month view' (#32067)

barryhughes:
We need to respect events that have been set as 'sticky in month view' (ie, they have a negative menu_order to float them to the top). Like [PR/259](https://github.com/moderntribe/the-events-calendar/pull/259) this targets an existing feature branch.


bordoni:
Looks good.

PR #263, opened 2015-08-18
Don't override ORDER BY with a blank string. That's silly.

borkweb:
This filter was added during the development cycle of release/120 [by me](https://github.com/moderntribe/the-events-calendar/pull/148) and it was really stupid.

See: https://central.tri.be/issues/38329


bordoni:
:8ball:

PR #264, opened 2015-08-19
Flip order of coordinates per RFC2445 4.8.1.6 (spotfix)

barryhughes:
As highlighted by a customer (props owed to [@Jon](http://theeventscalendar.com/support/forums/users/dclodging/) here) we were flipping coordinates in the ical feed.


bordoni:
:8ball:

PR #267, opened 2015-08-20
Add $postId parameter to the tribe_get_gcal_link filter

MZAWeb:


src/functions/template-tags/ical.php
```diff
@@ -44,7 +44,7 @@ function tribe_get_gcal_link( $postId = null ) {
 		$postId    = Tribe__Events__Main::postIdHelper( $postId );
 		$output    = Tribe__Events__Main::instance()->googleCalendarLink( $postId );
 
-		return apply_filters( 'tribe_get_gcal_link', $output );
+		return apply_filters( 'tribe_get_gcal_link', $output, $postId );
```
barryhughes: Needn't block merging, but if there was time to add a docblock in here that would be :+1:

PR #272, opened 2015-08-24
Transform _n() call into _nx() (spotfix for translation issue)

barryhughes:
Minor formatting changes, but the key difference is addition of context to the translated string (should make it easier for GP to differentiate between these strings and others set elsewhere in the codebase).


zbtirrell:
:+1:

PR #273, opened 2015-08-25
Deprecate JS event names and replace with the new standard

borkweb:
Switching to the new [event naming standard](https://github.com/moderntribe/the-events-calendar/wiki/JS-event-namespacing). The old event names have been left in place for backward compatibility.

Related:
* https://github.com/moderntribe/events-pro/pull/87

See: https://central.tri.be/issues/36301


bordoni:
It looks good, but it would be nice a developer did the smoketest on this one.

PR #275, opened 2015-08-25
Rework "noindex" logic so it doesn't depend on a 404 status being set

barryhughes:
A `noindex` should now be set for a) month view, always b) other views, when empty, regardless of returned HTTP status. An avenue to modify this behaviour has also been provided through a new hook.


bordoni:
:8ball:

PR #277, opened 2015-08-25
Month view day classes should work regardless of WP timezone (#38398)

barryhughes:
Please see [#38398](https://central.tri.be/issues/38398) for reference - should avoid problems where the incorrect day is highlighted in month view due to timezone issues.


bordoni:
Looks good.

PR #278, opened 2015-08-26
Feature/ical custom event links

lucatume:
This proposed change allows for the generation of iCal links for a custom combination of events; the IDs of those events to be passed in the `$_GET` parameter under the proposed `event_ids` key.
Use cases: in custom templates where events are fetched with custom criteria like “all events from this organiser and this organiser” or “all upcoming events from this organiser in this category”.

### DIFF
--- a/src/Tribe/iCal.php
+++ b/src/Tribe/iCal.php
@@ -134,7 +134,14 @@ class Tribe__Events__iCal {
 		// hijack to iCal template
 		if ( get_query_var( 'ical' ) || isset( $_GET['ical'] ) ) {
 			global $wp_query;
-			if ( is_single() ) {
+			if ( isset( $_GET['event_ids'] ) ) {
+				if ( empty( $_GET['event_ids'] ) ) {
+					die();
+				}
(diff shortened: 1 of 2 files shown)


src/Tribe/iCal.php
```diff
@@ -203,8 +210,7 @@ public static function generate_ical_feed( $post = null ) {
 		$blogName    = get_bloginfo( 'name' );
 
 		if ( $post ) {
-			$events_posts   = array();
-			$events_posts[] = $post;
+			$events_posts = is_array( $post ) ? $post : array( $post );
```
bordoni: Just cast as `(array)` wouldn't work here??

```
$events_posts = (array) $post
```

zbtirrell: wouldn't the cast convert the post object to an array rather than creating an array with the post as the first element?  seems like different behavior

bordoni: yeah @zbtirrell  is right. My bad.

bordoni:
:8ball:

PR #279, opened 2015-08-26
Modify workaround for handling WP imports of recurring events (#33497)

barryhughes:
When importing posts, WP avoids duplicates by comparing post name, date and type (which are also identical for recurring events). We already had special handling to avoid those recurring events being rejected as dupes, this extends it to cover some other cases (ie, where an event is broken out of the chain).


bordoni:
That's a nice set of comments! Looks good.

PR #280, opened 2015-08-27
Apply filters to create the Language when _GET['lang'] is set

bordoni:
_Ref:_ [C#16797](https://central.tri.be/issues/16797)


barryhughes:
Looks good :+1:

PR #281, opened 2015-08-31
Prevent halting JS when the regex fails miserably

borkweb:
See: https://central.tri.be/issues/39072


src/resources/js/tribe-events-ajax-list.js
```diff
@@ -74,7 +74,7 @@
 
 			if ( $( this ).parent().is( '.tribe-events-past' ) ) {
 				ts.view = 'past';
-			} else if ( 'undefined' !== result[1] ) {
+			} else if ( result && 'undefined' !== result[1] ) {
```
bordoni: Shouldn't we also check for the `typeof` or `result.length`?

borkweb: Yeah, that line was missing the `typeof`, it seems. That's been fixed now.

bordoni:
:8ball:

PR #284, opened 2015-08-31
Remove the 'updated' class, 'tribe-updated' now rules!

bordoni:
_Ref:_ [C#26499](https://central.tri.be/issues/26499)


borkweb:
Try not to commit the minified .js files if you aren't changing any JS.

PR #286, opened 2015-08-31
Feature/33853 live date preview

barryhughes:
Adds live preview support for date formats in *Events > Settings > Display* similiar to (and using the same ajax action as) WP in the *Settings > General* screen.


src/resources/js/events-admin-date-preview.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Provides live preview facilities for the event date format fields, akin
+ * to (and using the same ajax mechanism as) the date format preview in WP's
+ * general settings screen.
+ */
+jQuery( document ).ready( function( $ ) {
+	// Whenever the input field for a date format changes, update the matching
+	// live preview area
+	$( ".live-date-preview" ).siblings( "input" ).change( function() {
+		var format_field  = $( this );
```
borkweb: A common convention for jQuery (and it is super handy!) is to prefix variables that hold jQuery objects with a `$`. It makes it readily obvious that you can perform jQuery operations on the var without sweating things. It isn't in our JS standards, but if others are OK with it, I'd love to make it one for Products.

As such, I'd suggest the following:

```js
var $format_field = $( this );
var new_format = $format_field.val();
var $preview_field = $format_field.siblings( '.live-date-preview' );
```

borkweb:
:+1: (Side note: ideally the only minified JS files to be committed would be the ones whose un-minified versions were changed)

PR #287, opened 2015-09-01
Remove the 'updated' class, 'tribe-updated' now rules!

bordoni:
_Ref:_ [C#26499](https://central.tri.be/issues/26499)
_Rel:_ #284


borkweb:
This escaped me! I'll :+1: it, though I think it is best (in the future) to avoid committing minified CSS files that weren't _actually_ modififed.

PR #292, opened 2015-09-02
Adjust default for $default_to_original arg to true (spotfix)

barryhughes:
`Tribe__Events__Template_Factory::getMinFile()` will fallback to the unminified asset whenever `$default_to_original` is true. All calls to this method I can find across our suite accordingly set that to true (as the default, bizarrely, is false). This PR changes that default to help avoid annoying bugs when laying out new code.


zbtirrell:
Missing ticket reference
> via jenkins/no-ticket-ref

bordoni:
Looks good. Are you sure we won't have any missing `.min` files?

barryhughes:
If the minified asset is missing it loads the original instead, so it's safe even if we are missing some .min files :-)

bordoni:
So we are all good! :snowman:

PR #294, opened 2015-09-03
Move some of the CSV import fields around so similar ones are neighbors

borkweb:
See: https://central.tri.be/issues/24221#note-52


bordoni:
:8ball:

PR #298, opened 2015-09-04
Fix/39180 list view queries pagination

barryhughes:
Fixes two unrelated problems that were both impacting list pagination (and list titles). A) the query wasn't consistently referencing the correct date fields (according to the timezone mode) and B) the list title didn't always use the *selected* date when appropriate.


borkweb:
:+1:

PR #299, opened 2015-09-04
API cannot rely on a timezone being posted in when setting up event datetime data | 39235

barryhughes:
When `Tribe__Events__API::saveEventMeta()` is called to create/update an event, timezone data should be created and default to the site settings if not explicitly provided.


borkweb:
:cactus: :dash:

PR #300, opened 2015-09-04
Don't null out date_name

borkweb:
See: https://central.tri.be/issues/39230


zbtirrell:
there appears to be some back-compat risk implied by Gustavo's commentary here: https://github.com/moderntribe/the-events-calendar/pull/111

If you are comfortable, I am too.

PR #301, opened 2015-09-04
Improve handling of errors where timezone conversion cannot be made (#39204)

barryhughes:
We're making use of procedural datetime functions for brevity: the try/catch isn't appropriate here and means we're potentially missing some errors during conversion that we could otherwise handle fairly gracefully.

See: https://central.tri.be/issues/39204


bordoni:
Looks good. but missing the Central Ref.

PR #309, opened 2015-09-10
Look up one more level to find the readme.txt file (#39386)

barryhughes:
Re [C#39386](https://central.tri.be/issues/39386) the changelog reader was failing to locate the default `readme.txt` file. This can be reviewed but I've applied the hold label since we may wait until MR2 before merging.

### DIFF
--- a/src/Tribe/Changelog_Reader.php
+++ b/src/Tribe/Changelog_Reader.php
@@ -11,7 +11,7 @@ class Tribe__Events__Changelog_Reader {
 	}
 
 	protected function default_readme_file() {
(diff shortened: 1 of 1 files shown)


zbtirrell:
This looks good.  Is this an issue in our other plugins, or unique to TEC?

barryhughes:
The problem with the default changelog path is specific to core, other plugins form their own path and pass it in - but I could check those as a follow-up action.

barryhughes:
(Merging in time for 3.12.1.)

PR #310, opened 2015-09-11
Use site timezone (not UTC) when selecting current day in month view

barryhughes:
Correct code for highlighting the current day (see [C#39420](https://central.tri.be/issues/39420)).


bordoni:
:performing_arts:

PR #312, opened 2015-09-15
updating tests to lose some garbage

zbtirrell:
removing some files that don't need to be in here, updating the .gitignore. This will help make for a clean environment for running automated tests


borkweb:
:+1:

PR #313, opened 2015-09-15
updated code to generate single event iCal link

lucatume:
addresses https://central.tri.be/issues/38756

### DIFF
--- a/src/functions/template-tags/ical.php
+++ b/src/functions/template-tags/ical.php
@@ -10,7 +10,10 @@ if ( ! function_exists( 'tribe_get_single_ical_link' ) ) {
 	 * @return string URL for ical for single event.
 	 */
 	function tribe_get_single_ical_link() {
-		return tribe_get_ical_link();
+		$tec    = Tribe__Events__Main::instance();
+		$output = trailingslashit( $tec->getLink( 'single' ) ) . '?ical=1';
+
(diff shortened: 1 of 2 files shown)


src/functions/template-tags/ical.php
```diff
@@ -10,7 +10,10 @@
 	 * @return string URL for ical for single event.
 	 */
 	function tribe_get_single_ical_link() {
-		return tribe_get_ical_link();
+		$tec    = Tribe__Events__Main::instance();
+		$output = trailingslashit( $tec->getLink( 'single' ) ) . '?ical=1';
+
+		return apply_filters( 'tribe_get_single_ical_link', $output );
```
borkweb: I wonder if it might be better to modify [`Tribe__Events__iCal::get_ical_link()`](https://github.com/moderntribe/the-events-calendar/blob/release/121/src/Tribe/iCal.php#L40) to accept a parameter (defaulted to `home`) to indicate which link it should return.

Then this function can more closely resemble `tribe_get_ical_link()`:

```php
function tribe_get_single_ical_link() {
  $output = Tribe__Events__iCal::get_ical_link( 'single' );
  return apply_filters( 'tribe_get_single_ical_link', $output );
}
```

lucatume: Yes, it would probably be better. I will do it as soon as I get back.

borkweb:
:neckbeard:

PR #316, opened 2015-09-18
Optionally support the exact minute of an event/break out of 5min intervals

barryhughes:
Re [C#39725](https://central.tri.be/issues/39725) this change makes it possible to _maintain_ (but not set) minute-level precision in the event editor where required for specific events - without otherwise breaking out of the default 5 minute increments for all events.


borkweb:
:+1:

PR #317, opened 2015-09-18
Fix pathinfo support (#39736)

barryhughes:
Re [C#39736](https://central.tri.be/issues/39736), ensure there are slashes in all the right places for pathinfo-based permalinks.


borkweb:
:rocket:

PR #289, opened 2015-09-02
respect post type perms for venue and org creation in the event meta box

jbrinley:
This PR address central ticket https://central.tri.be/issues/39179 (which in turn was spawned from https://central.tri.be/issues/39092). If a user has caps to edit an event, but not to create an organizer or venue, then the event form should only allow the user to choose one, not create one.

PR #318, opened 2015-09-20
remove empty default organizer for restricted users

jbrinley:
If a user cannot create organizers, remove the default organizer 0 from the list of current organizers so that an event with no organizer does not automatically get the first one (alphabetically) re-assigned to it every time the user saves the event - updates pr #289, central https://central.tri.be/issues/39179


bordoni:
Looks good. :wine_glass:

PR #320, opened 2015-09-21
Feature/39614 remove microformat classes

jazbek:
Microformats removed from the templates in core. 

Ref: [C#39614](https://central.tri.be/issues/39614)


bordoni:
Looks good :wine_glass:

PR #321, opened 2015-09-22
Move base google data markup to abstract class; extend to event

jazbek:
Moved google data markup to an abstract class, base "thing" properties are set in the abstract class, events, venues, and organizers will extend and add to the base data. Most properties are shared between schema.org objects and are pulled the same way from different wp post types, so this is keeping it DRY.

Also made it not be a singleton anymore because we have to support php 5.2 which doesn't have late static binding.

Ref: [C#39616](http://central.tri.be/issues/39616)

### DIFF
--- a/src/Tribe/Google_Data_Markup.php
+++ b/src/Tribe/Google_Data_Markup.php
@@ -3,52 +3,48 @@
 /**
  * Handles output of Google structured data markup
  */
-class Tribe__Events__Google_Data_Markup {
+abstract class Tribe__Events__Google_Data_Markup {
 
-	/**
-	 * @var $instance
-	 */
-	private static $instance = null;
+	protected $filter = 'tribe_google_data';
 
 	/**
 	 * Compile the schema.org event data into an array
 	 */
-	private function build_data() {
-
+	protected function build_data() {
 		global $post;
-		$id  = $post->ID;
-
-		$events_data = array();
+		$id             = $post->ID;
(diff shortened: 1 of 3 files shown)


src/Tribe/Google_Data_Markup.php
```diff
@@ -56,27 +52,16 @@ private function build_data() {
...
-		$html        = '';
-		if ( ! empty( $events_data ) ) {
+		$data = $this->build_data();
+		$data = $this->filter_data( $data );
+
+		$html = '';
+		if ( ! empty( $data ) ) {
 			$html .= '<script type="application/ld+json">';
-			$html .= str_replace( '\/', '/', json_encode( $events_data ) );
+			$html .= str_replace( '\/', '/', json_encode( $data ) );
```
bordoni: Could we have some filter to change the final markup?

jazbek: @bordoni It is passed through a filter 2x - once with $data and once with the script block string:

Data: 
https://github.com/moderntribe/the-events-calendar/blob/feature/39616-venue-org-json-ld/src/Tribe/Google_Data_Markup.php#L41

Script block string:
https://github.com/moderntribe/the-events-calendar/blob/feature/39616-venue-org-json-ld/src/Tribe/Template/Single_Event.php#L32

borkweb:
This looks good to go! :+1:

PR #325, opened 2015-09-23
Feature/39045 daynum ids

lucatume:
Addresses https://central.tri.be/issues/39045

“On pages with both the main calendar and mini calendar widget, there will be multiple daynum-2, daynum-3, etc. style ID attributes in the markup which causes W3C validation to fail :(

It can also happen with months that don't fill up the full square grid of month view or the mini calendar, i.e. for months that have days from the preceding and following month on the calendar grid with it. So, e.g., a daynum-1 in two places on main calendar and in two places in the mini calendar widget, all 4 ids potentially on the same page.

Maybe better as a data attribute or something?”

### DIFF
--- a/src/Tribe/Template/Month.php
+++ b/src/Tribe/Template/Month.php
@@ -644,7 +644,7 @@ if ( ! class_exists( 'Tribe__Events__Template__Month' ) ) {
 
 				$days[] = array(
 					'daynum'       => $day,
-					'daynum-id' => Tribe__Events__Utils__Id_Generator::generate_id( $day, 'css-ids' ),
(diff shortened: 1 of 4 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$day' | [single-day.php:21](https://github.com/moderntribe/the-events-calendar/blob/572e6f430ec8a279afd0f4c0113ab3e22d046aee/src/views/month/single-day.php#L21)

borkweb:
:+1:

lucatume:
I did not delete the branch as it might prove useful to follow along after the TDD session.

PR #329, opened 2015-09-24
add translation for 2 strings

zbtirrell:
See: https://central.tri.be/issues/35801

(worth reviewing with `?w=1`)


borkweb:
:+1:

PR #344, opened 2015-09-29
Do not overwrite the query's found_posts param with the wrong total

barryhughes:
Ensure we have accurate daily counts in month view. [C#39994](https://central.tri.be/issues/39994)


bordoni:
By the way, do we have reason for this to be there? (that count)
This looks good.

barryhughes:
> By the way, do we have reason for this to be there? (that count)

Yep we do - if there are more events than can be displayed in a single month view cell we add some "See all _x_ events" text. This just makes sure that count is accurate (and respects filters etc).

bordoni:
Thansk :wink

PR #345, opened 2015-09-29
Revert change to admin slug made in ed9e9b6c1d (spotfix)

barryhughes:
Reverting a change to the admin slug made inadvertently in https://github.com/moderntribe/the-events-calendar/commit/ed9e9b6c1d494afde54b067873ffe591a455ecab: Filter Bar tests against this and expects the original value of `tribe_events_calendar` - and so it's also probable other addons/customizations do the same.

### DIFF
--- a/src/Tribe/Settings.php
+++ b/src/Tribe/Settings.php
@@ -136,7 +136,7 @@ if ( ! class_exists( 'Tribe__Events__Settings' ) ) {
 			// set instance variables
 			$this->menuName    = apply_filters( 'tribe_settings_menu_name', esc_html__( 'The Events Calendar', 'the-events-calendar' ) );
 			$this->requiredCap = apply_filters( 'tribe_settings_req_cap', 'manage_options' );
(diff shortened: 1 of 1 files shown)


bordoni:
haha I didn't know that. I changed the EventBrite thinking that this might be a thing for `r/121`

barryhughes:
Yeah it crossed my mind I could update Filter Bar, too - but then again we don't know for sure what other 3rd party stuff may also be testing this/setting things up against it `¯\_(ツ)_/¯`

bordoni:
Our add-ons should be just using that variable, or a constant for the "plugin-slug".

barryhughes:
Yeah, true, but we could still break 3rd party code with this change (which was really only intended for text domains, unless I'm misunderstanding).

bordoni:
I mean, this is good. I just mentioned the _fun_ fact.

barryhughes:
:wink:

PR #346, opened 2015-09-30
providing a new link for iconv

zbtirrell:
See: https://central.tri.be/issues/40069


bordoni:
Sounds good, but these sites are ugly... :dragon

PR #347, opened 2015-09-30
Move options definition to correct method (spotfix)

barryhughes:
We moved the setup of core rewrite rules to its own method but failed to move the $options definition, which they rely on.


bordoni:
:8ball:

PR #348, opened 2015-10-01
Explicitly set no_found_rows to false so we get a useful found_posts …

barryhughes:
Try to ensure `set_found_posts()` works its magic and provides us with meaningful `found_posts` counts in our month view queries. 

[C#39994](https://central.tri.be/issues/39994)


borkweb:
:+1:

PR #349, opened 2015-10-01
don't escape html...

zbtirrell:
Noticed while investingating a different bug

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -1331,7 +1331,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 
 			$messages[ self::POSTTYPE ] = array(
 				0  => '', // Unused. Messages start at index 1.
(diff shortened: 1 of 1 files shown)


barryhughes:
:ballot_box_with_check:

afragen:
Question: Isn't the point of escaping translated strings to protect against malicious translations? If so, shouldn't **all** the text be escaped?

If not, can someone please explain the entire logic again. I'm confused, not an uncommon predicament. ;)

zbtirrell:
you are absolutely right.  The *right* thing to do is escape all the translations.  Especially as the oversight and approval of these is no longer under Modern Tribe's control (as of https://theeventscalendar.com/release-the-events-calendar-3-12-1/).

That being said, we are not at 100% on these yet (not just here but all over the place).  I took the short route on fixing the bug in front of us rather than doing it the right way...

We're better than that: https://github.com/moderntribe/the-events-calendar/pull/350

afragen:
You guys aren't just "better"' you're awesome!

PR #350, opened 2015-10-01
fixing escaping properly

zbtirrell:
Thanks @afragen for keeping us honest
https://github.com/moderntribe/the-events-calendar/pull/349#issuecomment-144769450

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -1331,7 +1331,12 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 
 			$messages[ self::POSTTYPE ] = array(
 				0  => '', // Unused. Messages start at index 1.
-				1  => sprintf( __( '%1$s updated. <a href="%2$s">View %3$s</a>', 'the-events-calendar' ), esc_html( $this->singular_event_label ), esc_url( get_permalink( $post_ID ) ), esc_html( $this->singular_event_label ) ),
+				1  => sprintf(
+					esc_html__( '%1$s updated. %2$sView %3$s', 'the-events-calendar' ),
+					esc_html( $this->singular_event_label ),
+					'<a href="' . esc_url( get_permalink( $post_ID ) ) . '">',
+					esc_html( $this->singular_event_label ) . '</a>'
+				),
 				2  => esc_html__( 'Custom field updated.', 'the-events-calendar' ),
 				3  => esc_html__( 'Custom field deleted.', 'the-events-calendar' ),
 				4  => sprintf( esc_html__( '%s updated.', 'the-events-calendar' ), $this->singular_event_label ),


afragen:
@zbtirrell I'm just trying to keep my understanding correct. :+1: 

Now, do all the parameters that are escaped as `esc_html` need to be escaped again as the HTML as a whole is escaped? I would keep the `esc_url`.

zbtirrell:
they do each need escaping because they are inserted into the translated string via the `sprintf`.  

Once the string is all composed, we are fine though, because each individual piece was known to be good.

afragen:
That seems to imply that the parameters are not processed by the escaping of the main HTML? I have no idea if my statement is correct or not. If parameters ar processed independently of the primary string then the parameter in `4` also needs escaping. :wink:

ggwicz:
Hey @afragen , parameter `4` is escaped fine in my opinion. `$this->singular_event_label` gets placed where `%s` is in the string, which then as a whole is passed through the `esc_html__()` function. But then `$this->singular_event_label` itself is a variable generated by the method `get_event_label_singular()`. This method itself _also_ passes the string through `esc_html__()` to begin with.

Big asterisk to all of this is that it's in _my opinion_. More escaping might indeed be deemed necessary by folks higher up the chain than myself.

zbtirrell:
@afragen is correct.  `$this->singular_event_label` is not high risk, but it should be escaped as well.  There is no solid justification for assuming it is safe.

This is the line in question:
```
4  => sprintf( esc_html__( '%s updated.', 'the-events-calendar' ), $this->singular_event_label ),
```

That is a public class variable and it's possible that something could have manipulated prior to the point in the code.  Again, this is unlikely, but this goes back to my statement in the other PR, we need to tighten this up in *many* remaining places.

ggwicz:
@zbtirrell right on – so then changing to this would make it pass code review?

`sprintf( __( '%s updated.', 'the-events-calendar' ), esc_html( $this->singular_event_label ) )`

zbtirrell:
nope, because while you have now properly escaped `$this->singular_event_label` the translated string is now potentially unsafe.  It should be: 
```
4  => sprintf( esc_html__( '%s updated.', 'the-events-calendar' ), esc_html( $this->singular_event_label ) ),
```

Note: none of this has a bearing on this specific code review as the line in question in this comment is not being changed at this time.

barryhughes:
We could also invert that and escape the result of sprintf(), saving ourselves an extra function call and making the code a little shorter and more readable. Just to be picky :-p

zbtirrell:
@barryhughes totally!  This example differs from the line I originally changed as it does not have any HTML to work around

barryhughes:
For sure, I'm just enjoying the banter :beer:

bordoni:
:8ball:

afragen:
Wow, didn't realize we would have so much discussion on this. Glad its stimulating such a discourse.

afragen:
But escaping the sprintf, while it works, isn't technically correct as that's not the string. It's a function manipulating  the string.

zbtirrell:
it is the final result of the concatenated string, so that is a completely reasonable place to do the escaping.  You can't go that route if there is HTML in the string, but it this example it would be the simplest solution.

PR #352, opened 2015-10-05
updating tests for new WP_Browser version

zbtirrell:
This reduces the number of errors and issues


borkweb:
Nothing looks scary to me but I'll have to trust that this will work as you intended because I don't know enough about these changes yet :)

:+1:

PR #353, opened 2015-10-05
Filter attendees CSV items

bostondv:
This adds a filter to allow customizing the attendees csv export contents.

Internal reference: https://central.tri.be/issues/40224


bostondv:
Do I need to create a ticket somewhere? If so, where can I do this?

zbtirrell:
@bostondv nope.  That's an internal tag for us.  We need to create a ticket so that this patch gets tracked through QA and eventually credited in the changelings. :)

bostondv:
Ok awesome :)

PR #355, opened 2015-10-07
Remove plugin namespace from our JQ events but retain .tribe namespace

borkweb:
The jQuery namespacing wasn't working as I had intended. Instead, let's keep our event names to `[action].tribe`

Related: https://github.com/moderntribe/events-pro/pull/154

See: https://central.tri.be/issues/36301


bordoni:
Looks good.

PR #358, opened 2015-10-09
fixed issue with autoloader finding common deprecated files

zbtirrell:
I added a test for the 4 basic ways we expect the autoloader to work and removed a duplicate test


bordoni:
From my perspective it's good.

PR #359, opened 2015-10-09
fixing a deprecated method call in this test

zbtirrell:
Also, created a place to test deprecated methods going forward

### DIFF
--- a/tests/functional/Updater_Test.php
+++ b/tests/functional/Updater_Test.php
@@ -5,8 +5,10 @@
  */
 class Tribe__Events__Updater_Test extends Tribe__Events__WP_UnitTestCase {
 
+
 	public function test_constant_updates_applied() {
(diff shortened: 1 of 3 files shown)


tests/functional/Updater_Test.php
```diff
@@ -1,12 +1,12 @@
...
 
-
 /**
  * @group updates
  */
 class Tribe__Events__Updater_Test extends Tribe__Events__WP_UnitTestCase {
+
 	public function test_constant_updates_applied() {
-		Tribe__Events__Main::instance()->setOption( 'schema-version', 0 );
+		Tribe__Settings_Manager::instance()->set_option( 'schema-version', 0 );
```
bordoni: On `Tribe__Settings_Manager` the method `set_option` is a Static Method. Does it make a diff here?

zbtirrell: yep! good catch

PR #363, opened 2015-10-09
We are not interested in empty costs values (spotfix)

barryhughes:
Query was returning costs where nothing had been specified, causing breakages downstream.


zbtirrell:
I assume this does not screw up zero cost / RSVP stuff?

barryhughes:
We still get costs back when they are explicitly set to zero.

zbtirrell:
:shipit:

zbtirrell:
:shipit:

PR #365, opened 2015-10-13
fix bad array generated by tribe_get_organizer_ids sometimes

zbtirrell:
See: https://central.tri.be/issues/39868

### DIFF
--- a/src/functions/template-tags/organizer.php
+++ b/src/functions/template-tags/organizer.php
@@ -55,6 +55,15 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 				$organizer_ids[] = $event_id;
 			} else {
 				$organizer_ids = tribe_get_event_meta( $event_id, '_EventOrganizerID', false );
+
+				// for some reason we store a blank "0" element in this array.
+				// let's scrub this garbage out
+				if ( is_array( $organizer_ids ) ) {
+					$key = array_search( '0', $organizer_ids );
+					if ( false !== $key ) {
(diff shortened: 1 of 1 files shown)


src/functions/template-tags/organizer.php
```diff
@@ -55,6 +55,15 @@ function tribe_get_organizer_ids( $event_id = null ) {
 				$organizer_ids[] = $event_id;
 			} else {
 				$organizer_ids = tribe_get_event_meta( $event_id, '_EventOrganizerID', false );
+
+				// for some reason we store a blank "0" element in this array.
+				// let's scrub this garbage out
+				if ( is_array( $organizer_ids ) ) {
+					$key = array_search( '0', $organizer_ids );
```
bordoni: Can't we just `array_filter`?

zbtirrell: `array_filter` requires a callback which feels less awesome to me, but I'm open to it if in general we think it is better

bordoni: if you pass just `array_filter( (array) $organizer_ids );` and 0 would be excluded, I'm like 90% sure of it...

zbtirrell: I tested with this:
```
<?php
$bad_array = array('0','monkey');
$bad_array = array_filter( (array) $bad_array );
print_r($bad_array);
```
sure enough, it works as you suggest.  Changing.

bordoni:
:wink:

PR #368, opened 2015-10-13
Only activate the ticket plugin if other ticket plugins are active

borkweb:
In testing the various activated states of our plugins, @zbtirrell and I discovered that tickets was _always_ active.


zbtirrell:
:+1:

PR #370, opened 2015-10-14
Form correct strings for start/end date (spotfix | props: jazbek)

barryhughes:
As Jessica highlighted - we were testing for the existence of properties but weren't forming the property name correctly (ie, instead of checking for `EventStartDate` we were checking for `EventStartStartDate`) ... this fixes that.


borkweb:
:+1:

PR #371, opened 2015-10-15
Prevent fatal when TEC is upgraded and other plugins are not

borkweb:
TEC's plugins_loaded was firing at the same time as some of our other plugins which wasn't early enough to initialize the new autoloader for 4.0. This ensures that TEC is initialized before any other plugin regardless of 3.x or 4.x release.

See: https://central.tri.be/issues/40281


bordoni:
:+1:

PR #373, opened 2015-10-15
Bring back tribe_events_resource_url()

borkweb:
When The Events Calendar and Event Tickets are both active, the common library's tribe_resource_url() function returns an incorrect URL for TEC resources when Tickets is the plugin that runs common libs.

See: https://central.tri.be/issues/40451

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -1435,7 +1435,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 		$url = apply_filters( 'tribe_events_resource_url', $url, $resource );
 
 		if ( $echo ) {
-			echo $url;
+			echo esc_url( $url );
 		}
 
 		return $url;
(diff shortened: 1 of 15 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$url' | [general.php:1438](https://github.com/moderntribe/the-events-calendar/blob/504e84e3d228582cbaa3a8fec01fe7cf60459923/src/functions/template-tags/general.php#L1438)

src/functions/template-tags/general.php
```diff
@@ -1392,4 +1392,52 @@ function tribe_get_render_context( $query = null ) {
...
+
+		$url  = plugins_url( Tribe__Events__Main::instance()->plugin_dir . $path );
+
+		/**
+		 * Deprected the tribe_events_resource_url filter in 4.0 in favor of tribe_resource_url. Remove in 5.0
+		 */
+		$url = apply_filters( 'tribe_events_resource_url', $url, $resource );
+
+		if ( $echo ) {
+			echo $url;
```
barryhughes: Might we want an esc_url(), here or earlier?

barryhughes:
:boom:

PR #377, opened 2015-10-18
Improved German translation and typos.

ohli:
The expression "Veranstaltungen VOM Oktober" is wrong, instead this should be a IM.
Also found and fixed a typo.
First public pull request, please be kind.


zbtirrell:
@ohli Translations are now being managed by WordPress.org - https://translate.wordpress.org/projects/wp-plugins/the-events-calendar/stable

PR #383, opened 2015-10-20
Don't double-nest initialization of ticket plugin under 2 plugins_loaded

borkweb:
Let's check if we need to initialize the tickets plugin when we're bootstrapping all of the hooks and libraries for TEC.

Related: https://github.com/moderntribe/event-tickets/pull/17

See: https://central.tri.be/issues/40590


bordoni:
:wink:

PR #388, opened 2015-10-23
we were checking a class that was not deprecated

zbtirrell:


bordoni:
It was deprecated for a little while! :8ball: merge!

PR #389, opened 2015-10-23
Deprecated getOption + Importer Type

bordoni:
_Ref:_ [C#38847](https://central.tri.be/issues/38847)

### DIFF
--- a/src/io/csv/admin-views/general.php
+++ b/src/io/csv/admin-views/general.php
@@ -33,6 +33,9 @@ require 'header.php';
 			if ( strpos( $key, 'imported_post_status' ) === false ){
 				$value = Tribe__Settings_Manager::get_option( $key, null );
 			} else {
+				/**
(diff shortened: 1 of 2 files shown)


src/io/csv/admin-views/general.php
```diff
@@ -33,7 +33,11 @@
 			if ( strpos( $key, 'imported_post_status' ) === false ){
 				$value = Tribe__Settings_Manager::get_option( $key, null );
 			} else {
-				$type = isset( $field_args['importer_type'] ) ? $field_args['importer_type'] : 'csv';
+				if ( preg_match( '/\[([^\]]+)\]/i', $key, $match ) ) {
```
barryhughes: Could we add a comment giving an example of what the regex aims to match?

barryhughes:
You are a scholar and a gentleman.

PR #391, opened 2015-10-23
cleanup and consistent code for acceptance tests

zbtirrell:
With this, all the acceptance tests should work without requiring any known preconfiguration of the local instance


bordoni:
:+1:

PR #398, opened 2015-10-26
Ensure rewrite rules included translated forms for "tag" and "category" (#40579)

barryhughes:
Previously we were forming links where tag/category was translated, but our rewrite rules only covered the English terms - leading to unexpected results when users try to load taxonomy views in a non-English language setup.

[C#40579](https://central.tri.be/issues/40579)


borkweb:
:racehorse: :dash:

PR #397, opened 2015-10-26
adding tests for editing venues and organizers

zbtirrell:
See: https://central.tri.be/issues/40953

PR #401, opened 2015-10-27
Venues and Organizer caps back on track

bordoni:
_Ref:_ [C#40953](https://central.tri.be/issues/40953)
_Rel:_ #397


zbtirrell:
someday, I want us to use filtered caps and not persistent ones, but today is not that day.  

This looks good.

PR #402, opened 2015-10-27
Use the declared $calendar_day rather than the non-existent $current_day

borkweb:
This block of code changes in `release/121` and `$current_day` becomes a declared variable with useful data. But not in `release/120`

See: https://central.tri.be/issues/40980


barryhughes:
Just checking to be absolutely safe ... but is `release/120` the correct target branch (isn't CRMR going out post-4.0)?

borkweb:
The next CR is the release of Community Tickets which will be TEC 3.12.4

barryhughes:
:+1:

PR #406, opened 2015-10-28
Single Event title tag should be an h1

wpexplorer:
For SEO reasons the title to be an H1 tag and not an h2.


borkweb:
Can this change be Pull Requested against our working branch (`release/121.1`) rather than `master`? Thanks!  I'll close this one for now.

PR #408, opened 2015-10-29
a few testing and test-ability changes

zbtirrell:
* bugfix in the Updater so it does not run the upgrade functions on a new install
* changing functions from `protected` to `public` to make them more testable
* capturing the updater object in a `static` variable to make it more testable
* fix the Capabilities Venue/Organizer tests to use the right data provider
* add a bunch of Updater tests


bordoni:
Solid!

PR #413, opened 2015-11-03
ran all our images through ImageOptim for lossless compression

zbtirrell:
Since this is lossless, it should be totally safe.


borkweb:
:+1:

PR #414, opened 2015-11-03
Cancel was not getting translated

borkweb:
Fix that and clean up some markup.

See: https://central.tri.be/issues/39134


barryhughes:
:zap:

PR #419, opened 2015-11-04
Events title tag should be an h1

wpexplorer:
For SEO reasons it makes more sense for the title to be an H1 tag and not an h2.


borkweb:
Since this has merge conflicts, I've re-done the work and pull-requested it here: https://github.com/moderntribe/the-events-calendar/pull/425

borkweb:
Thank you for the change!

PR #420, opened 2015-11-05
Restore expected day classes in month view | #41258

barryhughes:
Restores the expected CSS classes for each day in month view:

* Compares timestamps with timestamps (was comparing a timestamp with an array)
* Drops unneeded use of `date_i18n()` for day numbers (was effectively converting them all to `01`)

[C#41258](https://central.tri.be/issues/41258)


borkweb:
:+1:

PR #421, opened 2015-11-05
Mobile month view - restore back compat fix | #41258

barryhughes:
Restore backwards compatibility measure which had been removed prematurely.

[C#41258](https://central.tri.be/issues/41258)


borkweb:
:+1:

PR #422, opened 2015-11-06
scaling improvements

codearachnid:
added scaling improvements for large event based sites to prevent extraneous database lookups for data that is cached by normal WP query caching.

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -1694,7 +1694,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 				if ( ! empty( $my_venues ) ) {
 					foreach ( $my_venues as $my_venue ) {
 						$my_venue_ids[] = $my_venue->ID;
-						$venue_title    = wp_kses( get_the_title( $my_venue->ID ), array() );
+						$venue_title    = wp_kses( $my_venue->post_title, array() );
 						$my_venue_options .= '<option data-address="' . esc_attr( $this->fullAddressString( $my_venue->ID ) ) . '" value="' . esc_attr( $my_venue->ID ) . '"';
 						$my_venue_options .= selected( $current, $my_venue->ID, false );
 						$my_venue_options .= '>' . $venue_title . '</option>';
@@ -1741,7 +1741,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 				if ( $venues ) {
 					echo $my_venues ? '<optgroup label="' . esc_attr( apply_filters( 'tribe_events_saved_venues_dropdown_optgroup', sprintf( esc_html__( 'Available %s', 'the-events-calendar' ), $this->plural_venue_label ) ) ) . '">' : '';
 					foreach ( $venues as $venue ) {
-						$venue_title = wp_kses( get_the_title( $venue->ID ), array() );
+						$venue_title = wp_kses( $venue->post_title, array() );
 						echo '<option data-address="' . esc_attr( $this->fullAddressString( $venue->ID ) ) . '" value="' . esc_attr( $venue->ID ) . '"';
 						selected( ( $current == $venue->ID ) );
 						echo '>' . $venue_title . '</option>';
@@ -1783,7 +1783,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 				if ( ! empty( $my_organizers ) ) {
 					foreach ( $my_organizers as $my_organizer ) {
 						$my_organizer_ids[] = $my_organizer->ID;
-						$organizer_title    = wp_kses( get_the_title( $my_organizer->ID ), array() );
+						$organizer_title    = wp_kses( $my_organizer->post_title, array() );
 						$my_organizers_options .= '<option value="' . esc_attr( $my_organizer->ID ) . '"';
 						$my_organizers_options .= selected( $current, $my_organizer->ID, false );
 						$my_organizers_options .= '>' . $organizer_title . '</option>';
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space after the first parenthesis | [Main.php:3815](https://github.com/moderntribe/the-events-calendar/blob/82836f9cda7ad6516f243f810c8ec51620f79512/src/Tribe/Main.php#L3815)
:x: | There must be a single space before the last parenthesis | [Main.php:3815](https://github.com/moderntribe/the-events-calendar/blob/82836f9cda7ad6516f243f810c8ec51620f79512/src/Tribe/Main.php#L3815)
:x: | There must be a single space after the first parenthesis | [Main.php:3834](https://github.com/moderntribe/the-events-calendar/blob/82836f9cda7ad6516f243f810c8ec51620f79512/src/Tribe/Main.php#L3834)
:x: | There must be a single space before the last parenthesis | [Main.php:3834](https://github.com/moderntribe/the-events-calendar/blob/82836f9cda7ad6516f243f810c8ec51620f79512/src/Tribe/Main.php#L3834)

barryhughes:
Ticketed: [C#41442](https://central.tri.be/issues/41442)

barryhughes:
Hey Tim,

Thanks for the PR!

I'm ever so reluctant to remove the calls to `get_the_title()` since that also removes the ability for those titles to be filtered by custom code (and I wouldn't want to trigger breakages for other users). 

Also - though we would definitely save ourselves a few function calls by accessing the `post_title` property directly - it doesn't look to me as if we'll be preventing a database lookup: the post will already be cached at this point and `get_the_title()` takes advantage of that via `get_post()` and ultimately `WP_Post::get_instance()`.

That said, you're clearing hitting a problem and we should address that. Here are a few questions/thoughts in no particular order:

* I'd love to get a better sense of the scale we're talking about here - how many venues and organizers are you dealing with here?
* If we do want to optimize things within these methods, maybe there is other fat to trim besides removing `get_the_title()` (since that has the undesirable result of removing filtering) ... I'm wondering if we really need the relatively expensive `wp_kses()` calls for instance and whether alternative means of sanitization could be used there
* Ultimately, perhaps we're looking in the wrong place and could arrive at a better solution by dynamically populating these dropdowns on demand via ajax, limiting the number of posts that are queried for at any given time

What are your thoughts?

codearachnid:
Hey @barryhughes thank you for the detailed follow up - I too am loathe to drop the get_the_title method - however it does add an additional db lookup. So if the filtering is still desired using `apply_filters('the_title', $obj->post_title, $obj->ID )` might be more appropriate - the issue is that a smaller object is returned in the venue and organizer queries within the plugin code and by passing through only the ID it forces the method to do a lookup against the DB for the title field again... This seems quite wasteful.

Also I wonder if you weren't privy to the testing I've run in regards to this PR? I sent a link along to the plugin twitter account with a link to the gist I am using to track my testing/improvements https://gist.github.com/codearachnid/58bff629062829f06297

barryhughes:
Hey Tim,

I hadn't seen the gist but that's a great summary and I think we're all on board with the idea of lazy/ajax-driven approach to solving this (and we have a ticket logged interally for exactly that) :+1: 

> the issue is that a smaller object is returned in the venue and organizer queries within the plugin code and by passing through only the ID it forces the method to do a lookup against the DB for the title field again... 

I could easily be missing something, but I just don't see this:

* We query for the venues (or organizers) at the top of each method
* Behind the scenes, WP takes care of caching these for the remainder of the request 
* Though we only pass an ID to `get_the_title()` this doesn't force a further db lookup - it pulls in the post object via a call to `get_post()` which takes advantage of the cache

If I'm following your description correctly it sounds like in your case there is an extra query for every single venue and organizer that is listed in those dropdowns ... I don't see this and applying your patch does not reduce the number of queries when I test it. Is this something you can replicate in a 'plain vanilla' test site?

If so, we can definitely get another team member to review this :smiley:

codearachnid:
So where the extra query comes into play is with the way `get_the_title` determines the title. 

Straight from WP core code:
```
function get_the_title( $post = 0 ) {
    $post = get_post( $post );
    $title = isset( $post->post_title ) ? $post->post_title : '';
    $id = isset( $post->ID ) ? $post->ID : 0;
...
```

Thus it seems wasteful to force get_post to execute a lookup again - potentially uncached or flushed (given how WPEngine & Page.ly caches). This becomes an expensive lookup against the object that could just be passed through. Potentially instead of loosing the `get_the_title` filter opportunity if the venue/organizer object were passed through it would be enough to sate the need for a lookup to get the title on the object that already looked up to provide the title.

I agree the ajax is the smart route - but even then from a perf standpoint it might make sense to pass the object instead of just the ID property only to fetch the object to display a sibling property.

As far as scale goes - we're looking at ~4500+ events, ~5500+ organizers, ~6000+ venues. We used to run around ~8k-10k events but began aggressively dropping events older than a year and cut down the event recurrence usage. The biggest problem right now is due to the amount of items loading that it bogs the loading of the event editor page with over 18,000 queries primarily due to the 2 select fields for organizer/venues. By commenting out that code I can drop the query count to < 200 queries for the page.

Side note: Given the manual moderation on the site (which I do not handle content myself, only code) it seems there is potential for duplicate venue/organizer entries given the community submission process along with orphaned venues/organizers as the events are cleaned out... for another topic/time I would be interested if there are better ways to moderate this content.

barryhughes:
> Thus it seems wasteful to force get_post to execute a lookup again 

Well, this is where we're seeing two different things. A call to `get_post()` doesn't necessarily mean it is going to perform another db query - and it really shouldn't do so here since the post objects are already in memory.

That said, I'm going to bow out and let someone else look this over with fresh eyes :smiley:

codearachnid:
As a bit of follow up - I have updated the gist with some additional findings I am using on the plugin:

I removed `wp_kses` and changed the argument sent to `get_the_title` from the ID to the item object. Based on the core functionality of `get_the_title` it will accept the object without additional DB lookups for sanity or value retrieval (see https://core.trac.wordpress.org/browser/tags/4.3.1/src/wp-includes/post-template.php#L110) I have confirmed that passing the object through reduces all the additional lookup calls even in a non-cached environment.

Removal of the kses methods does not resolve the issue of potential tags in the title but ideally that would be handled on save vs display of the content. Thus if the large majority use case is smaller sites it would make sense to continue down the current path; however, in the case of a large set of venues/organizers it might be better suited to allow performance tuning without modifying the plugin code via a filter. As identified in earlier conversation an AJAX call for this data would be most optimal to shove the processing to later stage while the admin page renders.

For background on wp_kses performanace see this article with pretty charts and numbers: https://www.tollmanz.com/wp-kses-performance/

PR #430, opened 2015-11-10
Fix inconsistency with Week icons didn't match others #000 vs #696969

borkweb:
See: https://central.tri.be/issues/41302


bordoni:
:+1:

PR #431, opened 2015-11-10
Strip out order_table logic (not needed until r/121) | #41327

barryhughes:
Strips out the `Tribe__Events__Tickets__Orders_Table` class (and some code that interacted with it): this probably shouldn't have made its way into `release/120` and was triggering fatal errors in some conditions.

[C#41327](https://central.tri.be/issues/41327)


borkweb:
That code is used in Community Events.

barryhughes:
Time for a new approach then.

PR #432, opened 2015-11-10
Hide the orders report for non-wootickets setups

borkweb:
See: https://central.tri.be/issues/41327


barryhughes:
That was easy :-p

PR #434, opened 2015-11-10
Ensure that EventStartDate isn't stripped from importer page queries

borkweb:
The Facebook importer lives on the "same" current screen as the event list in the dashboard, however, it _should not_ have the date filters removed like the event list requires. This fixes that so that duplicate Facebook imports can be identified.

This changeset includes tests.

~~This change set _requires_ [this PR](https://github.com/moderntribe/tribe-common/pull/28) to be merged into tribe-common (and the hash will need an update)~~ The hash for the `common/`directory has been updated.

Best reviewed with `?w=1`

See: https://central.tri.be/issues/41385


barryhughes:
:1234: :books:

PR #439, opened 2015-11-11
Re-structure Query::set_order() to allow better support on WP 4.2

bordoni:
_Ref:_ [C#38473](https://central.tri.be/issues/38473)


borkweb:
This is a good find. :+1:

PR #440, opened 2015-11-11
41435 make a js selector less specific for better compatibility with filters

ggwicz:
_Ref:_ [C#41435](https://central.tri.be/issues/41435)


src/resources/js/tribe-events-ajax-calendar.js
```diff
@@ -180,7 +180,7 @@
 
 			var $today = $wrapper.find( '.tribe-events-present' ),
 				$mobile_trigger = $wrapper.find( '.mobile-trigger' ),
-				$tribe_grid = $wrapper.find( '#tribe-events-content > .tribe-events-calendar' );
+				$tribe_grid = $wrapper.find( document.querySelector( '#tribe-events-content .tribe-events-calendar' ) );
```
barryhughes: Wrapping the result of `document.querySelector()` in a `find()` definitely jumps out at me as odd: what's the reason for doing this over say a straight `find()` (or wrapping the result of the querySelector call into a jQuery object)?

bordoni: I think to make it faster it would be better to do:
````
$wrapper.find( document. getElementById( 'tribe-events-content' ) ).find( '.tribe-events-calendar'  );
```

borkweb:
:+1:

PR #441, opened 2015-11-12
Smarter handling of tribe_events_month_has_events | #41332

barryhughes:
There can be more than one month view (and so more than one month template object) per request - ie one or more calendar widgets might be used alongside the main month view. In such a case, the widget-related object could remove a filter added by the main view object, causing unusual results.

This change should prevent that at least with regards to the `tribe_events_month_has_events` hook.

[C#41332](https://central.tri.be/issues/41332)


bordoni:
:+1:

PR #442, opened 2015-11-12
Hide the Orders report for Woo if version is less than 3.12

borkweb:
There's a fatal, otherwise.

See: https://central.tri.be/issues/41327#note-9


barryhughes:
:sandal:

PR #443, opened 2015-11-13
Don't change the visibility of these methods. Make new methods

borkweb:
To more easily test `get_updates()` and `constant_updates()`, those methods were converted to public. This causes fatals when Pro 3.12.x is active because Pro has a class that extends `Tribe__Events__Updater` and overrides the `protected` methods.

The fix is to retain those methods as protected and make new ones that are public.

See: https://central.tri.be/issues/40281

### DIFF
--- a/src/Tribe/Updater.php
+++ b/src/Tribe/Updater.php
@@ -59,9 +59,24 @@ class Tribe__Events__Updater {
 	 * and lower than $this->current_version will have its
 	 * callback called.
 	 *
+	 * This method has been deprecated in favor of a more testable public function
+	 *
+	 * @return array
+	 * @deprecated 4.0
+	 */
+	protected function get_updates() {
+		return $this->get_update_callbacks();
+	}
+
+	/**
+	 * Returns an array of callbacks with version strings as keys.
+	 * Any key higher than the version recorded in the DB
+	 * and lower than $this->current_version will have its
+	 * callback called.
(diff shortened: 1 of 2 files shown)


src/Tribe/Updater.php
```diff
@@ -59,9 +59,24 @@ public function update_version_option( $new_version ) {
 	 * and lower than $this->current_version will have its
 	 * callback called.
 	 *
+	 * This method has been deprecated in favor of a more testable public function
+	 *
+	 * @return array
+	 * @deprecated 4.0
+	 */
+	protected function get_updates() {
+		return $this->get_update_callbacks();
```
barryhughes: Optionally trigger a notice via `_deprecated_function()`?

src/Tribe/Updater.php
```diff
@@ -74,9 +89,22 @@ public function get_updates() {
 	 * Returns an array of callbacks that should be called
 	 * every time the version is updated
 	 *
+	 * This method has been deprecated in favor of a more testable public function
+	 *
+	 * @return array
+	 * @deprecated 4.0
+	 */
+	protected function constant_updates() {
+		return $this->get_constant_update_callbacks();
```
barryhughes: Consider `_deprecated_function()` here

barryhughes:
Good to go bar the above!

PR #444, opened 2015-11-13
Don't hide the default address fields if a default venue is selected

borkweb:
Just because the site owner wants a default venue to be selected when creating an event _doesn't_ mean that they don't want default address information when creating a new event in the dashboard.  Let's keep the address fields visible so defaults can be selected.

See: https://central.tri.be/issues/29366#note-17


barryhughes:
Hey Matt, just want to be sure I'm understanding this: Let's say:

* Default venue is _"A Nice Place"_
* Various default address fields are also set

If the user selects _"Use a New Venue"_ within the event editor should this change ensure those address fields (street, city etc) are filled with the defaults?

borkweb:
This specific changeset is to ensure that the address fields on Events > Settings > Default Content do not get hidden when a Default venue is selected.  However, I'm assuming you are asking about the overall goals associated with the Central ticket.

In the scenario you laid out, the following should be true:

1. When you go to add a new Event, the default Venue should be selected in the drop-down (with address fields hidden)
1. When you go to add a new Event and select _"Use a New Venue"_, the address fields should display but they should be blank.
1. When you go to add a new Venue (via Events > Venue > Add New), the default address information should be set to the defaults.

barryhughes:
It was the specific goal of the changeset I was interested in and your answer gives me everything I need :smile_cat:

PR #445, opened 2015-11-13
If we unhook posts_fields() we should also unhook posts_orderby()

barryhughes:
Within `Tribe__Events__Query` if `self::should_remove_date_filters( $query )` is true we unhook our `posts_fields` filter but not our `posts_orderby` filter (which depends on the aliases set up by the former). This change ensures they are all unhooked together - otherwise we hit things like:

*Unknown column 'EventStartDate' in 'order clause'*

[C#35479](https://central.tri.be/issues/35479)


borkweb:
:ok_hand:

PR #446, opened 2015-11-13
Remove date filters on import pages

borkweb:
With [this fix](https://github.com/moderntribe/the-events-calendar/pull/445), we don't really need filters on the import pages. In fact, excluding them is necessary. Previous tests returned false positives that were caused by the date of the event and the injected date aligning just so...through crazy random happenstance. When looking for importable events, we won't need to have a date-beefed-up query.

See: https://central.tri.be/issues/41385


bordoni:
:+1:

PR #452, opened 2015-11-17
Fixed broken acceptance test

lucatume:
ticket: https://central.tri.be/issues/38042
I've updated the broken acceptance test to use latest WPDb methods as such it will require a `composer update lucatume/wp-browser` call to pass


tests/acceptance/Issue_38042_Dropping_CategoryCest.php
```diff
@@ -4,16 +4,18 @@
...
+		$this->settings_backup = $I->grabOptionFromDatabase( 'tribe_events_calendar_options' );
+		codecept_debug( "Settings backup: " . json_encode( $this->settings_backup ) );
```
borkweb: Do we need to keep this debug statement?

borkweb:
:+1:

PR #453, opened 2015-11-19
Created the phantomjs suite

lucatume:
Ticket: https://central.tri.be/issues/40423
I've moved the only acceptance test we have in the new `phantomjs` suite and solved the problem (window size) that would prevent it from running.  
Along with it I've added a Tribe module to handle options in acceptance tests.
I've PRed on core as the issue under test (https://central.tri.be/issues/38042) was really a core issue. We currently have the same exact test on Pro and filters and I can easily replicate the test, the module and the suite set up there too should that be the case.

### DIFF
--- a/tests/_support/Modules/Acceptance/Options.php
+++ b/tests/_support/Modules/Acceptance/Options.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Tribe\Tests\Modules\Core\Acceptance;
+
+
+class Options extends \Codeception\Module {
+
+	/**
+	 * Returns an array of default options overriding the specified ones.
+	 *
+	 * @param array $overrides
+	 *
+	 * @return array
+	 */
+	public function grabDefaultCoreOptions( array $overrides = [ ] ) {
+		$defaults = [
+			'schema-version'                   => '4.0beta2',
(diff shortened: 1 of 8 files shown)


lucatume:
Just a side note: the debug message printing the backed up options is not showing not because options are not being backed up but because codeception runs too fast. An Xdebug run through the code will show the expected content (XDebug slows down exec time sensibly).

tests/_support/Modules/Acceptance/Options.php
```diff
@@ -0,0 +1,61 @@
...
+			'enable_month_view_cache'          => false,
+			'dateWithYearFormat'               => 'm/d/Y',
+			'dateWithoutYearFormat'            => 'F j',
+			'monthAndYearFormat'               => 'F Y',
+			'dateTimeSeparator'                => ' @ ',
+			'timeRangeSeparator'               => ' - ',
+			'datepickerFormat'                 => '0',
+			'tribeEventsBeforeHTML'            => '',
+			'tribeEventsAfterHTML'             => '',
+		];
```
borkweb: This beats the pants off of the serialized data. Nice change!

borkweb:
Good stuff! :cactus:

PR #456, opened 2015-11-21
Fix default Display data

bordoni:
_Ref:_ [C#41738](https://central.tri.be/issues/41738)

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -1611,15 +1611,10 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		public function displayEventVenueDropdown( $post_id ) {
 			$venue_id = get_post_meta( $post_id, '_EventVenueID', true );
 
-			$is_community_edit = false;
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -1610,15 +1610,21 @@ public function addVenueAndOrganizerEditor() {
 		 */
 		public function displayEventVenueDropdown( $post_id ) {
 			$venue_id = get_post_meta( $post_id, '_EventVenueID', true );
+
+			$is_community_edit = false;
+			if ( class_exists( 'Tribe__Events__Community__Main' ) ) {
+				$is_community_edit = Tribe__Events__Community__Main::instance()->isEditPage;
+			}
```
borkweb: Since this is 4 lines with the smallest amount of logic, it feels like it should be a method.

bordoni: Agreed, will work on dat.

bordoni:
Updated this code.

borkweb:
:bacon:

PR #458, opened 2015-11-24
Test for presence of events _without_ any cost

barryhughes:
Adds a helper to test for the existence of events _without_ a cost (needed for instance by Filter Bar to smooth out its own handling of cost ranges). Also adds a set of tests for our cost utils (including a failing test we'll need to solve as a further piece of work).

[C#41208](https://central.tri.be/issues/41208)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -55,16 +55,16 @@
         },
         {
             "name": "codeception/codeception",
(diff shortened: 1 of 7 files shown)


zbtirrell:
:+1:

PR #461, opened 2015-11-26
Allow plugins to do something once importer options have updated | 41714

barryhughes:
Fire an action once importer options have been updated.

[C#41714](https://central.tri.be/issues/41714)


bordoni:
:+1:

PR #463, opened 2015-11-29
add slash to past/upcoming redirect URL to prevent redirect loop

jbrinley:
With the existing redirect, http://example.com/events/upcoming would redirect to http://example.com/events/upcoming, since it lacks the trailing slash to match the `str_replace()`. We can be more accommodating with a `preg_replace()`.

https://central.tri.be/issues/41869
https://central.tri.be/issues/41937


borkweb:
:+1:

PR #467, opened 2015-12-01
Relocating date format and time separator fields from tribe-common

borkweb:
See: https://central.tri.be/issues/41980#note-3


bordoni:
:100:

PR #468, opened 2015-12-02
Remove unneeded (and problem-causing) code | 41574

barryhughes:
Removes a section of code that was wiping the title of the list which displays below month view in mobile mode. Props will be due to Jason on the forum for this one.

[C#41574](https://central.tri.be/issues/41574)


borkweb:
:+1:

PR #470, opened 2015-12-02
Default the enabled views to "month"

borkweb:
We set the default view to `month`, however, our default list of enabled views was empty. On a new install, this caused a 404 until the user set the views they wanted to enable via Events > Settings > Display.

This change sets `month` as an enabled view by default so that `/events/` can render before the user hits _Save_ on Events > Settings > Display

See: https://central.tri.be/issues/41981

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -356,6 +356,7 @@ At no point during the 3.0 lifecycle will the major version change. But you can 
 * Fix - Fixed a problem where iCal data was generated with incorrect dates in the case of some all days events (thanks to Matt for highlighting this)
 * Fix - Resolved a problem causing the previous month view to appear when it should not
 * Fix - Fixed issue in mobile month view where date was missing from heading
+* Fix - Resolved issue that caused /events/ to 404 if it was visited on a new install before hitting "Save" on the Events > Settings > Display page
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -356,6 +356,7 @@ At no point during the 3.0 lifecycle will the major version change. But you can
 * Fix - Fixed a problem where iCal data was generated with incorrect dates in the case of some all days events (thanks to Matt for highlighting this)
 * Fix - Resolved a problem causing the previous month view to appear when it should not
 * Fix - Fixed issue in mobile month view where date was missing from heading
+* Fix - Resolved issue that caused /events/ to 404 if it was visited on a new install before hitting "Save" on the Events > Settings > Display page
```
zbtirrell: was this an existing bug that we fixed, or something caused by the work in 4.0 (in other words, is a changelog entry required here?)

barryhughes:
:monkey:

readme.txt
barryhughes: I don't have a Central #ref to hand to back me up, but I definitely recall this/variations of it in the past.

borkweb: It was pre-existing

PR #471, opened 2015-12-02
Fixing some textdomain issues that were introduced in late changes

borkweb:
I did some changes yesterday to move settings fields from common to TEC.  In doing so, the textdomain was not updated from tribe-common to the-events-calendar. This was noticed while testing translations for tribe-common.


zbtirrell:
:+1:

PR #472, opened 2015-12-02
Exclude more stuff from packaging

borkweb:
Some of the unwanted dirs in `vendor/` were making their way into the packaged zip

Best if viewed with `?w=1`


zbtirrell:
:+1:

PR #473, opened 2015-12-02
Avoid hardcoded table prefixes | 42021

barryhughes:
A query with hardcoded `wp_` table prefixes had snuck in, breaking the admin events list when ordered by tag or category. This change should fix (for any sites that don't use `wp_` as their prefix).

[C#42021](https://central.tri.be/issues/42021)


bordoni:
:100:

PR #474, opened 2015-12-02
Admin events list: ensure date ordering is applied | 42022

barryhughes:
Previously, ordering the admin event list by date failed because we were *appending* the extra rules. This change prepends the rules instead, restoring proper ordering by start/end date.

Example, instead of:

`ORDER BY wp_posts.post_date DESC, tribe_event_start_date.meta_value DESC, tribe_event_end_date.meta_value DESC`

We now get:

`ORDER BY tribe_event_start_date.meta_value DESC, tribe_event_end_date.meta_value DESC,  wp_posts.post_date DESC`

[C#42022](https://central.tri.be/issues/42022)


borkweb:
:i like it: :neckbeard:

PR #475, opened 2015-12-03
Enable the list view for fresh installs in addition to month view

borkweb:
See: https://central.tri.be/issues/42052


bordoni:
:+1:

PR #477, opened 2015-12-04
Don't inject the _EventHideFromUpcoming on CSV import if it isn't needed

borkweb:
Events imported via CSV were getting the `_EventHideFromUpcoming` meta key inserted blank. That's pretty silly, and the next/previous link generation dies a horrible death if the meta key even exists.

See: https://central.tri.be/issues/42054


bordoni:
:100:

PR #479, opened 2015-12-04
Correct asset url (spotfix)

barryhughes:
This class was moved to tribe-common and back to core again, but on restoring it to core we didn't fix the path for a script we were enqueueing - so it was 404ing :sob:


bordoni:
:100:

PR #481, opened 2015-12-07
Observe the end-of-day cutoff when building the date range for multi-day

borkweb:
When running multi-day all-day events, we were not looking at the end of day cutoff when building the date range. Instead, we blindly took the end date/time and displayed that. Example:

End of day cut-off: 3am
Event: All-day, December 8 to December 10

*With the bug:*

* Event dates would be stored as 2015-12-08 03:00:00 to 2015-12-11 02:59:59
* Event would be rendered as December 8 - December 11

*With the fix:*

* Event dates would be stored as 2015-12-08 03:00:00 to 2015-12-11 02:59:59
* Event would be rendered as December 8 - December 10

See: https://central.tri.be/issues/40555


barryhughes:
:stars: 
:alien:

PR #482, opened 2015-12-08
Don't use a timestamp when calling tribe_beginning_of_day()

borkweb:
I had nicely functioning code and then I made a last minute change with timestamps to shorten the code needed to generate the correct date range. I obviously didn't test it well enough with multi-day all-day events with a non 12am end of day cutoff. This update rolls back the 12th hour timestamp tweaks and renders the dates as they should.

See: https://central.tri.be/issues/40555#note-10


zbtirrell:
:+1:

PR #484, opened 2015-12-08
Multi-day All-day events may require an adjusted date by eod cutoff

borkweb:
All-day Multi-day events have their end dates displayed individually in a few locations - when viewing a single event and viewing the event in the admin dashboard. In those instances, the true end date is not what is desired, but, rather an end date that observes the end of day cutoff.

This changeset adds a new method ( `tribe_get_display_end_date()` ) that calls `tribe_get_end_date()` and adusts the output based on the multi-day/eod status of the event whose end date is being fetched.

**Note:** I'm also removing `tribe_get_start_date()` and `tribe_get_end_date()` from TEC because those are duplicated in tribe-common now. Which meant that the ones in TEC were never declared because the functions already existed. That was a fun frustration as I worked on stuff.

See: https://central.tri.be/issues/40555#note-16

### DIFF
--- a/src/functions/template-tags/date.php
+++ b/src/functions/template-tags/date.php
@@ -103,14 +103,14 @@ if ( ! function_exists( 'tribe_get_display_end_date' ) ) {
 	 * Returns the event end date that observes the end of day cutoff
 	 *
 	 * @category Events
-	 * @param int    $event       (optional)
-	 * @param bool   $displayTime If true shows date and time, if false only shows date
-	 * @param string $dateFormat  Allows date and time formating using standard php syntax (http://php.net/manual/en/function.date.php)
-	 * @param string $timezone    Timezone in which to present the date/time (or default behaviour if not set)
+	 * @param int    $event        (optional)
+	 * @param bool   $display_time If true shows date and time, if false only shows date
+	 * @param string $date_format  Allows date and time formating using standard php syntax (http://php.net/manual/en/function.date.php)
+	 * @param string $timezone     Timezone in which to present the date/time (or default behaviour if not set)
 	 *
 	 * @return string|null Date
 	 */
-	function tribe_get_display_end_date( $event = null, $displayTime = true, $dateFormat = '', $timezone = null ) {
+	function tribe_get_display_end_date( $event = null, $display_time = true, $date_format = '', $timezone = null ) {
 		$end_date = tribe_get_end_date( $event, true, 'U', $timezone );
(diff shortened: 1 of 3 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not ')' | [Admin_List.php:284](https://github.com/moderntribe/the-events-calendar/blob/3850a833c65be091a064516f24c300d293b0d442/src/Tribe/Admin_List.php#L284)

src/functions/template-tags/date.php
```diff
@@ -146,25 +110,14 @@ function tribe_get_start_date( $event = null, $displayTime = true, $dateFormat =
 	 *
 	 * @return string|null Date
 	 */
-	function tribe_get_end_date( $event = null, $displayTime = true, $dateFormat = '', $timezone = null ) {
-		if ( is_null( $event ) ) {
-			global $post;
-			$event = $post;
-		}
+	function tribe_get_display_end_date( $event = null, $displayTime = true, $dateFormat = '', $timezone = null ) {
```
zbtirrell: a brand new function with a camelCase parameter.  gross.

borkweb: Ha! I copy-pasted the `tribe_get_end_date()` parameter list. I'll fix it.

barryhughes: I feel sick.

zbtirrell:
:hammer:

src/functions/template-tags/date.php
```diff
@@ -96,76 +96,29 @@ function tribe_get_end_time( $event = null, $dateFormat = '', $timezone = null )
...
-		if ( is_numeric( $event ) ) {
-			$event = get_post( $event );
-		}
+	function tribe_get_display_end_date( $event = null, $display_time = true, $date_format = '', $timezone = null ) {
+		$end_date = tribe_get_end_date( $event, true, 'U', $timezone );
+		$beginning_of_day = tribe_beginning_of_day( date( Tribe__Date_Utils::DBDATETIMEFORMAT, $end_date ) );
 
-		if ( ! is_object( $event ) ) {
-			return '';
+		if ( tribe_event_is_multiday( $event ) && $end_date < strtotime( $beginning_of_day ) ) {
```
barryhughes: If it was not a multiday event, would we not want to make this adjustment?

borkweb: The follow-up to this question is in Slack :D https://tribe.slack.com/archives/products-dev/p1449609622002735

The short answer is, no. We don't need to make that adjustment for the bug as described.

PR #486, opened 2015-12-09
Don't use the default country for new venues within the event editor | 42170

barryhughes:
The venue meta box (within the event meta box) was attempting to use the default country value, making it the odd man out amongst its fellow fields.

https://central.tri.be/issues/42170 but see also https://central.tri.be/issues/29366


borkweb:
:+1:

PR #487, opened 2015-12-09
Respect default venue

barryhughes:
Strangely, the default venue wasn't respected in the admin event editor (but was in the frontend submission form if CE was active). This change corrects that.

https://central.tri.be/issues/42170


borkweb:
:shipit:

PR #488, opened 2015-12-09
In WP 4.4, the $post_type is no longer an array but an object

borkweb:
Tested against WP 4.4, the `$post_type` variable is passed in as an object whereas in WP 4.3 it is passed in as an array where the `args` index holds the object we care about. This changeset supports both structures to prevent the fatal that was being thrown.

See: https://central.tri.be/issues/42220


barryhughes:
:seedling: :cow2:

PR #491, opened 2015-12-09
Make iCal generator respect the timezone settings | 42213

barryhughes:
Use the event-specific or sitewide timezone as appropriate when generating iCal data.

[C#42213](https://central.tri.be/issues/42213)


bordoni:
Sounds Reasonable :100:

PR #493, opened 2015-12-09
If cost holds non-numeric text, use 0 for the cost range

borkweb:
When the cost meta holds something other than a numeric string (parseable by our regex), there are a few array operations that throw notices. This changeset sets some defaults in the event that the regex fails.

See: https://central.tri.be/issues/42174


barryhughes:
:wolf:__/
'''"/  \ / \

PR #494, opened 2015-12-10
Allow 0-cost (free) events. Don't consider non-numeric as '0'

borkweb:
During 4.0, we started tossing out explicit 0-cost events when, in fact, we want them (and need them) identified separately so we can throw the `Free` label on events.

See: https://central.tri.be/issues/42173

### DIFF
--- a/src/Tribe/Cost_Utils.php
+++ b/src/Tribe/Cost_Utils.php
@@ -273,11 +273,11 @@ class Tribe__Events__Cost_Utils {
 			return array();
 		}
 
-		// remove any empty prices
-		$costs = array_filter( (array) $costs );
+		// make sure costs is an array
+		$costs = (array) $costs;
 
-		// If it's empty returns 0
-		if ( empty( $costs ) ) {
+		// If there aren't any costs, return a blank array
+		if ( 0 === count( $costs ) ) {
 			return array();
 		}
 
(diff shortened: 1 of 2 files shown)


src/Tribe/Cost_Utils.php
```diff
@@ -273,11 +273,11 @@ public function parse_cost_range( $costs, $max_decimals = null ) {
 			return array();
 		}
 
-		// remove any empty prices
-		$costs = array_filter( (array) $costs );
```
borkweb: This tosses out events with costs of `0`, which we need to retain. Hence the reason this is tossed out.

src/Tribe/Cost_Utils.php
```diff
@@ -290,8 +290,8 @@ public function parse_cost_range( $costs, $max_decimals = null ) {
 			if ( preg_match_all( '/' . $price_regex . '/', $cost, $matches ) ) {
 				$cost = reset( $matches );
 			} else {
-				$matches = array();
-				$cost = array( 0 );
+				$cost = array();
```
borkweb: Rather than defaulting strings to `0`, instead default them to a non-cost (an empty array).

src/Tribe/Cost_Utils.php
```diff
@@ -311,7 +311,7 @@ public function parse_cost_range( $costs, $max_decimals = null ) {
 
 		foreach ( $costs as $cost ) {
 			// Creates a Well Balanced Index that will perform good on a Key Sorting method
-			$index = str_replace( '.', '', number_format( str_replace( $this->get_separators(), '.', $cost ), $max ) );
+			$index = str_replace( array( '.', ',' ), '', number_format( str_replace( $this->get_separators(), '.', $cost ), $max ) );
```
borkweb: `number_format` adds commas. Let's make sure those don't exist in the index

bordoni:
:100:

PR #495, opened 2015-12-10
Better handling of internationalized slugs

barryhughes:
The venue and organizer CPT slugs can already be localized - but the end result with non-Roman alphabets like Greek is not always a happy one. This change corrects that, potentially paving the way for us to finally offer more complete support for Shavian.

[C#40385](https://central.tri.be/issues/40385)


bordoni:
:+1:

PR #500, opened 2015-12-11
Make it possible to order the admin events list by organizer/venue

barryhughes:
This change better isolates the logic for ordering by venue/organizer from the logic used to order by date (which doesn't always run, depending on the context) - letting users successfully order the admin events list by organizer/venue (such as if APM is enabled and has made those particular columns sortable).

[C#37286](https://central.tri.be/issues/37286)

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -164,7 +164,7 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 				do_action( 'log', 'multi_posttype', 'default', $query->tribe_is_multi_posttype );
 				add_filter( 'posts_fields', array( __CLASS__, 'multi_type_posts_fields' ), 10, 2 );
 				add_filter( 'posts_join', array( __CLASS__, 'posts_join' ), 10, 2 );
-				add_filter( 'posts_join', array( __CLASS__, 'posts_join_orderby' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


src/Tribe/Query.php
```diff
@@ -719,6 +687,86 @@ public static function posts_orderby( $order_sql, $query ) {
...
+			 * it being overwritten by Tribe__Events__Admin_List methods.
+			 *
+			 * @see Tribe__Events__Admin_List
+			 */
+			add_filter( 'posts_clauses', array( __CLASS__, 'posts_orderby_venue_organizer' ), 100, 2 );
+
+			/**
+			 * @deprecated since 4.0.2
+			 */
+			$join_sql = apply_filters( 'tribe_events_query_posts_join_orderby', $join_sql );
```
bordoni: What do you think of adding a `_doing_it_wrong` when we deprecate stuff like filters?

barryhughes: Yeah I was swithering about doing just that.

In the end I was reluctant to make it a 'noisy' deprecation because it feels like a hook name we may wish to restore and re-use in a context that isn't venue/organizer specific.

Happy to add it though if you think it's worth it?

barryhughes: Actually, that doesn't make too much sense - I guess we still need to draw attention to it. Will update.

bordoni:
:100:

PR #503, opened 2015-12-14
Avoid notice-level errs (set defaults) | 42156

barryhughes:
Avoid notice level errors caused by trying to access non-existent array elements. 

[C#42156](https://central.tri.be/issues/42156)


bordoni:
:crystal_ball:

PR #505, opened 2015-12-14
Explicitly set order for past event queries (ajax) | 42371

barryhughes:
This change ensures that a) the `$query->tribe_is_past` flag is set during past event requests made by ajax and b) the order is set to `DESC` (a recent-ish change to the default args used in `tribe_get_events()` requires this).

[C#42371](https://central.tri.be/issues/42371)


bordoni:
Looks good, I would only maybe change the new variables to 'is_' structures. But it's up to you. ☃

barryhughes:
I'll go with a merge - thanks man :+1:

PR #506, opened 2015-12-15
Mod the WHERE clause when fetching next/prev events to avoid exclusions

borkweb:
I've commented the code heavily, but the gist is - when we were fetching next/previous events, the query that was being generated excluded events that had identical times to the event being viewed. Additionally, sometimes events with duplicate times wouldn't show up in the single-event navigation at all.

This fix reworks the WHERE clause meta query for the date selection to add additional meta/post_id statements that aren't possible with the standard WP_Query parameters. It includes a heavily commented regex that I tried to make _very_ targeted to avoid accidental butchery.

This code includes tests.

See: https://central.tri.be/issues/41290

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -3931,7 +3931,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		 *
 		 * return string
 		 */
-		public function get_closest_event_where( $where_sql, $query ) {
+		public function get_closest_event_where( $where_sql ) {
 			// if we are in this method, we KNOW there is a section of the SQL that looks like this:
 			//     ( table.meta_key = '_EventStartDate' AND CAST( table.meta_value AS DATETIME ) [<|>] '2015-01-01 00:00:00' )
 			// What we want to do is to extract all the portions of the WHERE BEFORE that section, all the
@@ -4064,9 +4064,9 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 			 * @var WP_Post $post
 			 */
 			$args = (array) apply_filters( "tribe_events_get_{$mode}_event_link", $args, $post );
-			add_filter( 'posts_where', array( $this, 'get_closest_event_where' ), 10, 2 );
+			add_filter( 'posts_where', array( $this, 'get_closest_event_where' ) );
 			$results = tribe_get_events( $args );
-			remove_filter( 'posts_where', array( $this, 'get_closest_event_where' ), 10, 2 );
+			remove_filter( 'posts_where', array( $this, 'get_closest_event_where' ) );
 
 			$event = null;
 
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -3919,17 +3919,113 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		}
 
 		/**
-		 * Get a "previous/next post" link for events. Ordered by start date instead of ID.
+		 * Modify the WHERE clause of query when fetching next/prev posts so events with identical times are not excluded
+		 *
+		 * This method ensures that when viewing single events that occur at a given time, other events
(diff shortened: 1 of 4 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $query is never used | [Main.php:3934](https://github.com/moderntribe/the-events-calendar/blob/c548b3e862a783492b1b41fd02fa1250272d1f47/src/Tribe/Main.php#L3934)

src/Tribe/Main.php
```diff
@@ -3919,17 +3919,113 @@ public function isOrganizer( $postId = null ) {
...
+			// The end-game is to change the known SQL line (from above) into the following:
+			//
+			//  (
+			//    ( table.meta_key = '_EventStartDate' AND CAST( table.meta_value AS DATETIME ) [<|>] '2015-01-01 00:00:00' )
+			//    OR (
+			//      ( table.meta_key = '_EventStartDate' AND CAST( table.meta_value AS DATETIME ) = '2015-01-01 00:00:00' )
+			//      AND
+			//      table.post_id [<|>] POST_ID
+			//    )
+			//  )
```
barryhughes: Bear with me here because I'm only on :coffee: #1, but let's say we start with the following data:

```
ID  EventStartDate
1   2016-01-01 12:00:00
2   2016-02-10 12:00:00
3   2016-03-09 15:00:00
4   2016-03-09 15:00:00
5   2016-03-10 16:00:00
```

If we start with the user landing on event `#2` - this works well. The next event link will take them to event `#3`, then to event `#4` without skipping ahead to event `#5`. All good there - but what if the event data is then edited to the following (see the change in `#1`'s date/time):

```
ID  EventStartDate
1   2016-03-09 15:00:00
2   2016-02-10 12:00:00
3   2016-03-09 15:00:00
4   2016-03-09 15:00:00
5   2016-03-10 16:00:00
```

Again starting with a user visiting event `#2`, the first bit of the amended where clause will be matched (a date time greater than current) and we'll go to event `#3` and then `#4`, `#5` ... but we skip `#1`. 

Does that make sense and is it a concern for us to address here? Or am I just confused :confounded:

tests/functional/Tribe/Events/Navigation_Test.php
```diff
@@ -0,0 +1,59 @@
...
+		// your set up methods here
+	}
+
+	public function tearDown() {
+		// your tear down methods here
+		// then
+		parent::tearDown();
+	}
+
+	public function test_closest_event() {
```
barryhughes: Looks like we're setting up a linear set of events for this test - should we additionally simulate a pre-existing event having it's datetime amended (re my above comment)?

borkweb: I've added a `test_closest_event_non_linear()` method that sets up the data in the same way you suggested and it passes!

barryhughes: _Noyce._

src/Tribe/Main.php
borkweb: When the query is executed, the `ORDER BY` is set to sort by `EventStartDate` and then `post_id` (ASC/DESC depending on the state of next/prev), which causes ID 1 to be fetched in the last example.

tests/functional/Tribe/Events/Navigation_Test.php
```diff
@@ -0,0 +1,147 @@
...
+	 * The order should be:
+	 *
+	 *   ID  EventStartDate
+	 *   2   2015-12-01 12:00:00
+	 *   1   2015-12-02 15:00:00
+	 *   3   2015-12-02 15:00:00
+	 *   4   2015-12-02 15:00:00
+	 *   5   2015-12-03 16:00:00
+	 */
+	public function test_closest_event_non_linear() {
```
barryhughes: :+1:

PR #507, opened 2015-12-15
Avoid accidentally fetching event category #1 

barryhughes:
If $term is a boolean value we shouldn't cast it to an int and try to fetch that term. See also [PRO PR 248](https://github.com/moderntribe/events-pro/pull/248).

[C#42381](https://central.tri.be/issues/42381)


zbtirrell:
:+1:

PR #509, opened 2015-12-16
Don't include string-based costs when fetching max/min costs in range

borkweb:
This fix ensures that the min/max cost ranges only include numeric values while allowing single events to preserve their string-based cost formatting.

Cost range tests pass.

Related: https://github.com/moderntribe/events-filterbar/pull/54

See: https://central.tri.be/issues/39041


zbtirrell:
:+1:

PR #511, opened 2015-12-17
Only append the original orderby if it is an orderby post_date

borkweb:
This partial reversion/hybrid is a best-of-both-worlds of a few previous renditions of this code. This allows any non-post_date field to take precedence over the start/end date of the event _*except*_ the post_date, which largely doesn't matter for Events.

See: https://central.tri.be/issues/38331


bordoni:
Looks good :+1:

PR #513, opened 2015-12-21
Preserve HTML in month view tooltips

ggwicz:
Re [C#42556](https://central.tri.be/issues/42556)
This is me throwing my hat in the ring to try and fix our various excerpt-related issues that keep coming up on the forums – even after 4.0.1, 2, etc.

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -1254,7 +1254,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	 *
 	 * @param  WP_Post|int|null $post The Post Object|ID, if null defaults to `get_the_ID()`
 	 * @param  array $allowed_html The wp_kses compatible array
-	 *
+ 	 *
 	 * @return string|null Will return null on Bad Post Instances
 	 */
 	function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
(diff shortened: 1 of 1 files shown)


ggwicz:
Fixed the Central ticket reference.

ggwicz:
Re [C#42556](https://central.tri.be/issues/42556)

src/functions/template-tags/general.php
```diff
@@ -1344,8 +1322,24 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
 		 */
 		$excerpt_more = apply_filters( 'excerpt_more', ' [&hellip;]' );
 
-		// Now we actually trim it
-		$excerpt = wp_trim_words( $excerpt, $excerpt_length, $excerpt_more );
+		// Trim the excerpt. The following is just wp_trim_words() without its use of wp_strip_all_tags().
+		if ( strpos( _x( 'words', 'Word count type. Do not translate!', 'the-events-calendar' ), 'characters' ) === 0 && preg_match( '/^utf\-?8$/i', get_option( 'blog_charset' ) ) ) {
```
barryhughes: This approach is a little odd and although you've already linked to `wp_trim_words()`, where people can find the rationale, I feel like it makes sense to have the same explanatory note to translators here in our own method, since A) it means devs don't need to jump to that function to figure out why we're doing this and B) as we're using our own text domain, it may help translators to figure out what's going on here. We wouldn't want someone to accidentally translate `'words'`, for instance.

src/functions/template-tags/general.php
```diff
@@ -1344,8 +1322,24 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
...
+
+		if ( count( $words_array ) > $excerpt_length ) {
+			array_pop( $words_array );
+			$excerpt = implode( $sep, $words_array );
+			$excerpt = $excerpt . $excerpt_more;
+		} else {
+			$excerpt = implode( $sep, $words_array );
+		}
 
 		return wpautop( $excerpt );
```
barryhughes: Should we provide a filter here before returning the excerpt, basically echoing what `wp_trim_words()` does?

src/functions/template-tags/general.php
```diff
@@ -1344,8 +1322,24 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
 		 */
 		$excerpt_more = apply_filters( 'excerpt_more', ' [&hellip;]' );
 
-		// Now we actually trim it
-		$excerpt = wp_trim_words( $excerpt, $excerpt_length, $excerpt_more );
+		// Trim the excerpt. The following is just wp_trim_words() without its use of wp_strip_all_tags().
```
barryhughes: > The following is just wp_trim_words() __without its use of wp_strip_all_tags()__

So ... I realize we can't strip tags __and__ preserve HTML - but looking through the remainder of the function there doesn't seem to be anything to handle elements that are chopped in two. For example, let's say the cut off point is half way through the link here:

```html
Blah blah apples and <a href="...">look here</a> to learn about oranges.
                                       ^ 
                                 Cutoff point
```

Aren't we going to end up with an unclosed element?

PR #515, opened 2015-12-27
Change superclass for metaTest (spotfix)

barryhughes:
Run the `template-tags/metaTest.php` test without a class not found error.


bordoni:
:+1:

PR #516, opened 2015-12-27
Add note about PhantomJS to test instrucs (spotfix)

barryhughes:
Apart from some whitespace tidies, just adds a note about PhantomJS to steer folks in the right direction (lines 50-55).

### DIFF
--- a/tests.md
+++ b/tests.md
@@ -12,20 +12,20 @@ when Composer finished the update process (might take a while) set up your own [
 
 	vendor/bin/wpcept bootstrap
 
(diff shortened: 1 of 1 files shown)


bordoni:
Should we add this to all the add-ons?

barryhughes:
Hmm, all public ones maybe? Figured I'd start here and maybe longer term we'll eliminate the step to manually start it (phantomjs, I mean).

bordoni:
I mean, we internally need to be running these tests, so I guess it would be a good call to remove that step ASAP or Add that on all the add-ons. But for sure on the public ones is nice first move.

barryhughes:
Def :+1:

PR #517, opened 2015-12-29
Add helper to get venue URL (raw), plus tests | 26901

barryhughes:
Ref [C#26901](https://central.tri.be/issues/26901) - adds helper to obtain the venue URL (rather than just a formed HTML link).

### DIFF
--- a/src/functions/template-tags/venue.php
+++ b/src/functions/template-tags/venue.php
@@ -398,8 +398,8 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	 * @return string Formatted link to the venue website
 	 */
 	function tribe_get_venue_website_link( $post_id = null, $label = null ) {
-		$post_id = tribe_get_venue_id( $post_id );
-		$url     = tribe_get_event_meta( $post_id, '_VenueURL', true );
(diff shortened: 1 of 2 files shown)


src/functions/template-tags/venue.php
```diff
@@ -422,6 +422,22 @@ function tribe_get_venue_website_link( $post_id = null, $label = null ) {
...
 	/**
+	 * Returns the venue website URL related to the current post or for the optionally
+	 * specified post.
+	 *
+	 * @param int|null $post_id
+	 *
+	 * @return string
+	 */
+	function tribe_get_venue_website_url( $post_id = null ) {
+		return (string) tribe_get_event_meta(
```
bordoni: Can we make a note to talk about how to split function calls like this one in a Code Review meeting? It can get really weird, I mean we should decide on something, because some of us are doing X and others are doing Y, and on different cases.

bordoni:
All good, and loved the references in the tests.

src/functions/template-tags/venue.php
barryhughes: True! (cc @zbtirrell for next code review mtg :grin:)

PR #518, opened 2016-01-06
modified ical file name to use 11 digit hash

lucatume:
Ticket: https://central.tri.be/issues/39455

I've set the length of the has to 11 for the aestethic reason that the generated filename will not be cut with ellipsis in many browsers download windows.
I'm not that concerned about uniqueness as any system, being it a file, will attach a `(1)` or something to it to distinguish.


barryhughes:
:bread: :fork_and_knife:

PR #519, opened 2016-01-07
Fixed known range build and update

lucatume:
Ticket: https://central.tri.be/issues/33205

### DIFF
--- a/src/Tribe/Dates/Known_Range.php
+++ b/src/Tribe/Dates/Known_Range.php
@@ -31,7 +31,7 @@ class Tribe__Events__Dates__Known_Range {
 
 		$_stati = array( 'publish', 'private', 'protected' );
 		$_stati = apply_filters( 'tribe_events_known_range_stati', $_stati );
-		$stati  = "('" . implode( "'", $_stati ) . "')";
+		$stati  = "('" . implode( "','", $_stati ) . "')";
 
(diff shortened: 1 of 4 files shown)


src/Tribe/Dates/Known_Range.php
```diff
@@ -0,0 +1,150 @@
...
+	public function rebuild_known_range() {
+		global $wpdb;
+		remove_action( 'deleted_post', array( $this, 'rebuild_known_range' ) );
+
+		$earliest = strtotime( $wpdb->get_var( $wpdb->prepare( "
+				SELECT MIN(meta_value) FROM $wpdb->postmeta
+				JOIN $wpdb->posts ON post_id = ID
+				WHERE meta_key = '_EventStartDate'
+				AND post_type = '%s'
+				AND post_status IN ('publish', 'private', 'protected')
```
bordoni: We should allow people to hook to the post_status.

lucatume: Good idea, it was not supported on the original code but that's a good call. Notice I'm not using the stati when setting limit markers events to have a conservative approach: better to rebuild the known range on any possible limit event change than having to add an overlay of tracking limit events and stati.

bordoni:
:+1:

PR #522, opened 2016-01-15
Prevent fatal if import/continue screen is refreshed | 37186

barryhughes:
[C#37186](https://central.tri.be/issues/37186) - prevent a fatal/WSOD that can occur in some edge case-like scenarios.


borkweb:
:neckbeard:

PR #523, opened 2016-01-15
Replace "ical" and "events" (from the suggested filename for .ics downloads) | 38124

barryhughes:
Per [C#38124](https://central.tri.be/issues/38124) `{sitename}-{hash}.ics` is probably better than `ical-events-{hash}.ics` since even if we ignore the .ics extension the user is probably well aware that it is iCal format and that it contains events ;-)


borkweb:
I like it. :octopus:

PR #527, opened 2016-01-20
Same Slug Archive Dismissible

bordoni:
_Ref:_ [C#39617](https://central.tri.be/issues/39617)

### DIFF
--- a/src/resources/js/events-admin.js
+++ b/src/resources/js/events-admin.js
@@ -693,3 +693,33 @@ jQuery( document ).ready( function( $ ) {
 		return false;
 	} );
 } );
+
+
+/**
+ * Notice Dismiss
+ */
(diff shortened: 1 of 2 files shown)


src/resources/js/events-admin.js
```diff
@@ -693,3 +693,33 @@ jQuery( document ).ready( function( $ ) {
 		return false;
 	} );
 } );
+
+
+/**
+ * Notice Dismiss
+ */
+( function( $ ) {
```
barryhughes: Does it make sense to have this is common?

bordoni: I like that idea. I will change it up.

bordoni:
Updated.

borkweb:
:+1:

PR #530, opened 2016-01-25
Add Publicize to the CPT

bordoni:
_Ref:_ [C#42326](https://central.tri.be/issues/42326)


borkweb:
I'm making the assumption that you've tested this with Jetpack disabled and there aren't any notices. :+1:

PR #533, opened 2016-02-01
modified month event ordering to by start time

lucatume:
Ticket: https://central.tri.be/issues/41383
The events, on the Month view, were being sorted by menu order causing some mispositioning.


bordoni:
It looks good, but my question is, should we be worried that this change will bring people complaining because there were using this menu ordering?

lucatume:
yours is a good point but this change only do applies to the Month view and not to the other views that are sorted the way any `tribe_get_events` call works: by start date.
I've looked into the code and the `menu_order` is used to set an event to sticky.
My guess is a third party plugin is using that to do something on the events.

bordoni:
:100:

PR #539, opened 2016-02-04
Wrap the code to make a good usaged of Select2 with AJAX

bordoni:
_Ref:_ [C#41617](https://central.tri.be/issues/41617)

### DIFF
--- a/src/Tribe/Ajax/Select2.php
+++ b/src/Tribe/Ajax/Select2.php
@@ -80,12 +80,15 @@ class Tribe__Events__Ajax__Select2 {
 			'order' => 'ASC',
 			'paged' => $page,
 
-			's' => $_GET['q'],
-
 			// Speed up stuff
 			'update_post_meta_cache' => false,
 			'update_post_term_cache' => false,
 		);
+
+		if ( ! empty( $_GET['s'] ) ) {
+			// No scaping required since it's a WP_Query Argument
+			$query_args['s'] = $_GET['q'];
+		}
 
 		// Actually do the Search
 		$search = new WP_Query( $query_args );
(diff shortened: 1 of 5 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$current_attr' | [Main.php:1759](https://github.com/moderntribe/the-events-calendar/blob/194531bb1d78b5eabac2b644fb46e145f387ba87/src/Tribe/Main.php#L1759)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$current_attr' | [Main.php:1783](https://github.com/moderntribe/the-events-calendar/blob/194531bb1d78b5eabac2b644fb46e145f387ba87/src/Tribe/Main.php#L1783)

src/Tribe/Ajax/Select2.php
```diff
@@ -0,0 +1,213 @@
...
+	 * Response to Select2 to search Venues
+	 * This will produce a HTTP response with a JSON in the Select2 format
+	 *
+	 * @return HTTP wp_send_json
+	 */
+	public function search_venues() {
+		$response = (object) array(
+			'items' => array(),
+		);
+
```
borkweb: Should we bail here if `$_GET['q']` is empty?

bordoni: It never will be but, we can make sure it doesn't show a notice.

bordoni:
Just updated, we want to be able to have results without the search that's why we don't bail.

borkweb:
Sounds good! :cactus:

PR #544, opened 2016-02-09
Fixed the offset so it would take the difference of just one hour

kgneil:
Before the offset would be given and instead of giving the hour difference it was making it a 2 hour difference. What you were suppose to be taking away an hour from was adding an extra hour.

I found this error when trying to run the DST test.


barryhughes:
:+1:

PR #545, opened 2016-02-09
Hook to the_content on 9 rather than 10 for better plugin compatibility

borkweb:
So, I made this change and activated all of our plugins and loaded our single event page (with tickets) and nothing seemed awry.

Here's [the request](https://wordpress.org/support/topic/the-load_ecp_into_page_template-method-overrides-page-builder-content)

See: https://central.tri.be/issues/37991


barryhughes:
:+1:

PR #547, opened 2016-02-10
Fix/42756 mobile month selector compatibility

jesseeproductions:
_Ref:_ [#42756](https://central.tri.be/issues/42756)

Changed mobile month view selector to a TEC generated class instead:

.type-tribe_events

This will help with themes that might decide to remove it, here is why Genesis removed it:
Added .entry class to all content, in preparation for the potential absence of .hentry in a HTML5-flavoured Genesis that prefers Microdata over Microformats.

Tested in Genesis, Twenty Sixteen, Divi, X Theme, and Canvas


jesseeproductions:
ops gotta remove some of this first

PR #548, opened 2016-02-10
Esc for Regular Expression on Rewrite Terms

bordoni:
_Ref:_ [C#42789](https://central.tri.be/issues/42789)

### DIFF
--- a/src/Tribe/Rewrite.php
+++ b/src/Tribe/Rewrite.php
@@ -250,7 +250,7 @@ if ( ! class_exists( 'Tribe__Events__Rewrite' ) ) {
 				'today' => array( 'today', $tec->todaySlug ),
 				'day' => array( 'day', $tec->daySlug ),
 				'tag' => array( 'tag', $tec->tag_slug ),
-				'tax' => array( 'category', $tec->category_slug ),
+				'tax' => array( 'category-testing', $tec->category_slug ),
 				'page' => (array) 'page',
 				'all' => (array) 'all',
 				'single' => (array) Tribe__Settings_Manager::get_option( 'singleEventSlug', 'event' ),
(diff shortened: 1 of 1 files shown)


src/Tribe/Rewrite.php
```diff
@@ -250,7 +250,7 @@ public function get_bases( $method = 'regex' ) {
 				'today' => array( 'today', $tec->todaySlug ),
 				'day' => array( 'day', $tec->daySlug ),
 				'tag' => array( 'tag', $tec->tag_slug ),
-				'tax' => array( 'category', $tec->category_slug ),
+				'tax' => array( 'category-testing', $tec->category_slug ),
```
barryhughes: Is this a hangover from testing?

src/Tribe/Rewrite.php
```diff
@@ -487,11 +492,38 @@ public function remove_percent_placeholders( array $rules ) {
...
 		}
+
+		/**
+		 * A function to Add an \ before the required caracters in Regex
+		 * Avoid using it elsewhere
+		 *
+		 * @param  string $value
+		 * @return string
+		 */
+		public function _build_replace_esc_regex( $value ) {
```
barryhughes: Why the `_` prefix on the method name (ditto `_esc_regex_elements()`) ... might an `@internal` doc tag be better?

src/Tribe/Rewrite.php
```diff
@@ -487,11 +492,38 @@ public function remove_percent_placeholders( array $rules ) {
...
+		}
+
+		/**
+		 * Escapes the Required characters on a Regular Expression Word
+		 * Avoid using it elsewhere
+		 *
+		 * @param  string $base
+		 * @return string
+		 */
+		public function _esc_regex_elements( $base ) {
```
barryhughes: Seems really close to what [preg_quote()](http://php.net/manual/en/function.preg-quote.php) does - is there a reason not to use that (can we add a note in the docblock if so)?

src/Tribe/Rewrite.php
bordoni: Yes thanks!

src/Tribe/Rewrite.php
bordoni: I never used `preg_quote` good catch. :wink:

bordoni:
Thanks for the CR @barryhughes, just updated the code.

barryhughes:
yisss

PR #550, opened 2016-02-11
Feature/33491 cost range merge

lucatume:
Ticket: https://central.tri.be/issues/33491

This builds the base used to solve the ticket issue, review this first.

### DIFF
--- a/src/Tribe/Cost_Utils.php
+++ b/src/Tribe/Cost_Utils.php
@@ -391,7 +391,7 @@ class Tribe__Events__Cost_Utils {
 		$this->_current_original_cost_separator = empty( $matches[1][0] ) ? '.' : $matches[1][0];
 		$matches[0]                             = empty( $matches[0] ) ? $matches[0] : array_map( array(
 			$this,
-			'convert_decimal_separator'
+			'convert_decimal_separator',
 		),
 			$matches[0] );
 		$numeric_orignal_costs                  = empty( $matches[0] ) ? $matches[0] : array_map( 'floatval',
--- a/src/Tribe/Cost_Utils.php
+++ b/src/Tribe/Cost_Utils.php
@@ -8,7 +8,19 @@ if ( ! defined( 'ABSPATH' ) ) {
 	die( '-1' );
 }
 
+
 class Tribe__Events__Cost_Utils {
+
+	/**
+	 * @var string
+	 */
+	protected $_current_original_cost_separator;
+
+	/**
+	 * @var string
+	 */
+	protected $_supported_decimal_separators = '.,';
+
 	/**
 	 * Static Singleton Factory Method
 	 *
@@ -53,19 +65,22 @@ class Tribe__Events__Cost_Utils {
 	public function get_separators() {
 		/**
 		 * Allow users to create more possible separators, they must be only 1 char
+		 *
 		 * @var array
 		 */
 		return apply_filters( 'tribe_events_cost_separators', array( ',', '.' ) );
 	}
 
 	public function get_cost_regex() {
-		return apply_filters( 'tribe_events_cost_regex', '(([\d]+)[\\' . implode( '\\', $this->get_separators() ) . ']?([\d]*))' );
+		return apply_filters( 'tribe_events_cost_regex',
+			'(([\d]+)[\\' . implode( '\\', $this->get_separators() ) . ']?([\d]*))' );
 	}
 
 	/**
 	 * Check if a String is a valid cost
 	 *
-	 * @param  string  $cost String to be checked
+	 * @param  string $cost String to be checked
+	 *
 	 * @return boolean
 	 */
 	public function is_valid_cost( $cost, $allow_negative = true ) {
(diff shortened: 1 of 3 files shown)


src/Tribe/Cost_Utils.php
```diff
@@ -338,4 +338,78 @@ public function parse_cost_range( $costs, $max_decimals = null ) {
...
+		$merging_cost               = (array) $merging_cost;
+		$numeric_merging_cost_costs = array();
+		foreach ( $merging_cost as $c ) {
+			if ( is_numeric( $c ) ) {
+				$numeric_merging_cost_costs[] = $c;
+			}
+		}
+
+		$matches = array();
+		preg_match_all( '!\d+(?:\.\d+)?!', $original_string_cost, $matches );
```
barryhughes: Given `$10.99 - $59.95` looks like we'd get two matches back - `10.99` and `59.95` - but if the original string cost was `€10,99 - €59,95` we'd get four matches - `10`, `99`, `59` and `95`. Is that a concern or is that handled upstream/elsewhere?

src/Tribe/Cost_Utils.php
```diff
@@ -338,4 +338,78 @@ public function parse_cost_range( $costs, $max_decimals = null ) {
...
+	 * @param array              $sorted_maxs          An array of non numeric price maximums sorted smaller to larger,
+	 *                                                 e.g. `['Donation min $10', 'Donation min $20', 'Donation min
+	 *                                                 $100']`.
+	 *
+	 * @return string|array The merged cost range.
+	 */
+	public function merge_cost_ranges( $original_string_cost, $merging_cost, $with_currency_symbol, $sorted_mins = array(), $sorted_maxs = array() ) {
+		if ( empty( $merging_cost ) || $original_string_cost === $merging_cost ) {
+			return $original_string_cost;
+
```
barryhughes: Extra line we can remove!

src/Tribe/Cost_Utils.php
```diff
@@ -338,4 +338,78 @@ public function parse_cost_range( $costs, $max_decimals = null ) {
 
 		return (array) $ocost;
 	}
+
+	/**
+	 * @param       string       $original_string_cost A string cost with or without currency symbol,
+	 *                                                 e.g. `10 - 20`, `Free` or `2$ - 4$`.
+	 * @param       array|string $merging_cost         A single string cost representation to merge or a an array of
```
barryhughes: Typo: _or a an array_ should probably be _or an array_

tests/functional/Tribe/Events/Cost_UtilsTest.php
```diff
@@ -0,0 +1,110 @@
...
+	public function test_merge_cost_ranges_into( $expected, $original_string_cost = '', $merging_cost = '', $with_currency_symbol = false, $sorted_mins = array(), $sorted_maxs = array() ) {
+
+		add_filter( 'tribe_currency_symbol',
+			function () {
+				return '$';
+			},
+			100 );
+		add_filter( 'tribe_reverse_currency_position', '__return_false');
+
+		$sut = $this->make_instance();
```
barryhughes: Curious - why `$sut` as the var name here?

tests/functional/Tribe/Events/Cost_UtilsTest.php
```diff
@@ -0,0 +1,110 @@
...
+	 * @test
+	 * it should be instantiatable
+	 */
+	public function it_should_be_instantiatable() {
+		$sut = $this->make_instance();
+
+		$this->assertInstanceOf( 'Tribe__Events__Cost_Utils', $sut );
+	}
+
+	public function merge_cost_ranges_into_ranges() {
```
barryhughes: Is it worth covering values with a decimal point (and different formats of decimal points) and should we cover use of thousands separators?

tests/functional/Tribe/Events/Cost_UtilsTest.php
lucatume: Subject Under Test

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Comma required after last value in array declaration | [Cost_Utils.php:394](https://github.com/moderntribe/the-events-calendar/blob/beb7c84d3218f09688b8497efc834a5e80005085/src/Tribe/Cost_Utils.php#L394)

lucatume:
I've added tests and handling code for the cases you pointed out.
I will take 15' to look into the thousands separator and see what can be made of it.
All good points yours Barry, thanks.

lucatume:
Ok, there is no easy way I can think of to handle the intricacies of the thousands separator.
A way can be found but not in time budget; I've tried the [`number_format_i18n` function](https://codex.wordpress.org/Function_Reference/number_format_i18n) but that requires numbers to be numbers already.  
Sadly on PHP 5.2 the [NumberFormatter](http://php.net/manual/en/class.numberformatter.php) class is not available.
I'm open to any idea.

barryhughes:
Hmm, not sure.

Maybe as a minimum we document the fact (that we don't handle a mixture of thousands separators and decimals) either as a remark in one of the docblocks or as a failing test that we skip as something to possibly address at some future point?

Totally trust your judgement.

lucatume:
I've added some skipping tests, this is more of a memo for us than something users can know but not control.

barryhughes:
Perfect, I think that leaves us in a reasonable spot :+1:

PR #555, opened 2016-02-15
removed updated hentry from meta details start time

lucatume:
Ticket https://central.tri.be/issues/42530

I've removed the `.updated` class anywhere it appeared near the start time meta details


borkweb:
It looks like this is pull requested against the wrong branch.

PR #565, opened 2016-02-18
added handling of currency prefixed merging costs

lucatume:
Ticket: https://central.tri.be/issues/33491

This handles merging costs coming with currency. EB differs in return values...


lucatume:
P.S.: I've run the tests and those are still green.

src/Tribe/Cost_Utils.php
```diff
@@ -449,4 +450,12 @@ private function filter_numeric_values( $value ) {
 	private function restore_original_decimal_separator( $value ) {
 		return str_replace( '.', $this->_current_original_cost_separator, $value );
 	}
+
+	private function numerize_numbers( $value ) {
```
borkweb: Can you add a docblock to this method to describe what it does?

lucatume: Ok, will do.
For future reference: are `private` and `protected` methods too to be documented?

borkweb:
:+1: Looks good!

PR #566, opened 2016-02-19
Fix/37494 filter events only

lucatume:
Ticket: https://central.tri.be/issues/37494#note-20

While working on solving the ticket not above I've noticed many query filters are running taking the fact that the query is on the tribe_events post types for granted; this would block or interfere with accessory queries made on different post types (e.g. the query made for tickets while filtering the events list by tickets).

The alternative solution is to suppress_filters on all accessory queries: it works but it's not flexible and does not solve the problem.


borkweb:
:+1:

PR #567, opened 2016-02-19
fixed early call to Date_Utils

lucatume:
Ticket: https://central.tri.be/issues/42077

Small adjustments to get CORE functional tests back to green


borkweb:
:turtle:

PR #574, opened 2016-02-23
Restore PR 519 (fix known range build and update) | #33205

barryhughes:
Seems [this work](https://github.com/moderntribe/the-events-calendar/pull/519/commits) was also lost to weird merge issues. This PR restores it (the test cases added in PR 519 survived so didn't need to be restored).

[C#33205](https://central.tri.be/issues/33205)


lucatume:
OK, merging the PR as it's needed to fix a number of issues

PR #576, opened 2016-02-24
Create a DarkMagic to solve the_content madness

bordoni:
_Ref:_ [C#44253](https://central.tri.be/issues/44253)

### DIFF
--- a/src/Tribe/Templates.php
+++ b/src/Tribe/Templates.php
@@ -280,18 +280,50 @@ if ( ! class_exists( 'Tribe__Events__Templates' ) ) {
 
 			if ( self::is_main_loop( $query ) && self::$wpHeadComplete ) {
 
+				/**
+				 * To solve all the Filters problems happening on `the_content` hook we saw the need to free the default
+				 * space used before from the new priority (5) and the old default (10) to prevent HTML bugs but also
+				 * allow other Content Managers to control the output on The Events Calendar pages
+				 */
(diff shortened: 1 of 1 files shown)


src/Tribe/Templates.php
```diff
@@ -280,18 +280,50 @@ public static function setup_ecp_template( $query ) {
...
+				$hooks = array();
+			}
+
+			foreach ( $hooks as $priority => $hook ) {
+				if ( $priority <= $start || $priority > $end ) {
+					continue;
+				}
+				unset( $GLOBALS['wp_filter'][ $hook_name ][ $priority ] );
+
+				$GLOBALS['wp_filter'][ $hook_name ][ $priority - $start ] = $hooks[ $priority ];
```
borkweb: It feels like this oversteps some boundaries.  This seems like it could cause some real confusion and difficulties for other plugins/devs - especially when it comes to other plugins hooking to `the_content` after this method executes.

If priority `9` isn't working out, then maybe shifting back to `10` is the simple solution and accept some compatibility problems with some plugins. [C#37991](https://central.tri.be/issues/37991) is where the request to lower to 9 was made - one user vs a pile of others doesn't seem like a good trade-off for destructively adjusting hook priorities.

bordoni:
Ready to CR again. :wind_chime:

borkweb:
:black_square_button:

PR #578, opened 2016-02-24
rebuild known range when event goes over boundaries

lucatume:
Ticket: https://central.tri.be/issues/33205

Rebuild the known date range anytime an event goes over boundaries


borkweb:
:ok:

PR #579, opened 2016-02-24
Display Organizer Name if Pro is not active

jesseeproductions:
_Ref:_ [C#44285](https://central.tri.be/issues/44285)

If there is no pro then we should call the original function and return the Organizer Name only 

I believe this will enable us to keep the changes made for this ticket:

https://central.tri.be/issues/29941


borkweb:
:+1:

PR #581, opened 2016-02-24
Improve our Core JS to also look for fields on the FilterBar

bordoni:
_Ref:_ [C#41662](https://central.tri.be/issues/41662)


borkweb:
:racehorse: :dash:

PR #582, opened 2016-02-25
Make our Default Methods respect the Default options a little bit better

bordoni:
_Ref:_ [C#42470](https://central.tri.be/issues/42470)

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -3073,12 +3073,12 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 				 * Used to Filter the default value for a Specific meta
 				 *
 				 * @since 4.0.7
-				 * @var $default
+				 * @var $value
 				 * @var $id
 				 * @var $meta
 				 * @var $single
 				 */
-				$value = apply_filters( 'tribe_get_meta_default_value_' . $filter, $default, $id, $meta, $single );
+				$value = apply_filters( 'tribe_get_meta_default_value_' . $filter, $value, $id, $meta, $single );
 			}
 			return $value;
 		}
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -3042,10 +3042,20 @@ protected function dateHelper( $date ) {
...
+			// If the post doesn't exist just bail the get_post_meta
+			if ( is_string( $status ) && 'auto-draft' !== $status ) {
+				$value = get_post_meta( $id, $meta, $single );
+			} else {
+				$value = false;
+			}
+
 			if ( $value === false ) {
-				$method = str_replace( '_Event', '', $meta );
+				$method = str_replace( array( '_Event', '_Organizer', '_Venue' ), '', $meta );
```
borkweb: There's a few duplicated `$metaTags`/`$venueTags`/`$organizerTags` that - when the prefix is stripped off of all three types - will result in identical filters below. Example:

* `_EventURL`
* `_VenueURL`
* `_OrganizerURL`

When stripped of the prefix, they all become just `URL`. Then the filter on line 3059 will be `filter_eventsDefaultURL` for all three, giving none of them distinction.

Is that intended?

bordoni: I don't think so, didn't think of that angle. might need to refactor the defaults class

src/Tribe/Main.php
```diff
@@ -3042,11 +3042,43 @@ protected function dateHelper( $date ) {
...
+				/**
+				 * Used to Filter the default value for a Specific meta
+				 *
+				 * @since 4.0.7
+				 * @var $default
+				 * @var $id
+				 * @var $meta
+				 * @var $single
+				 */
+				$value = apply_filters( 'tribe_get_meta_default_value_' . $filter, $default, $id, $meta, $single );
```
borkweb: Should `$default` change to `$value` here?

bordoni:
Ready for CR again :crystal_ball:

borkweb:
:octopus:

PR #585, opened 2016-02-25
Add back month view daily event fitler

jesseeproductions:
Add back in month daily event filter that got removed along with readme.txt update for it

_Ref:_ [C#38137](https://central.tri.be/issues/38137)


borkweb:
:+1:

PR #586, opened 2016-02-25
Change the license key save and check to the singular option instead …

jesseeproductions:
_Ref:_ [C#43736](https://central.tri.be/issues/43736)

Since the license keys are saved as their own individual option this change is to just detect and save it to that option per plugin instead of the serialized main option. 

This change is in Core and Common


barryhughes:
:+1:

PR #593, opened 2016-02-26
Fix description wrap in month mobile and tooltip

jesseeproductions:
_Ref:_ [C#42701](https://central.tri.be/issues/42701)

The excerpt comes with p tags and our wrap was p tags so it would close our wrap , changed to a div to prevent this.


borkweb:
:train:

PR #601, opened 2016-03-11
Let's do gulp in 4.1 so we don't have to wait for 4.2

borkweb:
These are cherry-picked commits from the `develop` branch. Let's go ahead and move this change forward.


zbtirrell:
should also grab the commit that updated CONTRIBUTING.md

borkweb:
Ah, thank you. Added.

barryhughes:
Good move!

PR #520, opened 2016-01-07
Prevent strtotime warning for `new DateTime()` obj

iamhexcoder:
Prevents PHP warning produced from passing DateTime() object into `start_date` or `end_date` args of `tribe_get_events`

PR #602, opened 2016-03-11
Fix handling of datetime objects passed in custom event queries | #44736

barryhughes:
Resolves a problem introduced in [PR 520](https://github.com/moderntribe/the-events-calendar/pull/520) which saw us trying to access the `date` property of `DateTime` objects, which isn't actually a real thing.

[C#44736](https://central.tri.be/issues/44736)


borkweb:
Whoopsie. That's my bad. :+1:

barryhughes:
Yeah I missed it too, good catch by Sydney though :+1:

PR #604, opened 2016-03-14
Separate function return and object reference into two lines for PHP 5.2

borkweb:
See: https://central.tri.be/issues/44790


bordoni:
:+1:

PR #605, opened 2016-03-14
Support costs that begin with a decimal

borkweb:
This modifies the regex to allow for the cost separators to _begin_ the cost regex in addition to appearing in the middle. This _may_ cause issues with costs entered as `0.0.0`, but I'm willing to deal with that problem later if a user raises it in the interest of keeping the regex simpler.

_Note: This passes our functional tests for cost ranges_

See: https://central.tri.be/issues/44711


bordoni:
:+1:

PR #606, opened 2016-03-15
Feature/39038 csv import excerpt

lucatume:
Ticket: https://central.tri.be/issues/39038

This PR deals with the excerpt import.
Of note is some refactoring on the tests to extract useful methods from the featured image related test code.


src/Tribe/Importer/File_Importer_Events.php
```diff
@@ -229,4 +230,14 @@ private function translate_terms_to_ids( array $terms ) {
 		return $term_ids;
 	}
 
+	private function get_post_excerpt( $event_id, $import_excerpt ) {
+		if ( $event_id ) {
+			$post_excerpt = get_post( $event_id )->post_excerpt;
```
bordoni: What would happen if a non valid ID got here?

lucatume: `event_id` would be `false` at that moment

bordoni:
:+1:

PR #609, opened 2016-03-15
If there is no default country, return an empty string | #44829

barryhughes:
In the absence of a default country, let's return an empty string.

[C#44829](https://central.tri.be/issues/44829)


borkweb:
:ok:

PR #612, opened 2016-03-16
Feature/39038 csv import boolean fields

lucatume:
Tciket: https://central.tri.be/issues/39038

This PR adds support for event boolean fields:
* allow comments
* allow trackbacks
* hide from event listing
* sticky in month view


barryhughes:
:+1:

PR #615, opened 2016-03-16
Fix/41383 month order sticky

lucatume:
I had the default ordering switched from menu_order to date in the Month view but not taking menu_order into account will break sticky events.
This fix will sort events by menu_order and then date in the Month view.
Please keep in mind that the database used in the ticket has events with menu_order values all over the place (see http://glui.me/?i=nnlskdxlexoa8j7/2016-03-16_at_12.21.png/) and this code modification will restore the issue the user has in this particular instance.

I'm not aware of any of our plugins setting the menu order to anything else but 0 or -1 but I might simply ignore the fact. If it's like that then the issue might be a third party plugin modifying the menu order.

What should we do?

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -45,6 +45,10 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 				$can_inject = false;
 			}
 
+			if (isset($query->query_vars['eventDoNotInjectDate']) && $query->query_vars['eventDoNotInjectDate']){
+				$can_inject= false;
+			}
+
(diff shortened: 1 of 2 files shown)


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *Current Release*, but this was pull requested against *release/4.1.1*.

lucatume:
Ticket: https://central.tri.be/issues/44863

borkweb:
This was auto closed when the release/4.1.1 branch was removed.

PR #618, opened 2016-03-17
Hotfix to resolve breaking changes with the German lang file

borkweb:
While updating the german .mo/.po files, I went ahead and grabbed all of the latest language files.

See: https://central.tri.be/issues/44920 and https://central.tri.be/issues/44917


bordoni:
:+1:

PR #619, opened 2016-03-18
If $_VenueVenue holds the default value of 0, don't fill the input value

borkweb:
The `Tribe__Events__Default_Values` class returned a default ID of 0 if there was no default venue specified. Let's do two things:

1. Let's avoid placing non-truthy values in the field.
1. Let's not return 0 as a default venue ID - null is more appropriate if there isn't a venue

See: https://central.tri.be/issues/44826


barryhughes:
:+1:

PR #622, opened 2016-03-21
Bring back the default country array

borkweb:
This was removed in #609, however, it is relied on by Pro. The exected data stored for countries is the country id (i.e. US) and the country name (i.e. United States). When the default country is blank, let's provide the expected array.

Then, in `tribe_get_event_country()`, make sure we handle the array and return the country _name_.

See: https://central.tri.be/issues/44929


barryhughes:
:+1:

PR #627, opened 2016-03-24
Only set the Timezone Abbr and UTC dates if they aren't already set

borkweb:
Related: https://github.com/moderntribe/events-importer-ical/pull/65

See: https://central.tri.be/issues/43743


barryhughes:
:ok_woman:

PR #628, opened 2016-03-24
Don't quote TZID in DTSTART/END

borkweb:
The iCal spec doesn't call for it and it breaks some parsers (including the one we use in our iCal importer)

See: https://central.tri.be/issues/42795


barryhughes:
:ok_hand:

PR #630, opened 2016-03-24
Remove the Duplication

bordoni:
_Ref:_ [C#42572](https://central.tri.be/issues/42572)


barryhughes:
Hey man, I see from the ticket the problem was showing up in the context of the CE submission form and possibly other related pages - but if we merge this change it will kill any before/after HTML on regular event pages when the Default Event Template is in use.

bordoni:
Yeah... You are right, looking at that.

bordoni:
I removed the lines, now this became a _hotfix_, but it's worth merging. It was throwing a notice.

borkweb:
:+1:

PR #632, opened 2016-03-26
Improve the Organizer Metabox to allow public template on community

bordoni:
_Ref:_ [C#39026](https://central.tri.be/issues/39026)

### DIFF
--- a/src/Tribe/Admin/Organizer_Chooser_Meta_Box.php
+++ b/src/Tribe/Admin/Organizer_Chooser_Meta_Box.php
@@ -55,17 +55,12 @@ class Tribe__Events__Admin__Organizer_Chooser_Meta_Box {
 		$this->render_dropdowns();
 		$this->render_add_organizer_button();
 
-		if ( class_exists( 'Tribe__Events__Community__Main' ) ) {
-			ob_start();
-			include Tribe__Events__Templates::getTemplateHierarchy( 'community/modules/organizer-multiple.php' );
-			$community_html = trim( ob_get_clean() );
-		}
-
-		if ( empty( $community_html ) ) {
-			include $this->tribe->pluginPath . 'src/admin-views/new-organizer-meta-section.php';
-		} else {
-			echo $community_html;
-		}
+		/**
+		 * Make this Template filterable, used for Community Facing templates
+		 *
+		 * @var string $file_path
+		 */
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$community_html' | [Organizer_Chooser_Meta_Box.php:67](https://github.com/moderntribe/the-events-calendar/blob/efb5733c438ce99383ec449af05109e07f2a6c7e/src/Tribe/Admin/Organizer_Chooser_Meta_Box.php#L67)

src/Tribe/Admin/Organizer_Chooser_Meta_Box.php
```diff
@@ -50,7 +54,18 @@ protected function get_event( $event = null ) {
 	public function render() {
 		$this->render_dropdowns();
 		$this->render_add_organizer_button();
-		include $this->tribe->pluginPath . 'src/admin-views/new-organizer-meta-section.php';
+
+		if ( class_exists( 'Tribe__Events__Community__Main' ) ) {
```
barryhughes: Wondering: should core need to be aware of Community's existence in this way and could the same result be achieved by making the template path filterable (which Community could then take advantage of)?

bordoni: My only concern is introducing a Filter here only for this specific template, I would love to add something a little bit more broad but the current way the `getTemplateHierarchy` works would prevent me from doing it. I mean we have it duplicated on Tickets and Core, I'm not sure what would be the best approach here.

I think we could go for the `tribe_events_template_src/admin-views/new-organizer-meta-section.php` and filter it to be `community/modules/organizer-multiple.php`, but I felt this approach would look cleaner.

borkweb: If someone held me at NERF-gun point and asked if I'd prefer a filter or baking in awareness to Events Community, I'd probably say "filter". Luckily, nobody is pointing a NERF gun at me suction cup-tipped or otherwise...so I'm safe, guys. Don't worry. But, yeah. +1 for filter.

bordoni: :older_man: :gun:

bordoni:
Updated the PR

borkweb:
:+1:

PR #634, opened 2016-03-28
Allow imports to save mappings for each different type

jamesgol:
I was testing out some stuff and got tired of resetting these mappings each time I changed an import type.  This changes the option name to include the type, so you can switch between venue, event, and organizer without having to reset field mappings.

Also adds Tribe__Events__Importer__File_Importer::get_type() which might be useful for something.

### DIFF
--- a/src/Tribe/Importer/Admin_Page.php
+++ b/src/Tribe/Importer/Admin_Page.php
@@ -317,7 +317,7 @@ class Tribe__Events__Importer__Admin_Page {
 			return false;
 		}
 
-		update_option( 'tribe_events_import_column_mapping', $column_mapping );
+		update_option( 'tribe_events_import_column_mapping_' . $importer->get_type(), $column_mapping );
 		return true;
 	}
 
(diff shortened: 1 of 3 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Space found before comma in function call | [columns.php:16](https://github.com/moderntribe/the-events-calendar/blob/aef15fca6c6be62264957a3958c78fe872ac12bc/src/io/csv/admin-views/columns.php#L16)

borkweb:
Howdy! Thank you for this pull request! We are _definitely_ interested in this changeset, but it was pull requested against our `release/4.1.1` branch (which has just been deleted). Would you mind pull requesting this against `develop`?

jamesgol:
I will submit another PR when I have some free time, but you might consider changing the directions in README.md to specify that they should be against develop (which seems sane) instead of the latest release

borkweb:
For sure. I've already done that :)

PR #639, opened 2016-04-04
Load l10n using new Helper

bordoni:
_Ref:_ [C#36246](https://central.tri.be/issues/36246)


borkweb:
:+1:

Add a changelog entry and then this is mergeable.

PR #643, opened 2016-04-06
For some weird reason the dimissal method was removed

bordoni:
_Ref:_ [C#45010](https://central.tri.be/issues/45010)


borkweb:
:+1:

PR #644, opened 2016-04-06
Removing unneeded hook

borkweb:
While looking into the WP 4.5 breaking changes surrounding admin pages for tags, I discovered that our one hook to `load-edit-tags.php` didn't do anything of note.

The associated functions attempt to add a `post_type` argument to the term cloud links on the `edit-tags.php` page for our Event post type. I commented out the hook and then rolled back my WP through the 4.x versions and all the way back to 3.9 - the earliest that we claim to support - and the `post_type` argument was still present.

As such...let's deprecate the two functions and get rid of the hook rather than keeping useless code around.

See: https://central.tri.be/issues/45431

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -514,7 +514,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 
 			add_action( 'load-tribe_events_page_' . Tribe__Settings::$parent_slug, array( 'Tribe__Events__Amalgamator', 'listen_for_migration_button' ), 10, 0 );
 			add_action( 'tribe_settings_after_save', array( $this, 'flushRewriteRules' ) );
-			add_action( 'load-edit-tags.php', array( $this, 'prepare_to_fix_tagcloud_links' ), 10, 0 );
+
 			add_action( 'update_option_' . Tribe__Main::OPTIONNAME, array( $this, 'fix_all_day_events' ), 10, 2 );
 
 			// Check for a page that might conflict with events archive (but we only need to do that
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -4334,8 +4334,10 @@ public function addHelpAdminMenuItem() {
 		 * When the edit-tags.php screen loads, setup filters
 		 * to fix the tagcloud links
 		 *
+		 * @deprecated 4.1.2
 		 */
 		public function prepare_to_fix_tagcloud_links() {
+			_deprecated_function( __METHOD__, '4.1.2' );
```
barryhughes: `__CLASS__ . '::' . __METHOD__`? Just a thought, I don't care that much :smile:

src/Tribe/Main.php
```diff
@@ -4355,6 +4359,7 @@ public function prepare_to_fix_tagcloud_links() {
 		 * @return string
 		 */
 		public function add_post_type_to_edit_term_link( $link, $term_id, $taxonomy, $context ) {
+			_deprecated_function( __METHOD__, '4.1.2' );
```
barryhughes: Optionally throw in the __CLASS__ here too (if you also do it in `prepare_to_fix_tagcloud_links()` that is).

src/Tribe/Main.php
borkweb: `__METHOD__` includes the class already, so this one would be: `Tribe__Events__Main::prepare_to_fix_tagcloud_links`

borkweb: I think you may be thinking of `__FUNCTION__`?

barryhughes: I think you're right!

barryhughes:
Then merge, sir!

PR #645, opened 2016-04-06
Added featured image size and link filters.

fabianmarz:
Added ability to override the size and link parameters of the featured image without the need for a whole template override.

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -844,11 +844,11 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 			$post_id = get_the_ID();
 		}
 
-		$image_html     = get_the_post_thumbnail( $post_id, $size );
+		$image_html     = get_the_post_thumbnail( $post_id, apply_filters( 'tribe_event_featured_image_size', $size ) );
 		$featured_image = '';
 
 		//if link is not specifically excluded, then include <a>
-		if ( ! empty( $image_html ) && $link ) {
+		if ( ! empty( $image_html ) && apply_filters( 'tribe_event_featured_image_link', $link ) ) {
 			$featured_image .= '<div class="tribe-events-event-image"><a href="' . esc_url( tribe_get_event_link() ) . '">' . $image_html . '</a></div>';
 		} elseif ( ! empty( $image_html ) ) {
 			$featured_image .= '<div class="tribe-events-event-image">' . $image_html . '</div>';


src/functions/template-tags/general.php
```diff
@@ -844,11 +844,11 @@ function tribe_event_featured_image( $post_id = null, $size = 'full', $link = tr
 			$post_id = get_the_ID();
 		}
 
-		$image_html     = get_the_post_thumbnail( $post_id, $size );
+		$image_html     = get_the_post_thumbnail( $post_id, apply_filters( 'tribe_event_featured_image_size', $size ) );
```
barryhughes: It would be awesome if we could document this new hook :smile:

src/functions/template-tags/general.php
```diff
@@ -844,11 +844,11 @@ function tribe_event_featured_image( $post_id = null, $size = 'full', $link = tr
...
 		$featured_image = '';
 
 		//if link is not specifically excluded, then include <a>
-		if ( ! empty( $image_html ) && $link ) {
+		if ( ! empty( $image_html ) && apply_filters( 'tribe_event_featured_image_link', $link ) ) {
```
barryhughes: Again, it would be appreciated if we could add inline docs for this new hook :smile:

barryhughes:
Looks great, thanks for the PR @fabianmarz!

If you have time it would be really appreciated if you could add [inline documentation](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters) for those new hooks (it doesn't need to be too in depth, but adding this means we can potentially pick these up when building our technical documentation - which is refreshed each release cycle).

src/functions/template-tags/general.php
barryhughes: One other thought (also applies to the second hook you added): could you add `$post_id` as one of the vars we make available to callbacks? That way themers could modify the size only for specific posts, etc.

barryhughes:
Ticketed: [C#45603](https://central.tri.be/issues/45603)

src/functions/template-tags/general.php
fabianmarz: Yes, good catch! For the post thumbnail filter it would result in the following, because we need to pass two args to `the get_the_post_thumbnail` function I need to rely on `call_user_func_array`. Is there another (better) way to do it? 

```php
$image_html = call_user_func_array( 'get_the_post_thumbnail', apply_filters( 'tribe_event_featured_image_size', $post_id, $size ) );

function tribe_event_featured_image_size($post_id, $size) {
    return [$post_id, 'full'];
}
```

PR #649, opened 2016-04-07
wp_strip_all_tags on the lang var

borkweb:
wp_kses is expensive and there's no case where even partial HTML in the lang get var makes sense.

Follow up to: https://central.tri.be/issues/45516


bordoni:
Nice.

PR #652, opened 2016-04-13
JSON-LD new Structure

bordoni:
_Ref:_ [C#43595](https://central.tri.be/issues/43595)

_**Note:**  This PR is on Hold because I need to make sure I deprecated everything in the correct way_


bordoni:
@barryhughes as per your questions on moderntribe/event-tickets#167 about the `get_class( $this )` solution. If you do that on PHP 5.2 it just returns `false`. I managed to remove the ghost param, since it's a static method you don't need ot comply by the the abstract rules, unless you are dealing with an abstract method.

bordoni:
Rdy for CR

borkweb:
:+1:

PR #654, opened 2016-04-14
Datepicker automation - need help

rmacdon465:
See inline comments for the problems, but basically having trouble getting a hold of the select values.


rmacdon465:
Wrong area for PR

PR #655, opened 2016-04-14
Bug/32937 pre next nav

trishasalas:
https://central.tri.be/issues/32937


barryhughes:
Remember to add in a link back to the Central ticket ;-)

src/views/day/nav.php
```diff
@@ -13,13 +13,13 @@
 	die( '-1' );
 } ?>
 
-<h3 class="tribe-events-visuallyhidden"><?php esc_html_e( 'Day Navigation', 'the-events-calendar' ) ?></h3>
+<h3 class="screen-reader-text" tabindex="0"><?php esc_html_e( 'Day Navigation', 'the-events-calendar' ) ?></h3>
```
barryhughes: I could totally be showing my ignorance of various conventions here, but I don't think we have any rules of our own that take care of 'hiding' `.screen-reader-text` ... so my question is - outside of the usual set of standard WP themes - is it safe to assume most other themes have us covered here? Or should we also cover this in our own CSS?

trishasalas:
You're right and thanks for catching that.  I had it in there but lost it somehow in all my testing.

barryhughes:
:+1:

PR #656, opened 2016-04-14
Datepicker tests - need some help :)

rmacdon465:
Ticket: https://central.tri.be/issues/46515

This PR adds acceptance tests to make sure that the date picker format chosen in Events > Settings > Display > Datepicker Date Format is used and honoured on the site front-end.

Rework of previously done work. Please note that I've assumed the purpose of the tests was to check the FE and as such the date picker date format option is not set using the UI on the back-end but accessing the option directly using the WPDb module.

I've cut down a lot of redundancy and gone for something much simpler.


rmacdon465:
Tests now working.  Changing to code review.

lucatume:
Hell YEAH!

rmacdon465:
Hell YEAH is right!  Thanks for your help dude!

tests/wpwebdriveracceptance/DatePicker3Cept.php
```diff
@@ -0,0 +1,175 @@
...
+<?php
+
+$I = new WpwebdriveracceptanceTester($scenario);
+$I->wantTo('Test the datepicker cases');
+
+//Login to backend
+$I->amOnPage("/wp-login.php");
+$I->fillField('Username', 'test_user');
+$I->fillField('Password','test_user');
+//$I->fillField("#user_login", "admin");
```
barryhughes: Remove this and the next line if they're no longer needed?

tests/wpwebdriveracceptance/DatePicker3Cept.php
```diff
@@ -0,0 +1,175 @@
...
+//$I->fillField("#user_pass", "admin");
+$I->click('Log In');
+
+$I->wait(5);
+
+$output = '';
+
+$result = '';
+
+$array = array("2016-01-15","1/15/2016","01/15/2016","15/1/2016","15/01/2016","1-15-2016","01-15-2016","15-1-2016","15-01-2016");
```
barryhughes: Whitespace! Be nice to add space after opening and before closing parentheses, and after commas :smile:

tests/wpwebdriveracceptance/DatePicker3Cept.php
```diff
@@ -0,0 +1,175 @@
...
+$i = 0;
+
+foreach($array as $ar){
```
barryhughes: Whitespace! We'd normally format this as `foreach ( $array as $ar ) {` ... also, consider switching the var names? `$array` and `$ar` are very generic, `$date_formats` and similar could be nice here.

barryhughes:
I won't annotate each line, but whitespacing throughout could do with some tweaks to match our coding standards.

barryhughes:
On a similar note, though I realize you're working with a framework that prefers `snakeCase`, we almost always aim to use `snake_case` in our own code.

tests/wpwebdriveracceptance/DatePicker3Cept.php
```diff
@@ -0,0 +1,175 @@
...
+	$I->amOnPage('/events/today/');
+
+	$jQueryDay = 'jQuery(\'#tribe-bar-date\').val()';
+
+	$frontdate = $I->executeJS('return jQuery("#tribe-bar-date").val()');
+	
+	$I->wait(2);
+
+	//For Date Format Y-m-d
+	if(validateDate($option, 'Y-m-d') == 1){
```
barryhughes: Switch order to 'yoda-style' ie `( 1 == validateDate( $option, 'Y-m-d' ) )`

barryhughes: It looks as thouhg this and the next 8 if blocks essentially do the same thing - only the date format changes. How about we put the different formats in an array and then use a single foreach block instead? It could really shorten the length of this file and, if we want to change something, we'd change it once instead of in 9 separate locations.

bordoni:
I like how this looks, may I ask if we should add comments or docblocks on tests?

lucatume:
I would not add comments where what's happening is clear enough. 
Any place where you think the code is not self explaining?

barryhughes:
Mostly asking out of ignorance - but each of the 9 test methods are basically identical. 

Why is that preferred over either a single test method that uses a `foreach()` to traverse the formats array, or over much thinner individual methods that simply pass the format index/values to a 'fat' method that does the work? 

(Just seems like a lot of code duplication ... but if there's a testing-world reason for this I'd be super interested to hear.)

lucatume:
Using a loop would mean that if case #4 fails you will have one test, one assertion and one failure: which one did fail?
With 9 tests you get 9 tests, 9 assertions, 8 successes and 1 failure which is way more ablative and helping (think of reading the report in a Jenkins output).
Test code should go for readability, clarity and isolation over efficiency.

borkweb:
Let's get this merged.

lucatume:
Thanks Matt.

PR #662, opened 2016-04-20
Bug/32935 past events color

trishasalas:
https://central.tri.be/issues/32935


src/views/month/single-event.php
```diff
@@ -193,6 +193,6 @@
 ?>
 
 <div id="tribe-events-event-<?php echo esc_attr( $event_id ); ?>" class="<?php tribe_events_event_classes() ?>" data-tribejson='<?php echo esc_attr( tribe_events_template_data( $post ) ); ?>'>
-	<h3 class="tribe-events-month-event-title"><a href="<?php echo esc_url( $link ) ?>" class="url"><?php echo $title ?></a></h3>
+	<h3 class="tribe-events-month-event-title"><span><a href="<?php echo esc_url( $link ) ?>" class="url"><?php echo $title ?></a></span></h3>
```
zbtirrell: Was this extra span necessary, or could this style have been applied directly to the `<a>`?

trishasalas: It was necessary.  I first tired to apply the text decoration to the link but a previously declared `text-decoration=none` had a higher specificity.  I also tried using a pseudo element but that was just messy! 

It is admittedly a bit hacky and I'm not in love withe the code although I do (personally) like the visual it creates.  

There are enough other a11y issues and this is minor enough that if the span is a problem I don't mind revisiting this in a larger audit/scope.

trishasalas: ...or maybe I can remove the span and apply the line-through to the <h3> tag instead ;p

barryhughes:
* Not a big problem but we should avoid committing assets that haven't been changed (I see one `.css` file that changed but many `.min.css` files)
* Should the change made in `css/tribe-events-theme.css` have been applied to `postcss/tribe-events-theme.pcss` (and then be compiled _back_ to vanilla CSS)? 
* Do we need a matching rule in `postcss/tribe-events-full.pcss`?

barryhughes:
:+1:

PR #663, opened 2016-04-21
Fix/45081 js event names for filter bar

jesseeproductions:
_Ref:_ [C#45081](https://central.tri.be/issues/45081)

This is a fix in core to resolve an issue in Filter Bar due to change in event names in Filter Bar. 

Please see this ticket for a better explanation: [C#46262](https://central.tri.be/issues/46262)

I believe it is better for this ticket to start moving core to the newer standard and then complete that as part of this new ticket.


bordoni:
👍

PR #664, opened 2016-04-22
Timezone is one word

lucatume:
Ticket: https://central.tri.be/issues/46169

Not the biggest PR in history.


jesseeproductions:
:+1:

PR #666, opened 2016-04-22
Fix Importer due to change in WordPress 4.5

jesseeproductions:
_Ref:_ [C#46286](https://central.tri.be/issues/46286)

In WordPress 4.5 they wrapped the post_search filter in a conditional for not having a suppress_filters aregument:

https://github.com/WordPress/WordPress/commit/e6d2b6cab0702a480efc7de0c63467a858f32f8b#diff-8e20e94124dba46643814ef699330737

I believe the change I made by adding suppress_filters=false to the query should not cause earlier versions of WordPress to not have problems. 

I did this in both places we use the post_search filter


barryhughes:
:+1:

PR #668, opened 2016-04-25
Feature/44827 gmaps key api field

lucatume:
Ticket: https://central.tri.be/issues/44827

Builds on https://github.com/moderntribe/tribe-common/pull/84 (<<< review this before).

This PR adds the field to set an optional Google Maps JS API key and uses it, if set, on the FE.

I think I've found all the places where the JS API is used but any other information is welcome.

### DIFF
--- a/src/Tribe/Google/Maps_API_Key.php
+++ b/src/Tribe/Google/Maps_API_Key.php
@@ -1,0 +1,88 @@
+<?php
+
+
+/**
+ * Class Tribe__Events__Google__Maps_API_Key
+ *
+ * Handles support for the Google Maps API key.
+ */
+class Tribe__Events__Google__Maps_API_Key {
+
+	/**
+	 * @var string
+	 */
+	protected $api_key_option_name = 'google_maps_js_api_key';
+
+	/**
+	 * @var static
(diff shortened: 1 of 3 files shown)


src/Tribe/Google/Maps_API_Key.php
```diff
@@ -0,0 +1,88 @@
...
+	public function filter_tribe_events_google_maps_api( $js_maps_api_url ) {
+		$key = tribe_get_option( $this->api_key_option_name );
+		if ( ! empty( $key ) ) {
+			$js_maps_api_url = add_query_arg( 'key', $key, $js_maps_api_url );
+		}
+
+		return $js_maps_api_url;
+	}
+
+	public function filter_tribe_events_pro_google_maps_api($js_maps_api_url  ) {
```
barryhughes: 1) Need to tweak spacing (after opening parentheses) 2) Why is the method body empty - if there's a legit reason can we annotate it in a comment?

lucatume: No reason, just forgot it.

barryhughes:
:+1:

PR #669, opened 2016-04-26
Fix the order in Month View

jesseeproductions:
_Ref:_ [C#46113](https://central.tri.be/issues/46113)

Change so Month View single days events are in this order Sticky Events, then All Day Events, then Events by Start Time.

meta_value_datetime was not ordering the dates correctly, not sure why, but the meta_value worked in my tests

### DIFF
--- a/src/Tribe/Template/Month.php
+++ b/src/Tribe/Template/Month.php
@@ -609,7 +609,7 @@ if ( ! class_exists( 'Tribe__Events__Template__Month' ) ) {
 
 			// this  will skip updating term and meta caches - those were already
 			// updated in $this->set_events_in_month()
-			//Order of Events Sticky Events, All Day Events, then Events by start time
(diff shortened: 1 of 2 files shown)


src/Tribe/Template/Month.php
```diff
@@ -609,6 +609,7 @@ private function get_daily_events( $date ) {
 
 			// this  will skip updating term and meta caches - those were already
 			// updated in $this->set_events_in_month()
+			//Order of Events Sticky Events, All Day Events, then Events by start time
```
barryhughes: I'm not sure the comment is accurate - as things stand wouldn't a multiday event (that began before whatever the current day is) trump an all day event? Might be worth simplifying to `'Ordered by start time, sticky events first'` or something like that.

barryhughes:
Good to merge, tagging `enhance` only re the wording in a comment.

PR #670, opened 2016-04-27
Improve performance of query and make use of a transient | #46403

barryhughes:
Abandon `count(*)` - it's too expensive - and make use of a transient when testing to see if we have uncosted events. 

[C#46403](https://central.tri.be/issues/46403)


lucatume:
Ok!

PR #671, opened 2016-04-28
Make constructor public (class isn't using the singleton pattern) | spotfix

barryhughes:
Looks like the constructor was set to protected - but this class isn't using the singleton pattern so we need to undo that.

(Spotfix)


lucatume:
Could this be the case to add the missing `instance` method?

barryhughes:
Possibly but, at this stage in the release cycle, I don't want to refactor unnecessarily.

lucatume:
GO!

PR #673, opened 2016-04-29
Add new filter to tribe_events_get_the_excerpt | #46477

ggwicz:
The `tribe_events_get_the_excerpt()` function does a lot, and it's not rare that we get folks on the forums to alter the excerpt output in some way. It's unfortunately not always as easy a process as it might seem, because if they want the excerpt to do something that a code-stripping or word-trimming function *within* the function will override, then the only option is to make a template override _just_ so that they can use a different excerpt-getting function.

The introduction of this filter should hopefully help to simplify this process.

[C#46477](https://central.tri.be/issues/46477)


bordoni:
@ggwicz don't forget to add the `code review` label. This is good for a merge.

PR #674, opened 2016-05-02
PHP compat re empty() | Spotfix 

barryhughes:
Changes a use of `empty()` so we examine a variable instead of a function's return value, for compatibility with PHPs 5.2 -> 5.4.


zbtirrell:
👍

PR #679, opened 2016-05-02
Refine uncosted events query | #46403

barryhughes:
Further improve the uncosted events query (we only want it to look at events, not unrelated post types).

[C#46403](https://central.tri.be/issues/46403)


borkweb:
:boom:

bordoni:
On this, can we hold a sec?

barryhughes:
> Do we see any value for a multiple Post Types query, for tickets in the future of Cost Utils?

I don't know of any plans to treat non-event posts as events (except for ET/ET+ - but they handle things differently).

PR #681, opened 2016-05-03
use https for our site URL

zbtirrell:
See: https://central.tri.be/issues/46542

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -80,7 +80,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		public $postOrganizerTypeArgs = array();
 
 		public static $tribeUrl = 'http://tri.be/';
-		public static $tecUrl = 'http://theeventscalendar.com/';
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -80,7 +80,7 @@ class Tribe__Events__Main {
 		public $postOrganizerTypeArgs = array();
 
 		public static $tribeUrl = 'http://tri.be/';
-		public static $tecUrl = 'http://theeventscalendar.com/';
+		public static $tecUrl = 'https://theeventscalendar.com/';
```
barryhughes: Just a thought (doesn't stop us from merging this change though) - but would an [m.tri.be](http://m.tri.be) link afford more insights/tracking potential?

zbtirrell: This is used as a base url to concatenate into various places, so m.tri.be wouldn't work. We could review those locations and switch it out, but that's more of a lift than I think is worth it at the moment.

PR #685, opened 2016-05-06
Make organizer fields sticky

barryhughes:
Makes the organizer fields 'sticky', to avoid data loss during frontend event submission (such as if the submission is initially rejected due to missing required fields, etc).

[C#44457](https://central.tri.be/issues/44457)


bordoni:
👍

PR #687, opened 2016-05-13
Feature/38132 antispambot email

lucatume:
Ticket: https://central.tri.be/issues/38132

This PR modifies the code used to print an organizer email on the FE to filter it through the `antispambot` function; this happens after filtering the email to avoid feeding filtering functions HTML entities where a string might be expected.

I've also removed the notice about obfuscation as we are now taking care of it.


src/functions/template-tags/organizer.php
```diff
@@ -148,9 +148,18 @@ function tribe_has_organizer( $postId = null ) {
...
 
-		return apply_filters( 'tribe_get_organizer_email', $output );
+		/**
+		 * Allows for the organizer email to be filtered.
+		 *
+		 * Please note that obfuscation of email is done in subsequent line using the `antispambot` function.
+		 *
+		 * @param string $email The organizer email.
+		 */
+		$filtered_email = apply_filters( 'tribe_get_organizer_email', $email );
```
barryhughes: Might it be better to either to use antispambot() itself as a `tribe_get_organizer_email` filter (so it can be unhooked) or else use it before the apply_filters() call and pass the original and unmodified email address as a 2nd param?

barryhughes:
Could we get a changelog entry for this new feature?

zbtirrell:
Note: this PR is for the Next Release, so I've tagged it with `hold`

tr1b0t:
:bangbang: This Pull Request is against a Central ticket for "Next Release". It shouldn't be merged until the Central ticket has been changed to "Current Release" or "CR Maintenance Release".

PR #688, opened 2016-05-13
update rewrite rule generation to use default view

lucatume:
Ticket: https://central.tri.be/issues/31907

This PR simply removes the hard-coded default view option from the rewrite rules (so it's frozen in the rewrite rules) to use the default one specified in each moment.


barryhughes:
Nice! Can we get a changelog entry?

bordoni:
After this gets fully code reviewed, keep in mind this is for NR, so don't merge it yet.

tr1b0t:
:bangbang: This Pull Request is against a Central ticket for "Next Release". It shouldn't be merged until the Central ticket has been changed to "Current Release" or "CR Maintenance Release".

PR #689, opened 2016-05-13
Remove deprecated methods | spotfix

barryhughes:
Removes two deprecated methods:

* Been deprecated since 4.0, seems like a good time
* They are marked protected, I'd be surprised if 3rd party code uses them
* If other code did/does use them, fatals will occur anyway as they are themselves trying to call protected methods


bordoni:
Good stuff.

PR #695, opened 2016-05-17
modified GMaps API requests to always run on SSL

lucatume:
Ticket: https://central.tri.be/issues/44827#note-16

Small tweak: all Google Maps API requests should hit the `https` site.


bordoni:
I think with this we might need to introduce a Error if the server doesn't allow SSL connections, can happen quite a bit. Was happening on EB.

lucatume:
The PR was born out of a note so I'd simply can it: sounds fine?
In other plugins we take care of looking up if the site is on SSL or not and then make the request.
I'd leave it as it is then, thanks @bordoni for the feedback.

barryhughes:
Merging this one in to keep things moving.

PR #696, opened 2016-05-17
Improve date-based pagination | #42704

barryhughes:
If a user switches from month view to list view, then pages forwards/backwards, we should preserve the date of the month they navigated from when calculating pagination results.

https://central.tri.be/issues/42704


borkweb:
:+1:

PR #700, opened 2016-05-24
added padding to main #tribe div, set bkg color to white and increase…

trishasalas:
…d font size of h1 to match theme

https://central.tri.be/issues/61502


bordoni:
👍

PR #704, opened 2016-05-25
Fix infinite loop caused by Maybe iCal Method

jesseeproductions:
_Ref:_ [C#46320](https://central.tri.be/issues/46320)

This fixes the block that Trisha found when trying to create a shortcode for the Month View. Using have_posts() in the maybe_add_link causes the query to be reset. Changed to checked against the global instead. 

Also, changed the method to be and action as it hooks into an action not a filter.


borkweb:
🏩

borkweb:
Can you add a changelog entry for this before you merge?

PR #705, opened 2016-05-25
update submodule to save me from submodule purgatory

trishasalas:


trishasalas:
https://central.tri.be/issues/61425

PR #709, opened 2016-06-01
Document new hook and add to changelog | #46477

barryhughes:
Document new hook (also make the post object accessible to callbacks) and add changelog entry.

https://central.tri.be/issues/46477


bordoni:
Great addition.

PR #710, opened 2016-06-01
Document new filter, add changelog entry | #46476

barryhughes:
Adds changelog entry and inline docs for new hook, also makes the original $record array available to callbacks.

https://central.tri.be/issues/46476


borkweb:
:+1:

PR #712, opened 2016-06-02
Fix for single/multiple linked post submission handling | spotfix

barryhughes:
Found this en route to fixing [#61871](https://central.tri.be/issues/61871) though this isn't directly related and [#61847](https://central.tri.be/issues/61847) may be a better fit: ensures when a new single linked post (ie a venue) is submitted it retains the correct field values.

Previously, `$key` would be `0` in these situations so if `$submission[ $field_name ] == 'String'` then `$submission[ $field_name ][ $key ]` would reference the first character, `S`, effectively lopping off the remainder.

https://central.tri.be/issues/61847


borkweb:
:+1:

PR #713, opened 2016-06-05
Ensure that submitted venue/organizer fields are arrays

borkweb:
Eventbrite imports events and formats venues/organizers with the old pre-linked post method...where the fields are not arrays. This changes ensures that the fields are all arrays as the new way expects. This change should help any third party submissions to saveEventMeta in the API class as well!

See: https://central.tri.be/issues/61942


barryhughes:
:+1:

PR #716, opened 2016-06-09
Ignored Events

bordoni:
As I was working on the Ignored Events for Event Aggregator, I noticed that our structure to deliver notices was lacking a few features and it was duplicated in a few places.

_Rel:_ moderntribe/events-pro#334 moderntribe/tribe-common#97
_Ref:_ [C#61231](https://central.tri.be/issues/61231)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | String concat is not required here; use a single string instead | [Ignored_Events.php:69](https://github.com/moderntribe/the-events-calendar/blob/18b13d1afca91b342a97eaf1eaefb69432bc9afc/src/Tribe/Ignored_Events.php#L69)
:x: | String concat is not required here; use a single string instead | [Ignored_Events.php:69](https://github.com/moderntribe/the-events-calendar/blob/18b13d1afca91b342a97eaf1eaefb69432bc9afc/src/Tribe/Ignored_Events.php#L69)
:x: | String concat is not required here; use a single string instead | [Ignored_Events.php:70](https://github.com/moderntribe/the-events-calendar/blob/18b13d1afca91b342a97eaf1eaefb69432bc9afc/src/Tribe/Ignored_Events.php#L70)

bordoni:
We can have a look, a this again. There a few pieces of code here that Are still not fully ready, but we need to merge this and to avoid conflicts on other tickets that will come from EA.

src/Tribe/Ignored_Events.php
```diff
@@ -0,0 +1,526 @@
...
+		public static $legacy_deleted_post = 'deleted_event';
+
+		public static $legacy_origin = 'ical-importer';
+
+		/**
+		 * Static singleton variable
+		 *
+		 * @var self
+		 */
+		public static $instance;
```
barryhughes: No need to keep this public.

src/Tribe/Ignored_Events.php
```diff
@@ -0,0 +1,526 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	die( '-1' );
+}
+
+if ( ! class_exists( 'Tribe__Events__Ignored_Events' ) ) {
+	/**
+	 * Ignored Events are fully powered by this class
+	 */
+	class Tribe__Events__Ignored_Events {
+		public static $ignored_status = 'tribe-ignored';
```
barryhughes: I know you're not a fan of class constants, but these _could_ be class constants instead vs public, overwriteable fields #justSaying

src/Tribe/Ignored_Events.php
```diff
@@ -0,0 +1,526 @@
...
+
+			if ( empty( $_GET['post_status'] ) || $_GET['post_status'] !== self::$ignored_status ) {
+				return false;
+			}
+
+			if ( ! $this->has_legacy_deleted_posts() ) {
+				return false;
+			}
+
+			$html = '<p>' . '@TODO: Include a Cool message about why you are seen this Notice!' . '</p>';
```
barryhughes: Should we resolve this `@todo` before merging?

barryhughes: ...If not, can we add it as a @todo comment above this line rather than having it within the string? At least in terms of my own local tools, it's more likely to be visible that way.

src/Tribe/Ignored_Events.php
```diff
@@ -0,0 +1,526 @@
...
+
+			if ( self::$ignored_status !== $event->post_status ) {
+				return $actions;
+			}
+
+			$origin = get_post_meta( $event->ID, '_EventOrigin', true );
+
+			foreach ( $actions as $key => $html ) {
+				/**
+				 * @todo  Add logic to prevent removal of edit on EA origin
```
barryhughes: Not sure if we'd need to resolve this `@todo` prior to merge or not ... can you confirm?

src/Tribe/Ignored_Events.php
```diff
@@ -0,0 +1,526 @@
...
+			$event = get_post( $event );
+
+			if ( ! $event instanceof WP_Post ) {
+				return false;
+			}
+
+			// Update only what we need
+			$arguments = array(
+				'ID' => $event->ID,
+				'post_status' => 'publish',
```
barryhughes: Just a thought: if you trash a WP post (that had a status of 'private') then restore it, it is set back to private rather than being published to the world. Should we follow suit here?

src/Tribe/Ignored_Events.php
```diff
@@ -0,0 +1,526 @@
...
+			wp_send_json( $response );
+		}
+
+		/**
+		 * Notes:
+		 *
+		 * Working on this I notice that because `pre_delete_post` only exists after WP 4.4 we have a problem in terms of preventing that post from been deleted.
+		 * My workout for this is to hoot on `admin_init` of that page, change the `$_REQUEST` variable holding the IDs to remove any posts that fit the `ignored`
+		 *
+		 * On another problem, I realized that not WordPress still not ready to display custom status on the Admin, how should I display if we have a custom status
```
barryhughes: Could be a good discussion for team CR on Wednesday - I've got a related ticket where ideally we'd also be able to render custom statuses.

PR #717, opened 2016-06-09
Fix/broken tests fix

lucatume:
Ticket: https://central.tri.be/issues/43817

This PR fixed the broken tests plus:

* rename the `functional` suite to `integration` (because that's integration)
* create a new `nojsacceptance` suite to run tests that do not really need JS (that cuts time by 10)
* refactor/update tests to use a database dump (`/tests/_data/dump.sql` as a starting point; while WordPress hardcodes urls in the database latest WPBrowser iteration will swap those URLs on the fly and give us an easy-peasy way to kickstart the tests; current dump is an empty site (as in `wp site empty --yes` empty) with just TEC activated and default options saved.


lucatume:
on hold as I still have to fix the returning slug in url issue

PR #718, opened 2016-06-14
fix tribe_get_event_link function when passing 2nd param as true 

zbtirrell:
See: https://central.tri.be/issues/61891

Note: we never call the function in this manner within our own code, this would only be used for theme-ing and customizations

Side note: I'm also sneaking in a submodule hash update to the latest common.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -290,16 +290,17 @@ At no point during the 3.0 lifecycle will the major version change. But you can 
 = [4.2.1] Unreleased =
 
 * Fix - replace bad return type to avoid notices in the error log [62376]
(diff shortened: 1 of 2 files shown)


common
```diff
@@ -1 +1 @@
-Subproject commit f48693dcb0581d3766ada970383484d0c3004df7
+Subproject commit 9427d5e45ffc8b16301a16629531adf7b3371c28
```
borkweb: Was this hash commit intended?

zbtirrell: yeah, I noted it in the PR description.

borkweb: Well, I guess I should read that.

borkweb:
:+1:

PR #727, opened 2016-06-17
is_embed fix for older versions of WordPress

jesseeproductions:
_Ref:_ [C#62577](https://central.tri.be/issues/62577) - We use is_embed 4 different times in this template: the-events-calendar/src/Tribe/Templates.php 

Instead of wrapping those with a version check I added one to the general.php file 

Not sure if this is the best approach, but this would help prevent fatals in WordPress 3.9 to WordPress 4.3

Method for commit is in common: https://github.com/moderntribe/tribe-common/pull/108

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -291,17 +291,20 @@ At no point during the 3.0 lifecycle will the major version change. But you can 
 
 * Fix - replace bad return type to avoid notices in the error log [62376]
 * Fix - add missing styles for screen reader text [62500]
-* Fix - prevent fatal error of is_embed on WordPress 4.3 and lower [62577]
+* Tweak - adjust the month view caching rules [46581]
+* Fix - tribe_get_event_link() didn't work properly when passing second parameter as true [61891]
+* Tweak - reduce font size of titles in month view for skeleton styles [41461]
+* Fix - added styling to adjust list view description to 100% [62512]
 
 = [4.2] 2016-06-08 =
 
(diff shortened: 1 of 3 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a space between if and the first parenthesis '(' | [general.php:1543](https://github.com/moderntribe/the-events-calendar/blob/2bc0b3ee5b2e4c715dd95730a029408d5dcc1561/src/functions/template-tags/general.php#L1543)
:x: | Opening brace must be the last content on the line | [general.php:1545](https://github.com/moderntribe/the-events-calendar/blob/2bc0b3ee5b2e4c715dd95730a029408d5dcc1561/src/functions/template-tags/general.php#L1545)

bordoni:
This could be problematic, lets say that someone else does the same, but having the return be true? I mean I would rather make a wrapper for the function and do all the required checks.

jesseeproductions:
No problem, but I am not sure what you are looking for can you explain or give an example?

bordoni:
I mean, having a method that checks for the version + if doesn't function exists and if both are true then we use `is_embed`. Use that method instead of using `is_embed`.

zbtirrell:
I'm actually OK with this approach. @borkweb what do you think?

barryhughes:
Tagging `enhance` - looks like work is still in progress here (ie, moving to using new method `Tribe__Templates::is_embed()`).

borkweb:
🚆

PR #728, opened 2016-06-17
Bug/62512 list view description not full width

trishasalas:
https://central.tri.be/issues/62512


barryhughes:
:last_quarter_moon_with_face:

barryhughes:
Woops, setting back to `enhance`: can we get a changelog entry in place unless it was added via a previous PR? Good to merge once done/if in place already.

PR #731, opened 2016-06-22
Use array instead of [] for PHP 5.2 compatibility

jesseeproductions:
_Ref:_ [C#62699](https://central.tri.be/issues/62699) This is late in the game, but the only change is to remove the short array syntax and use the 5.2 and 5.3 compatible array().


tr1b0t:

zbtirrell:
:+1:

PR #734, opened 2016-06-23
Update the-events-calendar-pt_BR.po

ruanjsb:
In Portuguese the adjective usually comes after the substantive. Specially in this case, to be used in the export button, if you put the adjective before the substantive, makes no sense.


bordoni:
Hi @ruanjsb, you can use the [Official repository GlotPress](https://translate.wordpress.org/locale/pt-br/default/wp-plugins/the-events-calendar) to translate to pt_BR?

On every release we fetch the string from that repository so anything changed here on GitHub will be lost.

PR #735, opened 2016-06-24
Add a base Aggregator class and Aggregator__Service class

borkweb:
This begins the integration of the client and EA service.

_Ref:_ [C#62321](https://central.tri.be/issues/62321)


src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,123 @@
...
+
+	/**
+	 * Static Singleton Factory Method
+	 *
+	 * @return Tribe__Events__Aggregator
+	 */
+	public static function instance() {
+		if ( ! isset( self::$instance ) ) {
+			$className      = __CLASS__;
+			self::$instance = new $className;
```
bordoni: We could use: `self::$instance = new self;`, for a one liner.

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,125 @@
...
+
+	/**
+	 * Performs a GET request against the Event Aggregator service
+	 *
+	 * @param string $url Endpoint URL
+	 * @param array $data Array of parameters to send to the endpoint
+	 *
+	 * @return stdClass
+	 */
+	public function get( $url, $data = array() ) {
```
bordoni: Could `$url` be changed to `$endpoint` to avoid us having to call multiple methods all the time. Or maybe have a register method to setup all the possible endpoints ?

src/Tribe/Aggregator.php
borkweb: Oh yeah. I just did copy pasta here without really reading what I copied.

src/Tribe/Aggregator/Service.php
borkweb: Registering all the endpoints once sounds nice.

borkweb: Actually, no. I'll switch to `$endpoint` and build the URL from `get()`. There are only a small handful of endpoints - pre-registering them when we'd still need to do concatenation on a few of them is probably overkill.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $data is never used | [Service.php:112](https://github.com/moderntribe/the-events-calendar/blob/9a9fe76627e9659a0b182c2fde3bb75d142967d9/src/Tribe/Aggregator/Service.php#L112)

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,121 @@
...
+	 * Set up hooks
+	 */
+	protected function hooks() {
+	}
+
+	/**
+	 * Registers Event Aggregator with PUE so the license field shows up
+	 */
+	protected function register_with_pue() {
+		$this->pue_checker = new Tribe__PUE__Checker( $this->pue_url, $this->pue_slug, array(), '' );
```
bordoni: So, I was wondering ~~(after all this years)~~ if we could avoid having the `pue_url` and `pue_slug` as attributes.
Let me explain why, I don't really see any reasoning to have these be something outside of the `Tribe_PUE_Checker`, since after registering we will not have any usage for the variables, even if we needed we could fetch them from the PUE class it self.

```php
public function get_pue() {
    $url = 'http://tri.be/';
    $slug = 'events-aggregator';
    return new Tribe_PUE_Checker( $url, $slug, array() );
}
```

Then change the construct to reflect these same variables:
```php
$this->pue = $this->get_pue();
```

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	/**
+	 * Get the event-aggregator license key
+	 *
+	 * @return string
+	 */
+	public function get_license_key() {
+		return get_option( $this->license_meta_key );
```
bordoni: Same thing here, why store this on a variable since we have a method that does exactly what anyone would, fetch the information. Could have a default value from `get_option`? What do you think?

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * Static Singleton Factory Method
+	 *
+	 * @return Tribe__Events__Aggregator__Service
+	 */
+	public static function instance( Tribe__Events__Aggregator $aggregator ) {
+		if ( ! isset( self::$instance ) ) {
+			$className      = __CLASS__;
+			self::$instance = new $className( $aggregator );
```
bordoni: Can we make it a oneliner?

src/Tribe/Aggregator.php
```diff
@@ -58,8 +58,7 @@ class Tribe__Events__Aggregator {
 	 */
 	public static function instance() {
 		if ( ! isset( self::$instance ) ) {
```
afragen: Just trying to understand this line. As $instance is a declared variable won't it always be set? It may not have a value but that's not the same thing.

bordoni: So `isset()` returns `false` when we are dealing with a variable with `value === null`, so it's a safer way to check for `is_null`, prevents notices.

afragen: I usually just declare the `private static $instance = false` and check ` if ( false === self::$instance )`

Has the advantage of one less function call. :+1:

bordoni: Very true, the best approach to this would be to have the following declaration:
```php
	/**
	 * Static Singleton Holder
	 *
	 * @var self
	 */
	private static $instance;

	/**
	 * Static Singleton Factory Method
	 *
	 * @return self
	 */
	public static function instance() {
		return null === self::$instance ? self::$instance = new self : self::$instance;
	}
```

afragen: I'm not sure all on a single line is easier to read. :wink: 

I still think it makes more sense to declare the var with a value, especially if we're checking for the value. Otherwise I would probably use `empty()` and not `isset()` as it's a bit more forgiving.

bordoni: We are going to revisit these singleton methods in the future to kinda create a Std across the whole plugin suite. In this one it's up @borkweb to change. I would say that I like the ternary here because we are going to copy this across all classes, so it needs to be as small as possible, and behave the same way, that's why we don't declare a value also (PHP does it for us `null`).

borkweb: Good chat around this. Once again this was another copy pasta. I typically do the following (avoiding the ternary for readability):

```php
public static function instance() {
  if ( ! self::$instance ) {
    self::$instance = new self;
  }

  return self::$instance;
}
```

src/Tribe/Aggregator.php
borkweb: I'm down with eliminating some properties for this. I'll approach this in a slightly different way than what was suggested, but it's a nice direction.

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,183 @@
...
+		$filename = sanitize_file_name( $filename );
+
+		// The file hasn't been uploaded yet. Save it!
+		$filepath = $upload_dir['path'] . DIRECTORY_SEPARATOR . $filename;
+
+		// get the file type
+		$filetype = wp_check_filetype( basename( $filepath ), null );
+
+		// save the file to the filesystem in the upload directory somewhere
+		file_put_contents( $filepath, $results['body'] );
```
bordoni: Shouldn't we use WordPress methods?

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,246 @@
...
+	 * @param string $url Endpoint URL
+	 * @param array $data Array of parameters to send to the endpoint
+	 *
+	 * @return stdClass
+	 */
+	public function get( $endpoint, $data = array() ) {
+		$url = $this->build_url( $endpoint );
+		$url = esc_url_raw( add_query_arg( $data, $url ) );
+
+		$response = wp_remote_get( $url );
```
bordoni: We should have some sort of Error treatment for these requests right? Like for when our services are down or something.

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,180 @@
...
+		$attachment = array(
+			'guid' => $upload_dir['url'] . '/' . $filename,
+			'post_title' => preg_replace( '/\.[^.]+$/', '', $filename ),
+			'post_content' => '',
+			'post_status' => 'inherit',
+			'post_mime_type' => $filetype,
+		);
+
+		// insert the attachment
+		$attachment_id = wp_insert_attachment( $attachment, $upload_results['file'] );
```
barryhughes: `wp_insert_attachment()` can return `(int) 0` or a `WP_Error` - could be worth checking for that?

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,246 @@
...
+		$response = $this->get( 'import/' . $import_id );
+
+		return $response;
+	}
+
+	/**
+	 * Creates an import
+	 *
+	 * @param array $args {
+	 *     Array of arguments. See REST docs for details. 1 excpetion listed below:
```
barryhughes: Typo! (Excpetion)

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,194 @@
...
+
+		$extension = str_replace( 'image/', '', $response['headers']['content-type'] );
+
+		if (
+			preg_match( '/filename="([^"]+)"/', $response['headers']['content-disposition'], $matches )
+			&& ! empty( $matches[1] )
+		) {
+			$filename = $matches[1];
+		} else {
+			$filename = md5( $results['body'] ) . '.' . $extension;
```
barryhughes: Where is `$results` coming from?

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		$filename = sanitize_file_name( $filename );
+
+		// get the file type
+		$filetype = wp_check_filetype( basename( $filepath ), null );
```
barryhughes: I don't think `$filepath` is defined before this point; should it be `$filename` or something else here?

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,194 @@
...
+		return apply_filters( 'tribe_ea_origins', $origins );
+	}
+
+	/**
+	 * Fetches an image from the service and saves it to the filesystem if needed
+	 *
+	 * @param string $image_id EA Image ID
+	 */
+	public function get_image( $image_id ) {
+		$tribe_ea_meta_key = 'tribe_ea_image_id';
```
barryhughes: Could be nice to make this a class constant, or have a getter-function for it?

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,194 @@
...
+			$origin->text = __( $origin->name, 'the-events-calendar' );
+		}
+
+		return apply_filters( 'tribe_ea_origins', $origins );
+	}
+
+	/**
+	 * Fetches an image from the service and saves it to the filesystem if needed
+	 *
+	 * @param string $image_id EA Image ID
```
barryhughes: Add a `@return` tag indicating what users can expect on success/failure?

src/Tribe/Aggregator.php
```diff
@@ -0,0 +1,194 @@
...
+		// fetch an image
+		$response = $this->service->get_image( $image_id );
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		// if the reponse isn't an image then we need to bail
+		if ( ! preg_match( '/image/', $response['headers']['content-type'] ) ) {
+			return $response;
```
barryhughes: Are we sure we want to return `$response` here vs a WP_Error?

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,255 @@
...
+			self::$instance = new self( $aggregator );
+		}
+
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor
+	 */
+	public function __construct( $aggregator ) {
```
barryhughes: Should be private or protected?

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,255 @@
...
+		$url = "{$this->api_base_url}{$this->api_root}{$this->api_version}/{$endpoint}";
+		$url = add_query_arg( 'key', $this->api_key, $url );
+
+		return $url;
+	}
+
+	/**
+	 * Performs a GET request against the Event Aggregator service
+	 *
+	 * @param string $url Endpoint URL
```
barryhughes: The var name is `$endpoint` in the signature (rather than `$url`) :smile:

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,255 @@
...
+			$response = json_decode( wp_remote_retrieve_body( $response ) );
+		}
+
+		return $response;
+	}
+
+	/**
+	 * Performs a POST request against the Event Aggregator service
+	 *
+	 * @param string $url Endpoint URL
```
barryhughes: As above, either the signature or the doctag here are out of sync (`$url` \ `$endpoint`).

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,255 @@
...
+	 * @param array $data Array of parameters to send to the endpoint
+	 *
+	 * @return stdClass
```
barryhughes: Returns a stdClass _or_ a WP_Error.

barryhughes: ...ditto the `get()` method and anywhere else where this applies :smile_cat:

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,255 @@
...
+		if ( $response && 'success' === $response->status ) {
+			$origins = array_merge( $origins, $response->data->origin );
+		}
+
+		return $origins;
+	}
+
+	/**
+	 * Fetch import data from service
+	 *
```
barryhughes: Document the `$import_id` param?

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,255 @@
...
+	 * @return array
```
barryhughes: This method can return a `WP_Error` or the result of `$response` (which is returns a `stdClass` object or `WP_Error` only, unless I'm misreading).

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,255 @@
...
+		if ( ! $this->api_key ) {
+			return new WP_Error( 'invalid-ea-license', __( 'You must enter an Event Aggregator license key in Events > Settings > Licenses', 'the-events-calendar' ) );
+		}
+
+		$request_args = array(
+			'body' => $args,
+		);
+
+		if ( isset( $args['source_file'] ) ) {
+			$boundary = wp_generate_password( 24 );
```
barryhughes: Probably showing my ignorance here, but is there a reason to go to the cost of regenerating this boundary each time vs a static string?

src/Tribe/Aggregator/Service.php
```diff
@@ -0,0 +1,255 @@
...
+		}
+
+		$response = $this->post( 'import', $args );
+	}
+
+	/**
+	 * Fetches an image from the Event Aggregator service
+	 *
+	 * @param string $image_id Image ID to fetch
+	 */
```
barryhughes: Add a `@return` tag?

bordoni:
I've fixed all the Code Review points, but I'm getting this Error, I never touched the PUE code; Could you take a look at that @barryhughes?
```
Notice: Undefined index: event-aggregator.php in /Volumes/Korhal/dev/localhost/tribe/stable/wp-content/plugins/event-tickets/common/src/Tribe/PUE/Checker.php on line 184
```

src/Tribe/Aggregator/Service.php
bordoni: It's not much of big problem to regenerate it, really inexpensive in terms of performance, but it could be a problem in terms of trying to split this string again later, an action that I can't see a reason to use 😁

barryhughes:
Hey @bordoni! So re the undefined index, note that a warning is also being generated as part of the same problem:

```
opendir(/var/.../public/wp-content/plugins/event-aggregator): failed to open dir: No such file or directory
```

This happens if there isn't a valid Aggregator PUE key in place (or if Aggregator isn't yet set up on the PUE server).

The thing is, PUE is built to deal with plugin updates and it has code to inspect the plugin header (via [get_plugins()](https://codex.wordpress.org/get_plugins)). In the case of Event Aggregator, there is no  plugin directory/header file to inspect. So I think we're going to need some amendments to the PUE client code to solve this. We should definitely proceed with caution since it's PUE we're talking about ;-)

bordoni:
Ok, lets talk about this on the dev meeting today?

bordoni:
As we talked on the Hangout, Im merging this one.

PR #739, opened 2016-06-27
Only add events to archive tag queries

jesseeproductions:
_Ref:_ [C#45635](https://central.tri.be/issues/45635)


src/Tribe/Query.php
```diff
@@ -64,15 +64,15 @@ public static function parse_query( $query ) {
...
 
-			// Add tribe events post type to tag queries
+			// Add tribe events post type to tag queries only in tag archives
 			if ( $query->is_tag && (array) $query->get( 'post_type' ) != array( Tribe__Events__Main::POSTTYPE ) ) {
 				$types = $query->get( 'post_type' );
 				if ( empty( $types ) ) {
 					$types = array( 'post' );
 				}
-				if ( is_array( $types ) ) {
+				if ( is_array( $types ) && is_tag() ) {
```
barryhughes: Using `is_tag()` here might trigger `_doing_it_wrong()` notices if a custom query runs early enough. 

What do you think about `$query->is_main_query()` instead? We already know (from the outer if block) that this is a tag query, so limiting it to the _main_ query would seem to have pretty much the same effect without the potential for notices.

barryhughes:
:+1:

PR #740, opened 2016-06-27
Process shortcodes when filter is active

jesseeproductions:
_Ref:_ [C#42289](https://central.tri.be/issues/42289)


src/functions/template-tags/general.php
```diff
@@ -1380,8 +1380,11 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
 			$excerpt = wp_trim_words( $excerpt, $excerpt_length, $excerpt_more );
 		}
 
-		// Remove all shortcode Content before removing HTML
-		if ( ! $allow_shortcode ) {
+		// If shortcode filter is enabled lets process them
+		if ( $allow_shortcode ) {
+			$excerpt = do_shortcode( $excerpt );
```
barryhughes: If I'm remembering correctly - I may not be ;-) - the reason we added code to strip anything that even _looks_ like a shortcode is because during ajax requests we return the response so early it's not uncommon for some shortcodes to not have been registered.

In this approach, we'd still run that risk (ie, if `[3rd_party_magic]` won't be registered until after the point where we return our response, then it still won't be processed and we'll be left with what looks like an unparsed shortcode in the output).

What about this: instead of an either/or approach, we could default to always running `do_shortcode()` first and stripping unparsed shortcodes out second. That way shortcodes that have been registered early enough would still be rendered, but we'd clean up the detritus of shortcodes that weren't already registered (and we could still use filters/hooks to allow fine grained control over this policy).

bordoni: I like that idea @barryhughes, and one more thing I think we could add a test to verify that we are not changing behaviors in the future.

src/functions/template-tags/general.php
```diff
@@ -1380,11 +1380,15 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
 			$excerpt = wp_trim_words( $excerpt, $excerpt_length, $excerpt_more );
 		}
 
-		// Remove all shortcode Content before removing HTML
-		if ( ! $allow_shortcode ) {
-			$excerpt = preg_replace( '#\[.+\]#U', '', $excerpt );
+		// If shortcode filter is enabled lets process them
+		if ( $allow_shortcode ) {
+			$excerpt = do_shortcode( $excerpt );
 		}
```
jesseeproductions: I changed it so the shortcode is processed if the filter is true, do we want to change it so we always process now? I believe that could create support issues. Anytime we change how the display of the excerpts here it always create support so I would side with only processing shortcodes if the filter is active and otherwise strip them out.

barryhughes: > do we want to change it so we always process now?

We probably don't want to always process - but what I'm saying is if we _do_ process, we should probably also strip out any unprocessed shortcodes that remain, for the reasons described in my previous comment.

That secondary step could itself be turned on and off with a filter, to cover any unusual situations where cleanup isn't needed or desired.

barryhughes: ...misread your latest change: that is indeed what you're doing so we're all good here. Optionally, we could add a filter to control whether cleanup (`preg_replace( '#\[.+\]#U', '', $excerpt )`) takes place or not.

src/functions/template-tags/general.php
```diff
@@ -1355,6 +1355,13 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
 		 */
 		$allow_shortcode = apply_filters( 'tribe_events_excerpt_allow_shortcode', false );
 
+		/**
+		 * Prevent the removal of shortcode markup in the Excerpt
```
barryhughes: I wonder if we should make it clear that this will remove anything that even _looks_ like a shortcode, just to help people understand possible reasons to use it (such as if they have things like _"blah blah [footnote 5]"_ in their copy)?

src/functions/template-tags/general.php
```diff
@@ -1355,6 +1355,13 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
...
+		 *
+		 * @var bool
+		 */
+		$remove_shortcode = apply_filters( 'tribe_events_excerpt_prevent_shortcode_removal', false );
```
barryhughes: It confuses me a little that the hook is named _*prevent* shortcode removal_ and defaults to false while down on line 1396 the conditional is inverted (ie, we *do* remove anything looking like a shortcode, if it is false).

readme.txt
```diff
@@ -289,9 +289,10 @@ At no point during the 3.0 lifecycle will the major version change. But you can
 
 = [4.2.2] unreleased =
 
+* Add - Add the filter tribe_events_excerpt_prevent_shortcode_removal to prevent removal of shortcode markup in the excerpt [42289]
 * Fix - Add a check to avoid being unable to switch view when Filter Bar plugin is active and at least one field is set to auto-complete mode
 * Fix - Only add the events custom post type to the tag archive queries and not custom queries with tags [45635]
-
+* Fix - When using the filter tribe_events_excerpt_allow_shortcode shortcodes will now be processes in event excerpts in the list views [42289]
```
barryhughes: '...will now be _processed'_ maybe?

barryhughes:
Let's merge :zap:

PR #744, opened 2016-06-29
Fix/62532 multiple sticky organizers

lucatume:
Ticket: https://central.tri.be/issues/62532

This PR:

* modifies the `Tribe__Events__Linked_Posts__Chooser_Meta_Box` to conditionally print a field for each candidate linked post
* adds a check in the JS code to avoid running the `.shift()` method on the globally defined linked posts candidate

Question for reviewer: at line `Chooser_Meta_Box.php:125` I've left a lingering question about where/when this additional parsing makes sense; it could always make sense but I could use a second pair of eyes.


src/Tribe/Linked_Posts/Chooser_Meta_Box.php
```diff
@@ -121,6 +121,9 @@ public function render_dropdowns() {
 		}
 
 		?><script type="text/template" id="tmpl-tribe-select-<?php echo esc_attr( $this->post_type ); ?>"><?php $this->single_post_dropdown( 0 ); ?></script><?php
+		
+		// @todo: add more check here for the context: when/where should we do this?
```
borkweb: This feels like the right place/time to do this to me. I can't think of any context where this _shouldn't_ be executed.

borkweb:
:+1:

src/Tribe/Linked_Posts/Chooser_Meta_Box.php
lucatume: Ok, removing the line then, thanks. Needed a second pair of eyes.

PR #729, opened 2016-06-20
Fix/46149 featured image upload

lucatume:
Ticket: https://central.tri.be/issues/46149

This PR:

* refactors some code in the CSV Importer to handle featured image upload
* handles an hedge case with featured image upload that would see it removed for existing events
* adds a test for reimport behaviour

PR #746, opened 2016-06-29
Bug/46149 featured image import

lucatume:
Ticket: https://central.tri.be/issues/46149

This PR is the right target version of this: https://github.com/moderntribe/the-events-calendar/pull/729

It refactors some code in the CSV Importer to handle featured image upload
handles an hedge case with featured image upload that would see it removed for existing events
adds a test for reimport behaviour


barryhughes:
:+1:

PR #750, opened 2016-07-01
Stop using the GLOB_BRACE flag

MZAWeb:
The `GLOB_BRACE` flag for `glob()` is not available on all systems. Most notably it's not supported on Solaris and Alpine (a quite popular lightweight linux distribution used mostly for running docker containers). On this PR I propose a way of achieving the same result which doesn't use such flag and hence improves compatibility.

Related: https://github.com/moderntribe/event-tickets/pull/227

See: https://central.tri.be/issues/63172


borkweb:
Thanks - this looks good! I'll merge it. :+1:

PR #751, opened 2016-07-01
Fix iCal export.

jlambe:
Hi,

I encounter an issue regarding the iCal Export template button from a single event view.

The errors I got is "Cannot modify header information - headers already sent by..." This issue appears because you have some code inside your `Tribe__Events__iCal` class that is called after some rendered output in the browser.

Here are the lines the error report points to:
```php
$filename = apply_filters( 'tribe_events_ical_feed_filename', $site . '-' . $hash . '.ics', $post );
header( 'Content-type: text/calendar; charset=UTF-8' );
header( 'Content-Disposition: attachment; filename="' . $filename . '"' );
```
This code is run once the hook `tribe_tec_template_chooser` is called which is too late.
By changing the hook name to a more appropriate one: `template_redirect`, no more errors were triggered and the plugin runs smoothly with the iCal export.

I suggest you to test this. So inside your `Tribe__Events__iCal` class in the static method init, replace the hook `tribe_tec_template_chooser` to `template_redirect`. Test it and let us know when there is an update to this. The fix is available in this PR.

Here we keep the altered plugin on our development but it would be best if it is fixed generally.

The error and fix is working on plugin version 4.2.1 and 4.2.1.1. It runs on PHP 5.6.

Here is a link to changed code (image): https://cloudup.com/c5wYbNiQfnz

The project is under local development so no access to it but if you need more information, do not hesitate to ask.

Regards,

[C#64141](https://central.tri.be/issues/64141)

### DIFF
--- a/src/Tribe/iCal.php
+++ b/src/Tribe/iCal.php
@@ -13,7 +13,7 @@ class Tribe__Events__iCal {
 	public static function init() {
 		add_action( 'tribe_events_after_footer', array( __CLASS__, 'maybe_add_link' ), 10, 1 );
 		add_action( 'tribe_events_single_event_after_the_content', array( __CLASS__, 'single_event_links' ) );
-		add_action( 'tribe_tec_template_chooser', array( __CLASS__, 'do_ical_template' ) );
+		add_action( 'template_redirect', array( __CLASS__, 'do_ical_template' ) );
 		add_filter( 'tribe_get_ical_link', array( __CLASS__, 'day_view_ical_link' ), 20, 1 );
 		add_action( 'wp_head', array( __CLASS__, 'set_feed_link' ), 2, 0 );
 	}


barryhughes:
Hi @jlambe,

Thanks first of all for taking the time to report this and to put together a pull request :+1: 

> "Cannot modify header information - headers already sent by..."

This sort of problem commonly happens when display of errors is turned on and some code somewhere triggers a notice or warning (it can of course also happen if text is flushed to the browser prematurely). It's not something I find occurs with a stock installation of WordPress and The Events Calendar, however, nor are users generally complaining of this problem.

For that reason I can't help but wonder if the true source of the problem is specific to your own site (and I'd be reluctant to merge this change if that were the case). Do you have any further information you can share - what is the full error you are seeing and do you find you can reproduce the problem from within a clean installation with nothing else running besides our plugin and a default theme?

Thanks for your help!

barryhughes:
Hi @jlambe,

Just checking in here - any thoughts on the above? If it _is_ looking like a conflict rather than a straight up bug in our code, please do feel free to post on either of our support forums should you need further assistance on that count.

* [Support for free users](https://wordpress.org/support/plugin/the-events-calendar)
* [Support for active license holders](http://theeventscalendar.com/support)

Thanks!

jlambe:
Hi,

sorry I haven't had the time to test this on a fresh installation yet.

On our project, WordPress is installed on its own folder and we do have a `symfony-httpfoundation` package in use regarding http responses. I'm sure this error is not triggered on a regular WordPress installation. For your records, we do output content after the `template_redirect` hook custom conditions.

Our setup shows that the custom hook you use for iCal export is called too late. You're not safe in cases where people already have some content output before your custom hook. The fact of using `template_redirect` makes sure your headers are set before any templating conditions are met and so any custom output. Especially when it comes to headers, I will eventually suggests to set priority to `0` with the hook so it can cover all customizations.

barryhughes:
Thanks for the further detail, @jlambe.

That's fair and there doesn't seem to be any particular need to keep things on the current hook. I'll get this ticketed and we'll look at merging it in :+1:

barryhughes:
Ticketed: [C#64141](https://central.tri.be/issues/64141)

jlambe:
👍  Awesome 😃

PR #757, opened 2016-07-12
Fix/62523 jsonld event dates

lucatume:
Ticket: https://central.tri.be/issues/62523

Filter the `timezone_string` when building the JSON-LD data to output correct start and end times.

### DIFF
--- a/src/Tribe/JSON_LD/Event.php
+++ b/src/Tribe/JSON_LD/Event.php
@@ -20,11 +20,6 @@ class Tribe__Events__JSON_LD__Event extends Tribe__JSON_LD__Abstract {
 	public    $type = 'Event';
 
 	/**
-	 * @var int
-	 */
-	protected $current_post_id;
-
-	/**
 	 * On PHP 5.2 the child class doesn't get spawned on the Parent one, so we don't have
 	 * access to that information on the other side unless we pass it around as a param
 	 * so we throw __CLASS__ to the parent::instance() method to be able to spawn new instance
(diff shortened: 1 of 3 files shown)


src/Tribe/JSON_LD/Event.php
```diff
@@ -17,7 +17,12 @@ class Tribe__Events__JSON_LD__Event extends Tribe__JSON_LD__Abstract {
 	 * @see https://developers.google.com/structured-data/rich-snippets/
 	 * @var string
 	 */
-	public $type = 'Event';
+	public    $type = 'Event';
+
+	/**
+	 * @var int
+	 */
+	protected $current_post_id;
```
bordoni: Question, couldn't apply the timezone on the `get_data` method? To avoid having to set and unset a filter plus a new propriety been set here?

lucatume:
the `parent::get_data` method is not fetching the date and since we are filtering something quite pervasive to the website and the plugin I'd rather keep the filtering localised.

src/Tribe/JSON_LD/Event.php
```diff
@@ -51,13 +51,16 @@ public function get_data( $posts = null, $args = array() ) {
 			}
 
 			// Fetch first key
-			$post_id = key( $data );
+			$post_id = $this->current_post_id = key( $data );
```
bordoni: We no longer need `current_post_id` right?

lucatume: Not needed at all, thanks for pointing it out.

bordoni:
☑️

PR #759, opened 2016-07-13
Bug/62399 google map settings

jesseeproductions:
_Ref:_ [C#62399](https://central.tri.be/issues/62399) - Enable some fields with linked posts to not be hidden. Specifically the Google Map CheckBoxes


src/resources/js/events-admin.js
```diff
@@ -285,7 +285,8 @@ jQuery( document ).ready( function( $ ) {
 			if ( dropdown.length ) {
 				var value = dropdown.val();
 				if ( 0 !== parseInt( value, 10 ) ) {
-					fields.hide();
+					//hide all fields, but those with not-linked class i.e. Google Map Settings
+					fields.not('.not-linked').hide();
```
borkweb: Add some spaces in the function call as per [WordPress JS coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#arrays-and-function-calls):

```js
fields.not( '.not-linked' ).hide();
```

borkweb:
The class `not-linked` doesn't seem descriptive enough to me to indicate what it means. Maybe it is too cryptic? Dunno. That being said, I'm having a hard time thinking of an alternate class name so I'm going to give this a 👍 . If you can think of another class name that is a bit more communicative of what it is attempting to represent, I'd be eternally grateful. If you end up in the same boat as me and come up stumped, then I'll understand if you just merge it. 😸

jesseeproductions:
I am going to change it to remain-visible that seems more descriptive on what it does.

PR #760, opened 2016-07-14
Feature/62439 widgets jsonld

lucatume:
Ticket: https://central.tri.be/issues/62439

Builds on Common PR: https://github.com/moderntribe/tribe-common/pull/134

This PR adds JSON-LD markup printing to the list widget.

When reviewing keep in mind that the JSON-LD data generation class in core will take care not to print JSON-LD markup twice for the same event and that printing anywhere in `head` and `body` is fine with Google.


src/Tribe/JSON_LD/Event.php
```diff
@@ -87,4 +90,29 @@ public function get_data( $posts = null, $args = array() ) {
...
+		);
+
+		foreach ( $map as $normalization_regex => $normalized_price ) {
+			if ( preg_match( $normalization_regex, '' . $price ) ) {
+				$price = $normalized_price;
+			}
+		}
+
+		// normalize the decimal separator
+		$price = str_replace( ',', '.', $price );
```
barryhughes: Is there any risk of a cost such as `10,000` (ten thousand units) being converted to `10.000` (ten units) by this approach?

lucatume: Yes, there is and I had not considered that.
I'm thinking about pulling the decimal separator from WordPress but that's not an option and it really does not tell much.
Is there a way to get WordPress to tell me the decimal and thousands separator?

lucatume:
Ok, I've updated the code to use the [locale conventions](http://php.net/manual/en/function.localeconv.php) for the replacement, should do better.

src/Tribe/JSON_LD/Event.php
```diff
@@ -108,11 +108,13 @@ protected function normalize_price( $price ) {
 			}
 		}
 
+		$locale_conv          = localeconv();
```
barryhughes: Nice solution!

Thought for the future - maybe we also filter `$locale_conv[ 'mon_decimal_point' ]` for cases where the site admin can't influence `php.ini` for whatever reason (don't necessarily need that now though).

lucatume:
I've added some filtering to the game, made me feel better about it.

barryhughes:
:noice: !

PR #761, opened 2016-07-15
Fix WPML permalinks

BruceOnTheGo:


lucatume:
Thanks @BruceOnTheGo!

PR #762, opened 2016-07-15
take sitewide timezone setting into account when generating event jso…

lucatume:
Ticket: https://central.tri.be/issues/62523

Take site wide timezone setting into account when generating jsonld data


jesseeproductions:
:+1:

lucatume:
Thanks!

PR #763, opened 2016-07-18
Feature/32372 all link in recurrence view

lucatume:
Ticket: https://central.tri.be/issues/32372

This PR adds an action before the list view content is printed and updates the styles to cover the `.tribe-events-back` link, makes sense when in relation to the PRO PR here: https://github.com/moderntribe/events-pro/pull/362


src/views/list/content.php
```diff
@@ -16,6 +16,8 @@
 
 <div id="tribe-events-content" class="tribe-events-list">
 
+	<?php do_action('tribe_events_list_before_the_content'); ?>
```
barryhughes: Looks good! Be even sweeter if we could add a docblock in here though (and a dab of whitespace before and after the single quotes) :bow:

barryhughes:
Très bien!

PR #754, opened 2016-07-08
Event Aggregator Record

bordoni:
_Ref:_ [C#63149](https://central.tri.be/issues/63149)

PR #768, opened 2016-07-20
More UI work on EA: Facebook and friends

borkweb:
This is a followup to https://github.com/moderntribe/the-events-calendar/pull/754


bordoni:
Im good with the code here, minor things that we can polish later on.

PR #772, opened 2016-07-27
Feature/61992 attendee layout

barryhughes:
Revises and simplifies our integration with Event Tickets' attendee screen. See also [Event Tickets PR 232](https://github.com/moderntribe/event-tickets/pull/232) for new points of integration.

[C#61992](https://central.tri.be/issues/61992)


borkweb:
That is definitely a simplification :) 

🔲

PR #775, opened 2016-08-01
bug/64177-remove-unneeded-pipe

trishasalas:
remove pipe from single-events template also remove all related css

https://central.tri.be/issues/64177


barryhughes:
:8ball:

PR #777, opened 2016-08-01
fixing default state/province behavior

zbtirrell:
Default state/province should only be respected on the "add new venue" page, not on the "add new event page" where venues can be quick added.

See: https://central.tri.be/issues/64629


barryhughes:
:+1:

PR #781, opened 2016-08-02
Deprecate lingering meta factory code

barryhughes:
Formally deprecates disused meta factory code. Should be the penultimate step toward finally removing that system.


bordoni:
I like this change, simplify the codes a lot. 👍

PR #782, opened 2016-08-03
Tidy tribe_get_venue_details() to avoid using meta factory API | #39905

barryhughes:
Further update to remove meta factory related calls, can achieve the same using template tags.


bordoni:
👍

PR #784, opened 2016-08-04
Adds insertion of events/venues/organizers into the posts table

borkweb:
What this PR _does not do_:

1. It does not record the number of events imported during the import process
1. It does not...hrm...I think that's it.

See: https://central.tri.be/issues/61161


src/Tribe/Aggregator/Event.php
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	/**
+	 * Fetch all existing unique facebook IDs from the provided list that exist in meta
+	 *
+	 * @param array $ids Array of facebook ids
+	 *
+	 * @return array
+	 */
+	public function get_existing_facebook_ids( $ids = array() ) {
```
bordoni: What do you think about, instead of having `get_existing_(meetup|facebook|etc...)_ids` we do a cleaver thing with `get_existing_ids` like transforming the `facebook` key into the actual key for Facebook

src/Tribe/Aggregator/Event.php
```diff
@@ -0,0 +1,164 @@
...
+		foreach ( $values as &$value ) {
+			$value = preg_replace( '/[^a-zA-Z0-9]/', '', $value );
+		}
+
+		$sql = "
+			SELECT
+				meta_value,
+				post_id
+			FROM
+				{$wpdb->prefix}postmeta
```
bordoni: Should use `$wpdb->postmeta`

bordoni:
👍

PR #790, opened 2016-08-05
Add the groundwork for recurrence support. Fix bug with event selection

borkweb:
See: https://central.tri.be/issues/61161


bordoni:
👍

PR #802, opened 2016-08-11
replace "Clear Filters" with "Cancel"

zbtirrell:
See feedback doc

### DIFF
--- a/src/admin-views/aggregator/origins/ics.php
+++ b/src/admin-views/aggregator/origins/ics.php
@@ -42,8 +42,8 @@ $field->media_title = __( 'Upload an ICS File', 'the-events-calendar' );
 		<button type="submit" class="button button-primary tribe-preview">
 			<?php esc_html_e( 'Preview', 'the-events-calendar' ); ?>
 		</button>
-		<button type="button" class="button tribe-clear-filters">
-			<?php esc_html_e( 'Clear Filters', 'the-events-calendar' ); ?>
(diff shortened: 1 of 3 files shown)


src/admin-views/aggregator/origins/ics.php
```diff
@@ -42,8 +42,8 @@ class="tribe-ea-field tribe-ea-media_button tribe-dependent button button-second
 		<button type="submit" class="button button-primary tribe-preview">
 			<?php esc_html_e( 'Preview', 'the-events-calendar' ); ?>
 		</button>
-		<button type="button" class="button tribe-clear-filters">
-			<?php esc_html_e( 'Clear Filters', 'the-events-calendar' ); ?>
+		<button type="cancel" class="button tribe-cancel">
```
bordoni: Question; Is `cancel` a type for `<button>` I've never seen it. + There isn't any JS associated with the class?

zbtirrell: I matched this: https://github.com/moderntribe/the-events-calendar/blob/develop/src/admin-views/aggregator/tabs/import-form.php#L177

Behavior is identical to that button.

borkweb: Ha! I derped that and did `type="cancel" when I should have done `type="reset"`.

borkweb:
🐕 :dash:

PR #803, opened 2016-08-11
widen the URL field

zbtirrell:
example URL that should fit: https://www.facebook.com/BrianReganComedian/

This should match the width of two "large" fields with padding so it aligns with th right of the fields below.

Also, switched order on Location and Start Date fields

See feedback doc


bordoni:
👍

PR #807, opened 2016-08-11
Support import source names and add a meta box on imported events

borkweb:
See: https://central.tri.be/issues/62663


bordoni:
💯

PR #815, opened 2016-08-11
Update the text of the "Retain" option for the change authority metabox

borkweb:
See findings/feedback compilation doc


zbtirrell:
👍

PR #817, opened 2016-08-12
Remove the 'tribe_events_rewrite_rules' filter

Chouby:
Hi!

I propose to remove the filter 'tribe_events_rewrite_rules' because it conflicts with a WordPress filter of the same name. See https://github.com/WordPress/WordPress/blob/4.5.3/wp-includes/class-wp-rewrite.php#L1416

The WP filter has only one argument. Attempting to use the second argument of the TEC filter creates an error.

The WP filter sends only the rules related to the 'tribe_events' post type, while the TEC filter sends all rewrite rules. This can raise all sorts of conflicts in plugins using the WP filter and not aware of the existence of the TEC filter.

As is, the TEC filter is not very useful as the WP filter 'rewrite_rules_array' is already there to filter all rewrite rules if someone needs it.

Considering the inline comment, I propose to simply remove the existing filter, but IMHO, a different filter, only for `$this->rules` instead of `$this->rules + $wp_rewrite->rules`, and with a different name may be useful when a plugin needs to filter only the TEC rules.

_[Ticketed: [#66455](https://central.tri.be/issues/66455)]_

### DIFF
--- a/src/Tribe/Rewrite.php
+++ b/src/Tribe/Rewrite.php
@@ -112,6 +112,10 @@ if ( ! class_exists( 'Tribe__Events__Rewrite' ) ) {
 			 * @var Tribe__Events__Rewrite $rewrite
 			 */
 			do_action( 'tribe_events_pre_rewrite', $this );
+			
+			$this->rules = apply_filters( 'tribe_events_rewrite_rules_custom', $this->rules, $this );
+			
+			$wp_rewrite->rules = $this->rules + $wp_rewrite->rules;
 		}
 
 		/**
--- a/src/Tribe/Rewrite.php
+++ b/src/Tribe/Rewrite.php
@@ -112,12 +112,6 @@ if ( ! class_exists( 'Tribe__Events__Rewrite' ) ) {
 			 * @var Tribe__Events__Rewrite $rewrite
 			 */
 			do_action( 'tribe_events_pre_rewrite', $this );
-
-			/**
-			 * Backwards Compatibility filter, this filters the WP Rewrite Rules.
-			 * @todo  Check if is worth deprecating this hook
-			 */
-			$wp_rewrite->rules = apply_filters( 'tribe_events_rewrite_rules', $this->rules + $wp_rewrite->rules, $this );
 		}
 
 		/**


barryhughes:
Hi @Chouby,

Thanks for the pull request and our apologies for the delayed response. You've highlighted a valid problem and it would be great to solve it.

We can't merge this in 'as is' since this change would also prevent our custom rules from being appended to the existing set of WP_Rewrite rules, but we can definitely build on this and take it forward. Thoughts (some of which are just echoes of your own):

* As you suggested, we should add a new filter hook - perhaps `tribe_events_rewrite_rules_custom` - that is concerned only with `$this->rules` 
* We still need to append the resulting set of filtered rules to those of the `$wp_rewrite` object, so that operation needs to be restored
* We should ditch the existing filter hook and, given the situation, I think we can simply remove it completely and skip formal deprecation

If you're still interested in moving this forward and have the time to do so we'd love to see an update along the above lines and we can then look at merging it in.

Thanks again and apologies once more for the delay.

barryhughes:
Ticketed: [#66455](https://central.tri.be/issues/66455)

tr1b0t:

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [Rewrite.php:115](https://github.com/moderntribe/the-events-calendar/blob/f41e90edb60e9b1437ed8ba74b9aa0ad43e8bf05/src/Tribe/Rewrite.php#L115)
:x: | Whitespace found at end of line | [Rewrite.php:117](https://github.com/moderntribe/the-events-calendar/blob/f41e90edb60e9b1437ed8ba74b9aa0ad43e8bf05/src/Tribe/Rewrite.php#L117)

Chouby:
Sorry I indeed removed the most useful part of the code. I updated the PR to fix that and introduce the filter 'tribe_events_rewrite_rules_custom' to replace 'tribe_events_rewrite_rules'

tr1b0t:

barryhughes:
@Chouby - thanks for the update. This looks great and we'll get it merged in shortly - thank you for the contribution!

PR #824, opened 2016-08-13
Target the credentials hidden input correctly so subsequent fields show

borkweb:
See: findings compilation doc


jesseeproductions:
:+1:

PR #830, opened 2016-08-16
check license before showing upsell

zbtirrell:
This is only a partial fix. The data in the license info appears to be
stale and not being updated when the API key is changed for some reason.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not ')' | [import-form.php:265](https://github.com/moderntribe/the-events-calendar/blob/cb08fc5d1782e45296d9b4d02e89938f78c4b4e0/src/admin-views/aggregator/tabs/import-form.php#L265)

borkweb:
:+1:

PR #831, opened 2016-08-16
add an isset check

lucatume:
emerged while working on ticket https://central.tri.be/issues/61724 to restore ET tests:

add an `isset` check to avoid errors when a meta field is not set on `old`

### DIFF
--- a/src/Tribe/API.php
+++ b/src/Tribe/API.php
@@ -159,7 +159,7 @@ if ( ! class_exists( 'Tribe__Events__API' ) ) {
 					else {
 						update_post_meta( $event_id, $tag, $data[ $htmlElement ] );
 					}
-	//here
+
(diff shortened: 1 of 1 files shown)


src/Tribe/API.php
```diff
@@ -159,7 +159,7 @@ public static function saveEventMeta( $event_id, $data, $event = null ) {
 					else {
 						update_post_meta( $event_id, $tag, $data[ $htmlElement ] );
 					}
-
+	//here
```
borkweb: You should add 4 more lines just like this.

borkweb: Or remove this one, I guess.

lucatume:
And I do not know why I had added that.

borkweb:
:+1:

PR #835, opened 2016-08-16
Fixing bug with manual imports where they were no longer possible

borkweb:
Additionally rolling back width=100% issue because it broke wide, scrollable tables.


bordoni:
👍

PR #837, opened 2016-08-16
Add required info into CSV imports to enable the Import Meta box

borkweb:
See: https://central.tri.be/issues/62663

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -863,7 +863,7 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 				$results['created']++;
 			}
 
-			self::add_record_to_event( $event['ID'], $this->id, $this->origin );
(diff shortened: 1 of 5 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -915,6 +911,21 @@ public function insert_posts( $data = array() ) {
 	}
 
 	/**
+	 * Adds the import record and origin to the imported event
+	 *
+	 * @param int $id Event ID
+	 * @param int $record_id Import Record ID
+	 * @param string $origin Import Origin
+	 */
+	public static function add_record_to_event( $id, $record_id, $origin ) {
```
bordoni: Can we place this on the `Records` class to avoid needing to use a Static method here on an Abstract?

borkweb: Moved!

bordoni:
👍

PR #841, opened 2016-08-17
Reset polling counters so subsequent preview button clicks don't error

borkweb:
See: Findings compilation doc


bordoni:
Há! Nice 💯

PR #846, opened 2016-08-18
We don't need to sanitize in such a limited manner. wpdb prepare is fine

borkweb:
We were sanitizing aggressively for no real reason. We need to search for the uid values that are inserted.

See: Findings compilation doc


bordoni:
👍

PR #848, opened 2016-08-18
Add support for ET/ET+ CSV imports and prevent duplicate types

borkweb:
Related:

* https://github.com/moderntribe/event-tickets/pull/237
* https://github.com/moderntribe/event-tickets-plus/pull/210

See: https://central.tri.be/issues/61240#note-17

### DIFF
--- a/src/admin-views/aggregator/origins/csv.php
+++ b/src/admin-views/aggregator/origins/csv.php
@@ -9,7 +9,15 @@ $field->help        .= __( 'For the best results, import venue and organizer fil
 
 $field->source      = 'csv_content_type';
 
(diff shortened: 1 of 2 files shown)


src/admin-views/aggregator/origins/csv.php
```diff
@@ -9,7 +9,15 @@
 
 $field->source      = 'csv_content_type';
 
-$post_types = array_map( 'get_post_type_object', Tribe__Main::get_post_types() );
+$post_types = array_unique( Tribe__Main::get_post_types() );
```
bordoni: Can we do that on a Method inside the Origin class?

borkweb:
Code review addressed. Ready for another look-see.

PR #867, opened 2016-08-22
only show legacy import link for FB, iCal, and EB users

zbtirrell:
Leah and I discussed this and it seems like we'll cause less confusion by limiting the audience for this link to only those who need it.


borkweb:
:+1:

PR #869, opened 2016-08-22
Fix/39905 meta factory no no

barryhughes:
Further work to remove use of meta factory code.

### DIFF
--- a/src/functions/template-tags/organizer.php
+++ b/src/functions/template-tags/organizer.php
@@ -125,7 +125,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 		}
 
 		if ( $organizer_id && $link = tribe_get_organizer_website_link() ) {
-			$details[] = '<span class="link"> <a href="mailto:' . esc_attr( $link ) . '">' . $link . '</a> </span>';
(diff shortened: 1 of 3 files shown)


src/functions/template-tags/organizer.php
```diff
@@ -102,6 +102,49 @@ function tribe_get_organizer_label( $singular = true ) {
...
+		if ( $organizer_id && $tel = tribe_get_organizer_phone() ) {
+			$details[] = '<span class="tel">' . $tel . '</span>';
+		}
+
+		if ( $organizer_id && $email = tribe_get_organizer_email() ) {
+			$details[] = '<span class="email"> <a href="mailto:' . esc_attr( $email ) . '">' . $email . '</a> </span>';
+		}
+
+		if ( $organizer_id && $link = tribe_get_organizer_website_link() ) {
+			$details[] = '<span class="link"> <a href="mailto:' . esc_attr( $link ) . '">' . $link . '</a> </span>';
```
borkweb: This probably shouldn't be a `mailto` link.

barryhughes: It definitely should not!

borkweb:
:+1:

PR #872, opened 2016-08-23
hide the upsell when there is a value selected for origin

zbtirrell:


borkweb:
Rather than do this with JS, we can use the field dependency stuff to make this happen. In `Tribe/Aggregator/Tabs/New::maybe_display_aggregator_upsell()`, you can change the `<div>` to the following with the same effect:

```
<div class="notice inline notice-info tribe-dependent tribe-notice-tribe-missing-aggregator-license" data-ref="tribe-missing-aggregator-license" data-depends="#tribe-ea-field-origin" data-condition-empty>
```

borkweb:
:+1:

PR #873, opened 2016-08-23
import settings need to be available for non-Aggregator

zbtirrell:
Also some reorganization of the code there to make it more scannable


bordoni:
👍

PR #877, opened 2016-08-23
don't change post modified on schedule update

zbtirrell:
We set that field for the Last Import date, so we don't want to update it
if the settings change.

Also fixed a couple of notices.


bordoni:
👍

PR #880, opened 2016-08-24
Show disabled options and upsell to Aggregator

zbtirrell:
This pull request includes a number of significant changes spawning from the initial goal.

In order to make the disabled options work, this required an upgrade to Select2. Initially, I tried to go to the latest 4.x version, but that broke A LOT of things. I then rolled back to the latest 3.x version, but that was also troublesome. I finally landed on bumping it just far enough to get the disabled bug resolved, which should have the least impact.

This includes change in the way origins are fetched, but I believe it improves the translatability and future flexibility. **Note: due to this change, you will need to clear transients.**

This includes a change to present uneditable options (origin and import type) to be presented as static string rather than disabled select boxes. Along with that, during an edit, since the origin cannot be changed, we only need to load the relevant template.

Finally, this includes support for the `TRIBE_HIDE_UPSELL` constant for suppressing upsells with a snippet.


borkweb:
If this passes the @leahkoerper test, this is :+1: code-wise.

PR #883, opened 2016-08-24
Convert the Scheduled tab's "Run Now" link to use the new Queue process

borkweb:
This also fixes a bug where cron'd schedules were causing duplicate events

See: Findings doc


bordoni:
👍

PR #884, opened 2016-08-24
Only show Realtime Queue processing UI for imports that were interactive

borkweb:
See: Findings doc (sorta)

### DIFF
--- a/src/Tribe/Aggregator/Tabs/Scheduled.php
+++ b/src/Tribe/Aggregator/Tabs/Scheduled.php
@@ -237,6 +237,7 @@ class Tribe__Events__Aggregator__Tabs__Scheduled extends Tribe__Events__Aggregat
 
 			$child = $record->create_child_record();
 			$status = $child->queue_import();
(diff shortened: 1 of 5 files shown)


src/Tribe/Aggregator/Tabs/Scheduled.php
```diff
@@ -237,15 +237,14 @@ private function action_run_import( $records = array() ) {
...
+			$child->update_meta( 'interactive', true );
 			$child->finalize();
 			$child->process_posts();
 
 			if ( is_wp_error( $status ) ) {
 				$errors[ $record->id ] = $status;
 				continue;
 			}
-
-			//$success[ $record->id ] = $status;
```
zbtirrell: this isn't needed still?

borkweb: No, because interactively run schedules get the new polling notice rather than the "1 scheduled import has been queued" message.

zbtirrell:
:shipit:

PR #886, opened 2016-08-24
Return the WP_Error, not an ID when the daily limit is reached

borkweb:
This is a spotfix with the wrong branch name. Sue me.


zbtirrell:
yes please! 👍

PR #887, opened 2016-08-24
crank up the timeout when talking to Event Aggregator

zbtirrell:
This is important for slow connections and big file transfers


borkweb:
:+1:

PR #893, opened 2016-08-26
simple text revisions

zbtirrell:


borkweb:
:hot dog hot dog hot diggity dog:

PR #894, opened 2016-08-26
When an event is ignored, don't overwrite the post status

borkweb:
See: Findings doc

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -894,7 +894,7 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 
 		$unique_field = $this->get_unique_field();
 		$existing_ids = $this->get_existing_ids_from_import_data( $items );
-		$ignored_ids  = $this->get_ignored_ids( $existing_ids );
+		$ignored_ids  = Tribe__Events__Ignored_Events::instance()->get_ignored_posts_by_id( array( 'fields' => 'ids' ), $existing_ids );
 
 		$count_scanned_events = 0;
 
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1134,6 +1138,36 @@ protected function get_existing_ids_from_import_data( $import_data ) {
...
+	 */
+	protected function get_ignored_ids( $check ) {
+		if ( empty( $check ) ) {
+			return array();
+		}
+
+		$first = reset( $check );
+
+		if ( ! is_numeric( $first ) ) {
+			$check = wp_list_pluck( $check, 'post_id' );
```
bordoni: It would be cool wit we could pluck from an array of `WP_Post` too. What do you think?

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1134,6 +1138,36 @@ protected function get_existing_ids_from_import_data( $import_data ) {
...
 	}
 
+	/**
+	 * Gets all ids for events that have been ignored
+	 *
+	 * @param array $check Array of event IDs to check
+	 *
+	 * @return array
+	 */
+	protected function get_ignored_ids( $check ) {
```
bordoni: Also, we could move this to the `Ignored_Events` class.

zbtirrell:
Quick thought, do we ever want to overwrite post status?

Example: I import all events as "draft" in a scheduled daily import. Then I publish one. Shouldn't it stay published regardless of my authority setting?

borkweb:
That's a good thought, @zbtirrell. Maybe we don't! I'll go ahead an remove the post_status update for any post that has an ID!

borkweb:
Ok. This is ready for another look.

src/Tribe/Ignored_Events.php
```diff
@@ -481,13 +481,101 @@ public function has_ignored_posts( $check_legacy = true ) {
...
+		}
+
+		/**
+		 * Gets all ids for events that have been ignored
+		 *
+		 * @param null|array $data Array of event IDs/objects/data to check
+		 *
+		 * @return array
+		 */
+		protected function get_ignored_posts_query( $args = array() ) {
```
bordoni: Lets make this public, no need to call it `get_ignored` because of the class name. We can use `query_posts` no?

src/Tribe/Ignored_Events.php
```diff
@@ -481,13 +481,102 @@ public function has_ignored_posts( $check_legacy = true ) {
...
+					$ids = wp_list_pluck( $data, $id_field );
+				}
+			}
+
+			// if there aren't any IDs, let's make sure we don't get any results
+			if ( empty( $ids ) ) {
+				return array();
+			}
+
+			$args['post__in'] = $ids;
```
bordoni: We should merge if it previously exist right?

src/Tribe/Ignored_Events.php
```diff
@@ -481,13 +481,102 @@ public function has_ignored_posts( $check_legacy = true ) {
...
+
+			return $query->posts;
+		}
+
+		/**
+		 * Gets all ids for events that have been ignored
+		 *
+		 * @param null|array $data Array of event IDs/objects/data to check
+		 *
+		 * @return array
```
bordoni: Returns `WP_Query`

PR #899, opened 2016-08-26
improved styling of the upsell in the select

zbtirrell:


src/admin-views/aggregator/tabs/import-form.php
```diff
@@ -69,7 +69,11 @@ class="tribe-ea-field tribe-ea-dropdown tribe-ea-size-large"
 							$disabled = ( isset( $option->disabled ) ? $option->disabled : null );
 							if ( ! $upsell && $disabled ) {
 								$upsell = true;
-								?><option value="redirect"><?php esc_html_e( 'Buy Event Aggregator for more event sources and automatic imports!', 'the-events-calendar' ); ?></option><?php
+								?><option value="redirect"><?php
+									esc_html_e( 'Buy Event Aggregator', 'the-events-calendar' );
```
barryhughes: Asking this out of curiosity really, but why two translatable strings broken up by a pipe separator vs a single translatable string that includes the pipe separator?

zbtirrell: I'm only using the pipe separator to split the option into two lines in JavaScript. So,  there is no reason to expose the pipe to translators. Also, in the UI, these end up presented as two distinct elements to the user. See: 
screenshot 2016-08-29 08 08 56

barryhughes:
That makes sense :cactus:

PR #902, opened 2016-08-29
tribe_is_past_event conditional and ticket messages

jesseeproductions:
_Ref:_ [C#62303](https://central.tri.be/issues/62303) Add tribe_is_past_event() conditional to be used in Event Tickets and Single Event Templates

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -288,6 +288,33 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	}
 
 	/**
+	 * Detect if an Event Is Past
+	 *
+	 * Returns true if the current time is past the event end time
+	 *
+	 * @param null $event
+	 *
+	 * @return bool
+	 */
+	function tribe_is_past_event( $event = null ) {
+
+		if ( is_null( $event ) ) {
+			global $post;
+			$event = $post;
+		}
+		// Check if event has passed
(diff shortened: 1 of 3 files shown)


src/functions/template-tags/general.php
```diff
@@ -288,6 +288,33 @@ function tribe_event_is_multiday( $postId = null ) {
...
+
+		if ( is_null( $event ) ) {
+			global $post;
+			$event = $post;
+		}
+		// Check if event has passed
+		$gmt_offset = ( get_option( 'gmt_offset' ) >= '0' ) ? ' +' . get_option( 'gmt_offset' ) : ' ' . get_option( 'gmt_offset' );
+		$gmt_offset = str_replace( array( '.25', '.5', '.75' ), array( ':15', ':30', ':45' ), $gmt_offset );
+
+		if ( ! tribe_is_showing_all() && strtotime( tribe_get_end_date( $event, false, 'Y-m-d G:i' ) . $gmt_offset ) <= time() ) {
```
barryhughes: ... If we were to drop it from this lcoation, we could do the following back in `set_notices()`:

``` php
if ( ! tribe_is_showing_all() && tribe_is_past_event() ) {
    ...
}
```

jesseeproductions: Makes sense. I am making that change.

borkweb:
:+1:

PR #911, opened 2016-08-31
Now Facebook oAuth UX is using a full page instead of iFrame

bordoni:


src/admin-views/aggregator/origins/facebook.php
```diff
@@ -13,35 +13,26 @@
...
-			</div>
-		</div>
-	<?php else: ?>
-		<input id="tribe-has-facebook-credentials" type="hidden" value="1" />
-	<?php endif; ?>
+	<th scope="row">
+		<label for="tribe-ea-field-fb_credentials"><?php esc_html_e( 'Extra Credentials:', 'the-events-calendar' ); ?></label>
+	</th>
+	<td>
+		<a target="_blank" style="line-height: 28px;" href="<?php echo esc_url( Tribe__Events__Aggregator__Record__Facebook::get_auth_url() ); ?>">Login with Facebook</a>
```
zbtirrell: "Login with Facebook" needs to be translatable

zbtirrell:
other than that small adjustment (which you _can_ skip, since I'll be passing back through the text) this looks 👍

PR #924, opened 2016-09-01
added custom fields translations default setting

lucatume:
Ticket: https://central.tri.be/issues/40720#note-55

This PR scaffolds the Integrations section in Core too, as already done in Pro, and sets all sensible custom fields that we need translated to have a no non-sense translation of an event to be copied to the translated event.

Default would be to have them non translated and this would yield translated events being completely empty of all the expected meta.

The code will set just an *una tantum* default and will still allow users to have the last word on whic custom fields should be translated or not; we are just setting a sensible default.


src/Tribe/Integrations/WPML/Defaults.php
```diff
@@ -0,0 +1,150 @@
...
+			'_VenueOverwriteCoords',
+			'_VenuePhone',
+			'_VenueProvince',
+			'_VenueShowMap',
+			'_VenueShowMapLink',
+			'_VenueState',
+			'_VenueStateProvince',
+			'_VenueURL',
+			'_VenueVenue',
+			'_VenueZip',
```
borkweb: Are there fields missing from here? There seems to be a few fields [in Main](https://github.com/moderntribe/the-events-calendar/blob/develop/src/Tribe/Main.php) that are not in this list. Is that on purpose? Are there fields missing from Main that should be there? Should this list be built from a centralized collection of fields?

lucatume: Yes, ideally this list should contain any custom field we use in CORE to store relevant informations for an event that's supposed to be the copy/clone of another event, I've created the list using what WPML was picking up but something could be missing.

lucatume:
I've update  the list from Main and it seems now more complete

src/Tribe/Integrations/WPML/Defaults.php
```diff
@@ -0,0 +1,145 @@
...
+			'_EventShowMap',
+			'_EventCurrencySymbol',
+			'_EventCurrencyPosition',
+			'_EventCost',
+			'_EventCostMin',
+			'_EventCostMax',
+			'_EventURL',
+			'_EventOrganizerID',
+			'_EventPhone',
+			'_EventHideFromUpcoming',
```
borkweb: Let's add one more: `_EventRecurrenceRRULE` (which will be coming in with some aggregator records)

src/Tribe/Integrations/WPML/WPML.php
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Class Tribe__Events__Integrations__WPML__WPML
+ *
+ * Handles anything relating to The Events Calendar and WPML integration
+ *
+ * This class is meant to be an entry point hooking specialized classes and not
+ * a logic hub per se.
+ */
+class Tribe__Events__Integrations__WPML__WPML {
```
bordoni: We use the `Main.php` across our plugins, we could stick to that here.

lucatume: The case here is that we might end up with more integrations and having the main file of each called `Main` would make skimming the code quite a chore.
We use this convention elsewhere in the code where we call the main class `Tribe_..._Some_Some`

src/Tribe/Integrations/Manager.php
```diff
@@ -0,0 +1,53 @@
...
+	 * @var Tribe__Events__Integrations__Manager
+	 */
+	protected static $instance;
+
+	/**
+	 * The class singleton constructor.
+	 *
+	 * @return Tribe__Events__Integrations__Manager
+	 */
+	public static function instance() {
```
barryhughes: If we want to ensure only one instance of this class can be created we should probably define a protected or private constructor (with an empty body). Otherwise, accidents can happen in future where other code instantiates additional objects of this type (we just saw this with Tribe__Main for example).

src/Tribe/Integrations/WPML/Defaults.php
```diff
@@ -0,0 +1,146 @@
...
+	 */
+	protected $settings_manager;
+
+	/**
+	 * Tribe__Events__Integrations__WPML__Defaults constructor.
+	 *
+	 * @param SitePress|null               $sitepress
+	 * @param Tribe__Settings_Manager|null $settings_manager
+	 */
+	public function __construct( SitePress $sitepress = null, Tribe__Settings_Manager $settings_manager = null ) {
```
barryhughes: Should this be protected or private?

src/Tribe/Integrations/WPML/WPML.php
```diff
@@ -0,0 +1,53 @@
...
+	 * @var Tribe__Events__Integrations__WPML__WPML
+	 */
+	protected static $instance;
+
+	/**
+	 * The class singleton constructor.
+	 *
+	 * @return Tribe__Events__Integrations__WPML__WPML
+	 */
+	public static function instance() {
```
barryhughes: As with `Tribe__Events__Integrations__Manager` - if we want to enforce there being only one object of this type we should probably define a protected/private constructor, else it's implicitly public and instances can be created willy-nilly.

src/Tribe/Integrations/Manager.php
lucatume: Making the class constructor private makes it a pain to test; we are providing an ` instance` method for that reason. We have the same case repeating similarly throughout our code where the `__construct` method is public; this class has also an empty constructor so even instantiated twice it should cause no risks.

barryhughes: That's totally fine, but we should probably avoid describing it as a singleton in that case - it's not really a singleton.

src/Tribe/Integrations/WPML/Defaults.php
lucatume: Public, again for reasons of testing and for the same reason above: if we want to prevent devs from misusing the code we would end up locking it up in private and final methods and classes. Which is a thing that makes working with it almost impossible

PR #925, opened 2016-09-01
Better handling of CSV headers

borkweb:
This resolves a couple of issues:

* CSV saving when mapping fields failed to render fields that _must_ be mapped
* CSV field settings for organizers and venues were not auto-selected because the wrong option was being saved

See: https://central.tri.be/issues/65945

### DIFF
--- a/src/Tribe/Aggregator/Record/CSV.php
+++ b/src/Tribe/Aggregator/Record/CSV.php
@@ -205,19 +205,19 @@ class Tribe__Events__Aggregator__Record__CSV extends Tribe__Events__Aggregator__
 			$content_type = $this->get_content_type();
 		}
 
-		switch ( $content_type ) {
-			case 'event':
-			case 'events':
-				$content_type = 'events';
-				break;
-			case 'organizer':
-			case 'organizers':
-				$content_type = 'organizers';
-				break;
-			case 'venue':
-			case 'venues':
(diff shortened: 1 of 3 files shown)


src/Tribe/Aggregator/Record/CSV.php
```diff
@@ -193,6 +193,37 @@ public function get_content_type() {
...
+	 *
+	 * @return string CSV Importer compatible content type
+	 */
+	public function get_csv_content_type( $content_type = null ) {
+
+		if ( ! $content_type ) {
+			$content_type = $this->get_content_type();
+		}
+
+		switch ( $content_type ) {
```
bordoni: Hey can I suggest instead of the switch:
```php
$map = array(
     'event' => 'events'
     'events' => 'events'
     'organizer' => 'organizers'
     'organizers' => 'organizers'
     'venue' => 'venues'
     'venues' => 'venues'
);

if ( ! isset( $map[ $content_type ] ) ) {
     return false;
}

return $map[ $content_type ];
```

borkweb: I've tweaked the above slightly to account for the fact that some values passed in (or fetched from `get_content_type()`) are what is actually desired.

zbtirrell:
💵

PR #931, opened 2016-09-02
Force the first line of CSV to be a header

borkweb:
See: https://central.tri.be/issues/65945


zbtirrell:
Perhaps we should filter that? Not sure who'd want it though...

Whatever. Merge it.

borkweb:
It is innately filterable because of the `get/update_option`. Mergin!

PR #937, opened 2016-09-06
Show the available Tickets on Listview

masteradhoc:
Credit to https://gist.github.com/ggwicz
https://gist.github.com/ggwicz/bb71d468a8139b1918e830ea302bc7af

### DIFF
--- a/src/views/list/single-event.php
+++ b/src/views/list/single-event.php
@@ -41,6 +41,7 @@ $organizer = tribe_get_organizer();
 <?php do_action( 'tribe_events_after_the_event_title' ) ?>
 
 <!-- Event Available Tickets -->
+<?php
 $event_id = get_the_ID();
 
 $woo_tickets = Tribe__Tickets_Plus__Commerce__WooCommerce__Main::get_instance();
@@ -70,7 +71,7 @@ if ( $tickets_left == 0 ) {
 	if ( $stock_count - $sold_count == 0 ) {
 	      echo '<strong><em>Sold out!</em></strong><br>';
 	}
-}
+}?>
 
 <!-- Event Meta -->
 <?php do_action( 'tribe_events_before_the_meta' ) ?>
--- a/src/views/list/single-event.php
+++ b/src/views/list/single-event.php
@@ -40,6 +40,38 @@ $organizer = tribe_get_organizer();
 </h2>
 <?php do_action( 'tribe_events_after_the_event_title' ) ?>
 
+<!-- Event Available Tickets -->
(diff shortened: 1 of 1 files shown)


masteradhoc:
hi there. thanks for checking this PR. If its fine i'll add this also to the other views.

Uservoice:
https://tribe.uservoice.com/forums/195723-feature-ideas/suggestions/5605710--sold-out-only-x-positions-left-etc-type-label
and also first step of https://tribe.uservoice.com/forums/195723-feature-ideas/suggestions/13746300-to-get-stock-quantity-in-calender-view-and-add-to

barryhughes:
Thanks for the PR, @masteradhoc!

This could definitely be useful for lots of users but - as is - I'm afraid we probably can't merge this PR.

* It assumes `Tribe__Tickets_Plus__Commerce__WooCommerce__Main` will be defined, but for many users who do not have Event Tickets Plus installed that will not be the case so we'd need some sort of safety check
* The approach used in [George's snippet](https://gist.github.com/ggwicz/bb71d468a8139b1918e830ea302bc7af) (using the `tribe_events_after_the_event_title` hook) is a nice one: if we stick to that approach and perhaps move this to either Event Tickets or Event Tickets Plus I think we'd be on track
* Following from the above, we'd possibly also wish to genericize things a bit so that, for example, RSVP and EDD users aren't left out in the cold
* I'm also a little concerned that despite the popularity of the UserVoice request there may also be users who do not wish to see the extra output on their list views, so I do wonder about making this optional

Overall though, this is a great idea and the above points don't take away from that. @borkweb - what are your thoughts - do you think we'd be happy to see this reworked as a fresh PR targeting ET+?

masteradhoc:
any update here? @barryhughes :)

barryhughes:
Sorry for the delay, @masteradhoc.

There's definitely quite a bit of demand for this functionality but we do feel it would be better if the code was relocated to Event Tickets (and/or Event Tickets Plus) and made a little more generic. 

With that in mind, we're going to log this for inclusion in an upcoming release and we'll plan on developing it internally: there are a few things we need to hash out in more detail and I'd hate for you to work on a further PR only to have to make significant changes once more.

Thanks again for the contribution and for pushing this forward. If you have any other questions let me know, otherwise I'll go ahead and close this PR.

masteradhoc:
HI @barryhughes 
Thanks for the update.
I'll go ahead and close it, please include it in short timely manner. I have heard from many people that this for them would be a MUST for a great plugin like yours.

Thanks Brian

PR #940, opened 2016-09-07
reset all sub fields when the origin changes

zbtirrell:
This fixes weird artifact issues that occur if someone changes origins multiple times without submitting for an import.

Also fix a small typo in a dependency


borkweb:
Nice! :+1:

PR #946, opened 2016-09-07
Support appending a default category onto CSV imported events

borkweb:
Additionally fixing a tiny bug in the JS as it relates to resetting form fields.

See: https://central.tri.be/issues/61240#note-23


zbtirrell:
👍

PR #948, opened 2016-09-07
Set CSV default post status and categories correctly

borkweb:
Post statuses, when submitted via the New Import form, were not getting applied when inserting posts during CSV imports. This resolves that as well as ensures that default categories are set appropriately.

See: https://central.tri.be/issues/65358


barryhughes:
:snowboarder:

PR #949, opened 2016-09-07
Display a warning notice when legacy plugins are active when using EA

borkweb:
See: https://central.tri.be/issues/61948


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$active' | [Page.php:350](https://github.com/moderntribe/the-events-calendar/blob/e835e90f4a9412c7acae10bd7b8304bbe859b535/src/Tribe/Aggregator/Page.php#L350)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not 'isset' | [Page.php:351](https://github.com/moderntribe/the-events-calendar/blob/e835e90f4a9412c7acae10bd7b8304bbe859b535/src/Tribe/Aggregator/Page.php#L351)

src/Tribe/Aggregator/Page.php
```diff
@@ -304,4 +305,59 @@ public function template( $name, $data = array(), $echo = true ) {
...
+
+		if ( $ical_active ) {
+			$active[] = '<b>' . __( 'iCal Importer', 'the-events-calendar' ) . '</b>';
+		}
+
+		ob_start();
+		?>
+		<p>
+			<?php
+			echo sprintf(
```
barryhughes: `printf()` instead of `echo sprintf()`?

src/Tribe/Aggregator/Page.php
```diff
@@ -304,4 +305,59 @@ public function template( $name, $data = array(), $echo = true ) {
...
+		$facebook_active = $aggregator->is_legacy_facebook_active();
+
+		if ( ! $ical_active && ! $facebook_active ) {
+			return false;
+		}
+
+		$active = array();
+
+		if ( $facebook_active ) {
+			$active[] = '<b>' . __( 'Facebook Events', 'the-events-calendar' ) . '</b>';
```
barryhughes: `esc_html__()` (ditto for ical)?

src/Tribe/Aggregator/Page.php
```diff
@@ -304,4 +305,59 @@ public function template( $name, $data = array(), $echo = true ) {
...
+						'the-events-calendar'
+					)
+				),
+				$active[0],
+				isset( $active[1] ) ? $active[1] : ''
+			);
+			?>
+		</p>
+		<p>
+			<a href="<?php echo esc_url( admin_url( 'plugins.php?plugin_status=active' ) ); ?>"><?php echo esc_html_e( 'Manage Active Plugins', 'the-events-calendar' ); ?></a>
```
barryhughes: If we `esc_html_e()` we don't also need to echo it.

borkweb:
Derp. Ready for another look :)

src/Tribe/Aggregator/Page.php
```diff
@@ -119,6 +119,7 @@ private function __construct() {
 
 		tribe_asset( $plugin, 'tribe-ea-facebook-login', 'aggregator-facebook-login.js', array( 'jquery', 'underscore', 'tribe-dependency' ), 'admin_enqueue_scripts' );
 
+		tribe_notice( 'tribe-aggregator-legacy-import-plugins-active', array( $this, 'notice_legacy_plugins' ), 'type=warning' );
```
barryhughes: You've probably weighed this up already ... but is there a case for making this dismissable (or, is there a case where a user might conceivably wish to keep a legacy plugin active and can't deactivate them right away and would like to be rid of this notice)?

barryhughes:
Good to merge! :cactus:

src/Tribe/Aggregator/Page.php
borkweb: That's a good question, honestly. The original plan was to avoid permitting the use of Facebook/iCal if the legacy plugins were active. In the interest of simplicity, preventing the notice from being dismissed was a middle ground decision I made. @zbtirrell may have different ideas.

PR #952, opened 2016-09-08
Test to see if data.single is defined before accessing it | #65986

barryhughes:
We were trying to access a property of the `tribe_ignore_events.single` object on each (admin) page, however that object is not defined or needed on every request. This fix adds a safety check to avoid type errors.

[#65986](https://central.tri.be/issues/65986)


zbtirrell:
👍

PR #954, opened 2016-09-09
Skipped field is used by subclasses and so should not be private | spotfix

barryhughes:
Changes the `skipped` field's access modifier to protected: as subclasses like `Tribe__Events__Importer__File_Importer_Events` try to work with the property it cannot be private.


bordoni:
👍

PR #955, opened 2016-09-09
Remove exception for CSV-imported events re Hide & Ignore functionality

barryhughes:
Ensures that the `_EventOrigin` is reliably set for all events pulled in through EA, and removes the exception for CSV-sourced events re Hide & Ignore (ie, make it so they can be ignored rather than trashed).

[#61161](https://central.tri.be/issues/61161)


bordoni:
👍

PR #956, opened 2016-09-09
Always form the correct link to the ignored event list 

barryhughes:
Form the correct link to the ignored event list even if all we have are legacy, as yet unconverted deleted_posts. 

The problem this solves is that (if we have no EA-style ignored events, but we do have as yet unconverted 'legacy' ignored events) we don't present the user with a list of _all_ events when they click the _Ignored (X)_ link.

[#61161](https://central.tri.be/issues/61161)


bordoni:
👍

PR #959, opened 2016-09-12
Support both numeric and non-numeric categories when importing CSV

borkweb:
See: https://central.tri.be/issues/61240#note-27


bordoni:
👍

PR #962, opened 2016-09-13
small 5.2 compat changes on empty

lucatume:
Ticket: found while working on https://central.tri.be/issues/40720

Fix some uses of `empty` function on functions


barryhughes:
It looks like `Aggregator/Tabs/New.php` is now _empty_ of these problems ...

lucatume:
That must be English humor.

PR #963, opened 2016-09-13
Fix view categories link | spotfix

barryhughes:
*View your event categories* should link to event categories list, not the list of organizers.

### DIFF
--- a/src/Tribe/Aggregator/Tabs/New.php
+++ b/src/Tribe/Aggregator/Tabs/New.php
@@ -328,7 +328,7 @@ class Tribe__Events__Aggregator__Tabs__New extends Tribe__Events__Aggregator__Ta
 				_n( '%1$d new event category was created.', '%1$d new event categories were created.', $queue->activity->count( 'category', 'created' ), 'the-events-calendar' ),
 				$queue->activity->count( 'category', 'created' )
 			) .
-			' <a href="' . admin_url( 'edit.php?post_type=tribe_organizer' ) . '">' .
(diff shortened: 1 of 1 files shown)


src/Tribe/Aggregator/Tabs/New.php
```diff
@@ -328,7 +328,7 @@ public function get_result_messages( $queue ) {
 				_n( '%1$d new event category was created.', '%1$d new event categories were created.', $queue->activity->count( 'category', 'created' ), 'the-events-calendar' ),
 				$queue->activity->count( 'category', 'created' )
 			) .
-			' <a href="' . admin_url( 'edit.php?post_type=tribe_organizer' ) . '">' .
+			' <a href="' . admin_url( 'edit-tags.php?taxonomy=tribe_events_cat&post_type=tribe_events' ) . '">' .
```
borkweb: Should we check if the user is running WP 4.5+ and use the `term.php` page?

See: https://make.wordpress.org/core/2016/03/07/changes-to-the-term-edit-page-in-wordpress-4-5/

bordoni: I like that Idea.

barryhughes: > Should we check if the user is running WP 4.5+ and use the `term.php` page?

Not sure if I'm misunderstanding ... but it seems to me we wouldn't want to do that, as we're not linking to the edit page for an individual term.

borkweb: Move along. Nothing to see here. 👮 🚔

borkweb:
:+1:

PR #964, opened 2016-09-13
Remove mime type restrictions

borkweb:
Either Safari does not properly observe the .ics extension in plupload settings or the version of plupload in WP does not communicate well with Safari. As such, we're just going to remove the mime type restrictions on the EA import page.

See: Findings doc


bordoni:
Nice cool fix. 👍

PR #968, opened 2016-09-14
Don't add the Aggregator Record post type to Appearance > Menus

borkweb:
See: https://central.tri.be/issues/66516


barryhughes:
:partly_sunny:

PR #970, opened 2016-09-14
Bail if $response does not have a status property | spotfix

barryhughes:
Even if `$response` is not a `WP_Error`, there may be no `status` property (if the site doesn't have an EA key for example).


barryhughes:
Switched to `empty()`

PR #974, opened 2016-09-14
Close bumpdowns when the origin changes

borkweb:
See: https://central.tri.be/issues/66524


src/resources/js/aggregator-fields.js
```diff
@@ -105,6 +105,10 @@ tribe_aggregator.fields = {
 				obj.$.form.attr( 'data-origin', $( this ).val() );
 				obj.reset_preview();
 
+				// reset all bumpdowns
+				$( '.tribe-bumpdown-active' ).removeClass( 'tribe-bumpdown-active' );
```
bordoni: Could we use the selectors that are saved on the `selectors.dropdown` variable? (or update it)

borkweb: I'm not sure I know how `selectors.dropdown` relates to the bumpdowns.

bordoni: my bad...

PR #975, opened 2016-09-14
Make EA image import count correctly created images

bordoni:
_Ref:_ [C#66509](https://central.tri.be/issues/66509)

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -1025,7 +1025,7 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 			// If we have a Image Field from Service
 			if ( ! empty( $event['image'] ) ) {
 				// Attempt to grab the event image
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1014,19 +1015,44 @@ public function insert_posts( $items = array() ) {
...
+				if ( false === $image ) {
+					continue;
+				}
+
+				// Verify for more Complex Errors
+				if ( is_wp_error( $image ) ) {
+					continue;
+				}
+
+				if ( isset( $image->status ) && 'created' === $image->status ) {
```
barryhughes: This could be safe - but we had similar code elsewhere and it was possible that the response (`$image` here) was not a WP error but also didn't contain the expected fields.

So, maybe add a check to ensure status exists unless you're positive that's not a risk?

bordoni: In this case it's not a risk, because it's a thing we are checking on the original Get class.

PR #986, opened 2016-09-19
fix failed import detection

lucatume:
Ticket: https://central.tri.be/issues/66691

* avoid calling queue methods on WP_Error object
* fix the exit status check


bordoni:
🎱

PR #987, opened 2016-09-19
If EA licenses have notices, then consider the EA service inactive

borkweb:
Related: https://github.com/moderntribe/tribe-common/pull/206

See: https://central.tri.be/issues/65378


bordoni:
👍

PR #988, opened 2016-09-19
Clearly indicate that the preview of CSV imports is merely a preview

borkweb:
CSV previews do not contain the enterity of data from the CSV file.  Since that is the case, we need to ensure that the UI doesn't indicate that it does by:

* Removing the searching field
* Removing pagination and pagination helpers
* Removing the number from the import button

See: https://central.tri.be/issues/66580


bordoni:
👍

PR #992, opened 2016-09-20
Test if messages are set before generating notice | spotfix

barryhughes:
Resolves some notices/warnings where various message arrays are not set or are empty.


barryhughes:
May solve [#66808](https://central.tri.be/issues/66808).

bordoni:
👍

PR #994, opened 2016-09-20
Let scheduled imports be deactivated and reactivated

barryhughes:
Allows a scheduled import to be deactivated (so that automatic imports stop happening) and for deactivated imports to be reactivated (restoring the previously scheduled interval).

https://central.tri.be/issues/66838


bordoni:
Looks good

PR #996, opened 2016-09-21
DRY-up interval change code; avoid side effect of updating post_modifed_gmt

barryhughes:
This change has us using an update query instead of `wp_update_post()` when we modify the import interval, because the latter function has a side effect of updating `post_modified_gmt` (which we use to track the last import operation - and so is something we don't want to alter when the interval changes).

[#66838](https://central.tri.be/issues/66838)

### DIFF
--- a/src/Tribe/Aggregator/Cron.php
+++ b/src/Tribe/Aggregator/Cron.php
@@ -89,12 +89,12 @@ class Tribe__Events__Aggregator__Cron {
 			(object) array(
 				'id'       => 'on_demand',
 				'interval' => false,
-				'text'     => esc_html_x( 'On demand', 'aggregator schedule frequency', 'the-events-calendar' ),
+				'text'     => esc_html_x( 'On Demand', 'aggregator schedule frequency', 'the-events-calendar' ),
 			),
 			(object) array(
 				'id'       => 'every30mins',
 				'interval' => MINUTE_IN_SECONDS * 30,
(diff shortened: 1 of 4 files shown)


barryhughes:
Per @bordoni: will update and change to use `Tribe__Events__Aggregator__Record__Abstract::dont_change_post_modified()`, though it may make sense to relocate that to `tribe-common` at this point.

barryhughes:
Ready for another look. Per updates to [#66838](https://central.tri.be/issues/66838) this unwinds a number of the original changes (so we no longer have activation/reactivate row actions, etc).

src/Tribe/Aggregator/Cron.php
```diff
@@ -87,9 +87,9 @@ public function get_frequency( $search = array() ) {
 		 */
 		$found = $schedules = apply_filters( 'tribe_aggregator_record_frequency', array(
 			(object) array(
-				'id'       => 'inactive',
+				'id'       => 'on_demand',
 				'interval' => false,
-				'text'     => esc_html_x( 'Inactive (manual/on demand)', 'aggregator schedule frequency', 'the-events-calendar' ),
+				'text'     => esc_html_x( 'On demand', 'aggregator schedule frequency', 'the-events-calendar' ),
```
borkweb: I think this should be `On Demand` rather than the sentence case of `On demand` as per the Central ticket.

borkweb:
After that one casing change, this is 👍

PR #1004, opened 2016-09-21
Setup the proper data-depends when Facebook has been oauth'd

borkweb:
Simplify a bit of the markup and place the input field within a `<td>`


barryhughes:
:jack_o_lantern:

PR #1005, opened 2016-09-21
Communicate that a preview is a preview to the EA Service

borkweb:
This way we can track hard limits remotely.

See: https://central.tri.be/issues/66891


bordoni:
👍

PR #1006, opened 2016-09-21
Only show the legacy plugin migrator notice on EA pages

borkweb:
I _was_ going to also display the notice on the Legacy Import page, but there's already a huge pile of notices. We can add it if we _want_, but I don't think it is necessary. As soon as the legacy plugin is deactivated, the Legacy Import pages will go away.


bordoni:
👍

PR #1008, opened 2016-09-21
TEC should let tribe-common handle its own deprecated classes | #66872

barryhughes:
Builds on tribe-common pr 210: moves the task of loading tribe-common's deprecated classes from the dependent plugin to tribe-common itself.

https://github.com/moderntribe/tribe-common/pull/210
https://central.tri.be/issues/66872


bordoni:
👍

PR #1012, opened 2016-09-22
Bring over EA Service messages and translate them

borkweb:
This will allow us to cleanly translate the messages from the EA Service


bordoni:
👍

PR #1013, opened 2016-09-22
Add a way to disconnect from Facebook

borkweb:
This way we give folks an "out" _and_ allow us to more easily clear out Facebook tokens for testing :D


src/admin-views/tribe-options-addons-api.php
```diff
@@ -5,6 +5,26 @@
...
+if (
+	isset( $_GET['action'] )
+	&& isset( $_GET['_wpnonce'] )
+	&& 'disconnect-facebook' === $_GET['action']
+	&& wp_verify_nonce( $_GET['_wpnonce'], 'disconnect-facebook' )
+) {
+	tribe_update_option( 'fb_token', '' );
+	wp_redirect( $current_url );
+	die;
+}
```
bordoni: Can we move this to method outside of the `view`? Maybe put this on the `New.php` Tab where the Handling of the Authorization happens. 
Also, we need to delete: `fb_token_expires`and `fb_token_scopes`. Can set it to `null` too?

borkweb: I relocated some stuff to `Tribe__Events__Aggregator__Settings` - it seemed like the better place. As such, this is ready for another look-see.

src/admin-views/tribe-options-addons-api.php
```diff
@@ -5,6 +5,10 @@
  */
 $internal = array();
 
+$current_url = Tribe__Settings::instance()->get_url( array( 'tab' => 'addons' ) );
+
+Tribe__Events__Aggregator__Settings::instance()->handle_clear_fb_credential_submit( $_GET, $current_url );
```
bordoni: We could move this to an Action Right? Like `current_screen`?

borkweb:
Updated.

bordoni:
👍

PR #1016, opened 2016-09-23
Unhook iCal deactivation hook so recurring imports aren't purged

borkweb:
Events Importer: iCal does a silly thing. When you deactivate the plugin, it purges the saved recurring imports with a deactivation hook.  This PR removes iCal's action that does the purging.

See: https://central.tri.be/issues/61940#note-15


barryhughes:
Looks good

PR #1017, opened 2016-09-23
Write away from the flakey array_filter and check more deliberately

borkweb:
See: https://central.tri.be/issues/61940#note-18

### DIFF
--- a/src/Tribe/Aggregator/Migrate.php
+++ b/src/Tribe/Aggregator/Migrate.php
@@ -61,6 +61,9 @@ class Tribe__Events__Aggregator__Migrate {
 			return false;
 		}
 
+		do_action( 'debug_robot', '$this->is_ical_migrated() :: ' . print_r( $this->is_ical_migrated(), true ) );
+		do_action( 'debug_robot', '$this->has_ical_setting() :: ' . print_r( $this->has_ical_setting(), true ) );
+
 		if (
 			( $this->is_facebook_migrated() || ! $this->has_facebook_setting() )
 			&& ( $this->is_ical_migrated() || ! $this->has_ical_setting() )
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Don't leave debug_robot lying around | [Migrate.php:64](https://github.com/moderntribe/the-events-calendar/blob/c4ed8413eb6a1bcb3594ef4c4b858a2ae62380cd/src/Tribe/Aggregator/Migrate.php#L64)
:x: | Don't leave debug_robot lying around | [Migrate.php:65](https://github.com/moderntribe/the-events-calendar/blob/c4ed8413eb6a1bcb3594ef4c4b858a2ae62380cd/src/Tribe/Aggregator/Migrate.php#L65)

bordoni:
:10

src/Tribe/Aggregator/Migrate.php
```diff
@@ -245,6 +245,30 @@ public function is_ical_migrated() {
...
+	 * Filters out empty values
+	 *
+	 * NOTE: we aren't using array_filter because EVEN with an empty() alias, the results are
+	 * unpredictable
+	 *
+	 * @param array $original_values
+	 *
+	 * @return array
+	 */
+	public function filter_out_unwanted_values( $original_values ) {
```
barryhughes: Purely for my own ref I'd love to know more about when and where array_filter was letting us down here. That aside, as a general purpose helper that could be useful in all kinds of places, might it be better housed someplace inside tribe-common?

borkweb: Yeah, something along this vein may have a nice life in tribe-common - but I don't think this iteration as-is is as flexible as we could make it. In the interest of moving to Smoketest, I'm going to leave this here and puzzle over a better tribe-common method at a later date.

PR #1023, opened 2016-09-27
Fix various bugs around setting and updating the Show Map setting in EA

borkweb:
* Fix issues with default settings in EA
* Actually observe the Show Map setting for events
* Handle Show Map settings as if it were being updated from the origin.  In other words, make the Show Map settings observe the change authority settings

See: https://central.tri.be/issues/67228


bordoni:
💯

PR #1024, opened 2016-09-28
Do not register status section unless we are within the help tab [spotfix]

barryhughes:
Avoids registering the EA status section of the help page when something other than the help page has been registered. This helps prevent an HTTP request (to check the status) being made on each request.


bordoni:
👍

PR #1025, opened 2016-09-28
When FB creds are NOT SET, include a hidden field indicating as much

borkweb:
Without the field, `dependency.js` flips its lid and goes crazy-recursive

See: https://central.tri.be/issues/67214


bordoni:
Sounds good.

PR #1027, opened 2016-09-29
Prefer the activity() getter function over naked access to the (possibly null) property

barryhughes:
`$queue->activity` may be null; prefer `$queue->activity()` instead. Thanks to Jeff for [catching this!](https://tribe.slack.com/archives/quality-assurance/p1475188920001344)


barryhughes:
... moving beyond this, maybe we ought to make `activity` (the property) protected or private and/or implement it as a `@property-read` field via `__get()` - maybe something for us to consider more generally though during the next code review meet :philosoraptor:

bordoni:
@barryhughes I'm 100% towards us using more and more `__get()` on our plugins.

borkweb:
:+1:

PR #1031, opened 2016-10-01

elimn:

- Added call to tribe_register_plugin()
- Cleaned boostrap file to put this in line with our other plugins


tr1b0t:

elimn:
I just realized this should not exist yet. We had planned merge this into 4.3.1, but that branch does not exist yet. And it's possible that if you 4.3 scheduled changed someone would want to slip this in. Obviously we should wait on the PR until we know what release this should go in.

PR #1035, opened 2016-10-05
get_daily_events() is an instance method, call it appropriately | spotfix

barryhughes:
`get_daily_events()` is not a static method.


bordoni:
👍

PR #1036, opened 2016-10-05
If tribe_event_category is present in the baseurl, then keep it

borkweb:
Related: https://github.com/moderntribe/events-pro/pull/408

See: https://central.tri.be/issues/44800#note-57


src/resources/js/tribe-events.js
```diff
@@ -1253,18 +1253,42 @@ Date.prototype.format = function( mask, utc ) {
...
 				// failover to the baseurl of the event header
-				tribe_ev.state.params += '&baseurl=' + $header.data( 'baseurl' );
+				url = $header.data( 'baseurl' );
 			} else {
 				// use the current URL as a last ditch effort
-				tribe_ev.state.params += '&baseurl=' + window.location.origin + window.location.path;
+				url = window.location.origin + window.location.path;
 			}
 
+			tribe_ev.state.params += '&baseurl=' + url;
```
barryhughes: Should we test if `url` is null before concatenating `baseurl`? Or not matter?

barryhughes: ahem, ignore me

PR #1041, opened 2016-10-06
minor language change + small translation fix

zbtirrell:


src/admin-views/admin-welcome-message.php
```diff
@@ -16,18 +16,18 @@
...
 	</div>
 	<div class="tribe-half-column">
 		<h2><?php esc_html_e( 'Newsletter Signup', 'the-events-calendar' ); ?></h2>
-		<p><?php esc_html_e( 'Stay in touch with The Events Calendar Pro. We send out periodic updates, key developer notices, and even the occasional discount.', 'the-events-calendar' ); ?></p>
+		<p><?php esc_html_e( 'Stay in touch with The Events Calendar team. We send out periodic updates, key developer notices, and even the occasional discount.', 'the-events-calendar' ); ?></p>
 		<form action="http://moderntribe.createsend.com/t/r/s/athqh/" method="post">
-			<p><input id="listthkduyk" name="cm-ol-thkduyk" type="checkbox" /> <label for="listthkduyk">Developer News</label></p>
-			<p><input id="listathqh" name="cm-ol-athqh" checked type="checkbox" /> <label for="listathqh">News and Announcements</label></p>
-			<p><input id="fieldEmail" class="regular-text" name="cm-athqh-athqh" type="email" placeholder="Email" required /></p>
+			<p><input id="listthkduyk" name="cm-ol-thkduyk" type="checkbox" /> <label for="listthkduyk"><?php esc_html_e( 'Developer News', 'the-events-calendar' ); ?></label></p>
```
barryhughes: I realize they are autogenerated but we could (optionally) clean up the ID attribute values for these fields. But, if we don't care too much that is also fine.

PR #1050, opened 2016-10-11
Prevent notice when executing CSV import and created|updated|skipped = 0

borkweb:
Found this while hunting down a CSV state/province import bug


barryhughes:
:keyboard:

PR #1057, opened 2016-10-12
Failover to non-lowercase labels if lowercase not set

borkweb:
Also:

* Only show the "Show All" link if there's a page for viewing posts of that type (example: Tickets don't have a browsable page)
* Use the post type name being imported rather than always using "Events" when nothing has been imported/updated

See: https://central.tri.be/issues/67878


zbtirrell:
👍

PR #1060, opened 2016-10-12
Move tribe_assets for EA to init so csv_mappings can be filtered by ET+

borkweb:
See: https://central.tri.be/issues/67915


barryhughes:
:megusta:

PR #1062, opened 2016-10-12
Upgrade/compatibility spotfix

barryhughes:
Adds safety checks to avoid possible _plugin cannot be activated due to a fatal error_ problems when ET 4.2.7 is active and an attempt is made to activate TEC 4.3. Bit of an edge case, but leaving this here so we can merge if we want to.


bordoni:
👍

PR #1068, opened 2016-10-14
Make the Changes to prevent Duplicates from wrong FacebookID

bordoni:
_Ref:_ [C#68031](https://central.tri.be/issues/68031)

### DIFF
--- a/src/Tribe/Aggregator/Event.php
+++ b/src/Tribe/Aggregator/Event.php
@@ -168,9 +168,9 @@ class Tribe__Events__Aggregator__Event {
 			$keys[] = $key;
 			$keys[] = "_{$fields[ $origin ]['legacy']}";
 
-			$sql .= 'AND meta_key IN ( "' . implode( '", "', $keys ) .'" )';
+			$sql .= 'AND meta_key IN ( "' . implode( '", "', array_map( 'esc_sql', $keys ) ) .'" )';
 		} else {
-			$sql .= 'AND meta_key = ' . esc_sql( $key );
+			$sql .= 'AND meta_key = "' . esc_sql( $key ) . '"';
 		}
 
 		return $wpdb->get_results( $sql, OBJECT_K );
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/Event.php
```diff
@@ -158,11 +158,22 @@ public function get_existing_ids( $origin, $values ) {
...
 
-		return $wpdb->get_results( $wpdb->prepare( $sql, $key ), OBJECT_K );
+		/**
+		 * Allows us to check for legacy meta keys
+		 */
+		if ( ! empty( $fields[ $origin ]['legacy'] ) ) {
+			$keys[] = $key;
+			$keys[] = "_{$fields[ $origin ]['legacy']}";
+
+			$sql .= 'AND meta_key IN ( "' . implode( '", "', $keys ) .'" )';
```
barryhughes: Worth passing `$keys` through `esc_sql()` first, or safe as is?

src/Tribe/Aggregator/Event.php
```diff
@@ -158,11 +158,22 @@ public function get_existing_ids( $origin, $values ) {
...
+		} else {
+			$sql .= 'AND meta_key = ' . esc_sql( $key );
```
barryhughes: Should this also be single quoted besides being escaped?

src/Tribe/Aggregator/Event.php
bordoni: We could. Would make it safer, just gotta change the `implode`/

src/Tribe/Aggregator/Event.php
bordoni: `esc_sql` does that.

bordoni: Actually no, it's `$wpdb->prepare` that does that.

barryhughes:
:noice:

PR #1069, opened 2016-10-14
Feature/66436 list widget featured events

barryhughes:
Add featured event support to the (core) list widget.

https://central.tri.be/issues/66436


bordoni:
👍

PR #1071, opened 2016-10-14
Feature/65897 set as featured

barryhughes:
Update the _Event Options_ meta box to include a featured event checkbox.


barryhughes:
(Also reformats the _Event Options_ meta box to use `<p>` tags instead of `<br>` tags.)

borkweb:
:+1:

PR #1072, opened 2016-10-15
Add documentation for EventURL on tribe_create_event

bordoni:
_Ref:_ [C#44871](https://central.tri.be/issues/44871)


barryhughes:
Though a minor change, we should add a changelog entry thanking Keith (reporting user).

src/functions/advanced-functions/event.php
```diff
@@ -32,6 +32,7 @@
 	 * - EventShowMapLink bool - Set to true to display a link to the map in the Event view.
 	 * - EventShowMap bool - Set to true to embed the map in the Event view.
 	 * - EventCost string - Default cost of the Event.
+	 * - EventURL string - Link to the Event Website or Third-Party page
```
bordoni: Yeah maybe, I think it's worth kinda taking a second look at this docblock and use the WordPress Format here. I will log this for another MR, where we have more spare time. Sounds good?

barryhughes: Sure.

PR #1075, opened 2016-10-17

elimn:
In #66657 we are calling tribe_register_plugin() from each of our current plugins. We pass the full file path to the main plugin file, the main class name and plugin version. To facilitate that I added a plugin file constant to some of our bootstrap files, mimicking the elegant standard set in Event Tickets.


elimn:
Doh! I meant to open this for r/4.3.1.

PR #1076, opened 2016-10-17

elimn:
In #66657 we are calling tribe_register_plugin() from each of our current plugins. We pass the full file path to the main plugin file, the main class name and plugin version. To facilitate that I added a plugin file constant to some of our bootstrap files, mimicking the elegant standard set in Event Tickets.


the-events-calendar.php
```diff
@@ -27,11 +27,10 @@
 Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 */
 
+define( 'THE_EVENTS_CALENDAR_DIR', dirname( __FILE__ ) );
```
bordoni: I would avoid two variables. Just the file would do. You can Create the other variable on the Class no need to store one more constant.

the-events-calendar.php
```diff
@@ -27,11 +27,10 @@
...
+define( 'THE_EVENTS_CALENDAR_FILE', __FILE__ );
```
bordoni: Can I suggest that we stick with PHP Std for these type of Constants? `__TEC_FILE__`

src/Tribe/Main.php
```diff
@@ -283,6 +290,21 @@ public function maybe_set_common_lib_info() {
...
+		 * @return bool Indicates if Tribe Common wants the plugin to run
+		 */
+		public function register_active_plugin() {
+			if ( ! function_exists( 'tribe_register_plugin' ) ) {
+				return true;
+			}
+
+			return tribe_register_plugin( THE_EVENTS_CALENDAR_FILE, __CLASS__, self::VERSION );
+		}
+
```
bordoni: extra Space

the-events-calendar.php
elimn: Interesting idea. Would you include the abbreviation? Eitherway it is probably unique enough to not conflict in the global namespace. 

Many of our plugins already have the constant present in the same format as I used.

barryhughes: Fwiw, I'm not too big on `__MAGIC__` style notation here, because our own constants will never be magical (unlike `__FILE__` etc, which in a sense changes value depending on where you call it from).

src/Tribe/Main.php
bordoni: Empty new line yeah =D

PR #1079, opened 2016-10-18
Update the package whitelist  to include compser files

lucatume:
This PR whitelists the composer and di52 dependencies we need in production.

Please check my globbing is correct:

* `vendor/composer` only contains files
* `vendor/lucatume/di52/src` contains folders


borkweb:
:+1:

borkweb:
Ok. I hadn't looked at the `vendor/` directories close enough the first time. Now I'm good with a :+1:

PR #1082, opened 2016-10-19
Fetch locale translation for bases

lucatume:
Ticket: https://central.tri.be/issues/68062

This PR handles the case where we want to suport translated version of our rewrite bases, `/all` in the ticket instance but it would apply to *any*, when a site language is not English.  

What we were previously relying on WPML to do can be done by WPML *and* the available translations.  
In case you are wondering the WPML integration fires off on the `tribe_events_rewrite_i18n_slugs_raw` filter.


tr1b0t:

bordoni:
Check my comment above @lucatume

lucatume:
removed the tickets ref, thanks for keeping track @bordoni

bordoni:
👍

PR #1083, opened 2016-10-19
Avoid inserting null meta values into postmeta for EA Records

borkweb:
See: https://central.tri.be/issues/68290

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -139,7 +139,14 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 	 */
 	public function update_meta( $key, $value ) {
 		$this->meta[ $key ] = $value;
-		return update_post_meta( $this->post->ID, self::$meta_key_prefix . $key, $value );
+
+		$field = self::$meta_key_prefix . $key;
+
+		if ( null === $value ) {
+			return delete_post_meta( $this->post->ID, $field );
+		}
+
+		return update_post_meta( $this->post->ID, $field, $value );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -139,7 +139,14 @@ public function setup_meta( $meta ) {
 	 */
 	public function update_meta( $key, $value ) {
 		$this->meta[ $key ] = $value;
-		return update_post_meta( $this->post->ID, self::$meta_key_prefix . $key, $value );
+
+		$field = self::$meta_key_prefix . $key;
+
+		if ( null === $value ) {
```
barryhughes: You've got a better idea of what's happening in here than I do - but could it be worth testing if the value is empty rather than strictly null?

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -301,6 +308,21 @@ public function prep_post_args( $type, $args, $meta = array() ) {
...
+				continue;
+			}
+
+			// trim scalars
+			if ( is_scalar( $value ) ) {
+				$value = trim( $value );
+			}
+
+			// if the value is blank or null, let's avoid inserting it
+			if ( null === $value || '' === $value ) {
```
barryhughes: Could `empty()` be a better choice here? Might also let us ditch one `if()` block.

borkweb: We don't want to ditch values that are `0`

src/Tribe/Aggregator/Record/Abstract.php
borkweb: We only want to delete the meta if it is `null`. We'll not delete if it is a blank string or `0`.

PR #1084, opened 2016-10-19
Prevents On Demand Schedule to be triggered on Cron

bordoni:
_Ref:_ [C#68284](https://central.tri.be/issues/68284)


borkweb:
:+1:

PR #1086, opened 2016-10-19
Prune of Records for Event Aggregator, on 'tribe_aggregator_cron'

bordoni:
_Ref:_ [C#68283](https://central.tri.be/issues/68283)


src/Tribe/Aggregator/Cron.php
```diff
@@ -384,6 +385,56 @@ public function verify_fetching_from_service() {
 	}
 
 	/**
+	 * Checks if any Child Record needs to be created, this will run on the Cron every 15m
+	 *
+	 * @return void
```
barryhughes: May return boolean false.

bordoni:
@barryhughes ready for another look. Removed the return false, which is not needed.

barryhughes:
:banana:

PR #1087, opened 2016-10-19
Remove soft daily limit. Only attempt imports if daily limit unreached

borkweb:
Additionally, we are displaying a new error message on the Scheduled Imports tab if the last run scheduled import failed.

See: https://central.tri.be/issues/68286


barryhughes:
:trackball:

PR #1090, opened 2016-10-20
Remove two null overrides that were there for debug purposes

borkweb:
Additionally, make sure the origin gets pretty-fied.

See: https://central.tri.be/issues/68288


zbtirrell:
❤️

PR #1091, opened 2016-10-24
Dynamic Helper Text

jesseeproductions:
_Ref:_ [C#66484](https://central.tri.be/issues/66484)


src/Tribe/Asset/PHP_Date_formatter.php
```diff
@@ -0,0 +1,14 @@
...
+
+class Tribe__Events__Asset__PHP_Date_Formatter extends Tribe__Events__Asset__Abstract_Asset {
+
+	public function handle() {
+
+		$path = Tribe__Events__Template_Factory::getMinFile( $this->vendor_url . 'php-date-formatter/js/php-date-formatter.js', true );
+
+		$script_handle = $this->prefix . '-php-date-formatter';
+		wp_enqueue_script( $script_handle, $path, 'jquery', '1.3.4' );
+		Tribe__Events__Template_Factory::add_vendor_script( $script_handle );
```
barryhughes: If this is already in place and I've missed it, ignore me :smiley: ...but do we need to localize this script (ie to provide translated versions of [these defaults](https://github.com/kartik-v/php-date-formatter/blob/master/js/php-date-formatter.js#L56-L64))?

src/resources/css/events-admin.css
```diff
@@ -88,6 +88,11 @@
 #ticket_end_date {
 	width: 100px;
 }
+.event-dynamic-helper-text {
```
barryhughes: Also have gulp minify this asset?

src/Tribe/Asset/Admin.php
```diff
@@ -1,21 +1,86 @@
...
 		$path = Tribe__Events__Template_Factory::getMinFile( tribe_events_resource_url( 'events-admin.js' ), true );
 
 		wp_enqueue_script( $this->prefix . '-admin', $path, $deps, $this->filter_js_version(), true );
+
+		$data = array(
+			'date_with_year'    => tribe_get_date_option( 'dateWithYearFormat', Tribe__Date_Utils::DBDATEFORMAT ),
+			'date_no_year'      => tribe_get_date_option( 'dateWithoutYearFormat', Tribe__Date_Utils::DBDATEFORMAT ),
+			'datepicker_format' => Tribe__Date_Utils::datepicker_formats( tribe_get_option( 'datepickerFormat' ) ),
+			'days'              => array(
+				__( 'Sunday', 'the-events-calendar' ),
```
barryhughes: Though using our own textdomain is almost always the right decision, might it be better in this case to ditch it and so take advantage of the date translations used by WP itself?

If you have already discussed and dismissed that approach, feel free to ignore :-)

jesseeproductions: Are you talking about date_i18n? And calling it for each name?

barryhughes: I was actually just thinking drop the textdomain and use WP's translations - ie `__( 'Monday' )`, but I suppose there could be other ways to do the same thing.

jesseeproductions: I think dropping the textdomain if that works is the way to go.

src/resources/js/events-admin.js
```diff
@@ -774,4 +774,186 @@ jQuery( document ).ready( function( $ ) {
...
+		return false;
+	} );
+} );
+
+/**
+ * Event Date and Time Dynamic Helper Text
+ *
+ * @var object tribe_dynamic_helper_text
+ */
+var tribe_dynamic_helper_text = tribe_dynamic_helper_text || {};
```
jesseeproductions: 👍

src/Tribe/Asset/Admin.php
barryhughes: You could probably also do something like `$wp_locale->weekday[0]` (to get Monday) etc but I'm not sure that offers a great advantage, also isn't as readable, to me at least. I think I'd stick with just dropping the textdomain.

src/resources/js/events-dynamic.js
```diff
@@ -0,0 +1,155 @@
...
+
+	obj.field_class = '.event-dynamic-helper-text';
+	obj.date_fmt_settings = {
+     dateSettings: {
+         days: tribe_dynamic_help_text.days,
+         daysShort: tribe_dynamic_help_text.daysShort,
+         months: tribe_dynamic_help_text.months,
+         monthsShort: tribe_dynamic_help_text.monthsShort,
+     }
+ };
```
borkweb: Tabbing is weird for `obj.date_fmt_settings`

src/Tribe/Asset/Admin.php
```diff
@@ -1,21 +1,24 @@
...
 			$this->deps,
-			array(
-				'jquery',
-				'jquery-ui-datepicker',
-				'jquery-ui-sortable',
-				'tribe-bumpdown',
-				'underscore',
-				'wp-util',
-			)
+				array(
```
barryhughes: Sorry for not noticing this earlier - but the indentation is odd here: the second param should be tabbed back, ie:

```
$array_merge(
    $first_array,
    array (
        ...
    )
);
```

src/resources/js/events-dynamic.js
```diff
@@ -0,0 +1,155 @@
...
+	/**
+	 * Determine Message to Use based on Date and Time
+	 */
+	obj.msg_logic = function () {
+
+		//single date, different start and end time
+		if ( obj.start_date == obj.end_date && !obj.all_day && obj.start_time != obj.end_time ) {
+			obj.dynamic_text = obj.text[0];
+
+			//single date, same start and end time
```
barryhughes: The indentation is kind of odd here - I see that the comment is aligned with the inner contents of the `else if() {}` but it probably shouldn't be since we haven't yet entered that block (or, the comment could be moved inside the block).

barryhughes: Same applies for each of the else if comments that follow.

borkweb:
Did GitHub change how it displays submodule updates? I ask because it looks like `vendor/tickets` maybe be getting a hash commit at the bottom of the changesets. Is that what is happening there? If so, is that intended?

borkweb:
If that _is_ what is happening and it _is_ intended, then this is 🏒

jesseeproductions:
Not sure what happened, but the versions now match

borkweb:
Merge away!

PR #1093, opened 2016-10-25
implement support for asset aliasing

lucatume:
Ticket: https://central.tri.be/issues/43605

This PR solves the issue of two select2 scripts being queued on events admin and augments the `Asset` class to spot, track and use compatible script aliases to avoid this kind of conflict.

If the idea is good adding support for styles too is not difficult (although probably not useful).


borkweb:
It looks like the only thing committed here are `composer.lock` and files from `tests/`. Should there be more?

lucatume:
Yep, the actual files had not been pushed...

lucatume:
No, wait: let's not blame git. I had not pushed the files...

borkweb:
:+1:

PR #1097, opened 2016-10-26
bug/61742-search-queries-URL-encoded

trishasalas:
https://central.tri.be/issues/61742


src/resources/js/tribe-events.js
```diff
@@ -865,11 +865,11 @@ Date.prototype.format = function( mask, utc ) {
 				$tribe_bar.tribe_clear_form();
 			}
 
-			params = tribe_ev.fn.parse_string( params );
+			params = tribe_ev.fn.parse_string( decodeURIComponent( params.replace( /\+/g, '%20' ) ) );
```
borkweb: I still believe that this should be done within the `$.each()` on `value` but only when needed (text/hidden fields).

src/resources/js/tribe-events.js
```diff
@@ -865,11 +865,11 @@ Date.prototype.format = function( mask, utc ) {
...
 
 			$.each( params, function( key, value ) {
 				if ( key !== 'action' ) {
-					var name = decodeURI( key ),
+					var name = key,
```
borkweb: We should retain `decodeURI()` for the `key`

src/resources/js/tribe-events.js
borkweb: Looking at the contents of `value`, it appears that that variable is an array. So doing something like the following would be handy:

```js
if ( Array.isArray( value ) ) {
  value = decodeURIComponent( value[0].replace( /\+/g, '%20' ) );
}
```

borkweb:
:+1:

PR #1098, opened 2016-10-26
removed redundant meta key

lucatume:
Ticket: https://central.tri.be/issues/67495

This PR removes the `_EventStartDate` query args from the query.
At that point the events have been already pruned and sorted by the `_EventStartDate` meta and doing it again causes those that start at the start of day cutoff to "disappear".

This could very much use a second pair of eyes.


bordoni:
👍

PR #1099, opened 2016-10-28
Now ICS has it's default settings been returned correctly

bordoni:
_Ref:_ [C#67885](https://central.tri.be/issues/67885)


barryhughes:
:jack_o_lantern: 

Good to merge (though we'll probably need a changelog entry also).

PR #1100, opened 2016-10-28
Makes sure Tooltips are functional on Week and Month view

bordoni:
_Ref:_ [C#68094](https://central.tri.be/issues/68094)


barryhughes:
:bat:

Good to merge (though looks like we still need a changelog entry).

PR #1113, opened 2016-11-03
it's `tec.` not `events.` for TEC

lucatume:
Meh, silly me.


lucatume:
Because my mind plays tricks on me.

PR #1118, opened 2016-11-05
Timepicker defaults

bordoni:
_Ref:_ [C#66487](https://central.tri.be/issues/66487)

### DIFF
--- a/src/Tribe/Admin/Event_Meta_Box.php
+++ b/src/Tribe/Admin/Event_Meta_Box.php
@@ -202,7 +202,7 @@ class Tribe__Events__Admin__Event_Meta_Box {
 		// Fetch Status to check what we need to do
 		$status = get_post_status( $this->event->ID );
 
-		if ( ! is_string( $status ) || 'auto-draft' === $status ) {
+		if ( ! $status || 'auto-draft' === $status ) {
 			return true;
 		}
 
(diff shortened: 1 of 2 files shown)


src/Tribe/Admin/Event_Meta_Box.php
```diff
@@ -183,6 +184,70 @@ protected function set_end_date_time() {
...
+	public function is_auto_draft() {
+		if ( ! $this->event instanceof WP_Post ) {
+			return true;
+		}
+
+		// Fetch Status to check what we need to do
+		$status = get_post_status( $this->event->ID );
+
+		if ( ! is_string( $status ) || 'auto-draft' !== $status ) {
+			return true;
```
barryhughes: Seems like this will have us returning true if the status is `'publish'` and returning false if the status is `'auto-draft'` - we should either change the if condition or else flip the return statements.

src/Tribe/Admin/Event_Meta_Box.php
```diff
@@ -202,7 +202,7 @@ public function is_auto_draft() {
 		// Fetch Status to check what we need to do
 		$status = get_post_status( $this->event->ID );
 
-		if ( ! is_string( $status ) || 'auto-draft' !== $status ) {
+		if ( ! is_string( $status ) || 'auto-draft' === $status ) {
```
barryhughes: Now we're returning true here, is there still value in the `is_string()` test - especially as we're using `===` strict comparison?

bordoni: If it doesn't return a String it Should return `true`.

bordoni:
It's now on Code Review again.

barryhughes:
:camel:

PR #1119, opened 2016-11-06
Allowing for the usage of the timepicker field in other plugins

borkweb:
See: https://central.tri.be/issues/66713


bordoni:
Remove Common from this PR right?

borkweb:
That common hash has is the corresponding hash commit associated with this PR.

PR #1120, opened 2016-11-07
restored ical template tag load

lucatume:
Fix an error due to no more loaded ical template tag; I've also removed a deprecated function.


barryhughes:
:large_blue_diamond:

PR #1123, opened 2016-11-08
Use 'tribe-event-featured' CSS class for featured events | #65895

barryhughes:
By default, apply `.tribe-event-featured` if the event is featured.

https://central.tri.be/issues/65895


borkweb:
:+1:

PR #1124, opened 2016-11-08
Style featured events in the list widget

borkweb:
Here's an example of the styles:

screen shot 2016-11-08 at 1 07 52 pm

I've additionally made a fix to the List Widget "show only featured events" default.

See: https://central.tri.be/issues/65923


barryhughes:
:dragon:

PR #1130, opened 2016-11-09
#66484 Add Trigger Event on Time Picker Init

jesseeproductions:
_Ref:_ [C#66484 ](https://central.tri.be/issues/66484 ) - Add Trigger for Helper Text to detect setup of Time Picker


bordoni:
💯

PR #1132, opened 2016-11-09
If 'featured' option is selected, auto-select 'sticky' | #65897

barryhughes:
Per the relevant user stories, if a user checks the 'feature event' checkbox, let's automatically check the 'sticky in month view' checkbox (they can manually de-select if they wish).

Note: rather than cram more code into `events-admin.js` I've created a new file that is only interested in the event post editor with the intent that we add more to it and pull across any other event editor-specific items going forward.


src/Tribe/Main.php
```diff
@@ -1991,6 +1991,7 @@ public function add_admin_assets() {
 				Tribe__Events__Template_Factory::asset_package( 'ecp-plugins' );
 
 				if ( $admin_helpers->is_post_type_screen( self::POSTTYPE ) ) {
+					Tribe__Events__Template_Factory::asset_package( 'event-editor' );
```
bordoni: Can we use the new `tribe_asset` functions? Instead of this one?

src/resources/js/event-editor.js
```diff
@@ -0,0 +1,24 @@
...
+
+	var $sticky_in_month_view_checkbox = $( 'input[name="EventShowInCalendar"]' ),
+	    $featured_event_checkbox = $( 'input[name="feature_event"]' );
+
+	/**
+	 * If the 'feature event' box is checked, automatically check the
+	 * sticky-in-month-view box also.
+	 */
+	obj.auto_enable_sticky_field = function() {
+		if ( $( this ).prop( 'checked' ) ) {
```
bordoni: Can we use the inverted logic with `return`?

barryhughes: Return early if it isn't checked? I mean, we could, but in this case that would be more verbose as we're only interested in performing a single operation when the condition is met.

src/Tribe/Main.php
barryhughes: Updated!

src/resources/js/event-editor.js
```diff
@@ -0,0 +1,24 @@
+var tribe_events_event_editor = tribe_events_event_editor || {};
+
+/**
+ * Implements behaviours that are specific to the event editor.
+ */
+( function( obj, $ ) {
+
+	var $sticky_in_month_view_checkbox = $( 'input[name="EventShowInCalendar"]' ),
```
borkweb: As long as this file is enqueued in the footer, this should work fine. But in the event that it is moved to the header, the input fields may not exist on the page. My suggestion would be to add an `obj.init()` method and fire that off during document.ready.

borkweb:
⚡️

PR #1135, opened 2016-11-10
Avoids using LocalStorage on Private mode for OSX and iOS

bordoni:
_Ref:_ [C#40196](https://central.tri.be/issues/40196)


borkweb:
:+1:

PR #1137, opened 2016-11-10
Fix select2 script handle

lucatume:
Ticket: https://central.tri.be/issues/67122

The recurrence script requires `dropdowns.js` and that, in turn, reuqires `select2.js`; due to the improper naming `select2` was not being qued.


src/Tribe/Asset/Select2.php
```diff
@@ -21,9 +21,8 @@ public function handle() {
 		// we know of other plugins loading a version of select2 compatible with our needs
 		// let's not queue the script twice.
 		if ( ! $this->has_script_alias( 'select2' ) ) {
-			$script_handle = $this->prefix . '-select2';
-			wp_enqueue_script( $script_handle, $path, 'jquery', '3.2' );
-			Tribe__Events__Template_Factory::add_vendor_script( $script_handle );
+			wp_enqueue_script( 'select2', $path, 'jquery', '3.2' );
+			Tribe__Events__Template_Factory::add_vendor_script( 'select2' );
```
borkweb: With this change, I wonder if we even need the wrapping if statement. That was added by @bordoni to allow us to have a prefixed version of select2.

That being said, I wonder what the merit of having our own prefixed select2 really is. It seems kind of funky because there's no case where we'd want the script enqueued twice as far as I know (unless there's a no-conflict for select2 3.x and 4.x).

bordoni: Wait, 4.x? No we can't use that version, we are using 3.4.x

lucatume: I have no idea what you are talking about; the handle I've changed as `events-recurrence.js` would not q due to it, it does with this modification.
4.x what?

lucatume: "You" as in "you both"

lucatume:
this should be good to review, builds on https://github.com/moderntribe/tribe-common/pull/266

borkweb:
Can you add a tweak to `package-whitelist.json` to support the relocation?

lucatume:
done, removed select2 submodule and fixed the whitelist

package-whitelist.json
```diff
@@ -24,6 +24,8 @@
   "common/vendor/autoload_52.php",
   "common/vendor/composer/*",
   "common/vendor/lucatume/di52/src/**",
+  "common/vendor/select2/select2*.js",
+  "common/vendor/select2/select2.css",
```
borkweb: Oops! We need to also include the images.

Moving the `vendor/select2/*` lines from below up to here and prepending `common/` should catch the items that are missing here.

lucatume:
done!

borkweb:
:+1:

PR #1141, opened 2016-11-10
Set main events view as homepage (Feature #38757) 

barryhughes:
Ported from Event Rocket: https://central.tri.be/issues/38757


src/resources/js/front-page-setting.js
```diff
@@ -0,0 +1,27 @@
+var tribe_events_front_page_setting = tribe_events_front_page_setting || {};
```
bordoni: Can you create a structure like:

```js
var tribe_events_front_page_setting = tribe_events_front_page_setting || {};
( function( $, obj ) {
    'use strict';

} )( jQuery, tribe_events_front_page_setting );
```

barryhughes: I can - is there a clear advantage?

bordoni: Because you can apply the 'use strict' to the whole file without affecting anything else. And normally with this you prevent 1 script breaking the whole page.

PR #1147, opened 2016-11-11
Avoid breakages when front-page-events are enabled alongside Community Events

barryhughes:
The global `$wp_query` object for CE pages passes the `is_home()` test. This can lead to breakages when front page events are enabled, which this change aims to fix.

https://central.tri.be/issues/38757


bordoni:
👍

PR #1148, opened 2016-11-11
Update select2 slug for tribe-ea-fields dependencies | spotfix

barryhughes:
Further fix in the saga of select2 problems.

See also: https://central.tri.be/issues/69468


bordoni:
👍

PR #1151, opened 2016-11-14
Prevent notice on List Widget when there are no upcoming events

borkweb:


barryhughes:
:snowman:

barryhughes:
...Turns out (because we're extracting with the `EXTR_SKIP` flag) that this means `$no_upcoming_events` can only ever be true 😢

PR #1155, opened 2016-11-14
Restore enqueuing of event-editor.js in event edit screen | #65897

barryhughes:
Restores line responsible for enqueuing `event-editor.js` (which was clobbered by a later commit).


bordoni:
👍

PR #1156, opened 2016-11-14
Don't append /featured/ to single event slugs | #68051

barryhughes:
Do not modify single event slugs/URLs in the context of featured event archives.

https://central.tri.be/issues/68051


bordoni:
👍

PR #1162, opened 2016-11-17
add more styles for delete this button and confirmation

lucatume:
Ticket: https://central.tri.be/issues/67110 

This PR extends the styles of the `.tribe-delete-this` class to allow for an additional `.tribe-confirm-delete-this` class to use on related confirmation buttons.  
The PR also adds support for an `.hover-state` class to allow for transitional "hover-like" states.

An example below.

2016-11-17 at 14 15


barryhughes:
Looks good - do we need a changelog entry though?

lucatume:
this will go in PRO, the changelog will be there

PR #1163, opened 2016-11-17
Fix/69246 datepicker aesthetics

barryhughes:
Hide the button panel from the datepicker, style today as per the current _selected_ day.

No changelog entry added as this is essentially a child of [#66550](https://central.tri.be/issues/66550).

https://central.tri.be/issues/69246 (Item 5)


bordoni:
👍

PR #1164, opened 2016-11-18
Supply the datepicker with *full* month names (for the month dropdown)

barryhughes:
Supply the datepicker with _full_ month names in place of short month names (we want the full month names to render in the month selector dropdown within that widget).

No changelog entry added as this is essentially a child of [#66550](https://central.tri.be/issues/66550).

https://central.tri.be/issues/69246 (Item 1)


borkweb:
🔢

PR #1166, opened 2016-11-18
select2x2 path (spotfix)

barryhughes:
Our (post-)CSS was referencing a Select2 asset that no longer exists in that location.


borkweb:
📦

PR #1167, opened 2016-11-18
Re-add text that shouldn't have been removed. Whoopsie!

borkweb:
See: https://central.tri.be/issues/69243#note-5


barryhughes:
:ear_of_rice:

PR #1168, opened 2016-11-18
Improve alignment when event meta box inputs are arranged horizontally

barryhughes:
When inputs/selects are next to one another in the event meta box the alignment is suspect. This change lets us apply `.alignleft` where needed to get a better result.


borkweb:
:+1:

PR #1169, opened 2016-11-18
Change Customizer helper text single paragraph into two on List Settings

borkweb:
See: https://central.tri.be/issues/69243


barryhughes:
:ear_of_rice:

PR #1172, opened 2016-11-21
Adjust to use the EndDate instead of StartDate...

bordoni:
_Ref:_ [C#69688](https://central.tri.be/issues/69688)


borkweb:
:omg:

PR #1174, opened 2016-11-22
Datepicker: better differentiation between 'today' and the selected date

barryhughes:
'Today' and the current selected date in the datepicker share the same background colour. That can be confusing and might suggest to a user that a 'range' of dates have been selected. To avoid this lets:

* Use a light grey for 'today'
* Keep the current blue for the actual selected date
* Have the blue selected date styles trump over 'today', if they are the same date


borkweb:
💯

PR #1175, opened 2016-11-22
Use date_i18n() to get the date in the current site timezone | #69686

barryhughes:
When setting up the datepicker, if a date hasn't been selected let's use `date_i18n()` rather than `date()` to avoid discrepancies between the local timezone and UTC.


borkweb:
🔢

PR #1176, opened 2016-11-22
Datepicker: widen month select now it contains a full month | #69686

barryhughes:
The datepicker's month selector now contains the full month name rather than a three-letter abbreviation, and so needed to be widened.


bordoni:
👍

PR #1177, opened 2016-11-22
The main start/end datepickers generally should not care about changes made to other datepickers

barryhughes:
When PRO is enabled and a recurrence rule datepicker was set to some value, the end date for the event itself was being set to that same value. This change better isolates those datepickers to avoid that.

Additionally renamed some vars to use snake_case.


borkweb:
:+1:

PR #1180, opened 2016-11-23
fix the return value when a URL is expected

zbtirrell:
See: https://central.tri.be/issues/69821

As there isn't currently a `release/4.3.4` branch, I have PR'd this against `develop`. It should either get re-targeted if we plan to do 4.3.3 OR move the ticket and go out with the 4.4 changes.

### DIFF
--- a/src/functions/template-tags/organizer.php
+++ b/src/functions/template-tags/organizer.php
@@ -243,6 +243,11 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 				return apply_filters( 'tribe_get_organizer_link', $link, $postId, $full_link, $url );
 			}
 		}
+
+		if ( ! $full_link ) {
+			return '';
+		}
+
 		//Return Organizer Name if Pro is not Active
 		return tribe_get_organizer( $org_id );
 	}
--- a/src/functions/template-tags/organizer.php
+++ b/src/functions/template-tags/organizer.php
@@ -211,11 +211,11 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	/**
 	 * Organizer Page Link
 	 *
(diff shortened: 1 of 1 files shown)


src/functions/template-tags/organizer.php
```diff
@@ -243,6 +243,11 @@ function tribe_get_organizer_link( $postId = null, $full_link = true, $echo = fa
 				return apply_filters( 'tribe_get_organizer_link', $link, $postId, $full_link, $url );
 			}
 		}
+
+		if ( ! $full_link ) {
```
barryhughes: I'm _slightly_ concerned by this change.

* It doesn't take account of `$echo` (though really that's a structural problem with this function)
* If `$echo` *and* `$full_link` are both false the result will always be the same - an empty string - even when it could return the organizer name, but unlinked

Looking over the [linked issue](https://central.tri.be/issues/69821) to get a better handle on things, the experience reported there is also at odds with what I experience locally (develop branch). Could this be worth a second look?

Imho we should take this opportunity to also fix the phpdoc for this function, the notes for the `$full_link` param are misleading (it returns the organizer name, **not** the URL, when set to false - or at least that was the original intent before this change).

zbtirrell: 1. $echo is deprecated 
2. If PRO isn't active and you send $full_link as false, it'll always return the organizer name, which is not useful or expected in that scenario
3. I am able to reproduce the problem

barryhughes: OK, but at the least, we should update the docs for that param. Right now it says:

> If true outputs a complete HTML <a> link, otherwise only the URL is output

With this change, it's more like:

> If true outputs a complete HTML <a> link, otherwise an empty string

bordoni:
Where does this PR stands?

PR #1182, opened 2016-11-24
Remove a (no longer needed) select2 hack for the event template selector

barryhughes:
Removes a Select2 related hack [originally added for the 3.0 release](https://github.com/moderntribe/the-events-calendar/commit/b16141f0db81c9db5b8c676cc556afd815b3f3e4) that is no longer required and actually has undesirable side-effects with the currently-in-use version of Select2.

https://central.tri.be/issues/69649


bordoni:
👍

PR #1184, opened 2016-11-24
Fix/69835 featured events ajax

barryhughes:
* Adds a helper to test if `'featured'` is set to a positive value (either in the URL query or post data, if any)
* Updates the ajax response methods to use the new helper

https://central.tri.be/issues/69835

### DIFF
--- a/src/Tribe/Featured_Events.php
+++ b/src/Tribe/Featured_Events.php
@@ -75,18 +75,6 @@ class Tribe__Events__Featured_Events {
 	 * @return bool
 	 */
 	public function featured_events_requested() {
-		$featured_request_var = tribe_get_request_var( 'featured', false );
-
-		switch ( strtolower( $featured_request_var ) ) {
-			case 'true':
-				return true;
-				break;
-
-			case 'false':
-				return false;
-				break;
-		}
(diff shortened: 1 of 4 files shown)


src/Tribe/Featured_Events.php
```diff
@@ -67,4 +67,26 @@ public function is_featured_query( WP_Query $query = null ) {
...
+	 * Indicates if 'featured' is set to a positive value either in the URL query
+	 * or the posted data (if any).
+	 *
+	 * @return bool
+	 */
+	public function featured_events_requested() {
+		$featured_request_var = tribe_get_request_var( 'featured', false );
+
+		switch ( strtolower( $featured_request_var ) ) {
+			case 'true':
```
bordoni: Could we also check for stuff like `1` and `0`? I remember us having a method that would do that check for us somewhere in our code.

barryhughes: I added a [helper here](https://github.com/moderntribe/tribe-common/pull/288) as we don't seem to have an existing standalone helper for this that I could find.

bordoni:
👍

PR #1188, opened 2016-11-28
Remove defaults in the Customizer for links and buttons

borkweb:
With the defaults in place, they override all kinds of things. Let's let the theme's stylesheet (and TEC stylesheet) be the base and allow users to set/clear their selections rather than always forcing a default.

See: https://central.tri.be/issues/69755


barryhughes:
💨

PR #1189, opened 2016-11-29
Set the featured image on posts if there is an image

lucatume:
Ticket: https://central.tri.be/issues/69465

This PR fixes a check to make sure we set the featured image on EA imported posts if there is any image; whatever state of import the image is in.


bordoni:
It's good but can ask whoever is in QA to check if all the other importing images still work as expected in all possible cases? Like Cron, multiple batches of importing.

lucatume:
Ok, will note, thanks

PR #1197, opened 2016-12-01
Default end date should be based on site timezone | #69686

barryhughes:
Default end time in the event editor was 'today' in UTC, let's use 'today' in the local timezone instead.

https://central.tri.be/issues/69979


bordoni:
👍

PR #1201, opened 2016-12-02
Resolve the State Province not hiding problem

bordoni:
_Ref:_ [C#70138](https://central.tri.be/issues/70138)


barryhughes:
:+1:

PR #1203, opened 2016-12-06
Allow Venues and Organizers metas to be pre-populated by plugin

Chouby:
I am currently working on an integration between Polylang and The Event Calendar.

Generally when creating a post translation, we automatically duplicate the post meta in the new translation to avoid a boring manual duplicated work by the end user. In Polylang, this is done by duplicating the post meta and associating them to the 'auto-draft' post.

With the current code for venue and organizer metaboxes, even if the post meta is non empty and has been pre-populated by a plugin, it is not read because the test is not done on the post meta value itself but on the post status. There is a possible workaround for Venues, using the 'tribe_events_default_value_strategy' filter, but I don't like it because it would conflict with any other plugin using the same filter. There is no workaround for Organizers.

I propose to always read post metas when they are not empty (if pre-populated by another plugin such as Polylang).

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -3627,9 +3627,8 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 				}
 
 				foreach ( $this->venueTags as $tag ) {
-					$meta = get_post_meta( $event->ID, $tag, true );
-					if ( ! empty( $meta ) ) {
-						$$tag = esc_html( $meta );
+					if ( metadata_exists( 'post', $event->ID, $tag ) ) {
+						$$tag = esc_html( get_post_meta( $event->ID, $tag, true ) );
 					} else {
 						$cleaned_tag = str_replace( '_Venue', '', $tag );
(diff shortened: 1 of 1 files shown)


bordoni:
Hey @Chouby, thanks for submitting this Pull Request.
So you are using `get_post_metadata` filter to pre-populate the information here correct?
If so, it would be a safer check to use: `metadata_exists( 'post', $event->ID, $tag )` instead of `empty( $meta )`, because of things like `0` or an _empty string_.

Chouby:
I use `update_post_meta` to assign values based on the original post.
Yes, obviously you are right. `metadata_exist`s is much better!

bordoni:
@Chouby I will make sure this Pull Request gets included and tested for our next Release, not `v4.3.4` that should go out tomorrow morning if all goes according to plan, the one after that.

Chouby:
Great! Thank you!

bordoni:
_Ref:_ [C#70894](https://central.tri.be/issues/70894)

bordoni:
Closed this PR, in favor of one that we can merge to the next MR.

PR #1212, opened 2016-12-09
point to shortcodes in reference to pro

peterchester:
Geoff pointed out that there is no reference to shortcodes on our readme. I've added a note in the PRO section.


bordoni:
👍

PR #1215, opened 2016-12-12
#69887 Display Venue Link only if Pro Active

jesseeproductions:
_Ref:_ [C#69887](https://central.tri.be/issues/69887) - Only show link to Venue if Pro Active


bordoni:
👍

PR #1217, opened 2016-12-13
Fix/70614 stuck importing notice

lucatume:
Ticket: https://central.tri.be/issues/70614

This PR adds a `Queue_Cleaner` class to garbage collect stalling/failed pending records.
The garbage collection happens whenever the current record q is processed so it could be during an admin visit made from the user as well as during a cron job.

The cleaner will do 2 things:
* delete duplicate pending posts for the same import ID; the idea being that the user could attempt importing the same feed multiple times and one pending/fetching record is enough
* remove the `fetch` state from failed posts (this is what was actually stucking the notice) to avoid the wait lock of a failed record not running and hence its currenty activity, fetching, never being removed.

Please note: in the `Tribe__Events__Aggregator__Record__Queue::__construct` method I'm voiding subsequent calls to the the `process` method setting a `null_process` property: that's to get around the wrong previous existing use of `__construct` method return values, `__construct` will always return an instance of the object and the way to stop that is with exceptions. Putting in place `try/catch` blocks anywhere the constructor and `process` are used would require some major refactoring since in all use cases but one a valid `Tribe__Events__Aggregator__Record__Queue` instance is taken as granted. Voiding the `process` call shortcircuiting it is kind of an abuse of the "tell, don't ask" principle but still bearable.

### DIFF
--- a/src/Tribe/Aggregator/Record/Queue_Cleaner.php
+++ b/src/Tribe/Aggregator/Record/Queue_Cleaner.php
@@ -1,0 +1,118 @@
+<?php
+
+
+class Tribe__Events__Aggregator__Record__Queue_Cleaner {
+
+	/**
+	 * @var int The time a record is allowed to stall before having
+	 *          its status set to failed in seconds.
+	 */
+	protected $stall_limit = 1800;
+
+	/**
+	 * Removes duplicate records for the same import ID.
+	 *
+	 * While it makes sense to keep track of past import records it does not make sense
+	 * to keep more than one pending record for the same import ID.
+	 *
+	 * @param Tribe__Events__Aggregator__Record__Abstract $record A record object or a record post ID.
+	 *
+	 * @return int[] An array containing the deleted posts IDs.
+	 */
+	public function remove_duplicate_pending_records_for( Tribe__Events__Aggregator__Record__Abstract $record ) {
+		if ( empty( $record->meta['import_id'] ) ) {
+			return array();
+		}
+
+		$import_id = $record->meta['import_id'];
+
+		/** @var \wpdb $wpdb */
+		global $wpdb;
+
+		$pending_status = Tribe__Events__Aggregator__Records::$status->pending;
+
+		$query = $wpdb->prepare( "SELECT ID
+			FROM {$wpdb->postmeta} pm
+			JOIN {$wpdb->posts} p
+			ON pm.post_id = p.ID
+			WHERE p.post_type = %s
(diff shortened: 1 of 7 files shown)


src/Tribe/Aggregator/Record/Queue_Cleaner.php
```diff
@@ -0,0 +1,118 @@
...
+
+		$query = $wpdb->prepare( "SELECT ID
+			FROM {$wpdb->postmeta} pm
+			JOIN {$wpdb->posts} p
+			ON pm.post_id = p.ID
+			WHERE p.post_type = %s
+			AND p.post_status = %s
+			AND pm.meta_key = '_tribe_aggregator_import_id'
+			AND pm.meta_value = %s
+			ORDER BY p.post_modified_gmt DESC", Tribe__Events__Aggregator__Records::$post_type, $pending_status, $import_id );
```
bordoni: I would rather use a `WP_Query` given that it would provide us and third-party developers the power to change how this works.

src/Tribe/Aggregator/Record/Queue_Cleaner.php
```diff
@@ -0,0 +1,118 @@
...
+
+		$post_status = $record->post->post_status;
+
+		if ( ! in_array( $post_status, array( $pending, $failed ) ) ) {
+			return false;
+		}
+
+		$id = $record->post->ID;
+
+		if( $post_status === $failed){
```
bordoni: Spacing.

src/Tribe/Aggregator/Record/Queue_Cleaner.php
lucatume: the `WP_Query` is slow and prone to error due to indiscriminate third-party intervention. Here we are cleaning home with a non front-facing post type; I would allow someone to mess with this logic and query if there really is a request for it.

lucatume: I'm more than willing to make this a `WP_Query` use if there is a good reason and a use case: I cannot think of any in this case.

borkweb: I think a nice middle ground would be to add the ability to filter the entire SQL query string. Along mentioning that we don't suggest manipulating that query in the docblock.

lucatume:
Ok, I've updated to allow a measure of control on the query and this should be good for review.

barryhughes:
:star:

PR #1218, opened 2016-12-13
Fix notice issue due to missing source for ics files

lucatume:
Ticket: https://central.tri.be/issues/69960


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -501,7 +501,7 @@ public function queue_import( $args = array() ) {
 		$defaults = array(
 			'type'     => $this->meta['type'],
 			'origin'   => $this->meta['origin'],
-			'source'   => $this->meta['source'],
+			'source'   => ! empty( $this->meta['source'] ) ? $this->meta['source'] : '',
```
bordoni: I would rather have this be done inside of the default meta-creation, so that we don't have to perform this again in another place.

PR #1219, opened 2016-12-13
fixed code to avoid showing import comments in count

lucatume:
Ticket: https://central.tri.be/issues/68297


src/Tribe/Aggregator/Cron.php
```diff
@@ -19,7 +19,7 @@ class Tribe__Events__Aggregator__Cron {
 	 * Limit of Requests to our servers
 	 * @var int
 	 */
-	private $limit = 5;
+	private $limit = 0;
```
bordoni: Why `0`?

PR #1221, opened 2016-12-14
avoid notice during fetch phase

lucatume:
Tickets:

* https://central.tri.be/issues/69542
* https://central.tri.be/issues/69824

Fixes a notice popping up during the "fetch" phase of the import that would generate a PHP error.

### DIFF
--- a/src/Tribe/Aggregator/Record/Queue_Processor.php
+++ b/src/Tribe/Aggregator/Record/Queue_Processor.php
@@ -191,7 +191,10 @@ class Tribe__Events__Aggregator__Record__Queue_Processor {
 
 		$this->current_queue->set_in_progress_flag();
 		$processed = $this->current_queue->process( self::$batch_size );
-		$this->processed += $processed->activity->count( $this->current_queue->get_queue_type() );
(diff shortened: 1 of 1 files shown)


src/Tribe/Aggregator/Record/Queue_Processor.php
```diff
@@ -191,7 +191,10 @@ protected function do_processing() {
 
 		$this->current_queue->set_in_progress_flag();
 		$processed = $this->current_queue->process( self::$batch_size );
-		$this->processed += $processed->activity->count( $this->current_queue->get_queue_type() );
+		// in the 'fetch' phase this will not be a Queue object
+		if ( is_a( $processed, 'Tribe__Events__Aggregator__Record__Queue' ) ) {
```
bordoni: Avoid using `is_a`, use instanceof.

lucatume: Why is that? Any specific reason?

lucatume:
This should be ok

bordoni:
👍

PR #1222, opened 2016-12-14
avoid appending post type to network admin pages

lucatume:
Ticket: https://central.tri.be/issues/68599

When in the context of the network admin do not append the post type query arg: the menu is now in Settings.


bordoni:
👍

PR #1223, opened 2016-12-15
adjust top margin of tribe header on default page mobile view

trishasalas:
https://central.tri.be/issues/44702


borkweb:
If that sourcemappingURL should be in the CSS, then this is :+1:

PR #1226, opened 2016-12-15
Improve tribe_get_events_title() in respect of single event posts

barryhughes:
In essence, removes the `Upcoming Events` text found in single event post titles (ie `Upcoming Events | {Single Event Slug} | {Blog Name}` is simplified to just `{Single Event Slug} | {Blog Name}`), all other views are untouched.

https://central.tri.be/issues/46313


src/functions/template-tags/loop.php
```diff
@@ -128,7 +128,13 @@ function tribe_get_events_title( $depth = true ) {
 
 		$tribe_ecp = Tribe__Events__Main::instance();
 
-		$title = sprintf( esc_html__( 'Upcoming %s', 'the-events-calendar' ), $events_label_plural );
+		if ( is_single() && tribe_is_event() ) {
```
borkweb: Is the `tribe_is_event()` check necessary? If it happens to _not_ be an event, then failing over to "Upcoming %s" would be a little funky.

barryhughes: There are other single post possibilities, like venues (where we list upcoming events), was my thinking. Could certainly strip the `tribe_is_event()` check even so if you prefer?

bordoni: I would Say keep it in there.

barryhughes: :+1: it also keeps things in line with the ticket objectives

PR #1229, opened 2016-12-16
Allow Venues and Organizers metas to be pre-populated by plugin

bordoni:
_Ref:_ [C#70894](https://central.tri.be/issues/70894)
_Original:_ #1203


borkweb:
🆗

PR #1235, opened 2017-01-02
Do not allow access to uninitialized activity member | spotfix

barryhughes:
Ensure calls to `$queue->activity` are routed through `$queue->activity()` (so we can be sure a record activity object is returned to the caller).


bordoni:
👍

PR #1237, opened 2017-01-03
take possible activity return value into account

lucatume:
Take possible return values into account to avoid fatals

Ticket: https://central.tri.be/issues/69137#note-14


barryhughes:
This looks good, but tagging `on hold` as I'm guessing it's only targeting `master` in lieu of a not-yet-created release branch.

PR #1238, opened 2017-01-03
Adjustment for featured event styles in base list widget | #69978

barryhughes:
Fixes text colour for featured event styles (within regular events list widget).

https://central.tri.be/issues/69978


bordoni:
👍

PR #1239, opened 2017-01-03
add theme version to css override

berler:
Since the tribe-events.css override comes from the theme, it makes sense
to use the theme version when enqueuing the css. This makes it possible
for themes to release an update that changes the css override. Otherwise
browsers can have the old version of the theme's css cached.


barryhughes:
[Ticketed: #71973](https://central.tri.be/issues/71973)

barryhughes:
Updated to target `release/4.4.1`

PR #1240, opened 2017-01-03
Support a light featured-event accent in the core event list widget

barryhughes:
Further adjustment to support light featured event accents.

https://central.tri.be/issues/69978


borkweb:
:+1:

PR #1243, opened 2017-01-06
added check for post orders type plugin to add arg to ignore it's ord…

stormrockwell:
…ering

This will fix the conflict with the Post Types Order plugin that causes the ajax call for the list order to not be ordered correctly. The other plugin accepts the parameter ignore_custom_sort so I created a check to see if the plugin is active and if it is to add the parameter. There are numerous people with this issue throughout your forms and this would fix the issue.

https://theeventscalendar.com/support/forums/topic/post-types-order-plugin-compatability/

### DIFF
--- a/src/Tribe/Template/List.php
+++ b/src/Tribe/Template/List.php
@@ -95,6 +95,11 @@ if ( ! class_exists( 'Tribe__Events__Template__List' ) ) {
 				$args[ Tribe__Events__Main::TAXONOMY ] = $_POST['tribe_event_category'];
 			}
 
+			// if post types order plugin is installed, set args to ignore their custom order
+			if ( is_plugin_active( 'post-types-order/post-types-order.php' ) ) {
+				$args['ignore_custom_sort'] = true;
+			}
+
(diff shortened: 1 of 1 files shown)


stormrockwell:
I thought I did and the details link is a 404. This is my first pull request on this. Is there a proper way to do it?

here are a couple:
https://theeventscalendar.com/support/forums/topic/events-not-reordering-in-chronological-order-issue/
https://theeventscalendar.com/support/forums/topic/post-types-order-plugin-compatability/

barryhughes:
Hi @stormrockwell - thanks for the PR!

Generally speaking, we're not super keen on adding code to workaround quirks/issues with other plugins. Added to that, I could envision this being a problem for plugins other than our own so I wonder if it would be better to figure out a solution or workaround - whatever form that might take - that could be baked into the [Post Types Order plugin](https://wordpress.org/plugins/post-types-order/) itself?

Beyond that, the `tribe_events_listview_ajax_get_event_args` filter hook (right below your suggested change) could be used to achieve pretty much the same thing, for those users experiencing issues when integrating both plugins.

stormrockwell:
Completely understandable. I think their plugin should make it easier to exclude certain post types.

PR #1244, opened 2017-01-09
Add safety check 

barryhughes:
Helps prevent fatal errors during plugin activation (if an earlier than required version of Tribe Common is already present and loaded).


borkweb:
:aw_yeah:

PR #1246, opened 2017-01-12
Add Other URLs as a source to EA

lucatume:
Tickets:
* https://central.tri.be/issues/70395
* https://central.tri.be/issues/70530

This PR adds the code and UI to allow and handle the "Other URL" source on EA.

EA server response is supposed to be iCal-like.

Opening against `feature/rest-api` branch as the code supporting this PR is not on EA yet.


src/admin-views/aggregator/settings.php
```diff
@@ -365,6 +405,7 @@
 	$facebook,
 	$gcal,
 	$meetup,
+    $url,
```
borkweb: Tabbing is off here

src/Tribe/Aggregator/Record/Url.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+
+class Tribe__Events__Aggregator__Record__Url extends Tribe__Events__Aggregator__Record__Abstract{
```
borkweb: Add a space between `Abstract` and `{`

borkweb:
This looks pretty solid, @lucatume! Nice work. A few small spacing issues and that's it.

:1st_place_medal:

PR #1247, opened 2017-01-12
Smart event cost and website fields in the REST API response

lucatume:
Tickets:

* https://central.tri.be/issues/70386
* https://central.tri.be/issues/70385

This PR:

* will make sure that the REST response for an event website will be set to the event permalink itself if empty
* will set the cost of the event according to how we would format it on the front-end returning the complete cost string ("Free" or "$34") and the cost components in ASC order (`[0, 12, 15,30]` for an event that has tickets with prices ranging from free to $30)
* updates the tests (REST and Integration) accordingly

Notice that while the ticket deals with the other side of the import (the destination) setting the event data in the REST response makes sense.


bordoni:
👍

PR #1251, opened 2017-01-18
Attach wp_localize_script to the correct handle, jQuery is unsafe.

bordoni:
_Ref:_ [C#71369](https://central.tri.be/issues/71369)


barryhughes:
🖊 No changelog entry - not sure if you're adding now or at the end of the cycle though

PR #1253, opened 2017-01-19
Make sure we iCal Echo for single Events

bordoni:
_Ref:_ [C#71916](https://central.tri.be/issues/71916)


lucatume:
Solve the `readme.txt` conflict and good to merge!

PR #1254, opened 2017-01-19
Prevent Fatals on EA Image importing

bordoni:
_Ref:_ [C#70576](https://central.tri.be/issues/70576)

### DIFF
--- a/src/Tribe/Aggregator/API/Image.php
+++ b/src/Tribe/Aggregator/API/Image.php
@@ -67,16 +67,20 @@ class Tribe__Events__Aggregator__API__Image extends Tribe__Events__Aggregator__A
 			return $response;
 		}
 
+		// Get Headers and Body
+		$headers = wp_remote_retrieve_headers( $response );
+		$body = wp_remote_retrieve_body( $response );
+
 		// if the response isn't an image then we need to bail
-		if ( ! preg_match( '/image/', $response['headers']['content-type'] ) ) {
+		if ( ! preg_match( '/image/', $headers['content-type'] ) ) {
 			/**
 			 * @todo  See a way for Tribe__Errors to handle overwriting
 			 */
-			return new WP_Error( 'invalid-image', $response['body'] );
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/API/Image.php
```diff
@@ -67,16 +67,20 @@ public function get( $image_id ) {
...
+		$body = wp_remote_retrieve_body( $response );
+
 		// if the response isn't an image then we need to bail
-		if ( ! preg_match( '/image/', $response['headers']['content-type'] ) ) {
+		if ( ! preg_match( '/image/', $headers['content-type'] ) ) {
 			/**
 			 * @todo  See a way for Tribe__Errors to handle overwriting
 			 */
-			return new WP_Error( 'invalid-image', $response['body'] );
+			return new WP_Error( 'invalid-image', $$body );
```
lucatume: probably one `$` is fine

src/Tribe/Aggregator/API/Image.php
```diff
@@ -67,16 +67,20 @@ public function get( $image_id ) {
 			return $response;
 		}
 
+		// Get Headers and Body
+		$headers = wp_remote_retrieve_headers( $response );
+		$body = wp_remote_retrieve_body( $response );
+
```
lucatume: Probably too much of a defense but: we are making assumptions about the `$headers` array wishing any key we look for will be there (e.g. `content-type` and `content-disposition`); a lot of filtering could happen in that request and I'd advise a pre-emptive check on the keys we look for before trying to look them up.
Since the `Requests_Utility_CaseInsensitiveDictionary` class implements `ArrayAccess` use of an `array_intersect_key` is fast and fine here like:

```php
$go = count(array_intersect_key(array('content-type', 'content-disposition'), $headers)) === 2
```

lucatume:
Fix the `$` party and take my other note into consideration (optional though) and good to merge.

lucatume:
Ok!

PR #1264, opened 2017-01-26
Use the correct var name and double check the var before usage

borkweb:
See: https://wordpress.org/support/topic/fatal-error-in-4-4-1/


bordoni:
👍

PR #753, opened 2016-07-07
Don't call remove_action in a function hooked to the same action

Chouby:
When calling remove_action in a function hooked to the same action, all functions hooked to this action with a lower priority are never called, thus breaking 3rd party plugins.

Fortunately, there is an easy way to achieve the same goal using a static variable as proposed in this PR. 

[#63474](https://central.tri.be/issues/63474)

PR #1267, opened 2017-01-29
Don't call remove_action in a function hooked to the same action

Chouby:
This is a follow up of #753.

The goal of this PR was to avoid calling remove_action on 'save_post' while already running this action. The PR was merged in develop branch for the 4.3 main release. But the function was refactored in parallel in the 4.2.x branch and it looks like the merge lost the main purpose of #753, i.e removing the call to remove_action.

### DIFF
--- a/src/Tribe/Meta/Save.php
+++ b/src/Tribe/Meta/Save.php
@@ -109,9 +109,6 @@ class Tribe__Events__Meta__Save {
 			return false;
 		}
 
-		// Remove this hook to avoid an infinite loop, because saveEventMeta calls wp_update_post when the post is set to always show in calendar
-		remove_action( 'save_post', array( Tribe__Events__Main::instance(), 'addEventMeta' ), 15 );
-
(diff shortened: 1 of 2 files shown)


barryhughes:
Thanks for persevering with this &mdash; I'm sorry that happened!

I'm wondering though: with the advent of `WP_Hook` (introduced in WP 4.7) is there still an issue to resolve here? I haven't confirmed either way so could be off track, but I felt it was worth flagging.

Chouby:
I was believing that too. Maybe it works in some cases. But for our two plugins, nothing changes with WP4.7.

If you activate Polylang and The Event Calendar and you save an event, the [Polylang function hooked to 'save_post' with priority 21](https://github.com/polylang/polylang/blob/master/admin/admin-filters-post.php#L36) is never executed. If I comment the remove_action line, then it's ok.

barryhughes:
Logged as [#73122](https://central.tri.be/issues/73122) - we'll get this merged in however I'm placing on hold until the very next maintenance release has shipped, as we'll be unable to squeeze this one right now.

PR #1268, opened 2017-01-30
Use `get_page_template()` not `locate_template()`

QWp6t:
Using `locate_template()` effectively bypasses all of the normal template hierarchy filters that are normally made available to developers at this stage when loading a page. This PR ensures developers can still hook into those filters by using the same function that WordPress core uses in [**wp-includes/template-loader.php**](https://github.com/WordPress/WordPress/blob/master/wp-includes/template-loader.php#L56).

Full disclosure: I haven't tested this. A user was having difficulty with your plugin, and I noticed this issue when I was diving into your code. A workaround was possible for his particular issue, but I figured I would still bring this to your attention.

### DIFF
--- a/src/Tribe/Templates.php
+++ b/src/Tribe/Templates.php
@@ -127,7 +127,7 @@ if ( ! class_exists( 'Tribe__Events__Templates' ) ) {
 					add_action( 'loop_start', array( __CLASS__, 'setup_ecp_template' ) );
 				}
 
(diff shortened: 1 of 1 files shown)


borkweb:
@QWp6t, thank you for this pull request and suggestion! I've logged an internal ticket so that we can take a look at this more deeply. With a quick glance, this change seems positive!

borkweb:
Central ticket: https://central.tri.be/issues/72842

barryhughes:
Looks good, but placing on hold for the time being: we'll look at merging this in an upcoming release.

PR #1270, opened 2017-01-30
Change variable name to avoid confusion with PRO | spotfix

cliffordp:
`googleMapLink()` is in core, not PRO


src/functions/template-tags/google-map.php
```diff
@@ -24,8 +24,8 @@
 	 * @return string A fully qualified link to http://maps.google.com/ for this event
 	 */
 	function tribe_get_map_link( $postId = null ) {
-		$tribe_ecp = Tribe__Events__Main::instance();
-		$output    = esc_url( $tribe_ecp->googleMapLink( $postId ) );
+		$tec = Tribe__Events__Main::instance();
```
barryhughes: Should either fix alignment of `=` on this line or drop alignment on the following line.

PR #1273, opened 2017-02-01
Aggregator > global/source specific show-map setting fix

barryhughes:
Historically `'1'` was used to indicate a positive show map/map link value, though we began using `'yes'` in EA. This changes makes the editor UI tolerant of both possibilities.

https://central.tri.be/issues/67228


borkweb:
:+1:

PR #1275, opened 2017-02-01
Resolve Granularity Schedule CSS and Notices

bordoni:
_Ref:_ [C#68942](https://central.tri.be/issues/68942)


src/admin-views/aggregator/fields/schedule.php
```diff
@@ -40,7 +40,8 @@ class="tribe-dependent"
 				type="hidden"
 				name="aggregator[<?php echo esc_attr( $origin ) ?>][schedule_day]"
 				id="tribe-ea-field-<?php echo esc_attr( $origin ) ?>_schedule_day"
-				class="tribe-ea-field tribe-ea-dropdown tribe-ea-size-small"
+				class="tribe-ea-field tribe-ea-dropdown tribe-ea-size-medium"
+				style="margin-left: 0;"
```
borkweb: I see a few of these inline styles. Do those _have_ to be inline (and if so, I'd love to know why). If they don't need to be inline, can you move them to a `.pcss` file?

bordoni: Specifically because this item did not fit any of the other usages of `tribe-ea-field` next to a `tribe-ea-field-readonly`

bordoni: But I can add more rules to match this one.

borkweb: Yeah, adding more classes and moving that to .pcss would be 🍨

borkweb:
:+1:

PR #1277, opened 2017-02-01
Remove all instances of is_a

bordoni:
_Ref:_ [C#72812](https://central.tri.be/issues/72812)


borkweb:
Add a changelog entry. Otherwise, this is good to go

PR #1279, opened 2017-02-02
Feature/72655 ea notices refine

lucatume:

This PR:

* makes sure we use the localized version of EA server responses and not the English one
* marks re-scheduled imports as rescheduled in the history tab (was marked as an error)


borkweb:
:+1:

PR #1281, opened 2017-02-03
Update Copyright Validity

masteradhoc:


borkweb:
Good catch! Thank you, @masteradhoc!

PR #1282, opened 2017-02-03
Fix/72655 ea notices

lucatume:

This PR fixes a glitch that, depending on run times and db access times, would either mark any re-scheduled import (due to too many HTTPS requests) to failed or eternally pending.


bordoni:
👍

PR #1285, opened 2017-02-06
Register TEC with Tribe__Dependency

elimn:
Ticket: https://central.tri.be/issues/69583

We now have a plugin file constant like we do in our other plugins, and like our plugins this one will now register itself `Tribe__Dependency`. 

I rebased the changes in `feature/66657-register-plugins` to the MR's code. That branch was meant to be submitted a few MR's back, but slipped through the cracks.

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -139,6 +139,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		public $pluginName;
 
 		public $displaying;
+		public $plugin_file;
 		public $plugin_dir;
 		public $plugin_path;
 		public $plugin_url;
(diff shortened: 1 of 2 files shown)


barryhughes:
Looks good to me! My only question is if we're all happy with `THE_EVENTS_CALENDAR_PLUGIN_FILE` (I see other plugins have a comparable constant but it takes the form `*_MAIN_PLUGIN_FILE` etc), or if we'd like something shorter etc like `TEC_MAIN_FILE` ... (just checking as once this is added, we're pretty much stuck with it)

@bordoni @borkweb

elimn:
That makes sense. Personally, my ideal would be no constant. As you said once it's added. we can't effectively deprecate. I'd prefer that we have a set_file() method and a get_file() one. 

@borkweb and I were just talking about https://central.tri.be/issues/65190 which would be a rethinking of our bootstrap and init process for all of our plugins. Perhaps that would be an ideal time to consider the best way to do things? Until then we could stick with the current standard, which is a verbose constant. Just an idea.

bordoni:
I would go with `TRIBE_EVENTS_FILE` until we move away from this Constant usage. Me and @barryhughes just figured out a possible way to avoid this moving forward in the future.

barryhughes:
Let's merge this thing 💥

PR #1286, opened 2017-02-07
avoid undefined prop notice

lucatume:

Found out while workin on the ticket, this PR avoids the notice


barryhughes:
:goat:

PR #1289, opened 2017-02-11
community events list refresh - increase tribe template width

trishasalas:
https://central.tri.be/issues/70876

_related: https://github.com/moderntribe/events-community/pull/202_


trishasalas:
I made some changes here after a convo with Matt so this needs more eyes O_o

PR #1290, opened 2017-02-13
remove rest api settings

lucatume:
Ticket: https://central.tri.be/issues/72911

Remove the option to disable the REST API but leave the Settings structure in place as we might need it


bordoni:
🍸

PR #1291, opened 2017-02-13
Set the width of the map div within the column to 90% so it fills space

borkweb:
Setting to 90% to match that of the width being set when in mobile (< 768px)

See: https://central.tri.be/issues/69844


barryhughes:
🌟

PR #1293, opened 2017-02-13
Resolve EA Facebook image import problem caused by object vs array bug

borkweb:
`wp_remote_retrieve_headers()` returns an array of headers, not an object

See: https://central.tri.be/issues/72764


bordoni:
👍

PR #1296, opened 2017-02-14
When the map is paired with two columns, ensure the map appears BELOW

borkweb:
See: https://central.tri.be/issues/72272


barryhughes:
Looks good - is a changelog entry required here though?

borkweb:
Appended the ticket number to the changelog entry for 69844

barryhughes:
Shabba dabba 💥

PR #1297, opened 2017-02-14
Allow for negative numbers for "Month view events per day" setting

borkweb:
I know this allows for values less than -1, but I'm not concerned on that front.

Requires: https://github.com/moderntribe/tribe-common/pull/333

See: https://central.tri.be/issues/70497


bordoni:
👍

PR #1301, opened 2017-02-16
Fix /all/ slug for users of translation packs | #71996

barryhughes:
Ensure that the translated version of the /all/ slug is respected.

This is in the-events-calendar rather than events-pro simply because, for whatever historic reason, most of our code that relates to this sort of permalink lives here (and it wasn't within the scope of this fix to resolve that).

https://central.tri.be/issues/71996


borkweb:
:+1:

PR #1304, opened 2017-02-22
Use tribe_get_events() instead of calling Tribe__Events__Query::getEvents() directly

sc0ttkclark:
tribe_get_events() has it's own filter to allow customizing the results, but that's not used in three places across the codebase.


sc0ttkclark:
Props @strangerstudios for letting me know about this weirdness

PR #1307, opened 2017-02-28
preserve options on import

zbtirrell:
This fixes some options that weren't importing as expected.

For Aggregator, this covers: Facebook, Meetup, iCal, ICS, and gCal.

For CSV files, this specifically fixes 2 bugs
* saving show maps settings, rather than them always being overriden by the default setting
* expanded mime type support for CSV files

See: https://central.tri.be/issues/73619


borkweb:
🥇

PR #1316, opened 2017-03-14
Pass the Start and End of the EA search

bordoni:
_Ref:_ [C#74974](http://central.tri.be/issues/74974)

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -609,7 +609,16 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 
 	public function get_import_data() {
 		$aggregator = tribe( 'events-aggregator.main' );
-		return $aggregator->api( 'import' )->get( $this->meta['import_id'] );
+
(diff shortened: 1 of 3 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -609,7 +609,16 @@ public function queue_import( $args = array() ) {
...
+
+		// For now only apply this to the URL type
+		if ( 'url' === $this->type ) {
+			$data = array(
+				'start' => $this->meta['start'],
+				'end' => $this->meta['end'],
+			);
+		}
+
+		return $aggregator->api( 'import' )->get( $this->meta['import_id'], $data );
```
barryhughes: _Looks_ like `$data` may not be defined at this point (depending on the result of the above conditional). Worth defining it as an empty array, some point above the `if` block?

bordoni: Ohh very good point.

bordoni:
It's ready for Code Review again

borkweb:
:+1:

PR #1320, opened 2017-03-16
Enable month view cache by default

barryhughes:
For new installations/where the month view cache setting has not already been saved, default to enabling it.

https://central.tri.be/issues/74867


bordoni:
👍

PR #1322, opened 2017-03-17
Modify the wording for the Settings on EA ranges for Other URL

bordoni:
_Ref:_ [C#74974](http://central.tri.be/issues/74974)

### DIFF
--- a/src/Tribe/Aggregator/Settings.php
+++ b/src/Tribe/Aggregator/Settings.php
@@ -270,12 +270,12 @@ class Tribe__Events__Aggregator__Settings {
 	public function get_url_import_range_options( $title = true ) {
 		$options = array(
 			DAY_IN_SECONDS       => array(
-				'title' => __( 'One day', 'the-events-calendar' ),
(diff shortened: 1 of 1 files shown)


src/Tribe/Aggregator/Settings.php
```diff
@@ -270,12 +270,12 @@ public function default_settings_import( $origin ) {
 	public function get_url_import_range_options( $title = true ) {
 		$options = array(
 			DAY_IN_SECONDS       => array(
-				'title' => __( 'One day', 'the-events-calendar' ),
-				'range' => __( 'a day', 'the-events-calendar' ),
+				'title' => __( '24 hours', 'the-events-calendar' ),
```
barryhughes: I wonder if adding context would be useful here (`'title`' and `'range'`) or, if they'll always be identical now, if we should use a var instead of two l10n calls ... will leave to your discretion 🎩

bordoni: I really don't know why they exist, but I would say that this is up for discussion when @lucatume is back. He crafted this part of the code.

PR #1327, opened 2017-03-21
Fix/75379 correct days of week short value

niconerd:
https://central.tri.be/issues/75379


barryhughes:
Unsure why the Travis CI build could not complete (details won't load for me) - but this looks AOK. Merge away!

PR #1329, opened 2017-03-21
Respect settings preventing local changes being overriden when the ev…

barryhughes:
* Make it so that local changes to various fields including venues, organizers and categories are not overridden upon updates (when the change authority is `preserve_updates`)
* Make it so the event start/end dates don't reset to the Unix epoch in the same circumstances

[#72876](https://central.tri.be/issues/72876)

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -1070,7 +1070,18 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 				$event['post_parent'] = $k;
 			}
 
-			//if we should create a venue or use existing
+			// Do we have an existing venue for this event that we should preserve?
+			// @todo review: should we care about the potential for multiple venue IDs?
+			if (
+				! empty( $event['ID'] )
+				&& 'preserve_changes' === $update_authority_setting
+				&& $existing_venue_id = tribe_get_venue_id( $event['ID'] )
+			) {
+				$event['EventVenueID'] = $existing_venue_id;
+				unset( $event['Venue'] );
+			}
+
+			// if we should create a venue or use existing
 			if ( ! empty( $event['Venue']['Venue'] ) ) {
 				$v_id = array_search( $event['Venue']['Venue'], $found_venues );
 				if ( false !== $v_id ) {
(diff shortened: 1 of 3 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task &quot;review: should we care about the potential for multiple venue IDs?&quot; | [Abstract.php:1074](https://github.com/moderntribe/the-events-calendar/blob/315585faf188f2105896fa0c36b8c0396fa9fb6a/src/Tribe/Aggregator/Record/Abstract.php#L1074)

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1070,7 +1070,18 @@ public function insert_posts( $items = array() ) {
 				$event['post_parent'] = $k;
 			}
 
-			//if we should create a venue or use existing
+			// Do we have an existing venue for this event that we should preserve?
+			// @todo review: should we care about the potential for multiple venue IDs?
```
borkweb: Technically, we support multiple venue ids (if someone unhooks/re-hooks the post type from events). We can leave it as a To-Do for the future in the interest of time. Perhaps ticket the future change and we can prioritize it when there's a usecase or if time allows!

barryhughes: I can do that, for sure: I'd certainly prefer to address in the future as currently we lack a venue equivalent for tribe_get_organizer_ids(), amongst other things, and time is a-pressing.

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1070,7 +1070,18 @@ public function insert_posts( $items = array() ) {
...
+			if (
+				! empty( $event['ID'] )
+				&& 'preserve_changes' === $update_authority_setting
+				&& $existing_venue_id = tribe_get_venue_id( $event['ID'] )
+			) {
+				$event['EventVenueID'] = $existing_venue_id;
+				unset( $event['Venue'] );
```
borkweb: This appears to be using the `retain` authority settings rather than the `preserve_changes` setting. For proper adherence to `preserve_changes`, venues and organizers need local changes tracked (similar to [here](https://github.com/moderntribe/the-events-calendar/blob/51cf2cdcf4fd30d9c1f20201382b262ce95635fd/src/Tribe/Importer/File_Importer_Events.php#L88)) and observed on update, updating only non-changed fields.

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1091,6 +1102,16 @@ public function insert_posts( $items = array() ) {
 				unset( $event['Venue'] );
 			}
 
+			// Do we have an existing organizer(s) for this event that we should preserve?
+			if (
+				! empty( $event['ID'] )
+				&& 'preserve_changes' === $update_authority_setting
```
borkweb: Same `retain` vs `preserve_changes` comment

src/Tribe/Aggregator/Record/Abstract.php
barryhughes: [Logged for further action.](https://central.tri.be/issues/75600)

src/Tribe/Aggregator/Record/Abstract.php
barryhughes: [Logged for further action.](https://central.tri.be/issues/75601)

barryhughes:
Following discussion with @zbtirrell - merging this change as it represents an incremental improvement and we can follow-up in a future release with additional enhancements to complete the picture.

PR #1337, opened 2017-04-04
Purge import usage count on license update

lucatume:
Ticket: https://central.tri.be/issues/75561

Origins limit should be purged when the license string is updated.


bordoni:
👍

PR #1342, opened 2017-04-05
Never should display unix time on Filter Date

bordoni:
_Ref:_ [C#68942](http://central.tri.be/issues/68942)
---


lucatume:
:penguin:

PR #1344, opened 2017-04-06
Make it so the iCal description accounts for shortcode

KZeni:
I had a photo gallery's shortcode showing within the iCal Export's description area. This wasn't happening for the Google Calendar link so I made the iCal description filtering setup match the Google Calendar description filtering a bit more.

Also, I had tabs showing as "\t" within macOS Calendar... I replaced that within the str_replace array with an actual tab and it's working perfectly now.

:ticket: [#80452](https://central.tri.be/issues/80452)

### DIFF
--- a/src/Tribe/iCal.php
+++ b/src/Tribe/iCal.php
@@ -275,8 +275,8 @@ class Tribe__Events__iCal {
 			$item[] = 'CREATED:' . $tzoned->created;
 			$item[] = 'LAST-MODIFIED:' . $tzoned->modified;
 			$item[] = 'UID:' . $event_post->ID . '-' . $time->start . '-' . $time->end . '@' . parse_url( home_url( '/' ), PHP_URL_HOST );
-			$item[] = 'SUMMARY:' . str_replace( array( ',', "\n", "\r", "\t" ), array( '\,', '\n', '', '	' ), html_entity_decode( strip_tags( $event_post->post_title ), ENT_QUOTES ) );
-			$item[] = 'DESCRIPTION:' . str_replace( array( ',', "\n", "\r", "\t" ), array( '\,', '\n', '', '	' ), html_entity_decode( strip_tags( str_replace( '</p>', '</p> ', apply_filters( 'the_content', $event_post->post_content ) ) ), ENT_QUOTES ) );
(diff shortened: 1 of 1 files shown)


src/Tribe/iCal.php
```diff
@@ -275,8 +275,8 @@ public function generate_ical_feed( $post = null, $echo = true ) {
 			$item[] = 'CREATED:' . $tzoned->created;
 			$item[] = 'LAST-MODIFIED:' . $tzoned->modified;
 			$item[] = 'UID:' . $event_post->ID . '-' . $time->start . '-' . $time->end . '@' . parse_url( home_url( '/' ), PHP_URL_HOST );
-			$item[] = 'SUMMARY:' . str_replace( array( ',', "\n", "\r", "\t" ), array( '\,', '\n', '', '\t' ), html_entity_decode( strip_tags( $event_post->post_title ), ENT_QUOTES ) );
-			$item[] = 'DESCRIPTION:' . str_replace( array( ',', "\n", "\r", "\t" ), array( '\,', '\n', '', '\t' ), html_entity_decode( strip_tags( $event_post->post_content ), ENT_QUOTES ) );
+			$item[] = 'SUMMARY:' . str_replace( array( ',', "\n", "\r", "\t" ), array( '\,', '\n', '', '	' ), html_entity_decode( strip_tags( $event_post->post_title ), ENT_QUOTES ) );
```
barryhughes: Hi @KZeni!

I understand why you are making this change, but this effectively has us searching for tab characters and replacing them with tab characters. I feel like there's a simpler solution:

```php
// This...
str_replace( array( ',', "\n", "\t" ), array( '\,', '\n', '	' ), $text );

// Is no different to this...
str_replace( array( ',', "\n", "\t" ), array( '\,', '\n', "\t" ), $text );

// And can be simplified to this...
str_replace( array( ',', "\n" ), array( '\,', '\n' ), $text );
```

I simplified the code a little, but in essence we could remove the tabs from both arrays. Do you agree, would you be down to make this change?

KZeni: I only did that since I saw the "\t" text within the macOS Calendar app during real-world use with how it was set up previously, and my update to change `\t` to `	` seemed to fix that problem (for some reason).

The update to leave the tab as-is (remove from the str_replace array) & simplify the code as a result has been made.

barryhughes:
Excellent, thanks! We'll see about merging this change :+1:

barryhughes:
[Ticketed](https://central.tri.be/issues/80452) - merging.

PR #1347, opened 2017-04-07
timestamps are static, not timezone-dependent.

cliffordp:
https://central.tri.be/issues/70923

The entire logic of including a timezone is incorrect IMO.

This code does the following:
* get the datetime string (e.g. `2017-04-08 13:15:00`)
* get the timezone (e.g. `America/Los_Angeles`)
* combine them because strtotime() can handle it
* do strtotime()

If an event does not have a timezone in its postmeta, it will fallback to the WP timezone settings. This is handled by `self::get_event_timezone_string()`

### DIFF
--- a/src/Tribe/Timezones.php
+++ b/src/Tribe/Timezones.php
@@ -126,45 +126,17 @@ class Tribe__Events__Timezones extends Tribe__Timezones {
 	 *
 	 * @param int    $event_id
 	 * @param string $type (expected to be 'Start' or 'End')
-	 * @param string $timezone
 	 *
 	 * @return int
 	 */
-	protected static function get_event_timestamp( $event_id, $type = 'Start', $timezone = null ) {
+	protected static function get_event_timestamp( $event_id, $type = 'Start' ) {
 		$event    = get_post( Tribe__Events__Main::postIdHelper( $event_id ) );
-		$event_tz = get_post_meta( $event->ID, '_EventTimezone', true );
-		$site_tz  = self::wp_timezone_string();
+		$timezone = self::get_event_timezone_string( $event_id );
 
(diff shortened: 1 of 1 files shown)


src/Tribe/Timezones.php
```diff
@@ -126,45 +126,17 @@ public static function event_end_timestamp( $event_id, $timezone = null ) {
 	 *
 	 * @param int    $event_id
 	 * @param string $type (expected to be 'Start' or 'End')
-	 * @param string $timezone
 	 *
 	 * @return int
 	 */
-	protected static function get_event_timestamp( $event_id, $type = 'Start', $timezone = null ) {
+	protected static function get_event_timestamp( $event_id, $type = 'Start' ) {
```
barryhughes: You're eliminating the `$timezone` parameter - but you haven't updated this any place else. For example, `tribe_get_start_time()` also accepts a `$timezone` param and it would ultimately be passed along the chain to this one.

* If we're sure that param isn't needed (we need to think about 3rd party code, too) we should update all other related code accordingly
* We should consider if some sort of formal deprecation would make sense

cliffordp: As discussed via Slack, I intended this as a starting point toward fixing timestamps throughout our code base.

For example, implementing this fix alone without touching anything else causes the single event page's display of the start-end times to display UTC time alongside the correct timezone abbreviation, which makes the entirety of the displayed line incorrect.

bordoni:
Re-opened with `master` been the new branch which allows us to merge it into any MR later down the road.

bordoni:
This change will addressed later on down the line when we tackle our datepickers, this is code that I feel like to add more changes we want Unit tests first.

PR #1348, opened 2017-04-07
add taxonomy tracking in Main

lucatume:
Ticket: https://central.tri.be/issues/74

This PR makes sure we are tracking the `post_tag` and the `tribe_events_cat` taxonomies for our post types to avoid them being overriden in imports.


bordoni:
👍

PR #1351, opened 2017-04-11
Change Authority Datetime fix from 4.4.5

bordoni:
_Ref:_ [C#76534](http://central.tri.be/issues/76534)
---


borkweb:
Well, that makes sense. There doesn't seem to be an `$event` var to begin with! 🐙

PR #1353, opened 2017-04-13
Fix change Authority issues

lucatume:
Ticket: https://central.tri.be/issues/76534

Builds on: https://github.com/moderntribe/tribe-common/pull/358

### DIFF
--- a/src/Tribe/Aggregator/Event.php
+++ b/src/Tribe/Aggregator/Event.php
@@ -362,15 +362,15 @@ class Tribe__Events__Aggregator__Event {
 		}
 
 		// The start date needs to be adjusted from a MySQL style datetime string to just the date
-		if ( isset( $modified['_EventStartDate'] ) ) {
-			$start_datetime = strtotime( $data['EventStartDate'] );
+		if ( isset( $modified['_EventStartDate'] ) && isset( $post_meta['_EventStartDate'] ) ) {
+			$start_datetime = strtotime( $post_meta['_EventStartDate'] );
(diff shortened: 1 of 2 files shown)


zbtirrell:
It doesn't feel like this is respecting what is happening above - https://github.com/moderntribe/the-events-calendar/pull/1353/files#diff-f8677ebdc85d2610cfcd992b86b08790L360

src/Tribe/Aggregator/Event.php
```diff
@@ -349,27 +353,27 @@ public static function preserve_changed_fields( $data ) {
 				continue;
 			}
 
-			// if we don't have a field to reset to, let's unset the event meta field
-			if ( ! isset( $post_meta[ $field ] ) ) {
-				unset( $data[ $field ] );
```
bordoni: Don't we need to still unset the data field? I mean from what I remember without it would break.

lucatume: we set it to an empty string later in the code if the original is empty

lucatume: I mean here:  https://github.com/moderntribe/the-events-calendar/pull/1353/files#diff-f8677ebdc85d2610cfcd992b86b08790R364

bordoni: Why not Unset it? I mean avoid to trigger an update.

lucatume: Ok, latest version has an explicit unset

PR #1364, opened 2017-04-26
Add the no-results message for feeds

lucatume:
Ticket (contetxt): https://central.tri.be/issues/72634

This PR adds support for the `no-results` message when a fetch from a feed is successful but simply yields no results; this could happen when using keywords or dates or whatever.

This avoids the client hanging up on an empty events array after receiving the response.


borkweb:
:+1:

PR #1365, opened 2017-04-26
Removed duplicate date functions and added a filter.

elimn:
I noticed that more up to date versions of `tribe_get_start_time()` and `tribe_get_end_time()` are [in common](https://github.com/moderntribe/tribe-common/blob/ce0aadeffee590b2416a076ec45f00be5df80447/src/functions/template-tags/date.php#L156-L256). The ones here are duplicates, which is both unnecessary and going to make things harder to debug.

Then I added a filter to `tribe_get_display_end_date()` to bring it inline with our other template functions.

### DIFF
--- a/src/functions/template-tags/date.php
+++ b/src/functions/template-tags/date.php
@@ -14,92 +14,6 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 	return;
 }
 
-if ( ! function_exists( 'tribe_get_start_time' ) ) {
-	/**
-	 * Start Time
-	 *
-	 * Returns the event start time
-	 *
-	 * @category Events
-	 * @see http://php.net/manual/en/function.date.php
-	 *
-	 * @param int    $event       (optional)
-	 * @param string $dateFormat  Allows date and time formating using standard php syntax
-	 * @param string $timezone    Timezone in which to present the date/time (or default behaviour if not set)
-	 *
-	 * @return string|null Time
-	 */
-	function tribe_get_start_time( $event = null, $dateFormat = '', $timezone = null ) {
-		if ( is_null( $event ) ) {
-			global $post;
-			$event = $post;
-		}
-
-		if ( is_numeric( $event ) ) {
-			$event = get_post( $event );
-		}
-
-		if ( ! is_object( $event ) ) {
-			return;
-		}
-
-		if ( tribe_event_is_all_day( $event ) ) {
-			return;
-		}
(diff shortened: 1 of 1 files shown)


src/functions/template-tags/date.php
```diff
@@ -14,117 +14,48 @@
...
-	 * @param string $date_format  Allows date and time formating using standard php syntax
-	 * @param string $timezone     Timezone in which to present the date/time (or default behaviour if not set)
+	 * @param int|WP_Post $event        The event (optional).
+	 * @param bool        $display_time If true shows date and time, if false only shows date.
+	 * @param string      $date_format  Allows date and time formatting using standard php syntax.
+	 * @param string      $timezone     Timezone in which to present the date/time (or default behaviour if not set).
 	 *
 	 * @return string|null Date
 	 */
 	function tribe_get_display_end_date( $event = null, $display_time = true, $date_format = '', $timezone = null ) {
```
bordoni: Why name it `end_date` if we removed `start` lets make it `tribe_get_display_date`.

elimn: Are you suggesting it would be best if deprecated this function and replace it with another?

bordoni: By the looks of this PR we are deprecating one method already? Right?
I was suggesting that we "soft-deprecated" by pointing both the new one.

elimn: Hmm. This PR shouldn't be deprecating anything. It is removing some duplicate functions, those functions are already declared in Common and thus we don't need to redeclare them here. The ones in Common are already loaded before the ones here, and thus nothing is changing it's just removing orphaned code.

The other aspect of this PR is adding a little filter to `tribe_get_display_end_date()`. I was adding this filter simply because that function is missing one, yet most all of our template functions contain a filter. I was just trying to bring it inline with the rest. But, I could see an argument to be made for deprecating `tribe_get_display_end_date()`. And if that is something we're interested in doing, maybe adding a filter to it right now is a bad idea? 

Does that help clarify? Am I misunderstanding you?

elimn: After a hangout with @bordoni we decided to log this followup ticket:

https://central.tri.be/issues/77730

That will handle the possible deprecation of `tribe_get_display_end_date()`. In the mean time this spotfix will remove the two orphaned functions, but not enhance tribe_get_display_end_date() any until we decide whether or not that function should be deprecated.

PR #1366, opened 2017-04-26
Added continue if record instance is null

elimn:
Ticket: https://central.tri.be/issues/76923

[A user reported](https://theeventscalendar.com/support/forums/topic/events-calendar-scheduled-imports-not-saving/#post-1269111) that at times they were getting errors here because `$origin_instance` was null. It looks it can be null at times, thus we should account for this. User confirmed that this small change caused their errors to go away.


src/Tribe/Aggregator/Record/List_Table.php
```diff
@@ -286,6 +286,11 @@ protected function get_views() {
 
 		foreach ( $origins as $origin => $count ) {
 			$origin_instance = Tribe__Events__Aggregator__Records::instance()->get_by_origin( $origin );
+
+			if ( null === $origin_instance ) {
+				continue;
```
barryhughes: Wondering, do we want to log this with a note of the origin that was skipped?

elimn: That's probably a really good idea. I'm still not sure why the origin was skipped. My best guess was a conflict overrode the record somehow.

I'll enhance it.

bordoni:
👍

PR #1369, opened 2017-04-27
Unset event date meridian when preserving existing date

elimn:
Ticket: https://central.tri.be/issues/75787

The meridian from Facebook will get appended to the 24hr date that gets saved to the database, unless we unset it here.


bordoni:
👍

PR #1370, opened 2017-04-28
fix comments re: legacy full-address.php | spotfix

cliffordp:
https://theeventscalendar.com/support/forums/topic/full-address-php-template-file/#post-1275235


bordoni:
@cliffordp when the Next MR branch gets created you should move the base branch to `release/4.5.2` so it gets merged there.

PR #1373, opened 2017-04-28
change message slug to match REST one

lucatume:
Ticket: https://central.tri.be/issues/72634

Add the no-results message


lucatume:
Merging as already vetted.

PR #1380, opened 2017-05-09
Fix tests

lucatume:
Again cherry pick from the `bucket\recurrence` branch to fix the broken tests I've found there in `master` too


borkweb:
:+1:

PR #1381, opened 2017-05-09
cache the request reply to avoid hangouts

lucatume:
Ticket: not really

This PR adds a transient caching when trying to fetch EA origins; if our server is down or unresponsive each admin reques will hang for ~10 seconds making the admin UI unusable.
This change implements a logic like: if EA did not respond or responded "nope" keep that and try again in 5'

### DIFF
--- a/src/Tribe/Aggregator/Service.php
+++ b/src/Tribe/Aggregator/Service.php
@@ -238,7 +238,7 @@ class Tribe__Events__Aggregator__Service {
 
 		// If we have an WP_Error we return only CSV
 		if ( is_wp_error( $response ) || empty( $response->status ) ) {
-			set_transient('tribe_aggregator_origins_response', $origins, 300);
+			set_transient( 'tribe_aggregator_origins_response', $origins, 300 );
 
 			return $origins;
 		}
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space after the first parenthesis | [Service.php:241](https://github.com/moderntribe/the-events-calendar/blob/af0ac2a642bae90710e1cf7cff3e51c31d2003fb/src/Tribe/Aggregator/Service.php#L241)
:x: | There must be a single space before the last parenthesis | [Service.php:241](https://github.com/moderntribe/the-events-calendar/blob/af0ac2a642bae90710e1cf7cff3e51c31d2003fb/src/Tribe/Aggregator/Service.php#L241)

bordoni:
I remember us having something of that nature but removing it for some reason. Would like to try to dig why before we got forward with merging this.

lucatume:
Take your time, keep in mind this only kicks in when the fetching fails.

lucatume:
Closing this: we are modifying the target code base enough to find a better solution to the problem.

PR #1385, opened 2017-05-12
New hook to control inclusion/exclusion of year

barryhughes:
Following user feedback, makes it easier to exercise control over the inclusion or exclusion of the year in the event schedule details string.

:ticket: [#78070](https://central.tri.be/issues/78070)


borkweb:
:+1:

PR #1390, opened 2017-05-19
Change field initialization for pre-5.6 compatibility | #79208

barryhughes:
Initialize the `$time_to_live` property in a way that is compatible with PHP 5.5 and earlier.

:ticket: [#79208](https://central.tri.be/issues/79208)


zbtirrell:
👍

PR #1396, opened 2017-05-25
Pass $meta array as the error context | spotfix

barryhughes:
`$type` isn't actually defined, so it makes more sense to pass the `$meta` array (which will contain the type if set, besides potentially conveying other information).


bordoni:
👍

PR #1397, opened 2017-05-25
Fix/79344 get record erorrs

elimn:
Ticket: https://central.tri.be/issues/79344

 A customer had some records in her database with a blank origin. This is showing errors in many places because our Aggregator Records get methods all can return a Tribe_Error, or in some cases null. We need to properly check if this is happening whenever we call these.

### DIFF
--- a/src/Tribe/Aggregator/Cron.php
+++ b/src/Tribe/Aggregator/Cron.php
@@ -299,6 +299,10 @@ class Tribe__Events__Aggregator__Cron {
 		foreach ( $query->posts as $post ) {
 			$record = Tribe__Events__Aggregator__Records::instance()->get_by_post_id( $post );
 
+			if ( tribe_is_error( $record ) ) {
+				continue;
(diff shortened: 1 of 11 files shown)


elimn:
Would `Tribe__Error::is_error()` be a better alternative to a function? Only reason I went with a function is because WP does it with `wp_is_error()`.

ggwicz:
The use of a function seems appropriate to me given [this note from the engineering guide](http://moderntribe.github.io/products-engineering/php/):

> If there is a piece of code that is frequently referenced throughout the addons, or could be useful to users who want to create custom templates or addons, that’s when it’s appropriate to create a global function.

ggwicz:
@elimn probably stupid question: where is `tribe_is_error()` defined?

elimn:
@ggwicz My bad brother. tribe_is_error() is defined in Common. It just got merged seconds ago to the MR branch.

PR #1399, opened 2017-05-29
Make the update confirmation message easier to translate

barryhughes:
Keep markup and translatable elements separate.

:ticket: [#79729](https://central.tri.be/issues/79729)


elimn:
👍

PR #1404, opened 2017-06-01
Filter shortcodes before trimming excerpt

awbauer:
Ran into an issue on one of our sites with a shortcode partially showing in an excerpt. 

screenshot 2017-06-01 09 45 58

From looking at the `tribe_events_get_the_excerpt` function, trimming the text _after_ running the shortcodes seemed to make the most sense for this scenario. Not sure if there may be other implications I haven't seen. Tested this on our sites & the patch worked out great.


barryhughes:
:ticket: [#81226](https://central.tri.be/issues/81226)

barryhughes:
Thanks @awbauer!

PR #1405, opened 2017-06-02
Fix scheduled imports not running

lucatume:
Ticket: https://central.tri.be/issues/79587

This PR removes an over-zealous inconsistency check I have added (so yes: this issue is completely my fault) to allow scheduled imports to run free again.

### DIFF
--- a/src/Tribe/Aggregator/Cron.php
+++ b/src/Tribe/Aggregator/Cron.php
@@ -303,12 +303,6 @@ class Tribe__Events__Aggregator__Cron {
 				continue;
 			}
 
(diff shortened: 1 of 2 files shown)


zbtirrell:
I'm surprised there are no new tests or adjustments of existing tests to reflect this change.

lucatume:
Simply put: EA client code is not testable as is; it would require refactoring to put it in that condition.

zbtirrell:
OK - we can address the lack of testability in a future iteration :)

lucatume:
I'm totally in to do it, just in case.

PR #1410, opened 2017-06-05
Make our css targeting more specific for day view.

Camwyn:
Add id to selector for better targeting - so we don't accidentally pad the body!

Ticket: https://central.tri.be/issues/73421


Camwyn:
This has been overridden by https://github.com/moderntribe/the-events-calendar/pull/1415

PR #1414, opened 2017-06-06
bind the Known Range class in Main

lucatume:
Part of the work done for https://central.tri.be/issues/79791
The binding is needed to rebuild the known range in the Instance Engine on update.


bordoni:
👍

PR #1416, opened 2017-06-06
#80178 - Update Sold Out Message to be in a grey box matching stock box

jesseeproductions:
_Ref:_ [C#80178](https://central.tri.be/issues/80178)


bordoni:
Don't update common hash.

jesseeproductions:
Changed common back

barryhughes:
So long as that common ref is ok (not always 100% clear via GH's diff viewer), looks good to merge :+1:

PR #1426, opened 2017-06-13
Remove jQuery UI JS and remove from packaging whitelist | #80766

barryhughes:
* Removes unused JS from a vendor lib
* Removes from the packaging whitelist

Note that we do still appear to use the CSS and image assets, so the library hasn't been removed completely.

:ticket: [#80766](https://central.tri.be/issues/80766)


jesseeproductions:
👍

PR #1444, opened 2017-06-24
Fix/72241 update timepicker lib

elimn:
Ticket: https://central.tri.be/issues/72241

I don't necessarily want to merge this, I am just opening a PR to get some code review and feedback on this before deciding if we want to commit it here or in another release.

- In order to patch this bug, and many others that the Bootstrap Datepicker has patched, I am updating to version 1.7.0. This version is backwards compatible with 1.4.0, and was actually quite easy to update to.
- I wanted to move any modifications we were making to the lib to our own files, to make it even easier to update in the future. I did a file diff between our old version and the packaged vendor lib, finding any modifications we had made and moving them out. Most of the modifications we had made were actually patching bugs that got fixed in the newer version, and did not need moving.
- We are only using two files from the lib, a JS and a "standalone" CSS file. I made the directory structure and filenames match those of the packaged library, that way it will be really easy to see which file matches up in the future. The rest of the files in this folder were unused, and I stripped them out to save space.
- In theory, if we update in the future it should be as easy as copying those two files over from the packaged zip now that our code is nice and discrete.

**Questions**

1. The new version allows you to set the date format used for the "title" of the datepicker. This title appears between the Next/Prev month buttons at the top of the datepicker. By default it is in the YYYY MM format. It uses it's own unique formatting codes. I could not think of a good way to convert our current date-time formatting options to this Bootstrap one, so I just left it at the default. Can anyone come up with a better idea? This format can be changed using the tribe_bootstrap_datepicker_strings filter, so at the very least it can be customized it just doesn't honor any backend settings.
2. Should this be released in an MR or FR? It seems highly backwards compatible, and from what I have seen would be extremely unlikely to break customizations. But, I'm still leaning towards an FR just to be safe.


bordoni:
Interesting Pull Request, would be amazing to bring this into the _Dev Scrum_ correct?

src/resources/postcss/tribe-events-full.pcss
```diff
@@ -1422,8 +1422,17 @@ select.tribe-events-dropdown {
 			}
 		}
 	}
+
+	body.admin-bar & {
```
ggwicz: is the `&` character necessary here? to my (limited) knowledge, the ampersand can be used to _precede_ selectors, but not follow them like it is here...right?

elimn: It can even precede like is done here. Since it precedes the output CSS rule is `body.admin-bar .datepicker`, thus it will only target the datepicker when the admin bar is present.

The advantage of doing it this way is that is that now we only have one `.datepicker` selector and all of the CSS pertaining to it is grouped together. If we ever change the name of that selector we only have to change it in one place. And, actually it probably does make sense to change it to the more verbose `.datepicker-dropdown` selector at some point.

That said, it's mostly just a style thing and if you don't like it it can certainly be changed.

elimn:
@bordoni That's a good idea. Probably the best time for it.

ggwicz:
I would tag this `merge`, but since we're going to discuss it on Code Scrum etc., I'm just tagging it`hold` for now until some other folks can weigh in.

PR #1449, opened 2017-06-29
Load Tribe*common textDomain

sebastienserre:


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space after the first parenthesis | [Main.php:346](https://github.com/moderntribe/the-events-calendar/blob/5a007c5c4c775ccc516bf2dc8d4546f5e1800fe3/src/Tribe/Main.php#L346)

barryhughes:
Hi @sebastienserre - thanks for the PR! Can you provide some notes on why you need this change? If there is a problem with the loading of textdomains within Tribe Common, we'd _probably_ prefer to resolve it with a change [over here](https://github.com/moderntribe/tribe-common).

sebastienserre:
Hello,
I think this PR can be improved as ut won't work as I'd like...
As we already spoke, translate.WP.org is asking us (I'm PTE) to translate strings with the text domain 'Tribe-common' & these strings translated are not used in the WP website as the textdomain isn't loaded.
I think that if a `load_plugin_textdomain()` is loaded, the tribe common Stgring should be translated.
Today, TEC is 100% translated in French (https://translate.wordpress.org/locale/fr/default/wp-plugins/the-events-calendar) but in my website the part managed by tribe common are allways in English.

barryhughes:
There's certainly a need for more work to be done on the French translation of Tribe Common - currently it's sitting at 68% complete:

http://translations.theeventscalendar.com/projects/tribe-common/fr/default

Submitting translations via the above URL is the best way to have them included in our releases: though I do appreciate how frustrating it must be to translate strings from tribe-common via [translate.wordpress.org](https://translate.wordpress.org/projects/wp-plugins/the-events-calendar) only to find they aren't used, I don't think the change you've submitted here is the correct way to resolve that.

We'll certainly give the problem some thought, though.

PR #1450, opened 2017-06-30
Fix language switcher for WP installed in a folder

dgwatkins:
If WordPress is installed in a folder, we need to remove the folder from request_uri before using it.

Fixes the following support threads:
https://wpml.org/forums/topic/ruta-del-language-switcher-no-ok-con-el-plugin-the-events-managers/
https://wpml.org/forums/topic/translation-nav-menu-adds-another-path/

:ticket: [#81998](https://central.tri.be/issues/81998)


barryhughes:
Thanks @dgwatkins! Just wondering, would it be viable for us to use something like `WPML_Request::get_request_uri_lang()` for this (or is there any other method or function already baked into WPML that will provide this information)?

dgwatkins:
That function WPML_Request::get_request_uri_lang() is actually for getting the current language from the URL. 
We also have WPML_Request::get_request_uri() but it includes the folder like the $_SERVER variable so its no use here.

dgwatkins:
I can see how str_replace can be problematic there.
Let me look for a different solution

barryhughes:
I don't mind the `str_replace()` particularly - just crossed my mind there might be existing WPML functions we could build off of - if not, no problem :-)

dgwatkins:
The problem with WPML undocumented functions is that can change at any time.
I think its going to be better as it is.

Let me know if you need anything else.

barryhughes:
Ticketed: :ticket: [#81998](https://central.tri.be/issues/81998)

PR #1451, opened 2017-06-30
Fix/75363 eventbrite notice unsetup

elimn:
Ticket: https://central.tri.be/issues/75363

It became apparent during that ticket that when an item is disabled in the Import dropdown menu, it looks like it is part of the EA upsell we show when no EA key is present. So, I rejiggered things a bit to make the EA upsell visually distinct and always appear under the list of currently active origins.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $sort is never used | [Origins.php:80](https://github.com/moderntribe/the-events-calendar/blob/f7a47249d71becd9d2a1a1e49ec2d8b5a782ed3b/src/Tribe/Aggregator/API/Origins.php#L80)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$option' | [import-form.php:90](https://github.com/moderntribe/the-events-calendar/blob/f7a47249d71becd9d2a1a1e49ec2d8b5a782ed3b/src/admin-views/aggregator/tabs/import-form.php#L90)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$option' | [import-form.php:104](https://github.com/moderntribe/the-events-calendar/blob/f7a47249d71becd9d2a1a1e49ec2d8b5a782ed3b/src/admin-views/aggregator/tabs/import-form.php#L104)

src/admin-views/aggregator/tabs/import-form.php
```diff
@@ -1,5 +1,6 @@
 <?php
-$has_license_key = Tribe__Events__Aggregator::is_service_active();
+$has_license_key = tribe( 'events-aggregator.main' )->is_service_active();
```
barryhughes: Not sure why this changed - shouldn't it remain a static method call?

src/admin-views/aggregator/tabs/import-form.php
```diff
@@ -38,6 +39,23 @@
 		$field->placeholder = esc_attr__( 'Select Origin', 'the-events-calendar' );
 		$field->help = esc_attr__( 'Choose where you are importing from.', 'the-events-calendar' );
 		$field->options = tribe( 'events-aggregator.main' )->api( 'origins' )->get();
+		$field->upsell_options = array();
+
+		foreach ( $field->options as $key => $option ) {
+			$option->disabled = ( isset( $option->disabled ) ? $option->disabled : null );
```
barryhughes: Remove the outer parentheses?

src/admin-views/aggregator/tabs/import-form.php
```diff
@@ -38,6 +39,23 @@
...
+			$option->upsell = ( isset( $option->upsell ) ? $option->upsell : false );
+
+			$option->is_selected = false;
+			if ( ! empty( $_GET['ea-auth'] ) && $option->id === $_GET['ea-auth'] ) {
```
barryhughes: Yoda-ify

src/resources/js/aggregator-fields.js
```diff
@@ -711,10 +711,12 @@ tribe_aggregator.fields = {
 			}
 
 			args.upsellFormatter = function( option ) {
-				if ( 'redirect' == option.id ) {
-					var parts = option.text.split( '|' );
-					option.text = parts[0] + '<br><span class="tribe-upsell-subtitle">' + parts[1] + '</span>';
+				var $option = jQuery( option.element );
```
barryhughes: Could shorten to `$( option.element )`

src/admin-views/aggregator/tabs/import-form.php
elimn: Since it is the return of a function it can be called non statically. If I tried to call it statically I would get a `T_PAAMAYIM_NEKUDOTAYIM` error in PHP 5.2. I just tested this to double check. Should work in new and old version of PHP, and we're using the same line of code in a few other places.

bordoni: We have a bunch of other places that we use this structure, I would rather make `is_service_active` a non-static method.

barryhughes: I still feel like we _shouldn't_ (this implies it's an instance method and operates on an instance: it doesn't) but if it's the convention we're using elsewhere that's fine

barryhughes: Yep, or what @bordoni said (but that doesn't need to happen as part of this change)

PR #1461, opened 2017-07-13
Fix double anchor tag in tribe_get_organizer_details

f4w-pwharton:
`tribe_get_organizer_details()` wraps the result of `tribe_get_organizer_website_link()` in an anchor tag, however `tribe_get_organizer_website_link()` already returns an anchor tag resulting in broken HTML. 

This stops `tribe_get_organizer_details()` adding the second anchor tag.

Bug originally reported here in March: https://theeventscalendar.com/support/forums/topic/single-organizer-website-link-bug/

### DIFF
--- a/src/functions/template-tags/organizer.php
+++ b/src/functions/template-tags/organizer.php
@@ -125,6 +125,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 		}
 
 		if ( $organizer_id && $link = tribe_get_organizer_website_link() ) {
+			// $link is a full HTML string (<a>) whose components are already escaped, so we don't need create an anchor tag or escape again here
 			$details[] = '<span class="link">' . $link . '</span>';
 		}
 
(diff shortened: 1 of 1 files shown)


src/functions/template-tags/organizer.php
```diff
@@ -125,7 +125,7 @@ function tribe_get_organizer_details( $post_id = null ) {
 		}
 
 		if ( $organizer_id && $link = tribe_get_organizer_website_link() ) {
-			$details[] = '<span class="link"> <a href="' . esc_attr( $link ) . '">' . $link . '</a> </span>';
+			$details[] = '<span class="link">' . $link . '</span>';
```
ggwicz: Should add a comment here that explains why `$link` isn't escaped, etc. Something like:
```
// $link is a full HTML string whose components are already escaped, so we don't need to escape again here
```

ggwicz:
@f4w-pwharton, thanks for submitting this change. I flagged one small tweak that would enhance this enough for us to merge it. Thanks!

barryhughes:
:ticket: [#91074](https://central.tri.be/issues/91074)

PR #1462, opened 2017-07-13
Avoid overwriting FB Venues and Organizers when they have the same name

lucatume:
Ticket: https://central.tri.be/issues/75370

This PR uses, in the case of FB and Meetup origins, the ID provided by the origin to look for duplicates and avoid false positives while doing so.
What was happening in the ticket is that `Venue Foo` with a FB ID of `12345` and `Venue Foo` with a FB ID of `678910` would, incorrectly, be matched as same.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -334,6 +334,7 @@ Please see the changelog for the complete list of changes in this release. Remem
 * Fix - Add a safety check in isOrganizer() function (our thanks to Kevin for flagging this!) [81645]
 * Fix - Avoid EA Client hanging when no events are found while attempting an import from a Facebook source [82713]
 * Fix - Improve compatibility of The Events Calendar when operating with WPML from within a subdirectory (props: @dgwatkins) [81998]
+* Fix - Avoid overwriting Venues and Organizers when importing FB events with similarly named Venues and Organizers [75370]
 
 = [4.5.7] 2017-06-28 =
 
--- a/readme.txt
+++ b/readme.txt
@@ -329,7 +329,7 @@ Please see the changelog for the complete list of changes in this release. Remem
 
 = [4.5.9] TBD =
 
-* ...
+* Fix - Avoid overwriting Venues and Organizers when importing FB events with similarly named Venues and Organizers [75370]
 
 = [4.5.8] 2017-07-13 =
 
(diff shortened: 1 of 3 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | TRUE, FALSE and NULL must be lowercase; expected &quot;null&quot; but found &quot;NULL&quot; | [Abstract.php:1784](https://github.com/moderntribe/the-events-calendar/blob/c3efd2a53c5dd524a2e0ec84a1676d750829e9c8/src/Tribe/Aggregator/Record/Abstract.php#L1784)
:x: | Space found before square bracket; expected &quot;$fields[&quot; but found &quot;$fields [&quot; | [Abstract.php:1787](https://github.com/moderntribe/the-events-calendar/blob/c3efd2a53c5dd524a2e0ec84a1676d750829e9c8/src/Tribe/Aggregator/Record/Abstract.php#L1787)

src/Tribe/Aggregator/Event.php
```diff
@@ -231,11 +231,10 @@ public static function get_post_by_meta( $key = 'global_id', $value = null ) {
 			'global_id_lineage' => self::$global_id_lineage_key,
 		);
 
-		if ( ! isset( $keys[ $key ] ) ) {
-			return false;
+		if ( isset( $keys[ $key ] ) ) {
+			$key = $keys[ $key ];
```
borkweb: Should we also check `if ( empty( $key ) )` and return if so?

lucatume: I've pushed code to simply require the `$key` parameter.
We always use it with a key.

barryhughes:
Changing target branch: M17.13 set sail already :ship:

readme.txt
```diff
@@ -334,6 +334,7 @@ Please see the changelog for the complete list of changes in this release. Remem
 * Fix - Add a safety check in isOrganizer() function (our thanks to Kevin for flagging this!) [81645]
 * Fix - Avoid EA Client hanging when no events are found while attempting an import from a Facebook source [82713]
 * Fix - Improve compatibility of The Events Calendar when operating with WPML from within a subdirectory (props: @dgwatkins) [81998]
+* Fix - Avoid overwriting Venues and Organizers when importing FB events with similarly named Venues and Organizers [75370]
```
barryhughes: We'll need to sync with release/M17.14 and move this to the 4.5.9 section.

lucatume:
Ok, I've merged M17.14 code into the branch and moved the changelog entry to the 4.5.9 section

PR #1468, opened 2017-07-21
Cast error responses to WP_Error

lucatume:
Ticket: https://central.tri.be/issues/83344

This PR follows up the research work done by @borkweb (https://tribe.slack.com/archives/C1SMDBJQ7/p1500615585145005) to make sure we cast error responses coming from the service to proper WP_Errors.
Using the EA Mocker it observable that, without this, error responses would be queued and try to fetch forever.

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -1839,6 +1839,12 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 	}
 
 	/**
+	 * Cast error responses from the Service to WP_Errors to ease processing down the line.
+	 *
+	 * If a response is a WP_Error already or is not an error response then it will not be modified.
+	 *
+	 * @since TBD
+	 *
 	 * @param WP_Error|object $import_data
 	 *
 	 * @return array|\WP_Error
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -634,8 +634,9 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 	}
 
 	public function get_import_data() {
+		/** @var \Tribe__Events__Aggregator $aggregator */
 		$aggregator = tribe( 'events-aggregator.main' );
-		$data = array();
+		$data       = array();
 
 		// For now only apply this to the URL type
 		if ( 'url' === $this->type ) {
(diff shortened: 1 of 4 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1830,4 +1837,33 @@ protected function db_errors_happened() {
...
+	 * @param WP_Error|object $import_data
+	 *
+	 * @return array|\WP_Error
+	 */
+	protected function maybe_cast_to_error( $import_data ) {
+		if ( is_wp_error( $import_data ) ) {
+			return $import_data;
+		}
+
+		if ( ! empty( $import_data->status ) && 'error' === $import_data->status ) {
```
borkweb: I wonder if we should _also_ look and see if the response was something other than `200` and generate an error as well. Right now with our throttling, we return `403`.

src/Tribe/Aggregator/Service.php
```diff
@@ -172,7 +172,14 @@ public function get( $endpoint, $data = array() ) {
...
 		}
 
-		if ( isset( $response->data ) && isset( $response->data->status ) && '404' === $response->data->status ) {
+		if (
+			(
+				isset( $response->data )
+				&& isset( $response->data->status )
+				&& '404' === $response->data->status
+			)
+		     || '200' != wp_remote_retrieve_response_code( $response )
```
borkweb: I wonder if we should break the `200` out into its own if block with its own error message? And perhaps it would be a good plan to have a separate error message specifically for the `403` we are sending when we throttle.

src/Tribe/Aggregator/Record/Abstract.php
lucatume: I've pushed code to handle that.

src/Tribe/Aggregator/Service.php
borkweb: For `403`:

```php
return new WP_Error( 'core:aggregator:request-limit', esc_html__( 'The Event Aggregator service has blocked this request because it came too soon after a previous request.', 'the-events-calendar' ) );
```

For `! 200`:

```php
return new WP_Error( 'core:aggregator:invalid-service-response', esc_html__( 'There may be an issue with the Event Aggregator server. Please try your import again later.', 'the-events-calendar' ) );
```

lucatume: Latest push updates the HTTP codes handling

borkweb:
And we'll need a changelog entry once we figure out the end home for it, of course. (Just typing that so I don't forget)

PR #1470, opened 2017-07-21
Aggregator Insert Events Cron gets re-enqueued if it's not done

bordoni:
_Ref:_ [C#83382](http://central.tri.be/issues/83382)
---

### DIFF
--- a/src/Tribe/Aggregator/Record/Queue_Processor.php
+++ b/src/Tribe/Aggregator/Record/Queue_Processor.php
@@ -139,17 +139,17 @@ class Tribe__Events__Aggregator__Record__Queue_Processor {
 
 		while ( $this->next_waiting_record() ) {
 			if ( ! $this->do_processing() ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/Record/Queue_Processor.php
```diff
@@ -130,6 +142,15 @@ public function process_queue( $batch_size = null ) {
 				break;
 			}
 		}
+
+		$queue_items = get_post_meta( $this->current_record_id, Tribe__Events__Aggregator__Records::meta_prefix( Tribe__Events__Aggregator__Record__Queue::$queue_key ), true );
+
+		// We only get here if we done processing this batch.
```
borkweb: `[...] if we are not finished processing this batch`?

bordoni: If we are finished this batch of 5.

PR #1475, opened 2017-07-28
#80426 | Adjust "E-mail Address Obfuscation" Label When in Community Events

ggwicz:
_Ref:_ [C#80426](http://central.tri.be/issues/80426)

### DIFF
--- a/src/admin-views/create-organizer-fields.php
+++ b/src/admin-views/create-organizer-fields.php
@@ -16,6 +16,10 @@
 	</td>
 	<td class="organizer-email">
 		<input id="organizer-email" type='text' name='organizer[Email][]' class='organizer-email' size='25' value='' />
-		<p><?php esc_html_e( 'The e-mail address will be obfuscated on your site to avoid it getting harvested by spammers.', 'the-events-calendar' ); ?></p>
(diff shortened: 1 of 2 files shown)


src/admin-views/create-organizer-fields.php
```diff
@@ -16,6 +16,10 @@
 	</td>
 	<td class="organizer-email">
 		<input id="organizer-email" type='text' name='organizer[Email][]' class='organizer-email' size='25' value='' />
-		<p><?php esc_html_e( 'The e-mail address will be obfuscated on your site to avoid it getting harvested by spammers.', 'the-events-calendar' ); ?></p>
+		<?php if ( function_exists( 'tribe_is_community_edit_event_page' ) && tribe_is_community_edit_event_page() ) : ?>
```
barryhughes: While this works, imho TEC shouldn't know or care about the existence of Community Events functions (we shouldn't code and fix problems for our dependencies, it's a sign things are flowing in the wrong direction). 

I'd much rather the string be filterable (CE can then take advantage of the filter) or, simpler yet, how about simply using the first form of the string which works well in both situations?

ggwicz: > how about simply using the first form of the string which works well in both situations

The best solution by far! Will commit this change.

ggwicz:
@barryhughes, i've implemented your simple-but-perfect solution and this is ready for review again

PR #1481, opened 2017-08-02
Don't use HTML tags in translation strings

ramiy:
Moving the `HTML` tags out of the translation strings.

This makes it much easier to translate the strings to RTL languages.

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -941,7 +941,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 			// What the user can do
 			$edit_post_link = sprintf( __( 'Ask the site administrator to edit the %s slug', 'the-events-calendar' ), $name );
 			if ( isset( $post_type->cap->edit_posts ) && current_user_can( $post_type->cap->edit_posts ) ) {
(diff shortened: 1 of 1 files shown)


ramiy:
The old strings:

* `<a href="%s">Edit the %s slug</a>`
* `<a href="%s">edit Events settings</a>.`

The new strings:

* `Edit the %s slug`
* `edit Events settings`

ramiy:
One more bad example:

`<p>The following three fields accept the date format options available to the php date() function. <a href="http://codex.wordpress.org/Formatting_Date_and_Time" target="_blank">Learn how to make your own date format here</a>.</p>`

* The `<p>` tag should be removed from the translation string.
* The `date()` function should be replaced with a `%1$s` placeholder and moved outside the translation string.
* The URL `https://...` should be replaced with a `%2$s` placeholder and moved outside the translation string.

ramiy:
And seems like you have several other strings that need to be fixed. To make it easier to translate the plugin.

barryhughes:
:100: we're completely on-board with that and if you review recent code changes, I think you'll see this is a pattern we've made efforts to avoid in new code and wherever we need to modify existing strings.

We've generally avoided making sweeping changes of established strings, though, simply because there are lots of existing translations that are already in place for them and doing so would cause unnecessary disruption for many.

That approach is something we're open to revising, though, and we're always happy to receive solid community-sourced changes like this one: thanks for the contribution!

PR #1489, opened 2017-08-03
Update tribe-events-full.css for Bug #65136

jentheo:
Moving tooltips to display next to event titles (like with week view) instead of above to prevent them from being cut off


barryhughes:
Branch closed automatically when cleaning up M17.15 (and we'd decided not to merge this in as is). Work can continue in M17.16.

PR #1491, opened 2017-08-04
fix more issues found using the test plugin

lucatume:
Tickets:
* https://central.tri.be/issues/81691
* https://central.tri.be/issues/79740
* https://central.tri.be/issues/76730
* https://central.tri.be/issues/76729
* https://central.tri.be/issues/79737
* https://central.tri.be/issues/79739

More fixes found while using the testing plugin.

### DIFF
--- a/src/Tribe/Venue.php
+++ b/src/Tribe/Venue.php
@@ -495,7 +495,8 @@ class Tribe__Events__Venue extends Tribe__Events__Linked_Posts__Base {
 		                      ) );
 
 		if ( count( $args ) > 1 ) {
-			$tag = 'save_post_' . Tribe__Events__Main::VENUE_POST_TYPE;
+			$post_type = Tribe__Events__Main::VENUE_POST_TYPE;
+			$tag       = "save_post_{$post_type}";
 			remove_action( $tag, array( tribe( 'tec.main' ), 'save_venue_data' ), 16 );
 			wp_update_post( $args );
 			add_action( $tag, array( tribe( 'tec.main' ), 'save_venue_data' ), 16, 2 );
--- a/src/Tribe/Venue.php
+++ b/src/Tribe/Venue.php
@@ -318,10 +318,10 @@ class Tribe__Events__Venue extends Tribe__Events__Linked_Posts__Base {
 		}
 
 
-		update_post_meta( $venue_id, '_EventShowMapLink', isset( $data['ShowMapLink'] ) );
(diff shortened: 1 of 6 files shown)


src/Tribe/Venue.php
```diff
@@ -495,16 +495,19 @@ public function update( $venue_id, $data ) {
 		                      ) );
 
 		if ( count( $args ) > 1 ) {
+			$tag = 'save_post_' . Tribe__Events__Main::VENUE_POST_TYPE;
+			remove_action( $tag, array( tribe( 'tec.main' ), 'save_venue_data' ), 16 );
```
bordoni: `tribe_callback`?

src/Tribe/Venue.php
```diff
@@ -495,16 +495,19 @@ public function update( $venue_id, $data ) {
...
 			wp_update_post( $args );
+			add_action( $tag, array( tribe( 'tec.main' ), 'save_venue_data' ), 16, 2 );
```
bordoni: `tribe_callback`?

src/Tribe/Venue.php
```diff
@@ -495,16 +495,19 @@ public function update( $venue_id, $data ) {
...
 		}
 
 		if ( isset( $data['ShowMap'] ) && ! tribe_is_truthy( $data['ShowMap'] ) ) {
-			unset( $data['ShowMap'] );
+			$data['ShowMap'] = 'false';
```
bordoni: Should we use `no`/`yes` as is that The WordPress norm?

src/Tribe/Venue.php
```diff
@@ -495,16 +495,19 @@ public function update( $venue_id, $data ) {
 		                      ) );
 
 		if ( count( $args ) > 1 ) {
+			$tag = 'save_post_' . Tribe__Events__Main::VENUE_POST_TYPE;
```
bordoni: I think it would be nice if we could build the tag with a variable in the string. Normally easy to regexp for.

src/Tribe/Venue.php
lucatume: not sure... checking

src/Tribe/Venue.php
lucatume: not in this case: we'd need to update that on main too and I do not want to make that leap now.

src/Tribe/Venue.php
lucatume: no, it does not work: the rest of our code uses "false" and "1"

bordoni: Perfect.

PR #1493, opened 2017-08-04
Fix/76411 get link ld

elimn:
Ticket: https://central.tri.be/issues/76411

Related PR: https://github.com/moderntribe/tribe-common/pull/469

When pro isn't active the get linked post functions return the name of the organizer or venue :-/ So we should go with false here.

Took the opportunity to update tribe_get_event_link() a bit. Mostly it just needs to accept a WP_Post or ID, like everything else. But while I was adding that...

### DIFF
--- a/src/Tribe/JSON_LD/Event.php
+++ b/src/Tribe/JSON_LD/Event.php
@@ -140,4 +140,15 @@ class Tribe__Events__JSON_LD__Event extends Tribe__JSON_LD__Abstract {
 		return str_replace( $mon_thousands_sep, '', $price );
 	}
 
+	/**
+	 * Get a link to the event
+	 *
+	 * @param  int|WP_Post  $post The Post Object or ID
+	 *
+	 * @return false|string Link to the event or false
+	 */
+	protected function get_link( $post ) {
+		return tribe_get_event_link( Tribe__Main::post_id_helper( $post ) );
+	}
+
 }
--- a/src/Tribe/JSON_LD/Event.php
+++ b/src/Tribe/JSON_LD/Event.php
@@ -148,7 +148,7 @@ class Tribe__Events__JSON_LD__Event extends Tribe__JSON_LD__Abstract {
 	 * @return false|string Link to the event or false
 	 */
 	protected function get_link( $post ) {
-		return tribe_get_event_link( Tribe__Main::post_id_helper( $post ) );
+		return tribe_get_event_link( $post );
 	}
 
 }
--- a/src/Tribe/JSON_LD/Event.php
+++ b/src/Tribe/JSON_LD/Event.php
@@ -143,6 +143,8 @@ class Tribe__Events__JSON_LD__Event extends Tribe__JSON_LD__Abstract {
 	/**
 	 * Get a link to the event
 	 *
+	 * @since TBD
+	 *
 	 * @param  int|WP_Post  $post The Post Object or ID
 	 *
 	 * @return false|string Link to the event or false
--- a/src/Tribe/JSON_LD/Organizer.php
+++ b/src/Tribe/JSON_LD/Organizer.php
@@ -59,4 +59,21 @@ class Tribe__Events__JSON_LD__Organizer extends Tribe__JSON_LD__Abstract {
 		return array( $post_id => $data );
 	}
 
+	/**
+	 * Get a link to the event
+	 *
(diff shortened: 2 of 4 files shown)


src/Tribe/JSON_LD/Venue.php
```diff
@@ -79,4 +79,21 @@ public function get_data( $post = null, $args = array( 'context' => false ) ) {
...
 
+	/**
+	 * Get a link to the event
+	 *
+	 * @param  int|WP_Post  $post The Post Object or ID
+	 *
+	 * @return false|string Link to the event or false
+	 */
+	protected function get_link( $post ) {
+		if ( class_exists( 'Tribe__Events__Pro__Main' ) ) {
```
bordoni: I would rather us filter from Pro to change the Link. (if possible)

src/Tribe/JSON_LD/Organizer.php
```diff
@@ -59,4 +59,21 @@ public function get_data( $post = null, $args = array( 'context' => false ) ) {
...
 
+	/**
+	 * Get a link to the event
+	 *
+	 * @param  int|WP_Post  $post The Post Object or ID
+	 *
+	 * @return false|string Link to the event or false
+	 */
+	protected function get_link( $post ) {
+		if ( class_exists( 'Tribe__Events__Pro__Main' ) ) {
```
bordoni: I would rather us filter from Pro to change the Link. (if possible)

src/Tribe/JSON_LD/Event.php
```diff
@@ -140,4 +140,15 @@ protected function normalize_price( $price ) {
 		return str_replace( $mon_thousands_sep, '', $price );
 	}
 
+	/**
+	 * Get a link to the event
+	 *
+	 * @param  int|WP_Post  $post The Post Object or ID
+	 *
```
bordoni: `@since`

src/Tribe/JSON_LD/Organizer.php
```diff
@@ -59,4 +59,21 @@ public function get_data( $post = null, $args = array( 'context' => false ) ) {
 		return array( $post_id => $data );
 	}
 
+	/**
+	 * Get a link to the event
+	 *
+	 * @param  int|WP_Post  $post The Post Object or ID
+	 *
```
bordoni: `@since`

src/Tribe/JSON_LD/Venue.php
```diff
@@ -79,4 +79,21 @@ public function get_data( $post = null, $args = array( 'context' => false ) ) {
 		return array( $post_id => $data );
 	}
 
+	/**
+	 * Get a link to the event
+	 *
+	 * @param  int|WP_Post  $post The Post Object or ID
+	 *
```
bordoni: `@since`

src/Tribe/JSON_LD/Organizer.php
elimn: I completely agree. However, this MR does not affect Pro. Further, we need to make some overhauls to all three of these functions, which is ticketed: https://central.tri.be/issues/33734

Perhaps we can just add this item to that overhaul? Would that work @bordoni ?

borkweb: I'm not a huge fan of the half-step but for now I suppose it'll have to suffice in order to get this fix out there. Adding it to that overhaul would work. Could you also throw a `@TODO` on this and the venue if statement that points at the overhaul ticket and advertises that we don't like this hacky solution and it should be fixed at a later date?

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task &quot;Move this logic to Pro once #33734 is handled&quot; | [Organizer.php:72](https://github.com/moderntribe/the-events-calendar/blob/e3a90d47e60fd0d2a132388efe71191b006f0199/src/Tribe/JSON_LD/Organizer.php#L72)
:warning: | Comment refers to a TODO task &quot;Move this logic to Pro once #33734 is handled&quot; | [Venue.php:92](https://github.com/moderntribe/the-events-calendar/blob/e3a90d47e60fd0d2a132388efe71191b006f0199/src/Tribe/JSON_LD/Venue.php#L92)

src/Tribe/JSON_LD/Organizer.php
elimn: Verily. I added the `@TODO`.

PR #1495, opened 2017-08-04
Fail EA record fetches in a way that the queue can fail appropriately

borkweb:
We were only marking the record as failed if it was not part of a Queue.  We need to fail any time `prep_import_data()` is called and its fetch of the data is a WP_Error.

In addition to the above, I've also included a fix that allows import errors to properly get logged to the comment table.

See: https://central.tri.be/issues/83767


bordoni:
_Change Log_

PR #1497, opened 2017-08-06
Update Query.php to allow 60s query caching

garrettjohnson:
By including the seconds string in the default list query mysql has a difficult/impossible time caching the query results, because the query changes every second. By replacing `H:i:s` with `H:i:00` mysql can cache the query for 60s, which results in events being displayed with the same precision, since events do not use seconds, and better performance out of the box.


barryhughes:
:ticket: [#84841](https://central.tri.be/issues/84841)

barryhughes:
Thanks @garrettjohnson!

PR #1500, opened 2017-08-08
make sure date is fetched from the db if not available

lucatume:
Tickets:
* https://central.tri.be/issues/76730
* https://central.tri.be/issues/76729

### DIFF
--- a/src/Tribe/API.php
+++ b/src/Tribe/API.php
@@ -271,10 +271,14 @@ if ( ! class_exists( 'Tribe__Events__API' ) ) {
 
 			if ( isset( $data['EventStartDate'] ) ) {
 				$data['EventStartDate'] = Tribe__Date_Utils::datetime_from_format( $datepicker_format, $data['EventStartDate'] );
+			} elseif ( $existing_start_date = get_post_meta( $event_id, '_EventStartDate', true ) ) {
+				$data['EventStartDate'] = $existing_start_date;
 			}
 
 			if ( isset( $data['EventEndDate'] ) ) {
 				$data['EventEndDate'] = Tribe__Date_Utils::datetime_from_format( $datepicker_format, $data['EventEndDate'] );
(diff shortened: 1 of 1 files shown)


src/Tribe/API.php
```diff
@@ -271,10 +271,14 @@ protected static function prepare_event_date_meta( $event_id, $data ) {
 
 			if ( isset( $data['EventStartDate'] ) ) {
 				$data['EventStartDate'] = Tribe__Date_Utils::datetime_from_format( $datepicker_format, $data['EventStartDate'] );
+			} elseif ( $existing_start_date = get_post_meta( $event_id, '_EventStartDate', true ) ) {
+				$data['EventStartDate'] = $existing_start_date;
 			}
 
 			if ( isset( $data['EventEndDate'] ) ) {
 				$data['EventEndDate'] = Tribe__Date_Utils::datetime_from_format( $datepicker_format, $data['EventEndDate'] );
+			} elseif ( $existing_end_date = get_post_meta( $event_id, '_EventEndDate', true ) ) {
```
bordoni: Should we use `tribe_get_event_meta`?

src/Tribe/API.php
```diff
@@ -271,10 +271,14 @@ protected static function prepare_event_date_meta( $event_id, $data ) {
 
 			if ( isset( $data['EventStartDate'] ) ) {
 				$data['EventStartDate'] = Tribe__Date_Utils::datetime_from_format( $datepicker_format, $data['EventStartDate'] );
+			} elseif ( $existing_start_date = get_post_meta( $event_id, '_EventStartDate', true ) ) {
```
bordoni: Should we use `tribe_get_event_meta`?

lucatume: I think it adds an overhead we do not need in this case (post id helper and such); it's good in "combat zone": by the time we are here we know the `$event_id` is legit

src/Tribe/API.php
lucatume: same as above

PR #1501, opened 2017-08-08
fix REST API date filtering issues

lucatume:
Ticket: https://central.tri.be/issues/78375

This ticket fixes two issues:
1. EA client would send incorrect date limits to the EA Service
2. REST API server, the one baked into TEC, would incorrectly parse those incorrect date limits on its end

please note that the fixes to the REST API codes will not be ported over to the `bucket/full-rest-api` branch as the logic is completely refactored there


barryhughes:
We're through smoketesting for M17.15: let's target M17.16 instead.

barryhughes:
...and shift the changelog entry accordingly etc

lucatume:
Should be ok: changed base and moved changelog entry

PR #1505, opened 2017-08-10
Readme Spelling

garrett-eclipse:


readme.txt
```diff
@@ -329,11 +329,11 @@ Please see the changelog for the complete list of changes in this release. Remem
 
 = [4.5.10] 2017-08-09 =
 
-* Fix - Avoid duplicate events when importing from some iCal, Google Calendar and Facebook feeds in Event Aggregator (our thanks to Jason Sears, controlyours and many other users for highlighting these issues) [67038]
+* Fix - Avoid duplicate events when importing from some iCal, Google Calendar and Facebook feeds in Event Aggregator (our thanks to Jason Sears, control yours and many other users for highlighting these issues) [67038]
```
barryhughes: This is a username without spaces rather than two separate words, we could prefix it with `@` or some other suitable convention though

garrett-eclipse: That's my bad @barryhughes I didn't focus on the context well enough. It's fine without the @ but I'll leave that decision to you

PR #1522, opened 2017-08-18
Reschedule failed scheduled imports

lucatume:
Ticket: https://central.tri.be/issues/81639

This PR modifies the way EA Client detects if a scheduled import is in schedule or not to allow failed, not on-demand, scheduled imports to be rescheduled before their frequency.
While testing the ticket in question with the database attached to the ticket I noticed that yes, we would now pick up correct imports but a weekly import that failed 4 days ago would be given another chance to run in 3 days (a week); this modifications tries to address this issue putting in code the following reasoning:

* do this once a week...
* ...but keep trying until if it fails, do not wait the next week

Furthermore the code tries some fallback strategies to handle some incoherent records found in the db:

* missing `last_import_status` meta
* incoherent children records

and the like.

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -683,6 +683,11 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 
 		/** @var \Tribe__Events__Aggregator__API__Import $import_api */
 		$import_api  = $aggregator->api( 'import' );
+
+		if ( empty( $this->meta['import_id'] ) ) {
+			return tribe_error( 'core:aggregator:record-not-finalized' );
+		}
+
 		$import_data = $import_api->get( $this->meta['import_id'], $data );
 
 		$import_data = $this->maybe_cast_to_error( $import_data );
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -1004,16 +1004,14 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 		if ( empty( $status ) && $lookup_children ) {
 			$last_children_query = $this->query_child_records( array( 'posts_per_page' => 1, 'order' => 'DESC', 'order_by' => 'modified' ) );
 			if ( $last_children_query->have_posts() ) {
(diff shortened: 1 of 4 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -978,18 +992,37 @@ public function get_source_info() {
...
 
+		if ( empty( $status ) && $lookup_children ) {
+			$last_children_query = $this->query_child_records( array( 'posts_per_page' => 1, 'order' => 'DESC', 'order_by' => 'modified' ) );
+			if ( $last_children_query->have_posts() ) {
+				$last_children        = reset( $last_children_query->posts );
+				$last_children_status = str_replace( 'tribe-ea-', '', $last_children->post_status );
+
+				$map                  = array(
+					'failed'  => 'error:import-failed',
+					'success' => 'success:queued',
```
bordoni: Why double Success?

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -978,18 +992,37 @@ public function get_source_info() {
...
 	 */
-	public function get_last_import_status( $type = 'error' ) {
+	public function get_last_import_status( $type = 'error', $lookup_children = false ) {
 		$status = empty( $this->meta['last_import_status'] ) ? null : $this->meta['last_import_status'];
 
+		if ( empty( $status ) && $lookup_children ) {
+			$last_children_query = $this->query_child_records( array( 'posts_per_page' => 1, 'order' => 'DESC', 'order_by' => 'modified' ) );
+			if ( $last_children_query->have_posts() ) {
+				$last_children        = reset( $last_children_query->posts );
+				$last_children_status = str_replace( 'tribe-ea-', '', $last_children->post_status );
```
bordoni: There is no need to do this if we pass `tribe-ea-failed` and `tribe-ea-success` as the mapping strings.

lucatume: right!

src/Tribe/Aggregator/Record/Abstract.php
lucatume: I opt for "too much paste love"

lucatume:
I've undid my stupid things; good for another look

PR #1523, opened 2017-08-18
Altered tribe_events_get_the_excerpt() to set the global $post prior …

elimn:
Ticket: https://central.tri.be/issues/70473

Fixed issue with `tribe_events_get_the_excerpt()` returning a read more link that pointed to the current page when used inside of the inline event shortcode.

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -1444,7 +1444,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 
 		if ( ! has_excerpt( $post->ID ) ) {
 			// Temporarily alter the global post in preparation for our filters.
-			$global_post = $GLOBALS['post'];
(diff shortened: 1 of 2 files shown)


src/functions/template-tags/general.php
```diff
@@ -1443,6 +1443,10 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
 		$excerpt = wp_kses( $excerpt, $allowed_html );
 
 		if ( ! has_excerpt( $post->ID ) ) {
+			// Temporarily alter the global post in preparation for our filters.
+			$global_post = $GLOBALS['post'];
+			$GLOBALS['post'] = $post;
+
```
ggwicz: vertically align the `=`

elimn:
Thanks for the 👀

PR #1525, opened 2017-08-18
If the source URL for a new import lacks a scheme, add one | #76466

barryhughes:
Prepend `http://` to the source URL for new imports if the scheme/protocol is missing.

* Adds a filter to allow users to unwind this or otherwise make further adjustments
* I've preferred HTTP over HTTPS here from the POV that the former probably still has wider support (and most sites with HTTPS will automatically redirect HTTP requests anyway)

No community credit needed in the changelog entry.

:ticket: [#76466](https://central.tri.be/issues/76466)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | empty() and isset() must only contain variables (PHP 5.2 compatibility) | [Abstract.php:176](https://github.com/moderntribe/the-events-calendar/blob/4e7e9805193f2b8d19084dc0e23c208c1a1866b8/src/Tribe/Aggregator/Record/Abstract.php#L176)

PR #1527, opened 2017-08-19
Broken translation

cesare-montresor:
In the code the are mixedup functions that uses `__()` and `_x()` functions to translate, the second one don't get translated at all, replaced `esc_html_x()` with `esc_html__()` for the titles in the month and week view.
Feel free to ask for more details.


barryhughes:
Thanks for the pull request, @cesare-montresor!

I'm not quite following the reasoning for the change or why we would want to remove the context. Can you explain a little more?

zbtirrell:
The context seems essential here in distinguishing those two strings as different. 

I'm closing this, but feel free to reopen with an explanation if I am missing something.

PR #1529, opened 2017-08-22
remove merge aberration

lucatume:
Ticket: https://central.tri.be/issues/85636

Removes what I guess is a merge aberration.

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -169,7 +169,6 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 		}
 
 		// `source` will be empty when importing .ics files
-		$original_source = $this->meta['source'];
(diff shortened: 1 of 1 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -169,7 +169,6 @@ public function setup_meta( $meta ) {
 		}
 
 		// `source` will be empty when importing .ics files
-		$original_source = $this->meta['source'];
```
barryhughes: We need (/want) this variable for the `tribe_aggregator_meta_source` hook, below.

Is the problem that source may not be set? If so, can we move this down rather than removing it completely?

lucatume: correct; simply moving the line seems like a better approach

borkweb:
Merging on behalf of @lucatume!

PR #1530, opened 2017-08-23
fix EA client issues

lucatume:
Ticket: https://central.tri.be/issues/79975 

This PR is the client side of https://github.com/moderntribe/event-aggregator-site/pull/242.

This PR fixes some small glitches on EA Client post insertions system that would amount to failures when handling huge imports.  
The critical path times during queue processing, when happening on a cron on ajax request, is this:

1. fetch the queue data (might involve the Meta Chunker)
2. process its items
3. for each item insert or update a post
4. update queue activity

On step 3 the code looks up the database to see if we are inserting a new event or updating one looking for a unique identifier that is set **at the end of step 4** (and rightly so): step 4 from previous update/insertion or previous cron run (or ajax) might still not be complete due to DB times and step 3s, and 4s, of following operations would snowball into failure and inconsistency especially when processing large imports on AJAX.
This issue is augmented by the possible use of the Meta Chunker and its added DB times due to the number of events to import/update.
What this PR does:

1. add a lock acquisition and release system to the queue to avoid two concurrent processes (e.g. two AJAX handling, one AJAX handling and a cron run) to work on the same data; this would create the duplicates
2. add sanity checks throughout the queue processing system to avoid caching related issues from happening; e.g. some events would be deemed as "updated" when "created"; the reason being that WP and our cache would mark an event as existing before the database **actually** wrote it
3. assign a default user to events that are added when running via crons: the user would be set to `0` when cron runs and events would be created with no author; the fallback is to user the current user ID if available or that of the first user that can edit events (the admin probably); this could be refined assigning event to the user that scheduled the queue or to the one that is running the import manually but that seemed like a good first implementation; this would lead to some events created from imports not having an author
4. make sure some concurrency inconsistencies in the activity report are corrected: due to different states in different processes the activity log could show an event as both "created" and "updated"; the PR makes a sanity check and corrects that
5. fix a queue realtime issue where the AJAX based processing might stop and declare the import finished to resume when hitting the History tab; this was caused by concurrent access


lucatume:
putting this on hold as I've found an issue related to Venues

lucatume:
Duplicate venues and organizers issue fixed, good to review.

src/Tribe/Aggregator/Record/Queue.php
```diff
@@ -206,6 +215,7 @@ public function save() {
 		$chunker->register_chunking_for( $this->record->post->ID, $key );
 
 		if ( empty( $this->items ) ) {
+			do_action( 'tribe_aggregator_queue_finished', $this );
```
bordoni: Add Docblock

lucatume: I've removed this action as we do not use; will re-add it if we really need it in the future; it's a left-over from my testing.

src/Tribe/Aggregator/Event.php
```diff
@@ -188,15 +188,22 @@ public function get_existing_ids( $origin, $values ) {
 		}
 
 		$key = "_{$fields[ $origin ]['target']}";
+		$post_type = Tribe__Events__Main::POSTTYPE;
+		$interval = "('" . implode( "','", $values ) . "')";
```
borkweb: Should we `array_map( 'absint', $values )` for safety?

src/Tribe/Aggregator/Record/Activity.php
```diff
@@ -273,4 +275,18 @@ public function exists( $slug ) {
...
 	}
+
+	/**
+	 * Checks the activities for a slug to make sure there is no incoherency due to concurring processes.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $slug
+	 */
+	protected function sanity_check( $slug ) {
```
borkweb: This method name is pretty generic. Perhaps something obnoxiously descriptive like: `prevent_duplicates_between_item_actions()`

src/Tribe/Aggregator/Event.php
lucatume: better safe than sorry

src/Tribe/Aggregator/Record/Activity.php
lucatume: good call, what about `kill_unicorns`? We are in fact checking for a quite remote possibility.

borkweb: Other possibilities:
* `prevent_death_and_destruction()`
* `yo_dawg_i_heard_you_like_items_so_i_put_some_items_in_your_items_but_then_i_changed_my_mind()`
* `y_u_no_work_right()`

lucatume:
code updated, good for another pass

src/Tribe/Aggregator/Record/Activity.php
lucatume: I like the first one. Next time.

PR #1535, opened 2017-08-24
i18n: Avoid using HTML tags in translation strings

ramiy:
Old string:

* `<a href="%s">Edit the %s slug</a>`

New string:

* `Edit the %s slug`


barryhughes:
:ticket: [#88982](https://central.tri.be/issues/88982)

PR #1537, opened 2017-08-25
i18n: split translation strings

ramiy:
The following translation string:

`The settings below control the display of your calendar. If things don't look right, try switching between the three style sheet options or pick a page template from your theme.</p><p>There are going to be situations where no out-of-the-box template is 100&#37; perfect. Check out our <a href="%s">our themer's guide</a> for instructions on custom modifications.`

has `</p><p>` HTML tags is the string. It should be splited into two strings.

### DIFF
--- a/src/admin-views/tribe-options-display.php
+++ b/src/admin-views/tribe-options-display.php
@@ -45,8 +45,9 @@ $display_tab_fields = Tribe__Main::array_insert_before_key(
 		),
 		'info-box-description'               => array(
 			'type' => 'html',
-			'html' => '<p>' . sprintf(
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | String concat is not required here; use a single string instead | [tribe-options-display.php:48](https://github.com/moderntribe/the-events-calendar/blob/7c3cb2eda2abc52017fe8527ac28c11eabf6c1b7/src/admin-views/tribe-options-display.php#L48)

barryhughes:
:ticket: [#88982](https://central.tri.be/issues/88982)

PR #1540, opened 2017-08-25
Fix typo of "organizer" - should be "venue"

cliffordp:
Also capitalization and grammar


bordoni:
Would be nice to add a changelog

PR #1550, opened 2017-09-01
#84202 Undo Datepicker Format Code Change from M17.16

ggwicz:
_Ref:_ [C#84202](http://central.tri.be/issues/84202)

This commit may seem weird—but it is indeed all that’s required to undo
the support for “MM.DD.YYYY”-style date formats (i.e., formats
containing a dot instead of the usual dash or slash) in the Month View
Tribe Bar.

It thus also fixes the bug in #86937.


src/resources/js/tribe-events.js
```diff
@@ -1343,7 +1343,7 @@ Date.prototype.format = function( mask, utc ) {
 			ts.view = $tribe_events_header.data( 'view' );
 		}
 
-		if ( $tribe_events.tribe_has_attr( 'data-datepicker_format' ) ) {
+		if ( $tribe_events.tribe_has_attr( 'data-datepicker_format' ) && $tribe_events.attr( 'data-datepicker_format' ).length === 1 ) {
```
borkweb: I'd prefer to see this in Yoda format, but the change isn't required.

PR #1560, opened 2017-09-08
Add in randomization when scheduling and initiating scheduled imports

borkweb:
* When scheduling the next cron event, allow for +/- 5 minute adjustments
* When checking if a scheduled import is ready to run, allow for -5 to 0 minutes from the expected scheduled time

See: https://central.tri.be/issues/86628


barryhughes:
This is lovely: should it have a changelog entry though?

borkweb:
Added!

PR #1567, opened 2017-09-12
Fix/71994 patch jquery resize

niconerd:
https://central.tri.be/issues/84807

Not sure about the use of TBD inside the js. Will that be replaced by the packager or should I hardcode the version number?


borkweb:
I'd go ahead and hardcode the version number there. We don't analyze the JS files in our `mt tbd` script (though, we should probably add that at some point)

vendor/jquery-resize/jquery.ba-resize.js
```diff
@@ -191,6 +191,17 @@
...
           data = $.data( this, str_data );
+
+        /**
+         * The Events Calendar Patch
+         *
+         * @since TBD
+         * @see https://github.com/cowboy/jquery-resize/issues/1
+         * @see https://central.tri.be/issues/71994
+         */
+        if (!data) {
```
borkweb: Change spacing on this line to: `if ( ! data ) {`

vendor/jquery-resize/jquery.ba-resize.js
```diff
@@ -191,6 +191,17 @@
...
+          data = $.data( this, str_data, {})
```
borkweb: Add a space after the `}` and add a semicolon at the end of the line

niconerd:
thanks for the review @borkweb and for the help on this one!

PR #1576, opened 2017-09-15
Call `Tribe__Tickets__Tickets::modules()` directly instead of using `…

elimn:
Ticket: https://central.tri.be/issues/66421

I did not feel like this merited a changelog entry, it's basically a spotfix and is highly backwards compatible. I did put a changelog entry in the [related ET PR.](https://github.com/moderntribe/event-tickets/pull/476)

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -733,7 +733,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	 * @return bool
 	 */
 	function tribe_events_admin_show_cost_field() {
-		$modules      = apply_filters( 'tribe_events_tickets_modules', null );
+		$modules      = Tribe__Tickets__Tickets::modules();
 		$event_origin = get_post_meta( get_the_ID(), '_EventOrigin', true );
 		$show_cost    = empty( $modules ) ||
 						class_exists( 'Tribe__Events__Tickets__Eventbrite__Main' ) ||
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -733,7 +733,12 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	 * @return bool
 	 */
 	function tribe_events_admin_show_cost_field() {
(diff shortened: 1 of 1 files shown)


src/functions/template-tags/general.php
```diff
@@ -733,7 +733,7 @@ function tribe_events_get_days_of_week( $format = null ) {
 	 * @return bool
 	 */
 	function tribe_events_admin_show_cost_field() {
-		$modules      = apply_filters( 'tribe_events_tickets_modules', null );
+		$modules      = Tribe__Tickets__Tickets::modules();
```
barryhughes: What about deprecating this hook (rather than outright removal) and ensuring it still works, perhaps by wrapping within the `tribe_tickets_get_modules` hook?

elimn: That gets complicated. The only thing attached to `tribe_events_tickets_modules` is `Tribe__Tickets__Tickets::modules()`. If we ran that filter inside of the attached method, we would face recursion. In order to be fully backwards compatible though the method does need to stay attached to that filter until a non MR.

To account for this I used Tribe__Deprecation within the ET pull request. At this point, if anyone does try to hook into `tribe_events_tickets_modules` they will get a notice telling them its deprecated. I also have a PR for Common now to help facilitate this. https://github.com/moderntribe/tribe-common/pull/521

borkweb: The problem here is that `Tribe__Tickets__Tickets` may not exist when this function executes. It'd only exist if `event-tickets` is active.

borkweb: Removing the filter would probably be pretty low impact, but since the filter is being removed, it warrants a changelog entry and changing the code to the following:

```php
$modules = null;

if ( class_exists( 'Tribe__Tickets__Tickets' ) ) {
  $modules = Tribe__Tickets__Tickets::modules();
}
```

bordoni: I like that.

elimn: Nice catch. Fixed.

PR #1586, opened 2017-09-19
Make sure we are check if the Service is active to hide the Errors

bordoni:
_Ref:_ [C#78890](http://central.tri.be/issues/78890)
---


barryhughes:
:jack_o_lantern:

PR #1591, opened 2017-09-20
Set type on address JSON-LD to PostalAddress

sc0ttkclark:
https://central.tri.be/issues/85223

### DIFF
--- a/src/Tribe/JSON_LD/Venue.php
+++ b/src/Tribe/JSON_LD/Venue.php
@@ -55,6 +55,7 @@ class Tribe__Events__JSON_LD__Venue extends Tribe__JSON_LD__Abstract {
 
 		$data->address = array();
 
+		$data->address['@type'] = 'PostalAddress';
 		$data->address['streetAddress'] = tribe_get_address( $post_id );
 		$data->address['addressLocality'] = tribe_get_city( $post_id );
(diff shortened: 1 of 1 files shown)


src/Tribe/JSON_LD/Venue.php
```diff
@@ -55,6 +55,7 @@ public function get_data( $post = null, $args = array( 'context' => false ) ) {
 
 		$data->address = array();
 
+		$data->address['@type'] = 'PostalAddress';
```
bordoni: Why are we making this change here, instead of the prop `type`, which will be used to set the `@type`.
Also why `PostalAddress` instead of `Place`?

sc0ttkclark: Address isn't a full fledged JSON_LD object in TEC, so I just used the same way we set type in the other areas:

```
			$data->geo = (object) array(
				'@type' => 'GeoCoordinates',
				'latitude' => $geo['lat'],
				'longitude' => $geo['lng'],
			);
```

and 

```
				// Manually Include the Price for non Event Tickets
				$data->offers = (object) array(
					'@type' => 'Offer',
					'price' => $price,

					// Use the same url as the event
					'url' => $data->url,
				);
```

I've chosen PostalAddress because it's what's used in the example for Events as documented here:

https://developers.google.com/search/docs/data-types/events (click "See Markup")

PR #1596, opened 2017-09-20
#67482 Rename and Deprecate a Filter

ggwicz:
**Ticket:** [**#67482**](http://central.tri.be/issues/67482)

I _think_ I’m doing this right.

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -4355,7 +4355,21 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 
 			global $wp_query;
 
+			/**
+			 * Allows for customizing the "date search" field value.
+			 *
+			 * @deprecated TBD Use tribe_events_bar_date_search_default_value instead.
+			 *
+			 * @param string $value The "date search" field value, which defaults to an empty string.
+			 */
 			$value = apply_filters( 'tribe-events-bar-date-search-default-value', '' );
+
+			/**
+			 * Allows for customizing the "date search" field value.
+			 *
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -4355,8 +4355,22 @@ public function setup_date_search_in_bar( $filters ) {
...
+			 * @param string $value The "date search" field value, which defaults to an empty string.
+			 */
 			$value = apply_filters( 'tribe-events-bar-date-search-default-value', '' );
 
+			/**
+			 * Allows for customizing the "date search" field value.
+			 *
+			 * @param string $value The "date search" field value, which defaults to an empty string.
+			 */
+			$value = apply_filters( 'tribe_events_bar_date_search_default_value', '' );
```
barryhughes: If a 3rd party is using the former and now deprecated hook, we should still try to support them and avoid breakages. To that end, let's pass `$value` here instead of overwriting it by passing empty string `''`

barryhughes:
Merits a changelog entry?

ggwicz:
Both things you flagged are obviously necessary, @barryhughes — not sure how I totally blanked on them and missed those details. Apologies — will fix promptly and re-submit for review.

ggwicz:
[Unless there's some step I'm forgetting](https://tribe.slack.com/archives/C0258NY1C/p1505952603000149) here in the deprecation process, this one is ready for review again.

barryhughes:
Sublime :cactus: :sun_behind_small_cloud:

PR #1602, opened 2017-09-25
Fix issue with not running scheduled imports and missing prunes

lucatume:
Ticket: https://central.tri.be/issues/88395

This PR:
* fixes the issue where scheduled imports might not run if stuck in pending before the release of M17.18
* expired pending children records hanging around indefinitely


lucatume:
Targeting F17.7 as the base but could very well change to master should this not make the cut.

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -871,17 +871,17 @@ public function query_child_records( $args = array() ) {
...
+	public function get_child_record_by_status( $status = 'success', $qty = -1, array $args = array() ) {
 		$statuses = Tribe__Events__Aggregator__Records::$status;
 
 		if ( ! isset( $statuses->{ $status } ) && 'trash' !== $status ) {
 			return false;
 		}
 
-		$args = array(
-			'post_status'    => $statuses->{ $status },
+		$args = array_merge( $args, array(
```
Camwyn: Did we want to override any potentially passed values here or should we use `wp_parse_args`? These two values are passed individually so I'm just commenting as a double-check.

lucatume: we want to override so this is ok.

lucatume:
Rebase this on `master` after F17.7 is out of the door

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M17.19*, but this was pull requested against *release/F17.7*.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space after the first parenthesis | [general.php:1453](https://github.com/moderntribe/the-events-calendar/blob/9bde6b334f9a35aefddd16a8af1a632af567919e/src/functions/template-tags/general.php#L1453)
:x: | There must be a single space before the last parenthesis | [general.php:1453](https://github.com/moderntribe/the-events-calendar/blob/9bde6b334f9a35aefddd16a8af1a632af567919e/src/functions/template-tags/general.php#L1453)

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1035,7 +1035,7 @@ public function has_passed_retention_time() {
 		// Prevents Pending that is younger than 1 hour to be pruned
 		if (
 			Tribe__Events__Aggregator__Records::$status->pending === $this->post->post_status &&
-			$current > $created + HOUR_IN_SECONDS
+			$current < $created + HOUR_IN_SECONDS
```
barryhughes: 100% optional - it's not something you introduced in this change - but ideally the `&&` would be at the start of this line, not the end of the last line [(per standards)](http://moderntribe.github.io/products-engineering/code-standards/php/#multi-line-control-structures)

src/functions/template-tags/general.php
```diff
@@ -1450,7 +1450,7 @@ function tribe_events_get_the_excerpt( $post = null, $allowed_html = null ) {
 
 		if ( ! has_excerpt( $post->ID ) ) {
 			// Temporarily alter the global post in preparation for our filters.
-			$global_post     = $GLOBALS['post'];
+			$global_post = isset( $GLOBALS['post'] ) ? $GLOBALS['post'] : '';
```
barryhughes: I'll defer to your judgement here, but if `$GLOBALS['post']` isn't set, might 'restoring it' to null be safer and more reflective of what it was (than setting it to an empty string)?

Mostly, I'm sure it won't matter, but if there _is_ code out there that might run immediately after this and does it's own `isset( $GLOBALS['post'] )` then we'll be triggering a different path of execution in that code.

PR #1606, opened 2017-09-27
#46471 Fix Borked Tribe Bar Datepicker in Child Themes

ggwicz:
**Ticket:** [**#46471**](http://central.tri.be/issues/46471)

---
**Code Review Note:** There is a lot of JS in this PR, because when trying to figure this one out, I first looked at the CSS and failed to get something working. So I started looking through the JS to see if there was a culprit there, tweaking some formatting along the way.

I had missed something obvious earlier and ended up making a CSS change to resolve this after all — a very _simple_ one, to boot — but figured we should keep my JS changes anyways.


elimn:
There are a number of areas where you swapped out:
`$el = $( '#el' );'
for
' $el = $( document.getElementById( 'el' ) );`

Would you mind explaining your reasoning there? I don't think this is going to be noticeably more performany, but it is IMO a lot less legible. Were you thinking it was a half step between now and when we start weeding out unnecessary jQuery from our code?

ggwicz:
Hey @elimn 

First, the performance gains are _not_ totally trivial — [see here](https://jsperf.com/hide-with-and-without-jquery) for a commonly-cited example of the gains.

Also, re this point:
> Were you thinking it was a half step between now and when we start weeding out unnecessary jQuery from our code?

Yes — we get some non-meaningless performance gains super easily, while pushing the code more towards vanilla JS. I also personally don't think the legibility hit is that noticeable, especially given some of the other code-cleanup to make the formatting more consistent and breathable.

ggwicz:
@elimn We can certainly continue this conversation another time — I have thoughts on it, but for the sake of time, would you mind re-approving this since I just committed some new code to fix some conflicts (with readme.txt, just changelog entry info)?

Maybe we can talk about this on scrum if time allows and the desire exists.

PR #1612, opened 2017-09-29
Fix for allday multiday events

elimn:
Ticket: https://central.tri.be/issues/87775

I was not able to find anywhere in the spec where it stated that a multiday allday event should have DTEND set to the day after it ends. However, for single-day all day events it does specify this (although for those the DTEND field is also optional). Given that the spec it specifies this for single-day events it follows that you would need to do this for multiday ones as well. Furthermore, every calendar export I tested with, like Google Calendar, indeed adds a day to the DTEND of their all day multiday exports. They also properly import when we do this with our exports.


Camwyn:
This looks good, I'd note that `$type` is only used for all-day events, so you could just test for `$all-day` instead of setting it. Particularly not setting it for non-all-day events (since then it's just wasting space)

elimn:
Thank you for the feedback!

I noted that too. But given this extremely long spaghetti method I was hesitant to start using one variable as an indicator for two different things. That's also why I did not modify `$tzoned->end` and duplicated it instead, even though technically no other current code uses it after I would have modified it. I figured it was better to be safe and abundantly clear, especially considering the ical spec does allot DATE types in other circumstances ( we just don't currently support those ).

PR #1613, opened 2017-09-29
Update events-admin.js for Bug #45612

jentheo:
When a view is disabled, reset default view to an enabled view

### DIFF
--- a/src/resources/js/events-admin.js
+++ b/src/resources/js/events-admin.js
@@ -1,7 +1,7 @@
 var tribe_timepickers = tribe_timepickers || {};
 
 ( function ( $, obj ) {
-	'use strict';
+	"use strict";
 
 	obj.selector = {
 		container  : '.tribe-datetime-block',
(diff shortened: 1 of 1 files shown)


src/resources/js/events-admin.js
```diff
@@ -25,14 +25,14 @@ var tribe_timepickers = tribe_timepickers || {};
 	obj.$ = {};
 
 	obj.container = function( k, container ) {
-		var $container  = $( container );
-		var $all_day    = $container.find( obj.selector.all_day );
-		var $timepicker = $container.find( obj.selector.timepicker );
-		var $timezone   = $container.find( obj.selector.timezone ).not( obj.selector.input );
-		var $input      = $container.find( obj.selector.timezone ).filter( obj.selector.input );
+		var $container = $( container ),
+			$all_day = $container.find( obj.selector.all_day ),
```
barryhughes: This isn't 'wrong', but see our notes on [variable declarations in JS](http://moderntribe.github.io/products-engineering/code-standards/javascript/#variable-declaration) - the previous code where each variable is declared in its own statement is usually what we aim for.

src/resources/js/events-admin.js
```diff
@@ -25,14 +25,14 @@ var tribe_timepickers = tribe_timepickers || {};
...
+			$timepicker = $container.find( obj.selector.timepicker ),
+			$timezone = $container.find( obj.selector.timezone ).not( obj.selector.input ),
+			$input = $container.find( obj.selector.timezone ).filter( obj.selector.input ),
 
-		// Create the Link
-		var $timezone_link = $( obj.timezone.link( { label: $input.data( 'timezoneLabel' ), timezone: $input.data( 'timezoneValue' ) } ) );
+			// Create the Link
+			$timezone_link = $( obj.timezone.link( { label: $input.data( 'timezoneLabel' ), timezone: $input.data( 'timezoneValue' ) } ) );
```
barryhughes: As above, would be better as a standalone definition, ie: `var $timezone_link = $( whatever );`

src/resources/js/events-admin.js
```diff
@@ -69,8 +69,8 @@ var tribe_timepickers = tribe_timepickers || {};
 	obj.setup_timepickers = function( $timepickers ) {
 		// Setup all Timepickers
 		$timepickers.each( function() {
-			var $item = $( this );
-			var opts  = $.extend( {}, obj.timepicker.opts );
+			var $item = $( this ),
+				opts = $.extend( {}, obj.timepicker.opts );
```
barryhughes: Ditto re var declarations.

src/resources/js/events-admin.js
```diff
@@ -258,10 +254,10 @@ jQuery( document ).ready( function( $ ) {
 	 */
 	var $date_format      = $( '[data-datepicker_format]' ),
 		$view_select      = $( '.tribe-field-dropdown_select2 select' ),
-		viewCalLinkHTML   = $( document.getElementById( 'view-calendar-link-div' ) ).html(),
+		viewCalLinkHTML   = $( '#view-calendar-link-div' ).html(),
```
barryhughes: Personally, I like this style because it's far more readable, however it's arguably less performant and other reviews may ask you to change it back

barryhughes:
There are quite a few changes here outside of `set_selected_views( $this )` (which is where I think the 'real' changes as they relate to [#45612](https://central.tri.be/issues/45612) live) ... please do be wary of this as a) it increases the amount of code that needs to be reviewed b) some of those changes are out of keeping with our coding standards for JS

PR #1620, opened 2017-10-04
Use the latest "Requires PHP" header field

ramiy:
* Remove the "requirements" list.
* Required WordPress version is defined in "Requires at least" header field.
* Add the new "Requires PHP" header field.

See more info at: https://generatewp.com/minimum-required-php-version-plugins/


barryhughes:
:ticket: [#90285](https://central.tri.be/issues/90285)

PR #1621, opened 2017-10-05
Fixed issue with events that started before but ended within not bein…

elimn:
Ticket: https://central.tri.be/issues/89747

Not doing a readme entry just yet, we might want to make this a hotfix so I wanted to the readme entry commit to be separate for easy cherry picking.

Near as I can figure we only have three different scenarios which this query needs to account for.

- Events which start during the month: `tribe_event_start.meta_value >= '{$start_date_sql}' AND tribe_event_start.meta_value <= '{$end_date_sql}'`
- Events which end during the month: `tribe_event_end_date.meta_value >= '{$start_date_sql}' AND tribe_event_end_date.meta_value <= '{$end_date_sql}'`
- Events which start prior to the month and end after: `tribe_event_start.meta_value < '{$start_date_sql}' AND tribe_event_end_date.meta_value > '{$end_date_sql}'`

This updated query should handle all of those. In my tests it does. Let me know if you can think of anything I am missing.

The bug from the attached ticket was introduced with this commit: https://github.com/moderntribe/the-events-calendar/commit/5ca21bcf705cfcc2b42b3e714da53750cb1fa7d2 Oddly, I can not find a single functional difference between that commit and the one prior to it. But, bisect doesn't lie. Month view was working fine until this commit. :-/

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -932,18 +932,18 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 							WHERE tribe_event_start.meta_key = '_EventStartDate'
 							AND tribe_event_start.post_id IN ( {$sql_ids} )
 							AND (
-									(
-										tribe_event_start.meta_value >= '{$start_date_sql}'
-										AND tribe_event_start.meta_value <= '{$end_date_sql}'
-									)
-									OR (
-										tribe_event_end_date.meta_value >= '{$start_date_sql}'
-										AND tribe_event_end_date.meta_value <= '{$end_date_sql}'
-									)
-									OR (
-										tribe_event_start.meta_value < '{$start_date_sql}'
-										AND tribe_event_end_date.meta_value > '{$end_date_sql}'
-									)
+								(
+									tribe_event_start.meta_value >= '{$start_date_sql}'
+									AND tribe_event_start.meta_value <= '{$end_date_sql}'
+								)
+								OR (
+									tribe_event_end_date.meta_value >= '{$start_date_sql}'
+									AND tribe_event_end_date.meta_value <= '{$end_date_sql}'
+								)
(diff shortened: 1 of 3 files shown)


src/Tribe/Query.php
```diff
@@ -932,18 +932,18 @@ public static function getEventCounts( $args = array() ) {
...
-								)
-								OR (
-									tribe_event_start.meta_value <= '{$start_date_sql}'
-									AND tribe_event_end_date.meta_value >= '{$start_date_sql}'
-								)
-								OR (
-									tribe_event_start.meta_value >= '{$start_date_sql}'
-									AND tribe_event_start.meta_value <= '{$end_date_sql}'
-								)
+									(
```
bordoni: Avoid the extra Tab, it's not needed.

src/Tribe/Template/Month.php
```diff
@@ -560,18 +560,18 @@ protected function set_events_in_month() {
...
-						)
-						OR (
-							tribe_event_start.meta_value <= '{$start_date_sql}'
-							AND tribe_event_end_date.meta_value >= '{$end_date_sql}'
-						)
-						OR (
-							tribe_event_start.meta_value >= '{$start_date_sql}'
-							AND tribe_event_start.meta_value <= '{$end_date_sql}'
-						)
+							(
```
bordoni: Avoid the extra Tab, it's not needed.

elimn:
@bordoni Nice catch. Since the opening bracket has no statement before it, it lines up perfectly with the one above it. Which tricked me into thinking it needed to be indented.

This is the first time I have ever found a disadvantage to egyptian brace style. Oh well, it's still my favorite.

elimn:
This is ready for review, now that we know what branch to merge it into. I added a readme entry.

PR #1623, opened 2017-10-09
Send an X-EA-Site header with each request

lucatume:
Ticket: https://central.tri.be/issues/90489

Client-side work for PR: https://github.com/moderntribe/event-aggregator-site/pull/257 (ticket https://central.tri.be/issues/86606)

This PR modifies EA `Requests` class to make it so that each `GET` and `POST` request made by EA client will include a custom header the service can consume.


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *1.0.10*, but this was pull requested against *release/M17.20*.

lucatume:
I thought about adding a changelog entry but did not know how and if this change might be useful to clients; guess "better safe than sorry" applies here though.

PR #1626, opened 2017-10-11
#89918 Restore Functionality of Currency Position Options

ggwicz:
**Ticket:** [**#89918**](http://central.tri.be/issues/89918)

**Related PR:** https://github.com/moderntribe/tribe-common/pull/547


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [Cost_Utils.php:110](https://github.com/moderntribe/the-events-calendar/blob/4864909d71b5f280fad5946f3515e69a6e5dec77/src/Tribe/Cost_Utils.php#L110)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [Cost_Utils.php:110](https://github.com/moderntribe/the-events-calendar/blob/39648b407bff0602943051b875b5046a630c9d0a/src/Tribe/Cost_Utils.php#L110)

src/Tribe/Cost_Utils.php
```diff
@@ -101,9 +100,14 @@ public function get_formatted_event_cost( $event, $with_currency_symbol = false
...
 
 			if ( $with_currency_symbol ) {
-				$event_id = Tribe__Main::post_id_helper( $event );
-				$currency_symbol = get_post_meta( $event_id, '_EventCurrencySymbol', true );
+				$event_id          = Tribe__Main::post_id_helper( $event );
+				$currency_symbol   = get_post_meta( $event_id, '_EventCurrencySymbol', true );
 				$currency_position = get_post_meta( $event_id, '_EventCurrencyPosition', true );
+
+				if ( empty( $currency_position ) ) {
+					$currency_position = tribe_get_option( 'reverseCurrencyPosition', false );
```
barryhughes: This interests me: won't this typically set `$currency_position` to either **true** or **false** and looking at the [related PR you submitted](https://github.com/moderntribe/tribe-common/pull/547/files) doesn't `Tribe__Cost_Utils::maybe_format_with_currency()` test explicitly if it is the string `'prefix'`?

ggwicz: @barryhughes It _does_ test for that, which is a value that comes from the post meta on posts only when tickets plugins are not active, and the "Event Cost" field is used. (Instead of costs on individual tickets.)

In my tinkering, I boiled down the issue to that `tribe_format_currency()` template tag itself. There's a line in it like this:
```
$cost = $reverse_position ? $cost . $currency_symbol : $currency_symbol . $cost;
```
In a lot of cases I tried, `$reverse_position` would always be true at this point. I tried a bunch of different solutions before arriving at the changes in these two PRs, none of which worked — is this answer useful at all? 🤔

ggwicz: My thinking is that setting `$reverse_position` to `true` or `false` was the problem, and it should remain `null` so that `tribe_format_currency()` handles it like it's designed to.

https://github.com/moderntribe/tribe-common/blob/8497598d625e515200c316bcf1c07d84ebde51e1/src/Tribe/Cost_Utils.php#L132

ggwicz: @barryhughes Older versions of Tribe Common checked for `false` value of `$reverse_position` explicitly, somewhere along the way it switched to `null` (and from `or` to `and`):

* Old check in `tribe_format_currency()` → https://cldup.com/oJNe4kNQww.png
* New/current check in `tribe_format_currency()` → https://cldup.com/bCqF0-zc62-3000x3000.png

Any thoughts on this?

barryhughes: Not much more to add except we do know that the version of Common that will be in use is either the same one that will ship with TEC, or will be a more recent version. That to say, in this case backwards compatibility with older versions of Common isn't a concern.

ggwicz: Okay — do you think I should return to the drawing board and see if I can get the same result with a simpler change? Would be happy to try.

barryhughes: I think we can do some cleanup to this whole system right through to the `tribe_format_currency()` function - but it doesn't need to happen as part of this fix. If you're happy and confident let's move ahead and get it to QA.

ggwicz: Let's do it!

PR #1631, opened 2017-10-13
#82286 Disable Month View Caching [for now] when WPML is Active

ggwicz:
**Ticket:** [**#82286**](http://central.tri.be/issues/82286)


barryhughes:
:100:

PR #1637, opened 2017-10-20
Sometimes add linked post edit link for users with limited caps

elimn:
Ticket: https://central.tri.be/issues/89015

While working on the above ticket I accidentally fixed https://central.tri.be/issues/90107 and https://central.tri.be/issues/89995 . Basically, the edit venue/org links are not appearing when appropriate in Community. 
- When the page loads, the trigger event would not properly fire thus the links would not appear.
- We need a per post indication of whether or not a user can edit it. Further, it is highly advantageous and simpler to use WP' get_edit_post_link() than rebuilding our own half hearted version that won't always work. This implements that.


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M17.22*, but this was pull requested against *release/M17.21*.

PR #1641, opened 2017-10-25
Fix for recurring event label line break

mia-caro:
Fixes bug where .recurringinfo label would break to following line on single event pages


barryhughes:
PostCSS compiled and changelog added - merging.

PR #1645, opened 2017-10-26
#88589 More Nuanced Linked Posts Label i18n

ggwicz:
**Ticket:** [**#88589**](http://central.tri.be/issues/88589)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [Linked_Posts.php:972](https://github.com/moderntribe/the-events-calendar/blob/62bf9624bb2bdda9c4b0534adddc62d0209192eb/src/Tribe/Linked_Posts.php#L972)
:x: | Whitespace found at end of line | [Linked_Posts.php:979](https://github.com/moderntribe/the-events-calendar/blob/62bf9624bb2bdda9c4b0534adddc62d0209192eb/src/Tribe/Linked_Posts.php#L979)
:x: | Expected 1 space before &quot;!==&quot;; 2 found | [labels.php:16](https://github.com/moderntribe/the-events-calendar/blob/62bf9624bb2bdda9c4b0534adddc62d0209192eb/src/functions/utils/labels.php#L16)
:x: | Expected 1 space before &quot;!==&quot;; 2 found | [labels.php:17](https://github.com/moderntribe/the-events-calendar/blob/62bf9624bb2bdda9c4b0534adddc62d0209192eb/src/functions/utils/labels.php#L17)
:x: | Whitespace found at end of line | [labels.php:29](https://github.com/moderntribe/the-events-calendar/blob/62bf9624bb2bdda9c4b0534adddc62d0209192eb/src/functions/utils/labels.php#L29)
:x: | Expected 1 space before &quot;!==&quot;; 2 found | [labels.php:37](https://github.com/moderntribe/the-events-calendar/blob/62bf9624bb2bdda9c4b0534adddc62d0209192eb/src/functions/utils/labels.php#L37)
:x: | Expected 1 space before &quot;!==&quot;; 2 found | [labels.php:56](https://github.com/moderntribe/the-events-calendar/blob/62bf9624bb2bdda9c4b0534adddc62d0209192eb/src/functions/utils/labels.php#L56)

src/Tribe/Linked_Posts.php
```diff
@@ -959,6 +940,85 @@ public function render_meta_box_sections( $event ) {
...
+		// Here we render the main label string. The "core" linked post types (venue and organizer) are explicitly named to make
+		// translation a bit easier for the many languages where the words *around* the post type name may need to be different
+		// based on the specific post type name. For non-"core" post types, we just dynamically populate the post type name.
+		switch ( $post_type ) {
+
+			// Organizers
+			case Tribe__Events__Organizer::POSTTYPE :
+
+				if ( tribe_is_organizer_label_customized() ) {
+					$label = sprintf( esc_attr_x( 'Find %1$s %2$s', '"Find an Organizer", but when the word "Organizer" is customized to something else.', 'the-events-calendar' ), $indefinite_article, $singular_name );
```
barryhughes: Not sure the translation, escaping and substitution are in the correct order here. As it stands, we  escape the translation (good!) and then we apply the substitutions. 

However, if `$label` is indeed going to be used in an attribute that won't protect us - here and elsewhere we'd instead need to do things in this sort of order:

```php
$label = esc_attr( sprintf( _x( 'Foo %s', 'domain' ), $bar ) );
```

src/Tribe/Linked_Posts.php
```diff
@@ -959,6 +940,85 @@ public function render_meta_box_sections( $event ) {
...
+		$plural_name             = $this->linked_post_types[ $post_type ]['name'];
+		$singular_name           = ! empty( $this->linked_post_types[ $post_type ]['singular_name'] ) ? $this->linked_post_types[ $post_type ]['singular_name'] : $plural_name;
+		$singular_name_lowercase = ! empty( $this->linked_post_types[ $post_type ]['singular_name_lowercase'] ) ? $this->linked_post_types[ $post_type ]['singular_name_lowercase'] : $singular_name;
+
+		// First, determine what indefinite article we should use for the post type (important for English and several other languages).
+		$indefinite_article = _x( 'a', 'Indefinite article for the phrase "Find a {post type name}. Will be replaced with "an" if the {post type name} starts with a vowel.', 'the-events-calendar' );
+
+		$post_type_starts_with = substr( $singular_name, 0, 1 );
+		$post_type_starts_with = strtolower( $post_type_starts_with );
+		$english_vowels        = array( 'a', 'e', 'i', 'o', 'u' );
```
barryhughes: I've no idea as to how many other languages have a concept of indefinite articles ... however it seems like if `$indefinite_article` is translatable then the range of vowels we test against should also be translatable or at the very least filterable.

Fictional language Borkwebish - where the indefinite article is generally _zog_ unless the following word starts with _p, q_ or _r_ (in which case it is _zag)_ - provides a good example of the sort of problem I mean.

src/Tribe/Linked_Posts.php
```diff
@@ -959,6 +940,85 @@ public function render_meta_box_sections( $event ) {
...
+		switch ( $post_type ) {
+
+			// Organizers
+			case Tribe__Events__Organizer::POSTTYPE :
+
+				if ( tribe_is_organizer_label_customized() ) {
+					$label = sprintf( esc_attr_x( 'Find %1$s %2$s', '"Find an Organizer", but when the word "Organizer" is customized to something else.', 'the-events-calendar' ), $indefinite_article, $singular_name );
+
+					if ( $creation_enabled ) {
+						$label = sprintf( esc_attr_x( 'Create or Find %s', '"Create or Find Organizer", but when the word "Organizer" is customized to something else.', 'the-events-calendar' ), $singular_name );
```
barryhughes: Throughout this change, we've got some pretty long lines like this one. Though we don't want to break up translatable string literals we _can_ reformat the rest along these lines:

```php
$label = sprintf(
	esc_attr_x(
		'String %s',
		'Context',
		'domain'
	),
	$substitute
);
```

Up to you ... but it _might_ make this a tad easier to digest.

src/Tribe/Linked_Posts.php
```diff
@@ -959,6 +940,85 @@ public function render_meta_box_sections( $event ) {
...
+			case Tribe__Events__Venue::POSTTYPE :
+
+				if ( tribe_is_venue_label_customized() ) {
+					$label = sprintf( esc_attr_x( 'Find %1$s %2$s', '"Find a Venue", but when the word "Venue" is customized to something else.', 'the-events-calendar' ), $indefinite_article, $singular_name );
+
+					if ( $creation_enabled ) {
+						$label = sprintf( esc_attr_x( 'Create or Find %s', '"Create or Find Venue", but when the word "Venue" is customized to something else.', 'the-events-calendar' ), $singular_name );
+					}
+				} else {
+					$label = $creation_enabled ?  esc_attr__( 'Create or Find a Venue', 'the-events-calendar' ) : esc_attr__( 'Find a Venue', 'the-events-calendar' );
```
barryhughes: Another pretty long line - but the ternary gives us an opportunity to neatly multiline things along the following lines:

```php
$label = $some_condition
	? $this
	: $that;
```

src/functions/utils/labels.php
```diff
@@ -0,0 +1,63 @@
...
+ * A simple function for knowing if the "Event"/"Events" label has been changed.
+ *
+ * @since TBD
+ *
+ * @return bool
+ */
+function tribe_is_event_label_customized() {
+
+	if (
+		'Event' !== tribe_get_event_label_singular() ||
```
barryhughes: * This will probably work in most cases but I'll be a nerd and point out this would flop if filter functions for the corresponding hooks are not used throughout the whole request (ie, if someone listens for the calendar widget firing, and applies then unhooks one of those filters - to run once for a special effect, as it were - the result of this function wouldn't be reliable ... but I accept that's an edge case)
* We could probably simplify and condense to `return ( $cond_a || $cond_b /* || ... */ )` and drop the `if()` structure

src/Tribe/Linked_Posts.php
barryhughes: ... on second look, I see this logic isn't new, so don't worry

src/Tribe/Linked_Posts.php
ggwicz: good catch, fixed in all cases!

src/Tribe/Linked_Posts.php
ggwicz: cool. will leave for now

src/Tribe/Linked_Posts.php
ggwicz: good point; fixed all cases.

src/Tribe/Linked_Posts.php
ggwicz: rejoice! done in both instances

src/functions/utils/labels.php
ggwicz: that's a great point. i agree the failure case is an edge-case but definitely something we should improve upon if we think of it

ggwicz:
Great feedback Barry, this one's ready for another review! 🔍

PR #1646, opened 2017-10-27
Clear up next events link

barryhughes:
By appending `&raquo;` to the plural events label, in some languages such as French this meant we ended up with _Evénements » suivant_ instead of _Evénements suivant »_ ... this change alters the order in which we do things to avoid this.

:ticket: [#72097](https://central.tri.be/issues/72097)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -327,6 +327,7 @@ Please see the changelog for the complete list of changes in this release. Remem
 * Fix - Fixed issue where the value of an event's "Show Google Maps Link" option would not properly affect the displaying of the link on List View (props: @etechnologie) [75547]
 * Fix - Improve shortcode pagination/view change url so it is reusable (props: @der.chef and others) [70021]
 * Tweak - Fixed some display issues for the event schedule details (props @mia-caro)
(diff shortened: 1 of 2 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$events_label_plural' | [nav.php:34](https://github.com/moderntribe/the-events-calendar/blob/28848980541abb57e512e7f39d6ae05fb341bacb/src/views/list/nav.php#L34)

barryhughes:
Further change to clear up the previous events link in a similar manner and improve escaping for both.

barryhughes:
I simultaneously like the idea and flexibility but am unsure. _Should_ that be part of the translated text, or is it decoration (albeit taking the form of an HTML entity rather than being added via CSS)? How do those look in an RTL language with a different alphabet?

barryhughes:
Going to stick with what we have, though, as the most conservative option (apart from anything else, no breakage for languages where these strings have already been translated).

PR #1648, opened 2017-10-27
Prevent JS error when adding a Pro widget in the WP Customizer screen…

elimn:
Ticket: https://central.tri.be/issues/72127

Prevent JS error when adding a Pro widget in the WP Customizer screen .


barryhughes:
Going ahead with a merge to get this in QA's hands

PR #1656, opened 2017-11-02
Avoid unnecessary/unwanted extra iterations within tribe_get_template_part()

barryhughes:
Once we've successfully found and printed our template, we don't need to continue the loop.

:ticket: [#46630](https://central.tri.be/issues/46630)


Camwyn:
Looks pretty straightforward 🐬

PR #1660, opened 2017-11-06
Fix EA children records stuck in pending

lucatume:
Ticket: https://central.tri.be/issues/92614

Also: 
* https://central.tri.be/issues/91552
* https://central.tri.be/issues/92297
* https://central.tri.be/issues/91754

Related EA Service ticket: https://central.tri.be/issues/90501

This PR aims at fixing the issue where scheduled imports would not run unless manually triggered, before delving into the changes here is the nature of the issue:
1. when the `tribe_aggregator_cron` action is triggered (or its `single` alias) the code will verify and create the children scheduled records that needs to be created
2. since that children record creation involves a request to the Service, the one to create the import and get an import ID back, those request will undergo the soft HTTP limit we apply client side to avoid too many potentially time-consuming requests from going off at once
3. since those imports have been rescheduled the queue meta is not set and those would not be picked up in [the logic to check for waiting records](https://github.com/moderntribe/the-events-calendar/blob/28d50f611f748c01ee335cedc9dc9496594438a7/src/Tribe/Aggregator/Record/Queue_Processor.php#L178); this would lead to the scheduled imports not running
4. when 3 is fixed then, when `tribe_aggregator_process_insert_records` cron action runs, the client will try to process the queue for a number of queued scheduled imports (10 by default); usually 7 of those scheduled imports have been soft-throttled (see 2) and their corresponding import has never been created on the Service, this would make them fail; hence the message about the missing import id shown in the failure note

To cope with the above this PR puts in place the following changes:
a. make sure that children scheduled records always have a queue meta set; this solves 3 above
b. set a flag indicating a child scheduled record needs queueing on the Service (i.e. create the import)
c. when processing a record queue check that flag and, eventually, queue the import for that scheduled record; this solves 4
d. when doing `c` above make sure to handle the case where the lazy import queueing itself is rescheduled (soft http request limit still applies) and handle the record rescheduling or failure
e. add some checks and missing arguments on methods here and there to make sure non PHP warnings/errors/notices are generated


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *1.0.11*, but this was pull requested against *release/M17.22*.

lucatume:
putting this back to on hold as I've found a Redis related issue along the way

lucatume:
This is good for another look, please read the updated description.

lucatume:
This will be tested in branch, updating related tickets.

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M17.23*, but this was pull requested against *release/M17.22*.

PR #1668, opened 2017-11-09
#79079 - remove next scheduled import note for on the demand imports

jesseeproductions:
_Ref:_ [C#79079](https://central.tri.be/issues/79079)

### DIFF
--- a/src/Tribe/Aggregator/Tabs/New.php
+++ b/src/Tribe/Aggregator/Tabs/New.php
@@ -384,14 +384,17 @@ class Tribe__Events__Aggregator__Tabs__New extends Tribe__Events__Aggregator__Ta
 										 _x( ' at ', 'separator between date and time', 'the-events-calendar' ) .
 										 date( get_option( 'time_format' ), $scheduled_time );
 
-				$messages['success'][] = '<br/>' .
-										 sprintf( // add in timing
-											 __( 'The next import is scheduled for %1$s.', 'the-events-calendar' ),
-											 esc_html( $scheduled_time_string )
-										 ) .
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/Tabs/New.php
```diff
@@ -384,14 +384,17 @@ public function get_result_messages( $queue ) {
...
+
+                    $messages['success'][] = '<br/>' .
+                                             sprintf( // add in timing
+                                                 __( 'The next import is scheduled for %1$s.', 'the-events-calendar' ),
+                                                 esc_html( $scheduled_time_string )
+                                             ) .
+                                             ' <a href="' . admin_url( 'edit.php?page=aggregator&post_type=tribe_events&tab=scheduled' ) . '">' .
+                                             __( 'View your scheduled imports.', 'the-events-calendar' ) .
+                                             '</a>';
+                }
```
ggwicz: if it doesn't interfere with output, i would recommend putting this all in one `sprintf()` statement:

```
sprintf(
	__( 'The next import is scheduled for %1$s. <a href="%2$s">View your scheduled imports</a>' ),
	esc_html( $scheduled_time_string ),
	admin_url( 'edit.php?page=aggregator&post_type=tribe_events&tab=scheduled' )
);
```

jesseeproductions: Since the scope of this ticket is to remove the message, I would rather not create more issues by changing the message and make this more complex to translate.

ggwicz: okay 👍

ggwicz: approved and ready for merge!

PR #1693, opened 2017-11-15
ensure post_author is propagated to children records

lucatume:
Ticket: https://central.tri.be/issues/92947

This PR updates the code in charge of generating the children records to make sure the post author is propagated and applied to the imported events.


lucatume:
it's really a minor fix that adds a second layer of defense for something we had already patched; I would not add a changelog entry.

PR #1705, opened 2017-11-28
Fixed events gmap id index was translated.

LucaPipolo:
## Issue
Google Maps are not loaded due to wrong map ID.

## Details
1. The JS [see [here](https://github.com/moderntribe/the-events-calendar/blob/master/src/resources/js/embedded-map.js#L90)] is looking for an ID that starts with `tribe-events-gmap-` + index.
2. Events Gmap ID is assigned into the map view [see [here](https://github.com/moderntribe/the-events-calendar/blob/master/src/views/modules/map.php#L19)].
3. The original value of index is an integer, e.g. `0`, but, since `esc_attr_e` is used, It expects a string and it does not automatically casting it. So the plugin is doing `esc_attr_e(0)` instead of `esc_attr_e('0')`. This leads to very unexpected results, e.g, in my case I have `tribe-events-gmap-0/Email` instead of `tribe-events-gmap-0`.
3. Then I'm wondering why using `esc_attr_e` (so also applying a translation, see https://codex.wordpress.org/Function_Reference/esc_attr_e), instead of just `esc_attr`.

## Proposed solutions
- Use `esc_attr` instead of `esc_attr_e`.

### DIFF
--- a/src/views/modules/map.php
+++ b/src/views/modules/map.php
@@ -16,4 +16,4 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 $style = apply_filters( 'tribe_events_embedded_map_style', "height: $height; width: $width", $index );
 ?>
-<div id="tribe-events-gmap-<?php esc_attr_e( $index ) ?>" style="<?php esc_attr_e( $style ) ?>"></div><!-- #tribe-events-gmap-<?php esc_attr_e( $index ) ?> -->
+<div id="tribe-events-gmap-<?php esc_attr( $index ) ?>" style="<?php esc_attr_e( $style ) ?>"></div><!-- #tribe-events-gmap-<?php esc_attr( $index ) ?> -->


src/views/modules/map.php
```diff
@@ -16,4 +16,4 @@
 
 $style = apply_filters( 'tribe_events_embedded_map_style', "height: $height; width: $width", $index );
 ?>
-<div id="tribe-events-gmap-<?php esc_attr_e( $index ) ?>" style="<?php esc_attr_e( $style ) ?>"></div><!-- #tribe-events-gmap-<?php esc_attr_e( $index ) ?> -->
+<div id="tribe-events-gmap-<?php esc_attr( $index ) ?>" style="<?php esc_attr_e( $style ) ?>"></div><!-- #tribe-events-gmap-<?php esc_attr( $index ) ?> -->
```
fabianmarz: This will not display the index as you forgot to call echo in the php tag `<?php echo esc_attr( $index ) ?>`.

sc0ttkclark: We shouldn't use _e() when it's not necessary. It can't be officially translated when passed a variable for the translation string.

Just replace these with `<?php echo esc_attr( $index ); ?>` and we can call this golden.

fabianmarz:
@Ferddy789, why did you approve this PR? The code changes are not done yet so the code will not work as the attribute is not echoed correctly.

barryhughes:
Thanks, @LucaPipolo - we've merged your commit in in via [PR 1749](https://github.com/moderntribe/the-events-calendar/pull/1749) :+1:

PR #1711, opened 2017-12-04
No Google Maps Link if no address to link to

cliffordp:
https://central.tri.be/issues/94909

Went the route of modifying `tribe_get_map_link_html()` because it seemed the best way to go (if there's no URL, why output <a> tag?) but also because I found quite a few checks against `tribe_show_google_map_link()` without first checking `&& tribe_address_exists()`, as PRO does in one case.

### DIFF
--- a/src/functions/template-tags/google-map.php
+++ b/src/functions/template-tags/google-map.php
@@ -40,12 +40,18 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	 * @return string A fully qualified link to https://maps.google.com/ for this event
 	 */
 	function tribe_get_map_link_html( $postId = null ) {
-		$link = sprintf(
-			'<a class="tribe-events-gmap" href="%s" title="%s" target="_blank">%s</a>',
-			esc_url( tribe_get_map_link( $postId ) ),
-			esc_html__( 'Click to view a Google Map', 'the-events-calendar' ),
-			esc_html__( '+ Google Map', 'the-events-calendar' )
-		);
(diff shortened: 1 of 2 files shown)


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for **, but this was pull requested against *release/M17.25*.

src/functions/template-tags/google-map.php
```diff
@@ -40,12 +40,18 @@ function tribe_get_map_link( $postId = null ) {
...
-			esc_html__( '+ Google Map', 'the-events-calendar' )
-		);
+		$map_link = esc_url( tribe_get_map_link( $postId ) );
+
+		$link = '';
+
+		if ( ! empty( $map_link ) ) {
+			$link = sprintf(
+				'<a class="tribe-events-gmap" href="%s" title="%s" target="_blank">%s</a>',
+				$map_link,
```
ggwicz: small nitpick, but worth changing: escape as late as possible. so for this, would recommend making the `$map_link` definition just this:

`$map_link = tribe_get_map_link( $postId );`

and then down in the `sprintf()` call, use `esc_url( $map_link )`

cliffordp:
@ggwicz what if `esc_url()` causes it to go from `"gooooober"` to `""` -- therefore no actual link

ggwicz:
@cliffordp i think that's a rare enough case that we could probably not worry about it, but i see the point and reasoning so we can proceed without moving the location of `esc_url()` -- this is good to merge!

PR #1713, opened 2017-12-04
#66749 - preserve checkboxes and radio values when changing views

jesseeproductions:
_Ref:_ [C#66749](https://central.tri.be/issues/66749)


src/resources/js/tribe-events-bar.js
```diff
@@ -339,8 +339,18 @@ var tribe_events_bar_action;
 					}
 					else {
 						if ( $this.is( ':checkbox' ) ) {
+							if ( $this.is( ':checked' ) && 'undefined' === typeof ts.url_params[ $this.attr( 'name' ) ] ) {
+								// if checkbox and not defined setup as an array
+								ts.url_params[ $this.attr( 'name' ) ] = [];
+								ts.url_params[ $this.attr( 'name' ) ].push( $this.val() );
+							} else if ( $this.is( ':checked' ) && 'undefined' !== typeof ts.url_params[ $this.attr( 'name' ) ] ) {
+								// add value to array
+								ts.url_params[ $this.attr( 'name' ) ].push( $this.val() );
```
bordoni: To avoid the duplicated logic here; We could only setup the array in place inside of the conditional  and always push to it after.

PR #1736, opened 2017-12-22
Correctly handle organizer images

lucatume:
Ticket: https://central.tri.be/issues/96354

We do not show or use them... yet we should handle them correctly.


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -2266,4 +2247,32 @@ public function should_queue_import( $should_queue_import = null ) {
...
+		$args = array(
+			'ID'         => $organizer_id,
+			'image'      => $service_image,
+			'post_title' => get_the_title( $organizer_id ),
+		);
+		$image = $this->import_image( $args );
+
+		if ( ! is_wp_error( $image ) && ! empty( $image->post_id ) ) {
+			// Set as featured image
+			$featured_status = $this->set_post_thumbnail( $organizer_id, $image->post_id );
```
barryhughes: Could `$featured_image` or `$image_added` be a better name for this var?

barryhughes:
Looks good, but we should add a changelog entry.

lucatume:
Still working on this

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -2266,4 +2247,47 @@ public function should_queue_import( $should_queue_import = null ) {
...
 	}
+
+	/**
+	 * Attaches a service-provided image to an organizer.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $organizer_id The organizer post ID.
+	 * @param string $service_image
+	 * @param \Tribe__Events__Aggregator__Record__Activity $activity
```
barryhughes: No need to change it, but we don't really need a fully qualified class name here and could drop the leading `\`

PR #1739, opened 2017-12-28
Add new utility function

mitogh:
Add `tribe_is_home_events` function similar to `is_home` in WP to test if we are on the homepage of the events.

See Ticket: https://central.tri.be/issues/42195

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -1669,4 +1669,41 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 
 		return $field ? $body_and_separator . $field : $body;
 	}
+
+	/**
+	 * Utility function to test if we are on the home of events, it makes test the case when the page is set to be on
+	 * the frontpage of the site and if the user is on that page is on the homepage or if the user is on the events page
+	 * where the eventDisplay is set to default.
+	 *
+	 * @return bool
+	 */
+	function tribe_is_home_events() {
+		global $wp_query;
+
+		$events_as_front_page = tribe_get_option( 'front_page_event_archive', false );
+		// If the reading option has an events page as front page and we are on that page is on the home of events.
+		if (
+			$wp_query->tribe_is_event
+			&& $events_as_front_page
+			&& true === get_query_var( 'tribe_events_front_page' )
+		) {
+			return true;
+		}
+
+		// If the readme option does not has an event page as front page and if id the 'default' view on the main query
+		// as is going to set to 'default' when is loading the root of events/ rewrite rule also makes sure is not on
+		// a taxonomy or a tag.
+		if (
+			! $events_as_front_page
+			&& $wp_query->tribe_is_event // Make sure following conditionals operate only on events
+			&& ( isset( $wp_query->query['eventDisplay'] ) && 'default' === $wp_query->query['eventDisplay'] )
+			&& is_post_type_archive()
+			&& ! is_tax()
+			&& ! is_tag()
+		) {
+			return true;
+		}
+		// No condition was true so is not on home of events.
+		return false;
+	}
 }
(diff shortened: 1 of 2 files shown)


src/functions/template-tags/general.php
```diff
@@ -1669,4 +1669,41 @@ function tribe_separated_field( $body, $separator, $field ) {
 
 		return $field ? $body_and_separator . $field : $body;
 	}
+
+	/**
+	 * Utility function to test if we are on the home of events, it makes test the case when the page is set to be on
+	 * the frontpage of the site and if the user is on that page is on the homepage or if the user is on the events page
+	 * where the eventDisplay is set to default.
+	 *
```
bordoni: add `@since`

mitogh: Thanks, changes has been updated.

mitogh:
I added a new method here @bordoni is basically a separation of the same function in two as I'm going to need that one on #1748, let me know if you have any questions.

src/functions/template-tags/general.php
```diff
@@ -1669,4 +1669,60 @@ function tribe_separated_field( $body, $separator, $field ) {
 
 		return $field ? $body_and_separator . $field : $body;
 	}
+
+	/**
+	 * Function to test if we are on the front page of the events, as WordPress language front page is different from
+	 * is_home, if we have a front page on the reading options and if the uzer is on that page, this function will
```
barryhughes: * User

src/functions/template-tags/general.php
```diff
@@ -1669,4 +1669,60 @@ function tribe_separated_field( $body, $separator, $field ) {
 
 		return $field ? $body_and_separator . $field : $body;
 	}
+
+	/**
+	 * Function to test if we are on the front page of the events, as WordPress language front page is different from
```
barryhughes: This is good - but sometimes the information is easier to digest if we start with a nice and short sentence explaining very briefly what the function does - with any further explanations, notes or examples into additional paragraphs. For instance:

```
/**
 * Tests if we are on the site homepage and if it is set to display the main events page.
 *
 * <notes about the differences with WordPress's definition of front page, examples of
 * real usage etc>
 * 
 * @since TBD
 *
 * @return bool
 */
```

Note also the line break between 'meta' information such as `@since` and the `@return` tag.

barryhughes:
@mitogh - I like these functions but there's a possible bear-trap we ought to warn users of: if code _within an events template_ calls `tribe_is_events_front_page()`, for example, then it will work so long as the front page was loaded directly. If the user pages forward by ajax, though, it will cease to work. This could cause some confusion and is probably worth warning about.

We should also add a changelog entry noting the addition of these new template tags.

src/functions/template-tags/general.php
mitogh: Thanks.

src/functions/template-tags/general.php
mitogh: Thanks for the heads up.

mitogh:
Thanks for the feedback @barryhughes. However I'm not really sure about the changelog. just a couple of questions there:

1. Should we edit the `readme.txt` file?
2. What version and date should we add?

Thanks.

barryhughes:
> Should we edit the readme.txt file?

Yes :+1:

> What version and date should we add?

The date should generally be TBD at this stage, but the new version number can be found at the top of the readme file (though, if in doubt, TBD is fine for this, too).

In this case, you can add your entry [within this block](https://github.com/moderntribe/the-events-calendar/blob/release/M18.01/readme.txt#L311-L316).

mitogh:
Changes has been added diff is not able to pick the changes, on the missing changes.

cc @barryhughes @bordoni

barryhughes:
:+1:

PR #1740, opened 2017-12-28
Filter events if we are on a category

mitogh:
This will allow to have only events that are part of the current term
inside of the the list of events to avoid render extra events that are
not part of the same category.

See https://central.tri.be/issues/80551


src/Tribe/Template/Month.php
```diff
@@ -314,13 +319,41 @@ protected function produce_json_ld_markup( $echo = false ) {
...
 
 			if ( ! $echo ) {
 				return ob_get_clean();
 			}
 		}
 
+		/**
+		 * Remove any event that is not the same as the current term of the array of events, it return a modified events
+		 * array with the events that only has the current term or original events if the term is not valid.
+		 *
```
bordoni: Add a `@since` param to the method, even if it's a Private one.

mitogh: Added.

PR #1741, opened 2017-12-31
Enable translation of iCal calendar description

gafderks:
# Issue
The calendar description in a generated iCal feed is not translated.

# Details
The translation is not used.

# Proposed solutions
Use esc_html__ to enable translation of the description.

### DIFF
--- a/src/Tribe/iCal.php
+++ b/src/Tribe/iCal.php
@@ -420,7 +420,7 @@ class Tribe__Events__iCal {
 
 		$content .= 'X-WR-CALNAME:' . $x_wr_calname . "\r\n";
 		$content .= 'X-ORIGINAL-URL:' . $blogHome . "\r\n";
-		$content .= 'X-WR-CALDESC:Events for ' . $blogName . "\r\n";
+		$content .= 'X-WR-CALDESC:' . esc_html__( 'Events for ', 'the-events-calendar' ) . $blogName . "\r\n";
 
 		/**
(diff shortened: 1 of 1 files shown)


src/Tribe/iCal.php
```diff
@@ -420,7 +420,7 @@ public function generate_ical_feed( $post = null, $echo = true ) {
 
 		$content .= 'X-WR-CALNAME:' . $x_wr_calname . "\r\n";
 		$content .= 'X-ORIGINAL-URL:' . $blogHome . "\r\n";
-		$content .= 'X-WR-CALDESC:Events for ' . $blogName . "\r\n";
+		$content .= 'X-WR-CALDESC:' . esc_html__( 'Events for ', 'the-events-calendar' ) . $blogName . "\r\n";
```
barryhughes: Hey there @gafderks - love the idea here - but might it be better to use substitution via `sprintf()` and to also provide some context - something like this?

```php
sprintf( esc_html_x( 'Events for %s', 'iCal feed description', 'the-events-calendar' ), $blogname )
```

Adding context would make it easier for translators to target this string specifically and using the `%s` placeholder would allow us to cleanly support languages with a different word order.

gafderks: Thanks for the feedback! I have implemented these suggestions.

barryhughes:
:ticket: [#96677](https://central.tri.be/issues/96677)

barryhughes:
Thanks @gafderks - we'll get this merged in pronto!

PR #1744, opened 2018-01-01
spotfix | corrections from "venue" to "organizer"

cliffordp:
Only comments changed other than a filter name, which may warrant a readme or code comment / deprecation


barryhughes:
Thanks @cliffordp!

PR #1757, opened 2018-01-10
Use singular label instead of hardcode value

mitogh:
Make sure to use the singular label value of the post type insitead of
the hard code value of 'Event'

@ [Central ticket](https://central.tri.be/issues/70387)


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for **, but this was pull requested against *release/M18.02*.

PR #1758, opened 2018-01-10
check for the enabled rest api status

lucatume:
Ticket: https://central.tri.be/issues/97209

Make sure the TEC REST API can be correctly disabled.


lucatume:
I still have to update the readme.txt file as the changes from 18.01 have not been merged in there yet

PR #1762, opened 2018-01-11
Make sure the Map options are present on save venue

mitogh:
This will ensure the options for the map are always present when a venue is saved in order to allow the option to set
them to false if that's the case from the admin panel as they are not present inside of the `$_POST` request if the
checkbox are set to `false`.

It adds a new filter as well `tribe_events_venue_created_map_default` that will allow the community events the option
to fix a bug when `tribe_events_community_apply_map_defaults` is set to `false`.

[Central](https://central.tri.be/issues/91869)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space before the last parenthesis | [Main.php:3201](https://github.com/moderntribe/the-events-calendar/blob/f169bf02370e1ae02ffbb826ac9f2cc748d15e81/src/Tribe/Main.php#L3201)

PR #1768, opened 2018-01-16
For past events use current date instead of end of day

mitogh:
For any event that is past on the list view we should consider the end
of the event as the current date instead of the current end of the day
as will wrap events of today in the past.

See: https://central.tri.be/issues/71936


barryhughes:
Remember to add a changelog entry 😄

mitogh:
Changelog has been updated.

PR #1771, opened 2018-01-17
Allow the usage of `featured_color_scheme_custom` if present

mitogh:
This change allows the usage of `featured_color_scheme_custom` over `featured_color_scheme` if present, this will use the
color selected by the user instead of the default color scheme.

See https://central.tri.be/issues/96821


mitogh:
Closing this one as has `common` on it.

PR #1778, opened 2018-01-19
Better handling/querying support for hidden events | #91254

barryhughes:
This change supports queries which explicitly set `'hide_upcoming'` to false (in order to expose events that would otherwise be hidden).

:ticket: [#91254](https://central.tri.be/issues/91254)


tests/wpunit/Tribe/Events/QueryTest.php
```diff
@@ -57,6 +57,7 @@ public function it_should_not_remove_date_filters_on_non_event_edit_pages() {
 		] );
 
 		$this->assertFalse( Query::should_remove_date_filters( $query ), 'Date filters should not be removed on non-event edit pages' );
+		unset( $GLOBALS['current_screen'] ); // cleanup
```
barryhughes: Note - here and elsewhere I'm doing this because otherwise the change to the global state caused by `set_current_screen()` calls persists and scuppers other tests.

PR #1781, opened 2018-01-19
Add filters for Tribe Commerce to alter the TEC cost display

Camwyn:
🎫 96777 
There is no Tribe Commerce bucket for TEC and since this is just adding three new filters it didn't seem like one was necessary?


Camwyn:
I can use another filter without changing TEC...

PR #1782, opened 2018-01-22
Trigger an event when the embed map is created

mitogh:
The event is fired when an embedded map is attached as well.

See https://central.tri.be/issues/91984


bordoni:
➕ 💯 for this

PR #1787, opened 2018-01-25
Make sure the `/events` page is not hardcoded

mitogh:
🎟  https://central.tri.be/issues/98463
The URL might change due a different slug based on the options page,
make sure this URL is not hardcoded.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -309,7 +309,7 @@ The plugin is made with love by [Modern Tribe Inc](http://m.tri.be/2s).
 == Changelog ==
 
 = [4.6.10.1] 2018-01-25 =
-* Fix - Make sure rewrite rule for `/events` is not hardcoded and is based on dynamic option field [98463]
+* Fix - Make sure rewrite rule for `/events` is not hardcoded and is based on dynamic option field (thanks to @earnjam and others for flagging this problem) [98463]
 
 = [4.6.10] 2018-01-23 =
 
--- a/readme.txt
+++ b/readme.txt
@@ -309,8 +309,7 @@ The plugin is made with love by [Modern Tribe Inc](http://m.tri.be/2s).
 == Changelog ==
 
 = [4.6.10.1] 2018-01-25 =
-
-
+* Fix - Make sure rewrite rule for `/events` is not hardcoded and is based on dynamic option field [98463]
 
 = [4.6.10] 2018-01-23 =
 
--- a/src/Tribe/Rewrite.php
+++ b/src/Tribe/Rewrite.php
@@ -255,10 +255,11 @@ class Tribe__Events__Rewrite extends  Tribe__Rewrite {
 			'tag' => array( 'tag', $tec->tag_slug ),
 			'tax' => array( 'category', $tec->category_slug ),
 			'page'     => array( 'page', esc_html_x( 'page', 'The "/page/" URL string component.', 'the-events-calendar' ) ),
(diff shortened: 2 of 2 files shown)


src/Tribe/Rewrite.php
```diff
@@ -255,11 +255,12 @@ public function get_bases( $method = 'regex' ) {
 			'tag' => array( 'tag', $tec->tag_slug ),
 			'tax' => array( 'category', $tec->category_slug ),
 			'page'     => array( 'page', esc_html_x( 'page', 'The "/page/" URL string component.', 'the-events-calendar' ) ),
-			'single' => array( 'event', $tec->rewriteSlugSingular ),
-			'archive' => array( 'events', $tec->rewriteSlug ),
+			'single' => array( Tribe__Settings_Manager::get_option( 'singleEventSlug', 'event' ), $tec->rewriteSlugSingular ),
```
barryhughes: This improves things but as I think about it, shouldn't we do something like:

```php
array_filter( array(
	'event',
	Tribe__Settings_Manager::get_option( 'singleEventSlug', 'event' ),
	$tec->rewriteSlugSingular
) )
```

Looking at the comment above this block, we always want to support the original/English slug - but right now we potentially will miss that element.

bordoni: Uhnn I was thinking maybe the intention was to allow the users to completely remove `events/event` using a different option on the DB.

readme.txt
```diff
@@ -309,8 +309,7 @@ The plugin is made with love by [Modern Tribe Inc](http://m.tri.be/2s).
 == Changelog ==
 
 = [4.6.10.1] 2018-01-25 =
-
-
+* Fix - Make sure rewrite rule for `/events` is not hardcoded and is based on dynamic option field [98463]
```
barryhughes: Especially with a hotfix, it's a nice touch to acknowledge folks who highlighted the problem. Maybe add: _(thanks to elecevan, nickfetner and everyone else who reported this problem)_

src/Tribe/Rewrite.php
mitogh: This one should be only for the single view? as if we do it for `archive`  we are still going to have the same issue if we leave `events` @barryhughes

readme.txt
mitogh: For sure, will update this one.

src/Tribe/Rewrite.php
barryhughes: Hmm, don't think it is but if I'm wrong let's also update the comment above this section.

Seems though that we'd be creating a headache in scenarios where URLs using the original slug are promulgated via sharing links etc.

barryhughes: > This one should be only for the single view?

mitogh: if we leave `events` hardcoded is going to be present anyways  so we are going to have the `events` url taken by the plugin so I think we should update the documentation around this, plus is the default value if the option value is not changed if will be different only if the user change is intentionality. 

cc @bordoni @barryhughes

barryhughes: I think the sane default is for us to support:

barryhughes: ... filtering out any duplicates

barryhughes: Actually, I see what you're saying. Let's keep this as is.

PR #1791, opened 2018-02-06
Generate iCal feed with correct events on shortcode

mitogh:
🎟  https://central.tri.be/issues/86227

When the shortcode is being used it was using the post / page object as
the object to populate the list of events instead of using the data from
ths shortcode being used when the events where displayed via a
shortcode.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Usage of ELSE IF is discouraged; use ELSEIF instead | [iCal.php:167](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L167)
:x: | There must be a single space after the first parenthesis | [iCal.php:191](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L191)
:x: | There must be a single space before the last parenthesis | [iCal.php:191](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L191)
:x: | There must be a single space before the last parenthesis | [iCal.php:212](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L212)
:x: | Each line in an array declaration must end in a comma | [iCal.php:213](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L213)
:x: | There must be a space between foreach and the first parenthesis '(' | [iCal.php:232](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L232)
:x: | Missing space after array opener. | [iCal.php:236](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L236)
:x: | There must be a single space after the first parenthesis | [iCal.php:236](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L236)
:x: | There must be a single space after the first parenthesis | [iCal.php:236](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L236)
:warning: | Usage of ELSE IF is discouraged; use ELSEIF instead | [iCal.php:521](https://github.com/moderntribe/the-events-calendar/blob/31693535d6d97190645cf16c8b91a837a582c2d0/src/Tribe/iCal.php#L521)

src/Tribe/iCal.php
```diff
@@ -157,14 +157,89 @@ public function do_ical_template() {
 				$events    = tribe_get_events( array( 'post__in' => $event_ids ) );
 				$this->generate_ical_feed( $events );
 			} elseif ( is_single() ) {
-				$this->generate_ical_feed( $wp_query->post );
+				if ( is_singular( Tribe__Events__Main::POSTTYPE ) ) {
```
barryhughes: 🤔 in the interests of minimizing the nesting of if blocks, what do you think about shifting from:

```
} elseif ( is_single() ) {
    if ( is_singular( /*...*/ ) {
        # ...
    } else {
        # ...
    }
} else { #...
```

To something like:

```
} elseif ( is_singular( /*...*/ ) {
    # ...
} elseif ( is_single() ) {
    # ...
} else { #...
```

src/Tribe/iCal.php
```diff
@@ -157,14 +157,89 @@ public function do_ical_template() {
...
+					// If the single view is rendered we need to render the post only if is an Event
+					$this->generate_ical_feed( $wp_query->post );
+				} else {
+					// otherwise it might be a shortcode.
+					$this->generate_ical_feed( $this->find_events_in_shortcode() );
```
barryhughes: This looks like a solid working solution, _but,_ I'm a little uncomfortable with us adding special handling for a PRO feature inside of TEC. As a rule of thumb, the dependency (PRO) should have knowledge of the parent (TEC) but not the other way round.

Is it possible for us to approach this by making the list of post IDs used to generate the ical feed filterable, and filtering it from within the shortcode logic contained in PRO?

(cc @bordoni for any thoughts on this)

src/Tribe/iCal.php
mitogh: Seems more readable in that way ;)

src/Tribe/iCal.php
mitogh: Got it, so we just need to update as follows: 

1. Add a filter to filter list of events IDs.
2. Add a callback on PRO about this change
3. Move the logic all the way there (in PRO).

Let me know what do you think.

Thanks

barryhughes: Yes! Adding a filter would anyway be useful outside of this scenario, so it feels like a great way to go :+1:

mitogh:
Ready for another set of 👀 . 

PS: [PR for the PRO functionality](https://github.com/moderntribe/events-pro/pull/781)

mitogh:
Do you think we can run travis again here @bordoni ?

Thanks

PR #1792, opened 2018-02-06
Don't modify the main query if is in a single shortcode.

mitogh:
🎟  https://central.tri.be/issues/91787


src/Tribe/Query.php
```diff
@@ -128,9 +133,9 @@ public static function parse_query( $query ) {
 				: false;
 
 			$query->tribe_is_event_query = ( $query->tribe_is_event
-			                                 || $query->tribe_is_event_category
-			                                 || $query->tribe_is_event_venue
-			                                 || $query->tribe_is_event_organizer )
+				|| $query->tribe_is_event_category
```
barryhughes: (1/2) not sure if this was deliberate or automated formatting, but just wanted to flag as a) it's not a functional change and so is in a sense unwanted noise b) neither this new indentation nor the previous indentation really match how we want things to be

src/Tribe/Query.php
```diff
@@ -242,78 +247,78 @@ public static function pre_get_posts( $query ) {
...
 					switch ( $query->query_vars['eventDisplay'] ) {
-						case 'custom':
-							// if the eventDisplay is 'custom', all we're gonna do is make sure the start and end dates are formatted
-							$start_date = $query->get( 'start_date' );
-
-							if ( $start_date ) {
-								$start_date_string = $start_date instanceof DateTime
-									? $start_date->format( Tribe__Date_Utils::DBDATETIMEFORMAT )
-									: $start_date;
+					case 'custom':
```
barryhughes: (2/2) here also it looks like this has been reformatted for no particular reason and actually the previous formatting where each `case` block was indented was correct

mitogh: Woops seems like was due the editor let me rollback 💃

src/Tribe/Query.php
```diff
@@ -660,18 +665,18 @@ public static function set_orderby( $default = null, $query = null ) {
...
-				case 'tribe_sort_ecp_organizer_filter':
-					$orderby = 'organizer';
-					break;
-				case 'title':
-					$orderby = $url_param;
-					break;
-				default:
-					$orderby = $default;
-					break;
+			case 'tribe_sort_ecp_venue_filter':
```
barryhughes: Similar to the above, we should indent by one more tab when inside `{}` blocks

src/Tribe/Query.php
barryhughes: cool :+1:

mitogh:
Ready for another set of 👀 .

PR #1795, opened 2018-02-06
Checks for the message in two places instead of the message

bordoni:
_Ref:_ [C#94708](http://central.tri.be/issues/94708)
---


borkweb:
Looks like a REST test is failing here. This PR didn't cause the issue - but I'm just calling that out.

PR #1796, opened 2018-02-07
Feature/81831 add url validation on ea

bordoni:
_Ref:_ [C#81831](http://central.tri.be/issues/81831)
---
_Depends on:_ https://github.com/moderntribe/tribe-common/pull/608


src/Tribe/Aggregator/Record/Url.php
```diff
@@ -16,4 +16,37 @@ class Tribe__Events__Aggregator__Record__Url extends Tribe__Events__Aggregator__
...
 	}
+
+	/**
+	 * Matches which other origin this source url might be
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string $source Which source we are testing against
+	 *
+	 * @return string|bool
```
mitogh: I think it might be more consistent easier if this one always returns a string and when is `false` returns an empty string instead.

bordoni: That actually strikes as odd, I mean, it does make sense to simplify the results but it also empty string isn't the clearer "answer" in my view for what just happened.

mitogh: I think you make a good point. ;)

PR #1797, opened 2018-02-07
Prevent incompatibility with lodash if present

mitogh:
🎟  https://central.tri.be/issues/92205

This change makes sure everytime lodash is used with underscore they
both will fight for the usage of window._ on the global namespace.

This change makes sure it leaves this priority to lodash if both present
and use a fallback for the plugins like: window.underscore || window._


src/Tribe/Main.php
```diff
@@ -2171,6 +2172,39 @@ public function loadStyle() {
...
+				$dir = plugin_dir_url( TRIBE_EVENTS_FILE ) . 'src/resources/js/';
+
+				$before = file_exists( $path . 'underscore-before.js' ) ? file_get_contents( $dir . 'underscore-before.js' ) : '';
+				if ( $before ) {
+					$tag = "<script type='text/javascript'>$before</script>\n" . $tag;
+				}
+
+				$after = file_exists( $path . 'underscore-after.js' ) ? file_get_contents( $dir . 'underscore-after.js' ) : '';
+				if ( $after ) {
+					$tag = $tag . "\n<script type='text/Javascript'>$after</script>";
```
bordoni: If the Underscore is not async we can include it like a normal JS file, no need to make it inline.

src/Tribe/Main.php
```diff
@@ -2171,6 +2172,39 @@ public function loadStyle() {
...
+				return $tag;
+			}
+
+			if ( 'underscore' === $handle ) {
+				$path = plugin_dir_path( TRIBE_EVENTS_FILE ) . 'src/resources/js/';
+				$dir = plugin_dir_url( TRIBE_EVENTS_FILE ) . 'src/resources/js/';
+
+				$before = file_exists( $path . 'underscore-before.js' ) ? file_get_contents( $dir . 'underscore-before.js' ) : '';
+				if ( $before ) {
+					$tag = "<script type='text/javascript'>$before</script>\n" . $tag;
```
bordoni: If the Underscore is not async we can include it like a normal JS file, no need to make it inline.

src/Tribe/Main.php
```diff
@@ -2171,6 +2172,39 @@ public function loadStyle() {
...
+		 * @return string The <script> tag.
+		 */
+		public function prevent_underscore_conflict( $tag, $handle ) {
+
+			if ( is_admin() ) {
+				return $tag;
+			}
+
+			if ( 'underscore' === $handle ) {
+				$path = plugin_dir_path( TRIBE_EVENTS_FILE ) . 'src/resources/js/';
```
bordoni: Take a look at the code from `Tribe__Asset` there are methods there for that. And we should move this method to Common.

src/Tribe/Main.php
```diff
@@ -2171,6 +2172,39 @@ public function loadStyle() {
...
+		 * Before underscore is loaded to the FE we add two scripts on before and one after to prevent underscore from
+		 * taking place on the global namespace if lodash is present.
+		 *
+ 		 * @since TBD
+ 		 *
+ 		 * @param string $tag The <script> tag for the enqueued script.
+		 * @param string $handle The script's registered handle.
+		 * @return string The <script> tag.
+		 */
+		public function prevent_underscore_conflict( $tag, $handle ) {
```
bordoni: Also avoid adding new methods to `Main.php`

src/Tribe/Main.php
mitogh: Cool I'm going to take a look at this one.

src/Tribe/Main.php
mitogh: In this particular case we have to, as we don't know what script is going to be loaded before `underscore`, for the one after works but the one before we need to make sure is done before `underscore` is attached to the global namespace.

src/Tribe/Main.php
mitogh: I don't think there is a method there to do inline JS they are wrappers around `wp_enqueue_script`. What do you think if we add a new method in that class for inline JS.

Also which class do you think it might be the best one to move this logic (most of this PR).

Thanks.

bordoni: I would move it to a new Class `Load_Assets.php` in common.

src/Tribe/Main.php
bordoni: What I mean was to add a `<script src="">` these scripts instead of inline including them. Manually include those but avoid reading the file via the backend.

PR #1799, opened 2018-02-08
Resolve Settings for EA imports + Fix some alignments

bordoni:
_Ref:_ [C#81889](http://central.tri.be/issues/81889)
---


src/resources/postcss/aggregator-page.pcss
```diff
@@ -276,6 +276,10 @@
 		color: gray;
 	}
 
+	.tribe-ea-inline-label {
+		line-height: 30px;
```
borkweb: This is the only item in this PR that gives me pause. I think it'll do the job with spacing, but is it the right way to do it? If the label was a multi-word label that ends up wrapping on small screen sizes, a line height of 30px is pretty ugly. While line-height can accomplish what you want, perhaps padding and/or margins is more desirable?

The labels we are talking about here are short, so perhaps it doesn't matter all that much. I'll leave it up to you whether you change it or not.

bordoni: Very true, will fix that.

PR #1802, opened 2018-02-08
#95272 Allow Mixed-Content Event Costs

ggwicz:
**Ticket:** [**#95272**](http://central.tri.be/issues/95272)


ggwicz:
Gustavo and I chatted and decided against implementing this change altogether, and will instead publish a simple Extension for folks who want to add text labels to their Event Cost fields.

See [note 15 on the associated ticket](https://central.tri.be/issues/95272#note-15) for more information.

PR #1804, opened 2018-02-09
spotfix | $beginning_of_day should be midnight, not 1 sec after

cliffordp:
`00:00:00` is the beginning of the day, not `00:00:01`

Note that this makes it the same as Month View's logic a few lines above

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -287,7 +287,7 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 						case 'day':
 							$event_date = $query->get( 'eventDate' ) != '' ? $query->get( 'eventDate' ) : date( 'Y-m-d', current_time( 'timestamp' ) );
 							$query->set( 'eventDate', $event_date );
-							$beginning_of_day = strtotime( tribe_beginning_of_day( $event_date ) ) + 1;
+							$beginning_of_day = strtotime( tribe_beginning_of_day( $event_date ) );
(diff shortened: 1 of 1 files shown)


ggwicz:
Have you seen any issues arising from this, @cliffordp? Or did you just notice this in the code and submitted the patch?

I'm skeptical of making this change because, for example, the addition of that second may be an intentional change to _ensure_ a more reliable result here. You mention other logic in the code that _doesn't_ use the additional second; but that code may be doing something different than this code is.

---

- If there's no issues arising from this code, **I would personally vote against messing with this**.
- If there _is_ a bug or more fleshed-out reason for the change, can you log a ticket and explain the problem there?

cliffordp:
Firstly, in my opinion it is incorrect. If something else relies on it that would be the place to input the modification.

Secondly, I discovered it when building https://github.com/mt-support/tribe-ext-schedule-day-view/tree/initial, and I saw what ended up in $wp_query was off by a second so I couldn't use it and had to build my own "beginning of day"

barryhughes:
My memory of this is a little vague but I believe George is basically on the right track: we did this as a practical workaround for issues relating to handling of all day events in month view and possibly other views.

I'm definitely not opposed to us rationalizing this and making it saner, but despite the small size of the changeset I don't think this is safe to push forward as a spotfix. It needs to be ticketed (so that QA see and test the change); ideally we'd also add supporting tests.

barryhughes:
This PR was closed automatically following the cleanup of release/M18.03 (and the deletion of that branch).

cliffordp:
Moved to https://central.tri.be/issues/99641

PR #1806, opened 2018-02-09
Don't modify pages as well if is the main_query

mitogh:
Ticket: https://central.tri.be/issues/91787

This PR does not include a changelog as is a hotfix to a change already added.

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -78,6 +78,11 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 				return $query;
 			}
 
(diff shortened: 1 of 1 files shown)


mitogh:
Yeah however we can say that the `query` shouldn't be modified if: 

1. If is a single view and the post is not from an event post type so the view shouldn't be modified.
2. If is a page as the main calendar view set the view to is_post_type_archive() and page is not used unless is a shortcode, and shortcode attributes are provided by the shortcode itself. 

Following queries are not going to be the `main_query` as they might be part of the query inside of a content which can be pass in this one if that's the case as we are testing only if we are on the `main_query`

PR #1808, opened 2018-02-09
Make sure arguments are overwritten by the $args value

mitogh:
Ticket: https://central.tri.be/issues/86227

Make sure $args takes priority on default arguments, changelog is not
required as is a fix already present on the changelog.


bordoni:
We need to fix the Tests here.

mitogh:
Tests fixed ;)

PR #1821, opened 2018-02-16
Update submodule hash

mitogh:


bordoni:
Updating Submodule hash is not needed, @tr1b0t will do that for us on `release/` branches.

image

mitogh:
🐼

PR #1824, opened 2018-02-16
Cache HTTP call for a day to avoid multiple calls

mitogh:
**Ticket** https://central.tri.be/issues/99486

Makes sure next calls are based from the transient instead of the HTTP
call.


src/Tribe/Aggregator.php
```diff
@@ -3,6 +3,14 @@
 defined( 'WPINC' ) or die;
 
 class Tribe__Events__Aggregator {
+	/**
+	 * Cache key used to storage the services list returned by the call to:
+	 * - Tribe__Events__Aggregator__Service::instance()->get_origins();
+	 *
+	 * @since TBD
+	 */
+	const SERVICES_CACHE_KEY = 'tribe_aggregator_services_list';
```
bordoni: Avoid new Constants, use a public prop. Access by `tribe( 'tec.aggregator' )->key_cache_services`.
When saving keys let's try to always start with `key_...`

src/Tribe/Aggregator.php
```diff
@@ -582,4 +593,13 @@ public function hook() {
...
 		return true;
 	}
+
+	/**
+	 * Function used to remove cache stored in transients.
+	 *
+	 * @since TBD
+	 */
+	public function clear_services_list_cache() {
+		delete_transient( self::SERVICES_CACHE_KEY );
```
bordoni: `return` the result o the function call.

src/Tribe/Aggregator/Admin_Bar.php
```diff
@@ -80,7 +80,13 @@ public function init( WP_Admin_Bar $wp_admin_bar = null ) {
 			return;
 		}
 
-		$service_response = Tribe__Events__Aggregator__Service::instance()->get_origins();
+		$service_response = get_transient( Tribe__Events__Aggregator::SERVICES_CACHE_KEY );
+
+		// Save HTTP request into a transient
+		if ( false === $service_response ) {
+			$service_response = Tribe__Events__Aggregator__Service::instance()->get_origins();
+			set_transient( Tribe__Events__Aggregator::SERVICES_CACHE_KEY, $service_response, DAY_IN_SECONDS );
```
bordoni: I have a feeling like we have some cache in place for the origins, or we had it in the past and was removed for reason. Or at least we need to invalidate cache once the user adds a new license.

src/Tribe/Aggregator.php
mitogh: Curios here of the reasons behind this also wondering if this should be the approach from now on as in some places the `CONST` is the current way to do this.

src/Tribe/Aggregator/Admin_Bar.php
mitogh: 👍

src/Tribe/Aggregator.php
mitogh: 👍

src/Tribe/Aggregator/Admin_Bar.php
mitogh: Based on the history of the file there seems that never has been stored in some sort of caching mechanism  (the response.)

src/Tribe/Aggregator.php
bordoni: Yes, so the problem is that there is no way to deprecate `CONST` on PHP 5.2 so we are avoiding adding any new ones. I have a ticket out for me to update that on our Engineering Docs.

src/Tribe/Aggregator/Admin_Bar.php
bordoni: Pefecto!

PR #1829, opened 2018-02-21
Remove condition on mobile redirection

mitogh:
**Ticket** https://central.tri.be/issues/68716

The condition of same URL and current URL can be a false positive as in such cases the base URL and current URL might the same, but the redirection is still required as the conditions above ensure that if mobile and desktop view are different we should redirect even if current URL and base URL are the same, as the base URL can be the canonical URL of the events.


src/resources/js/tribe-events.js
```diff
@@ -776,10 +776,7 @@ Date.prototype.format = function( mask, utc ) {
 				! tribe_ev.state.view ||
 
 				// We are on the default mobile view
-				tribe_ev.data.default_mobile_view == tribe_ev.state.view ||
-
-				// We are with an defined view
-				tribe_ev.data.cur_url == tribe_ev.data.base_url
+				tribe_ev.data.default_mobile_view == tribe_ev.state.view
```
ggwicz: Did you run `gulp` command on this PR? To generate the minified JS file?

mitogh: I did not, thanks for flagging this will send in a moment.

mitogh: File has been added ;)

PR #1834, opened 2018-02-23
97344 Force Local TZ when Getting Event Abbr

ggwicz:
**Ticket:** [**#97344**](http://central.tri.be/issues/97344)
---
**Code Review Notes**
This change addresses an issue that happens **only** when the "Timezone Mode" option for events is set to _Use sitewide timezone everywhere_. This issue does not happen when the _Use local timezones for events_ option is selected. 

The best way to explain the reason for this change is with some screenshots.

- First, view [this screenshot](https://cldup.com/r0MNJGhXpx-3000x3000.png)—it shows the admin for two events occurring in the `Europe/Zurich` timezone, one _before_ Daylight Savings, the other _after_ Daylight Savings.
- Next, view [this screenshot](https://cldup.com/zXzmF1iSF8-3000x3000.png)—it shows the front-end output of these events.
   - The pre-Daylight Savings event should show `8am–5pm CET`.
  - The post-Daylight Savings event should show `8am–5pm CEST`.
  - The post-Daylight Savings event could also show `9am–6pm CEST`.
  - But **after looking at Google Cal and others**, it seems like the norm is to preserve the numeric time, and just adjust the Daylight Savings flag; so the expected output _should be_ `8am–5pm CEST`.

**With the above notes in mind**, I've found that making this change here addresses the problem—at least, in terms of how _I think of the problem_. 

**If we don't implement a change like this**, and keep things as-is, then I think we should:
- (1) Do a touch-up of the existing Knowledgebase article on timezones to help clarify how _Use sitewide timezone everywhere_ setting works
- (2) Add a "FAQs About Timezones" Knowledgebase article to more address some specific, nuanced use-cases like this one to help clarify to users what output to expect


ggwicz:
Setting ticket to `hold` while I flesh out some notes explaining this change on the PR.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [Timezones.php:45](https://github.com/moderntribe/the-events-calendar/blob/db8dba37efc8a1b4b7904350b724860574dcf648/src/Tribe/Timezones.php#L45)
:x: | Whitespace found at end of line | [Timezones.php:46](https://github.com/moderntribe/the-events-calendar/blob/db8dba37efc8a1b4b7904350b724860574dcf648/src/Tribe/Timezones.php#L46)
:x: | Whitespace found at end of line | [Timezones.php:53](https://github.com/moderntribe/the-events-calendar/blob/db8dba37efc8a1b4b7904350b724860574dcf648/src/Tribe/Timezones.php#L53)

ggwicz:
Alright—I've fleshed out the notes on this PR, so it's ready for code review. Removing the `hold` flag.

barryhughes:
@ggwicz - not sure if there are accompanying changes to this but by itself/as it currently stands, I'm not sure we should merge this in.

1) You're forcing a new behaviour that won't align with the configured timezone settings. I like the drift of where you're going with this change, but I think that makes for a confusing UX and I don't think we should include this without corresponding changes in *Events ‣ Settings ‣ General ‣ Timezone Settings* (else, what users configure there isn't going to match the reality of what you've done in the code)

2) If I'm reading this correctly, this change only impacts the timezone abbreviation - it doesn't stop the sitewide timezone from being enabled more generally. Suppose someone creates an event taking place at 0800 Eastern - this change means we'll potentially render the time on the frontend as 0500 Eastern (0500 being the same time on the Pacific coast, Eastern being the local timezone you are enforcing)

Am I misinterpreting this? If I'm not, this seems like it has the potential to really confuse our users.

zbtirrell:
due to the potential implications of a change like this, let's be sure to QA before merging

ggwicz:
Hey @barryhughes, you wrote this:
> Suppose someone creates an event taking place at 0800 Eastern - this change means we'll potentially render the time on the frontend as 0500 Eastern (0500 being the same time on the Pacific coast, Eastern being the local timezone you are enforcing)

To my knowledge, this is not right—can you elaborate or provide examples of this? My whole thinking on this issue is that I do not see our plugin adjusting the actual clock time—the numeric value of the hours/minutes. It only adjusts the timezone abbreviation.

barryhughes:
> To my knowledge, this is not right—can you elaborate or provide examples of this?

For sure. Using your branch:

* Set your WP timezone to America/Vancouver
* Create an event taking place at 0800 America/New York
* Configure TEC to use a sitewide timezone
* Configure TEC to append the timezone abbreviation
* Now view the single event you created on the frontend

barryhughes:
@ggwicz - as I review further, I'm also unsure how this solves the core problem described in #97344.

I'm reading the problem as, an event taking place in July - but viewed during the winter months - will show as taking place at 8am CET instead of 8am CEST and the converse - an event taking place in December but viewed during the summer months will display as taking place at 8am CEST instead of 8am CET.

ggwicz:
Thanks all for the feedback and for reading, thinking, etc.—Barry and I found a simpler and less-invasive/opinionated/bad approach for fixing this, which you can now see over on this Common PR: https://github.com/moderntribe/tribe-common/pull/630

I will be closing _this_ PR for now.

PR #1835, opened 2018-02-23
Update cache key for month template

mitogh:
**Ticket** https://central.tri.be/issues/99366

Use variables from the request to construct the query and then create a
s hash that is smaller than a serializable value from the array and the
variables from the request.


src/Tribe/Template/Month.php
```diff
@@ -186,7 +186,13 @@ public function __construct( $args = array() ) {
 			// Cache the result of month/content.php
 			if ( $this->use_cache ) {
 				$this->cache_expiration = apply_filters( 'tribe_events_month_view_transient_expiration', HOUR_IN_SECONDS );
-				$this->html_cache = new Tribe__Template_Part_Cache( 'month/content.php', serialize( $this->args ), $this->cache_expiration, 'save_post' );
+
+				$cache_id = serialize( $this->args );
+				if ( isset( $_REQUEST ) && is_array( $_REQUEST ) ) {
```
bordoni: Shouldn't we add those two?

mitogh: amm, not sure what you mean here would you mind clarify a bit more, thanks.

mitogh: oh! I think what it's 💃

bordoni: I mean the Args and Request

mitogh: 👍

PR #1839, opened 2018-02-27
Refactor pagination on list view to be extendable

mitogh:
**Ticket** https://central.tri.be/issues/97772

Make sure the logic around pagination is extendable so we can reuse in
other locations or by other plugins.


mitogh:
Code is reused over here: https://github.com/moderntribe/events-pro/pull/800

PR #1845, opened 2018-02-28
Use timezones to compare if an event is on the past

mitogh:
**Ticket**  https://central.tri.be/issues/91168

Make sure to use the Event or global event time zone to compare if an
event is on the past instead of relying all the time on global WP
settings.


mitogh:
Closing as includes `common` modules files.

PR #1851, opened 2018-03-01
Make sure data is always an object

mitogh:
**Ticket** https://central.tri.be/issues/99089

Make sure any string that looks like an object should be parsed as such
in order to access the properties of the object.


src/resources/js/tribe-events-ajax-calendar.js
```diff
@@ -125,9 +125,18 @@
 
 						if ( $this.tribe_has_attr( 'data-tribejson' ) ) {
 
-							var data = $this.data( 'tribejson' );
+							var data = $this.data( 'tribejson' )
```
barryhughes: Still ought to have the terminating `;` ... even though it's not strictly needed, that's our style/is what our coding standards expect

src/resources/js/tribe-events-ajax-calendar.js
```diff
@@ -125,9 +125,18 @@
...
+							if ( 'string' === typeof data ) {
+								try {
+									data = JSON.parse( data )
```
barryhughes: As above - throughout this change we should use semicolons to terminate statements

mitogh:
Changes has been applied ready for another set of eyes 👁

PR #1853, opened 2018-03-02
97978 Get Events-Homepage Selection to Persist in Admin and Customizer

ggwicz:
**Ticket:** [**#97978**](http://central.tri.be/issues/97978)

**Code Review Note:** A readme entry already exists for this ticket; this is QA follow-up dev.


src/Tribe/Customizer/Front_Page_View.php
```diff
@@ -45,14 +45,27 @@ public function setup() {
 	 */
 	public function add_events_page_option( $output, $r, $pages ) {
```
bordoni: Can we use a better variable name for `$r`?

ggwicz: that's the name of the var in the core function so i just kept it consistent, which i think is the right move -- thoughts?

bordoni: Change it to something more read-able if we can.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $pages is never used | [Front_Page_View.php:46](https://github.com/moderntribe/the-events-calendar/blob/13b64e8e00c533bc051206261188993c8211b266/src/Tribe/Customizer/Front_Page_View.php#L46)

PR #1868, opened 2018-03-15
Fix empty string used as event start date when generating adjacent links

jeremyfelt:
**Ticket:** [**#101757**](http://central.tri.be/issues/101757)

---
When building the query used to lookup adjacent links, it appears that an empty string is always used for the `_EventStartDate` value and unexpected events are used in the previous/next links
on single event pages.

When `$post_obj->EventStartDate` is called, post meta is checked for a meta key that does not exist. Adjusting this to the correct meta key, `$post_obj->_EventStartDate` provides a correct date and results in properly generated links.

The README mentions a `develop` branch, but I don't see on on this repo. I took a guess that this is the development branch for the upcoming release. Let me know if I should submit the PR against something else.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -311,6 +311,7 @@ The plugin is made with love by [Modern Tribe Inc](http://m.tri.be/2s).
 = [4.6.13] TBD =
 
 * Tweak - Added Gutenberg compatibility for the Extension
+* Fix - Prevent Adjacent Events links failing due to wrong meta key been used (props to @jeremyfelt for fixing this) [101757]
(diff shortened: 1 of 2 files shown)


bordoni:
Hi @jeremyfelt,
I've just updated out `readme.md` with the correct branch, thanks a lot for the warning on that and for your Pull Request.

You took the correct guess, the latest `release/MYY.XX` is the next Release that will be out, but you can always PR it to `master` and we will change the base branch to the correct one.

PR #1877, opened 2018-03-22
Purge expired Events

nefeline:
Ref. https://central.tri.be/issues/98826

### DIFF
--- a/src/Tribe/Event_Cleaner_Scheduler.php
+++ b/src/Tribe/Event_Cleaner_Scheduler.php
@@ -148,9 +148,9 @@ class Tribe__Events__Event_Cleaner_Scheduler {
 		$event_post_type = Tribe__Events__Main::POSTTYPE;
 
 		$posts_with_parents_sql = "SELECT DISTINCT post_parent
-				FROM {$wpdb->posts}
-				WHERE post_type= '$event_post_type'
-					AND post_parent <> 0
+		FROM {$wpdb->posts}
+		WHERE post_type= '$event_post_type'
+			AND post_parent <> 0
 		";
 
 		$sql = "SELECT post_id
--- a/src/Tribe/Event_Cleaner_Scheduler.php
+++ b/src/Tribe/Event_Cleaner_Scheduler.php
@@ -1,13 +1,13 @@
 <?php
 
 /**
- * Class Event_Scheduler
+ * Class Event_Cleaner_Scheduler
  *
  * Uses cron to move old events to trash and/or permanently delete them.
  *
  * @since TBD
  */
-class Tribe__Events__Event_Scheduler {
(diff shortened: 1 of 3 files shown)


src/Tribe/Event_Cleaner_Scheduler.php
```diff
@@ -150,7 +150,14 @@ public function select_events_to_purge( $month ) {
 		INNER JOIN {$wpdb->postmeta} AS t2 ON t1.ID = t2.post_id
 		WHERE t1.post_type = %d
 			AND t2.meta_key = '_EventEndDate'
-			AND t2.meta_value <= DATE_SUB( CURDATE(), INTERVAL %d MONTH )";
+			AND t2.meta_value <= DATE_SUB( CURDATE(), INTERVAL %d MONTH )
+			AND t1.post_parent = 0
+			AND t1.ID NOT IN (
+				SELECT DISTINCT post_parent
+				FROM wp_posts
+				WHERE post_type= 'tribe_events'
```
bordoni: Shouldn't we use a variable for the Post Type here too?

src/Tribe/Event_Cleaner_Scheduler.php
```diff
@@ -150,7 +150,14 @@ public function select_events_to_purge( $month ) {
 		INNER JOIN {$wpdb->postmeta} AS t2 ON t1.ID = t2.post_id
 		WHERE t1.post_type = %d
 			AND t2.meta_key = '_EventEndDate'
-			AND t2.meta_value <= DATE_SUB( CURDATE(), INTERVAL %d MONTH )";
+			AND t2.meta_value <= DATE_SUB( CURDATE(), INTERVAL %d MONTH )
+			AND t1.post_parent = 0
+			AND t1.ID NOT IN (
+				SELECT DISTINCT post_parent
```
bordoni: Maybe it would be worth saving this on a another SQL variable to make is more readable?
`AND t1.ID NOT IN ( $posts_with_parents_sql )`

src/Tribe/Event_Cleaner_Scheduler.php
```diff
@@ -1,13 +1,13 @@
...
 /**
- * Class Event_Scheduler
+ * Class Event_Cleaner_Scheduler
  *
  * Uses cron to move old events to trash and/or permanently delete them.
  *
  * @since TBD
  */
-class Tribe__Events__Event_Scheduler {
+class Tribe__Events__Event_Cleaner_Scheduler {
```
bordoni: Why the change in names?

barryhughes: I asked for this to make it less ambiguous (since the class exists specifically to deal with cleanup tasks whereas event scheduler is less explicit and could imply it handles publishing events on a schedule or something of that order).

PR #1879, opened 2018-03-25
Changed text string for consistency

andrasguseo:
Changed text string to match the one in "src/admin-views/create-organizer-fields.php".


bordoni:
@andrasguseo when you create these new Pull Requests don't forget to give the Branch a descriptive name; `spotfix/what-you-are-fixing`

PR #1880, opened 2018-03-26
Fix warning in redirect_past_upcoming_view_urls()

barryceelen:
The `Tribe::redirect_past_upcoming_view_urls()` method incorrectly calls `trim()` on the result of `wp_parse_url()`, which is an array or `false`.

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -2416,7 +2416,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 			}
 
 			// Prep strings so we can compare paths sans any leading/trailing slashes
-			$requested_path = trim( $requested_path, '/' );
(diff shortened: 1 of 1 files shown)


neillmcshea:
_Ref:_ [C#110463](http://central.tri.be/issues/110463)

Thanks @barryceelen for your submission! Sorry for the long wait we are working on a better process for our community PR's - we have this one added to our workflow and will let you know if we have any further questions.

jesseeproductions:
Thanks for the fix! I have merged it in for our 4.6.22 release. Thanks!

jesseeproductions:
Hi @barryceelen Unfortunately, after some more testing I found that I have to revert the fix.

You can see why here:
https://github.com/moderntribe/the-events-calendar/pull/2110

I am not sure why it is not working for you. Let me know if you have steps to reproduce and we can see about fixing it, thanks!

PR #1884, opened 2018-03-26
Fix queue processing reports

lucatume:
Ticket: #102323

This PR changes two parts of our AJAX powered queue real-time processing to improve it:
1. we would not display reports if the progress is not at least `1%`; this is a problem for huge files where it might take minutes to get to that `1%`; for a long time the user would have no visible feedback something is being processed (the file in the issues has ~10k rows)
2. we would update the `updated` and `skipped` numbers from the queue report; I've changed this to only update these numbers if the new one is gt the current one. Due to parrallel procession one queue that has an updated count of `4` might start before, and end after, another that has an update count of `5`; the use would see the number change from `3` -> `5` -> `4` and that's incosistent.
In the ticket file this glitch would happen often as the processing would spin up a number of parallel processes.


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *1.0.18*, but this was pull requested against *release/M18.06*.

PR #1886, opened 2018-03-27
Rework EA Scheduled imports re-scheduling policy

lucatume:
Ticket: #102489

This PR changes an important detail in how we reschedule failed scheduled imports:
* what we currently do is that, whenever a scheduled import fails for any reason, we reschedule it immediately; this approach has the problem that, if a scheduled import failed for a reason that is not resolved in 15' (when it will try to run again) then it will consume an import slot... and fail again. This would drain the daily imports.
* the change makes it so that we are still re-scheduling upon failure but try to do so less frequently using the scheduled frequency as a guideline; see the code for details. All of it filterable.

This PR also adds some more information in the client produced log.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Usage of ELSE IF is discouraged; use ELSEIF instead | [Cron.php:361](https://github.com/moderntribe/the-events-calendar/blob/514b6945d4a0becb609ffb76b0e8b2a19fbfacf0/src/Tribe/Aggregator/Cron.php#L361)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Usage of ELSE IF is discouraged; use ELSEIF instead | [Cron.php:361](https://github.com/moderntribe/the-events-calendar/blob/3348962cad008f78c0ee4f59fb431d57a179fc81/src/Tribe/Aggregator/Cron.php#L361)

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -2386,4 +2390,114 @@ public function import_event_image( $event, $activity ) {
...
+		return $last_child_post && $last_child_post instanceof WP_Post
+			? Tribe__Events__Aggregator__Records::instance()->get_by_post_id( $last_child_post->ID )
+			: $this;
+	}
+
+	/**
+	 * Returns this record last child post object.
+	 *
+	 * @since TBD
+	 *
```
bordoni: Add the $force param.

lucatume:
Doc, updated: good for another look.

PR #1887, opened 2018-03-27
Expand the size of the TimeZone input.

mitogh:
Ticket: https://central.tri.be/issues/100363


bordoni:
Would you add GIF and screenshots of how this should behave? With Diff size of screens on the Ticket?

mitogh:
For sure.

Adding those into the QA instructions. But just a summary over here.

By setting the `min-width` to 100px is going to behave exactly the same on smaller screens as is going to prevent to going on any lower size than the original value.

And on any larger screen is limited to 198px so as long as there some space the input is going to expand until `198px` and on smaller screen is going to contract until `100px`.

Adding the visual references on central once is on QA status.

PR #1889, opened 2018-03-28
Use a numeric ID to identify the Main Events Page instead

mitogh:
**Ticket**: https://central.tri.be/issues/100832

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -313,7 +313,7 @@ The plugin is made with love by [Modern Tribe Inc](http://m.tri.be/2s).
 * Fix - Prevent warnings on retrival of posts if "Include events in main blog loop" is checked (thanks to Colin Carmichael for report this problem) [97667]
 * Fix - Make the `tribe-ea-record` custom post type (used to store Event Aggregator record information) private [99106]
 * Fix - Expand the size of the TimeZone input on the admin to allow better visibility for long names [100363]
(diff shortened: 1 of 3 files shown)


bordoni:
Do we want to migrate users that save this in the past?

readme.txt
```diff
@@ -313,6 +313,7 @@ The plugin is made with love by [Modern Tribe Inc](http://m.tri.be/2s).
 * Fix - Prevent warnings on retrival of posts if "Include events in main blog loop" is checked (thanks to Colin Carmichael for report this problem) [97667]
 * Fix - Make the `tribe-ea-record` custom post type (used to store Event Aggregator record information) private [99106]
 * Fix - Expand the size of the TimeZone input on the admin to allow better visibility for long names [100363]
+* Fix - If Main Events page is set to be used on the front page the selection is now reflected on the customizer [100832]
```
ggwicz: I'd recommend tweaking this entry a bit, like as follows:
```
* Fix - If the main events page is set to be the site's front page, ensure it shows as "selected" in the Customizer's front page option [100832]
```

mitogh:
In terms of the migration is not required unless the user wants to see the "selected" attribute applied. 

But once this is saved changes are going to be saved, if this never happens it still work as expected on the Admin as the current mechanism for the `checked` attribute on WP is will fallback on the last available option if `post` is not checked which in this case is `page`.

PR #1890, opened 2018-03-28
Display a failed Scheduled record retry time

lucatume:
Ticket: https://central.tri.be/issues/102489

This PR adds a method to the Record class to return the timestamp of the record next retry (modified in PR https://github.com/moderntribe/the-events-calendar/pull/1886) to display it, for failed Scheduled records only, in the Scheduled tab: https://cl.ly/0G0R2M3c2L1g

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -2515,7 +2515,7 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 	 *
 	 * @return int
 	 */
-	protected function get_retry_interval() {
+	public function get_retry_interval() {
 		if ( $this->frequency->interval === DAY_IN_SECONDS ) {
 			$retry_interval = 6 * HOUR_IN_SECONDS;
 		} elseif ( $this->frequency->interval < DAY_IN_SECONDS ) {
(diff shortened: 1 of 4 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -2530,10 +2530,47 @@ protected function get_retry_interval() {
...
+
+		if ( empty( $retry_interval ) ) {
+			return false;
+		}
+
+		if ( ! $this->get_last_import_status( 'error', true ) ) {
+			return false;
+		}
+
+		$last_attempt_time = strtotime( $this->last_child()->post->post_modified_gmt );
```
bordoni: Should we do some validation on the date? It might be invalid.

lucatume: The code is this:
```php
$last_attempt_time = strtotime( $this->last_child()->post->post_modified_gmt );
$retry_time = $last_attempt_time + (int) $retry_interval;

if ( $retry_time < time() ) {
    $retry_time = false;
}
```

So if the `strtotime` function cannot parse the `post_modified_gmt` it will return `false`.
Then `retry_time` will be `0 + something`.
Since we're not at the beginning of times `retry_time` will be `< time()` and `retry_time` will be then set to `false`

bordoni: perfecto!

PR #1897, opened 2018-03-29
Add tribe_events_event_insert_args and tribe_events_event_update_args hooks

sc0ttkclark:
https://central.tri.be/issues/102545


lucatume:
I've fixed the issues that were preventing this from working in Common and modified the Venue and Organizer list validation.
The part I'm not fully understanding is the one where we look for and `OrganizerID` or `VenueID` array in the list specifically: that seems too specific a solution to be in TEC and that would require updates of the hardcoded keys whenever needed; I've changed that to deal with similar setups extracting values.

PR #1903, opened 2018-04-02
Adjusted setting label to match

andrasguseo:
Adjusted the label of the setting to match the one on line 159, as it is the same type of setting.


mitogh:
Maybe delete the branch as well @andrasguseo ?

PR #1907, opened 2018-04-03
Change HTTP return code of empty archives REST API requests to 200

lucatume:
Ticket: https://central.tri.be/issues/102283

This PR modifies the HTTP status code TEC REST API will return on empty archives to `200` (OK) and will make it return an empty array.
Scott correctly noted the WordPress REST API would not return `404` (as we currently do) when queries or filters would yield an empty archive page for events, venues, organizers, categories or tags. This PR fixes this misalignment.


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for **, but this was pull requested against *release/M18.06*.

PR #1915, opened 2018-04-05
Fix/42146 Remove strings from translation file

nefeline:
Ref.: https://central.tri.be/issues/42146


bordoni:
Based on conversation on the Ticket, we are not tackling this issue.

PR #1918, opened 2018-04-09
Change the check to avoid PHP notices

lucatume:
Ticket: https://central.tri.be/issues/99106#note-12

Since the `last_import_error` variable might not be defined use `empty` to check it.

### DIFF
--- a/src/Tribe/Aggregator/Record/List_Table.php
+++ b/src/Tribe/Aggregator/Record/List_Table.php
@@ -542,8 +542,9 @@ class Tribe__Events__Aggregator__Record__List_Table extends WP_List_Table {
 		$now = current_time( 'timestamp', true );
 
 		$retry_time = false;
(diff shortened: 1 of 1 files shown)


src/Tribe/Aggregator/Record/List_Table.php
```diff
@@ -542,8 +542,9 @@ public function column_imported( $post ) {
 		$now = current_time( 'timestamp', true );
 
 		$retry_time = false;
-		if ( $last_import_error ) {
-		    $retry_time = $record->get_retry_time();
+
+		if ( ! empty( $last_import_error ) ) {
```
zbtirrell: should it be changed on 526 as well?

lucatume: on line 526 we just set it, line 524, so it will be defined

PR #1924, opened 2018-04-12
#103823 - replace stored eb token with check to EA

jesseeproductions:
_Ref:_ [C#103823](https://central.tri.be/issues/103823)

### DIFF
--- a/src/Tribe/Aggregator/Service.php
+++ b/src/Tribe/Aggregator/Service.php
@@ -344,6 +344,8 @@ class Tribe__Events__Aggregator__Service {
 	/**
 	 * Fetch Eventbrite Extended Token from the Service
 	 *
+	 * @since TBD
+	 *
 	 *  @return stdClass|WP_Error
 	 */
 	public function has_eventbrite_authorized() {
--- a/src/Tribe/Aggregator/Service.php
+++ b/src/Tribe/Aggregator/Service.php
@@ -321,9 +321,10 @@ class Tribe__Events__Aggregator__Service {
 	 *
 	 * @return array
 	 */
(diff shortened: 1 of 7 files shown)


src/Tribe/Aggregator/Service.php
```diff
@@ -335,7 +336,7 @@ public function get_eventbrite_token() {
 		 */
 		$args = apply_filters( 'tribe_aggregator_eventbrite_token_callback_args', $args );
 
-		$response = $this->get( 'eventbrite/token', $args );
+		$response = $this->get( 'eventbrite/validate', $args );
```
jesseeproductions: It looks like this should be the endpoint to check if eventbrite has a token for the site, but I only get this message:

[12-Apr-2018 21:09:18 UTC] stdClass Object
(
    [status] => error
    [message_code] => error:eventbrite-token-not-found
    [message] => You do not have an active connection to Eventbrite through your account and Event Aggregator.
)

Route:
https://github.com/moderntribe/event-aggregator-site/blob/d3b1449144f9b1b6c6d1e6d23b4e890d81e49178/wp-content/plugins/event-aggregator/src/Tribe/Route/Eventbrite.php#L16

sc0ttkclark: Pass the license key as `key` and `url` (you have url param set correctly).

If you plan on supporting the TEC Eventbrite license key for this endpoint, you'll want to also pass in the `'origin' => 'eventbrite'`

sc0ttkclark: Sent you a DM in slack with the URLs that are working.

PR #1934, opened 2018-04-19
fix double display of the date/time section in the event metabox

lucatume:
Ticket: https://central.tri.be/issues/102196

Just a small fix to avoid the "Time and Date" tab displaying twice on the Event edit screen.

Question for @bordoni: we have used the template to render it in two places (see code), which one of the two should we keep?

### DIFF
--- a/tests/restv1/TagDeletionCest.php
+++ b/tests/restv1/TagDeletionCest.php
@@ -50,17 +50,17 @@ class TagDeletionCest extends BaseRestCest {
 	}
 
 	/**
-	 * It should return 401 if user cannot delete event tag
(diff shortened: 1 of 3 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task &quot;here we need to decide if we want to load the whole form below or section by section&quot; | [Event_Meta_Box.php:50](https://github.com/moderntribe/the-events-calendar/blob/adad2e7954659f2f44dfd991872140e05d81f8e7/src/Tribe/Admin/Event_Meta_Box.php#L50)

PR #1937, opened 2018-04-20
EA Eventbrite API updates

sc0ttkclark:
https://central.tri.be/issues/103823

Fixes import preview issues and problems from changing license keys around after token authorization flow.


jesseeproductions:
merging to release to QA

PR #1939, opened 2018-04-24
Set site param for import create on eventbrite origins

sc0ttkclark:
https://central.tri.be/issues/35454

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -730,6 +730,11 @@ abstract class Tribe__Events__Aggregator__Record__Abstract {
 				: date( Tribe__Date_Utils::DBDATETIMEFORMAT, $args['end'] );
 		}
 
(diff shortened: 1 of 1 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -730,6 +730,11 @@ public function queue_import( $args = array() ) {
 				: date( Tribe__Date_Utils::DBDATETIMEFORMAT, $args['end'] );
 		}
 
+		// Set site for origin(s) that need it for new token handling.
+		if ( 'eventbrite' === $args['origin'] ) {
+			$args['site'] = site_url();
```
bordoni: Could we fetch that from the Callback URL? or Other data that we might have previously?

sc0ttkclark: Previews don’t send callback. This just standardizes a param to send either way. I don’t want to send it for non EB token requests because it’d start adding lots of sites to the DB unnecessarily.

PR #1941, opened 2018-04-24
Fix: `[tribe_events]` month view pagination without the bar [95720]

juanfra:
https://central.tri.be/issues/95720

Add `tribe-bar-date` URL parameter when using the shortcode.

Tested with ugly and pretty permalinks, with and without the bar.


src/resources/js/tribe-events.js
```diff
@@ -1408,6 +1408,15 @@ Date.prototype.format = function( mask, utc ) {
 					}
 				}
 			}
+
+			if ( ( ts.view === 'month' ) && ( ! $( '#tribe-events-bar' ).length ) ) {
```
bordoni: Yoda first conditional, and remove extra parenthesis on each one of the conditionals

PR #1955, opened 2018-05-01
Honor the original return type instead of returning `null`

mitogh:


src/functions/template-tags/loop.php
```diff
@@ -256,12 +256,13 @@ function tribe_get_upcoming_link() {
 	 * @return bool
 	 */
 	function tribe_has_previous_event() {
+		$wp_query = tribe_get_global_query_object();
+		$has_previous = false;
 
-		if ( ! $wp_query = tribe_get_global_query_object() ) {
-			return;
+		if ( null === $wp_query ) {
+			return apply_filters( 'tribe_has_previous_event', $has_previous );
```
bordoni: Docblocks

src/functions/template-tags/loop.php
```diff
@@ -303,11 +303,13 @@ function tribe_has_previous_event() {
 	 */
 	function tribe_has_next_event() {
 
-		if ( ! $wp_query = tribe_get_global_query_object() ) {
-			return;
+		$wp_query = tribe_get_global_query_object();
+		$has_next  = false;
+
+		if ( null === $wp_query ) {
+			return apply_filters( 'tribe_has_next_event', $has_next );
```
bordoni: Docblocks

src/functions/template-tags/loop.php
mitogh: This is not part of a new filter but is the same filter on the function just in case the where `wp_query` is null the original filter is applied.

Same case for the other one let me know if you think it requires a doccblock as well.

mitogh:
`readme` was avoided as there's no ticket for this one.

Let me know about this @bordoni 

Thank you

bordoni:
We should have the Changelog event with ticket.

mitogh:
Thanks just one more question.

mitogh:
Would you mind take a look at the questions above @bordoni thanks.

bordoni:
We are good for changes, we might need to revise later how to docblock filters in two places

PR #1965, opened 2018-05-05
⌗106324 Adding Some Tests for Event Cost Formatting

ggwicz:
**Ticket:** [**⌗106324**](http://central.tri.be/issues/106324)

**Code Review Notes:**

- While the "milestone" for this PR is "Next Maintenance Release", we're late enough in the current MR's development cycle where [Gustavo agreed that I should just commit this code to `master`](https://tribe.slack.com/archives/C3XGB48QK/p1525369229000680) instead of possibly tying up the release branch more than needed.


tests/integration/Unticketed_Cost_Formatting_Test.php
```diff
@@ -0,0 +1,206 @@
...
+			'3.00 8.00 125.95', // Represents a range of values.
+			'Free',             // The cost field should not manipulate this, and just output it literally.
+			'*&^$@#%@',         // Ditto: Should not be manipulated, just output literally.
+			'東京は都会です',     // Ditto: Should not be manipulated, just output literally.
+			'1995.95',
+		];
+
+		$iterations = 0;
+
+		foreach ( $costs as $event_cost ) {
```
lucatume: Why not extend the `Events_TestCase` and use its very convenient factory? All this code would collapse into one line.
`$this->factory()->event->create(['meta_input' => ['_EventCost' => 5]]` and so on.

lucatume: Plus: not a fan of too smart setup methods when the setup can be shown in the test methods: ideally a test method contains Arrange, Act, Assert. Whenever you move one of the steps out of the test method it loses clarity.

ggwicz:
Going to implement some feedback from Luca here.

ggwicz:
I've implemented DataProvider per luca's feedback over Slack; this one's ready for another review! 🔍

PR #1969, opened 2018-05-07
Add some checks to see if the `$url` is the events page with pretty p…

juanfra:
https://central.tri.be/issues/94328

Add some checks to see if the `$url` is the events page with pretty permalinks params (starts with the events URL)

### DIFF
--- a/src/Tribe/Rewrite.php
+++ b/src/Tribe/Rewrite.php
@@ -412,6 +412,8 @@ class Tribe__Events__Rewrite extends Tribe__Rewrite {
 	 */
 	public function filter_url_to_postid( $url ) {
 
+		$events_url = Tribe__Events__Main::instance()->getLink();
+
 		// check if the site is using pretty permalinks
 		if ( '' !== get_option( 'permalink_structure' ) ) {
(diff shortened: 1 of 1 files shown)


src/Tribe/Rewrite.php
```diff
@@ -412,18 +412,25 @@ protected function add_hooks() {
...
 		// check if the site is using pretty permalinks
 		if ( '' !== get_option( 'permalink_structure' ) ) {
 			$url_query = @parse_url( $url, PHP_URL_QUERY );
 
 			// Remove the "args" in case we receive any
 			if ( ! empty( $url_query ) ) {
 				$url = str_replace( '?' . $url_query, '', $url );
+			} else {
+				// Check if they're viewing the events page with pretty params
+				if ( substr( $url, 0, strlen( $events_url ) ) === $events_url ) {
```
Camwyn: Could we just do `0 === strpos( $url, $events_url )`? (or `stripos` to be case-insensitive)

juanfra: Yes, we can do that. It's kind of the same.

Camwyn:
🍓

PR #1986, opened 2018-05-21
Add 2018 survey message/link | #106817

barryhughes:
Adds a dismissable, time-bound admin notice that will appear on event admin screens up until 2018-06-08. 

:ticket: [#106817](https://central.tri.be/issues/106817)

### DIFF
--- a/src/Tribe/Admin/Notice/Marketing.php
+++ b/src/Tribe/Admin/Notice/Marketing.php
@@ -1,0 +1,53 @@
+<?php
+/**
+ * @internal This class may be removed or changed without notice
+ */
+class Tribe__Events__Admin__Notice__Marketing {
+	/**
+	 * Register marketing notices.
+	 *
+	 * @since TBD
+	 */
+	public function hook() {
+		tribe_notice(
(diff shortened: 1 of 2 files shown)


src/Tribe/Admin/Notice/Marketing.php
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	/**
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function should_display() {
+		return tribe( 'admin.helpers' )->is_screen()
+			&& date_create()->format( 'Y-m-d' ) < '2018-06-08';
```
bordoni: Are you sure this is PHP 5.2 compat?

barryhughes: Yep, should be good: `date_create()` was introduced in 5.2.0 and `DateTime::format()` was introduced in 5.2.1, so within bounds (min supported version for us being 5.2.4).

bordoni: I meant the usage o `date_create` without variable assign, into a `->` usage.

PR #1987, opened 2018-05-22
Manage circular dependencies by requeuing items a limited amount of times

lucatume:
Ticket:  https://central.tri.be/issues/106772

This PR resolves an issue with circular dependencies while trying to import events.
Given data for 3 events (A, B and C) and given that A depends on a Venue/Organizer defined in C, B depends on an Venue/Organizer defined in A and C depends on a Venue/Organizer defined in A then a deadlock would take place requeing the items needlessly.

The reason we check dependencies is to avoid critical paths where two events will look for the same Venue/Organizer post, not find it and write it resulting in two copies of the same Venue/Organizer.

By allowing requeuing to happen a limited number of times one of the deadlocked items will be written unlocking the next and so forth.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | String &quot;tribe_aggregator_import_process_requeue_limit&quot; does not require double quotes; use single quotes instead | [Import_Events.php:59](https://github.com/moderntribe/the-events-calendar/blob/2cd8c5459da950fcbf2c9585c0b1280d6dd0b410/src/Tribe/Aggregator/Processes/Import_Events.php#L59)

PR #1991, opened 2018-05-23
Marketing notice should appear on tag screen

barryhughes:
Marketing notice should also appear on the (events) tag admin screen. Previously it appeared on all our screens except for this one.

:ticket: [#106817](https://central.tri.be/issues/106817)


barryhughes:
Note: CI failures all seem to be database errors entirely unconnected with this change.

PR #2002, opened 2018-05-29
Correct the "View All" link when using the events month view and plain permalinks.

juanfra:
https://central.tri.be/issues/72544

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -2643,7 +2643,7 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 				case 'day':
 					$eventUrl = add_query_arg( array( 'tribe_event_display' => $type ), $eventUrl );
 					if ( $secondary ) {
-						$eventUrl = add_query_arg( array( 'date' => $secondary ), $eventUrl );
+						$eventUrl = add_query_arg( array( 'eventDate' => $secondary ), $eventUrl );
 					}
 					break;
 				case 'week':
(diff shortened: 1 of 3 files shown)


src/Tribe/Template/Month.php
```diff
@@ -495,10 +495,13 @@ protected function get_title( $original_title, $sep = null ) {
 		private static function view_more_link( $date ) {
 			$day_link       = tribe_get_day_link( $date );
 			$tribe_bar_args = self::get_tribe_bar_args();
+
 			if ( ! empty( $tribe_bar_args ) ) {
+				if ( isset( $tribe_bar_args['tribe_event_display'] ) ) {
```
Camwyn: No need to check if the variable is set - just unset it. It won't throw an error if it's not set already. 😉

src/Tribe/Main.php
```diff
@@ -2643,7 +2643,7 @@ public function uglyLink( $type = 'home', $secondary = false ) {
 				case 'day':
 					$eventUrl = add_query_arg( array( 'tribe_event_display' => $type ), $eventUrl );
 					if ( $secondary ) {
-						$eventUrl = add_query_arg( array( 'date' => $secondary ), $eventUrl );
+						$eventUrl = add_query_arg( array( 'eventDate' => $secondary ), $eventUrl );
```
Camwyn: Nice catch! Should we be correcting the one in the month case as well?

juanfra: Yep, I've checked and it has the same problem. It usually uses the `tribe-bar-date` though. But one thing for sure is that `date` doesn't work. Will change that for `eventDate` as well.

PR #2004, opened 2018-05-31
Check for an active date, not always today.

Camwyn:
🎫 https://central.tri.be/issues/96511

Make sure when we resize we don't change the displayed events


src/resources/js/tribe-events-ajax-calendar.js
```diff
@@ -180,16 +180,21 @@
 
 		function tribe_mobile_month_setup() {
 
-			var $today          = $wrapper.find( '.tribe-events-present' );
+			var $active_day       = $wrapper.find( '.mobile-active' );
```
bordoni: Lets use the `$activeDay` since the new Standard is camelCase for JS.

Camwyn: Yep, changed most of the others as well 😉

PR #2014, opened 2018-06-01
Keep filter bar date when changing from month to list view, with all datepicker formats

juanfra:
https://central.tri.be/issues/83018


src/resources/js/tribe-events-bar.js
```diff
@@ -294,22 +294,16 @@ var tribe_events_bar_action;
...
-				if ( ts.datepicker_format !== '0' ) {
-					if ( day.length ) {
-						dp_date = tribeDateFormat( $tribedate.bootstrapDatepicker( 'getDate' ), 'tribeMonthQuery' );
-						$tribedate.val( dp_date + day );
-					}
-					else {
-						$tribedate.val( '' );
-					}
-
+				if ( '0' !== ts.datepicker_format ) {
```
bordoni: Should we convert it to a number or compare it with `!=`? Making it less likely to fail if for some reason we fetch the variable with `.data()` in the future.

juanfra: Yeah, we can :) This was like that already, the datepicker format is coming via options. What would you prefer to do from those two options?

PR #2019, opened 2018-06-05
Add venue name to full address string

TheLastProject:
This allows iCal exports to generate some location info and aids the end-user into more easily recognising where exactly they should be when they're close to the event location. It also makes sure that there is at least some location info in the iCal export, to prevent confusion both for the WordPress webmaster (why isn't the location showing up?) as to the end-user (what is the name of the place I need to be).

My username on WordPress.org is sylviavanos.

_Ref:_ [C#110464](http://central.tri.be/issues/110464)

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -2837,7 +2837,14 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		 */
 		public function fullAddressString( $postId = null ) {
 			$address = '';
+			if ( tribe_get_venue( $postId ) ) {
+				$address .= tribe_get_venue( $postId );
+			}
+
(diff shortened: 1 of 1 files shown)


neillmcshea:
_Ref:_ [C#110464](http://central.tri.be/issues/110464)

@TheLastProject - thank you for your first PR to the events calendar! We appreciate the time you took to help make the plugin even better. We have this now added to our workflow and we'll let you know if we have any questions.

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M18.12*, but this was pull requested against *release/M18.11*.

jesseeproductions:
Thanks @TheLastProject I have merged the coding for release with 4.6.22

PR #2026, opened 2018-06-06
Display the correct Import Settings when EB is enabled

nefeline:
Ref. https://central.tri.be/issues/106947

Eventbrite tickets (4.5) is enabling EA by adding a dummy/fake PUE key. This functionality invalidates our original check for the import settings because Tribe__Events__Aggregator::is_service_active() is always returning true (with TEC + EB enabled), even if EA is disabled. To fix this, added an additional validation to check if the pue_install_key_event_aggregator is empty, and if it is, don't display the EA fields.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | empty() and isset() must only contain variables (PHP 5.2 compatibility) | [settings.php:115](https://github.com/moderntribe/the-events-calendar/blob/2ee98b67a688b35b213294f9bea68e73c2b80b2d/src/admin-views/aggregator/settings.php#L115)

src/admin-views/aggregator/settings.php
```diff
@@ -112,7 +112,7 @@
 
 $global = $ical = $ics = $facebook = $gcal = $meetup = $url = array();
 // if there's an Event Aggregator license key, add the Global settings, Facebook, iCal, and Meetup fields
-if ( Tribe__Events__Aggregator::is_service_active() ) {
+if ( Tribe__Events__Aggregator::is_service_active() && ! empty( get_option( 'pue_install_key_event_aggregator' ) ) ) {
```
bordoni: We need to setup a method to get the _PUE License_ due to been a a little bit more complex than this.
See this code for why is that:
https://github.com/moderntribe/the-events-calendar/blob/master/src/Tribe/Aggregator/Service.php#L82-L86

Side note, avoid `empty()` directly to a method or function, it can only be called to variables.

src/Tribe/Aggregator.php
```diff
@@ -298,6 +298,27 @@ public static function is_service_active() {
...
 	}
 
+	/**
+	 * Verifies if user has a license key
+	 *
+	 * @return boolean
+	 *
+	 * @since TBD
+	 */
+	public static function has_license_key() {
```
bordoni: Lets avoid a Static , lets go with a normal method

PR #2029, opened 2018-06-06
Check the Events Aggregator license each time the EA page is accessed

juanfra:
https://central.tri.be/issues/67864

Add this check, so users with expired licenses cannot add new "premium" imports.


juanfra:
Absolutely, I thought of changing it to something like `check_license` or something, but I ended up going for the same name as PUE has. I'll go with your suggestion, I think it's better! Thanks @Camwyn! 🎉

PR #2035, opened 2018-06-07
Add `tribe_events_month_has_events_filtered` function and filter

juanfra:
🎫 https://central.tri.be/issues/67607

Not happy with the results. Been trying many different things, and this one seemed to be to most cost-efficient one. Not sure if you had something else in mind, @barryhughes?

### DIFF
--- a/src/Tribe/Template/Month.php
+++ b/src/Tribe/Template/Month.php
@@ -1243,40 +1243,41 @@ if ( ! class_exists( 'Tribe__Events__Template__Month' ) ) {
 		 */
 		public function has_events_filtered() {
 
-			if ( 'month' === Tribe__Events__Main::instance()->displaying ) {
-
-				if ( ! $wp_query = tribe_get_global_query_object() ) {
-					return;
-				}
-
-				// Get the date from the main query
-				$event_date = $wp_query->get( 'eventDate' );
-
-				// If we don't have the date, get the current date
-				$month = empty( $event_date )
-					? tribe_get_month_view_date()
-					: $wp_query->get( 'eventDate' );
-
-				// prepare the args for this month
-				$args = array(
-					'eventDisplay'   => 'custom',
-					'start_date'     => self::calculate_first_cell_date( $month ),
-					'end_date'       => self::calculate_final_cell_date( $month ),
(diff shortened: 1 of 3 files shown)


src/Tribe/Template/Month.php
```diff
@@ -1234,5 +1234,50 @@ public function has_events() {
...
+		/**
+		 * Check if the month has events when all the filters have been applied.
+		 *
+		 * @since TBD
+		 *
+		 * @return bool
+		 */
+		public function has_events_filtered() {
+
+			if ( 'month' === Tribe__Events__Main::instance()->displaying ) {
```
bordoni: Lets invert this logic, to `return false;` if `!==` than `month`, avoids indentation.

src/Tribe/Template/Month.php
```diff
@@ -1234,5 +1234,50 @@ public function has_events() {
...
+
+				if ( ! $wp_query = tribe_get_global_query_object() ) {
+					return;
```
bordoni: let's return `false` here.

src/functions/template-tags/month.php
```diff
@@ -195,6 +196,17 @@ function tribe_events_month_has_events() {
 		return apply_filters( 'tribe_events_month_has_events', false );
 	}
 
+	/**
+	 * Returns whether there are any events in the month,
+	 * with the filtered results.
+	 *
```
bordoni: Add a `@since`

src/functions/template-tags/month.php
```diff
@@ -195,6 +196,17 @@ function tribe_events_month_has_events() {
...
+	 * @return bool
+	 * @see Tribe__Events__Template__Month::has_events_filtered()
+	 **/
+	function tribe_events_month_has_events_filtered() {
+		return apply_filters( 'tribe_events_month_has_events_filtered', Tribe__Events__Template__Month::has_events_filtered() );
```
bordoni: Lets add a docblock.

src/functions/template-tags/month.php
juanfra: This function was already there. I will add the since of the version where it was introduced.

PR #2038, opened 2018-06-08
Fix: to drop the order from the query we need to unset orderby, not order.

dgwatkins:
I double checked this change works by looking at the queries with "Query Monitor"

- Before this change I could see the "order by" which we are trying to remove.
- And after this change I could see the same query without the "order by" as intended.


mitogh:
Hey @dgwatkins thanks a lot for this 👍

neillmcshea:
Ref:_ [C#110459](http://central.tri.be/issues/110459)

Thanks @dgwatkins for the PR. -  we're taking a look and will let you know if we have any questions or comments

jesseeproductions:
Thanks @dgwatkins I have merged the fix into our release for 4.6.22! Thanks Again!

PR #2049, opened 2018-06-12
Update `tribe_events_month_has_events_filtered()`

juanfra:
🎫 https://central.tri.be/issues/67607

Changes after QA response.


src/functions/template-tags/month.php
```diff
@@ -212,7 +212,9 @@ function tribe_events_month_has_events_filtered() {
 		 *
 		 * @since TBD
 		 */
```
bordoni: Move the docblock close to the Filter.

juanfra: My bad!

PR #2055, opened 2018-06-14
Fix the import settings for Eventbrite

nefeline:
Ref. https://central.tri.be/issues/106947

- Add the Eventbrite import settings on EA;
- Display the correct import settings when Eventbrite Tickets is enabled;

### DIFF
--- a/src/admin-views/aggregator/settings.php
+++ b/src/admin-views/aggregator/settings.php
@@ -565,8 +565,6 @@ $internal = array_merge(
 
 /**
  * If Eventbrite Tickets is enabled and Event Aggregator is disabled, display the correct import settings
- *
- * @since TBD
  */
 if ( class_exists( 'Tribe__Events__Tickets__Eventbrite__Main' ) && ! tribe( 'events-aggregator.main' )->has_license_key() ) {
 	$internal = array_merge(
@@ -642,8 +640,6 @@ if ( tribe( 'events-aggregator.main' )->is_service_active() ) {
 
 	/**
 	 * If Eventbrite Tickets is enabled and Event Aggregator is disabled, display the correct import links
-	 *
-	 * @since TBD
 	 */
 	if ( class_exists( 'Tribe__Events__Tickets__Eventbrite__Main' ) && ! tribe( 'events-aggregator.main' )->has_license_key() ) {
 		$ea_keys = array(
@@ -656,7 +652,7 @@ if ( tribe( 'events-aggregator.main' )->is_service_active() ) {
 		);
 
 		foreach ( $ea_keys as $key ) {
-			unset ( $import_setting_links[ $key ] );
+			unset( $import_setting_links[ $key ] );
 		}
 	}
 
--- a/src/admin-views/aggregator/settings.php
+++ b/src/admin-views/aggregator/settings.php
@@ -112,7 +112,7 @@ $ea_disable = array(
 
 $global = $ical = $ics = $facebook = $gcal = $meetup = $url = $eb_fields = array();
 // if there's an Event Aggregator license key, add the Global settings, Facebook, iCal, and Meetup fields
-if ( Tribe__Events__Aggregator::is_service_active() && tribe( 'events-aggregator.main' )->has_license_key() ) {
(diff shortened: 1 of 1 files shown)


jesseeproductions:
Is their a corresponding PR to remove the settings from Eventbrite Plugin then? 

https://github.com/moderntribe/events-eventbrite/blob/release/M18.09/src/Tribe/Admin/Settings.php

bordoni:
That is a very good question. We need to make sure we are not duplicating these.

nefeline:
Hey @jesseeproductions , no there isn't: I didn't remove the settings from Eventbrite Plugin because we still want them to be displayed when Event Aggregator is disabled, right? To completely remove them from Eventbrite tickets some additional checks would be necessary on EA, i.e. disable all import settings if there's no license key, except the Eventbrite import settings, if the plugin is active. I can do this change, but will need more time.

jesseeproductions:
Good call, yeah we need to support it as you said for both. I think we need to have the settings in TEC like you have here, but if Eventbrite is active display only those settings on the import page, Then remove them from Eventbrite so we only ever have one area of settings.

nefeline:
Ok, updated the code and removed the import settings from Eventbrite tickets. Also, added the necessary conditionals for when EA is disabled and Eventbrite tickets is enabled.

src/admin-views/aggregator/settings.php
```diff
@@ -575,8 +634,32 @@
...
 	);
 
+	/**
+	 * If Eventbrite Tickets is enabled and Event Aggregator is disabled, display the correct import links
+	 *
+	 * @since TBD
+	 */
+	if ( class_exists( 'Tribe__Events__Tickets__Eventbrite__Main' ) && ! tribe( 'events-aggregator.main' )->has_license_key() ) {
+		$ea_keys = array(
+			'ical-settings',
```
bordoni: Why are we unset everything?

src/admin-views/aggregator/settings.php
```diff
@@ -515,9 +559,24 @@
...
 	$meetup,
 	$url,
+	$eb_fields,
 	$ea_disable
 );
 
+/**
+ * If Eventbrite Tickets is enabled and Event Aggregator is disabled, display the correct import settings
+ *
+ * @since TBD
```
bordoni: No need for this since.

src/admin-views/aggregator/settings.php
```diff
@@ -575,8 +634,32 @@
...
+			'ical-settings',
+			'ics-settings',
+			'facebook-settings',
+			'google-settings',
+			'meetup-settings',
+			'url-settings',
+		);
+
+		foreach ( $ea_keys as $key ) {
+			unset ( $import_setting_links[ $key ] );
```
bordoni: space after `unset`

src/admin-views/aggregator/settings.php
nefeline: Hey @bordoni , thanks for the review :+1: ! This unset exists to remove only the import setting links related to Event Aggregator when it is disabled (the user doesn't have a license key) and Eventbrite Tickets is enabled.

bordoni: So this code probably needs to live in Eventbrite Tickets. Right? If it only comes into play if you have it active.

nefeline: It only comes to play if both things are happening simultaneously: EA is disabled + EBT is enabled.
What makes things more complex here is the fact that Eventbrite tickets is enabling EA by default (without a license key). This is my understanding:

Eventbrite import settings should be displayed by default on EA (because it is possible to import from EB without Eventbrite Tickets)
If the user enables EBT, nothing happens, the import settings are the same
If the user doesn't have a license key for EA, all import settings are removed
If the user doesn't have a license key for EA, but has EBT enabled, all import settings should be removed, except Eventbrite
I believe it makes more sense to leave the code on EA because in this case, we are still disabling the default EA import settings if the user doesn't have a license key and just adding an extra check in case of EBT is enabled (i.e. don't remove the default EA settings for Eventbrite if EBT is active).

PR #2062, opened 2018-06-18
Restore and display header navigation items

juanfra:
https://central.tri.be/issues/108754

Changes on the pcss file

### DIFF
--- a/src/resources/postcss/tribe-events-theme.pcss
+++ b/src/resources/postcss/tribe-events-theme.pcss
@@ -90,9 +90,6 @@ ul.tribe-events-sub-nav a,
 }
 
 /* General Template Header/Footer Nav */
-#tribe-events-header .tribe-events-nav-pagination {
-	display: none;
-}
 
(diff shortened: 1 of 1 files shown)


Camwyn:
Should ask @dpellenwood why he made that change - maybe you did? That was part of the a11y changeset.

dpellenwood:
Are the header nav items supposed to be displayed by default? If they are that's totally fine. IIRC, when I was doing the a11y work, there was some styles that basically hid the inner content of the nav items for the header nav, not the whole nav menu. This caused an a11y issue related to missing content. I just hid the whole nav menu element as a repair. But, if they are supposed to show by default, then this is fine. :)

juanfra:
Thank you @dpellenwood @Camwyn! Yes, these were being displayed so far. And given https://github.com/moderntribe/the-events-calendar/commit/4927305cbc18be2d64b4a2b353e196f01618790c and https://central.tri.be/issues/108754 we should keep displaying them.

PR #2063, opened 2018-06-18
Do not use global phrase if there's no EA license.

juanfra:
https://central.tri.be/issues/94426

### DIFF
--- a/src/admin-views/aggregator/settings.php
+++ b/src/admin-views/aggregator/settings.php
@@ -6,7 +6,6 @@
 $internal = array();
 $use_global_settings_phrase = __( 'Use global import settings', 'the-events-calendar' );
 $post_statuses = get_post_statuses( array() );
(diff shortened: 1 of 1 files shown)


src/admin-views/aggregator/settings.php
```diff
@@ -19,6 +18,8 @@
 );
 $events_aggregator_is_active = tribe( 'events-aggregator.main' )->is_service_active();
 
+$origin_post_statuses = $events_aggregator_is_active ? array( '' => $use_global_settings_phrase ) + $post_statuses : $post_statuses;
```
mitogh: Would you mind split this into multiple lines:

```php
$origin_post_statuses = $events_aggregator_is_active 
	? array( '' => $use_global_settings_phrase ) + $post_statuses 
	: $post_statuses;
```

PR #2072, opened 2018-06-26
Add event ID parameter to tribe_events_event_classes filter

niconerd:
https://central.tri.be/issues/64807

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -215,6 +215,10 @@ Still not happy? Shoot us an email to support@theeventscalendar.com or tweet to 
 
 == Changelog ==
 
+= [4.6.20] TBD =
+
+* Tweak - Added event ID parameter to `tribe_events_event_classes` filter to make it more useful [64807]
+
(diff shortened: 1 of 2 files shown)


niconerd:
@bordoni qq I just found this filter is being used here: https://github.com/moderntribe/the-events-calendar/blob/master/src/Tribe/Template_Factory.php#L105 and here https://github.com/moderntribe/the-events-calendar/blob/master/src/Tribe/Template_Factory.php#L460

Do you think those uses need amend? Or maybe we should roll back the change I made?

mitogh:
I think this is fine @niconerd as the filter call the method `event_classes` with only one argument the `$classes`.

By default `apply_filters` only send a single argument if you have multiple arguments in a filter when you use the filter you need to specify how many arguments do you need if non specified only one is passed.

So it should be fine

mitogh:
Let me know if there's anything else I can do for you @niconerd

niconerd:
Thanks for reviewing Cris 🥇

PR #2080, opened 2018-06-27
Don't filter $postarr for updateEvent

sc0ttkclark:
https://central.tri.be/issues/109482

Further addresses work done in #2077 and makes it work for REST API (testing was done on series, but did not address the full issue).


sc0ttkclark:
Adding actual tests for this one

PR #2082, opened 2018-07-02
Fix/109722 protect invalid tribe create event

cliffordp:
https://central.tri.be/issues/109722 (bug) and https://central.tri.be/issues/107323 (tests)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space before the last parenthesis | [event.php:61](https://github.com/moderntribe/the-events-calendar/blob/e326ac333184b9ed6e577840f0e7e7d87af6e127/src/functions/advanced-functions/event.php#L61)
:x: | There must be a single space before the last parenthesis | [event.php:61](https://github.com/moderntribe/the-events-calendar/blob/e326ac333184b9ed6e577840f0e7e7d87af6e127/src/functions/advanced-functions/event.php#L61)
:x: | No space before closing parenthesis is prohibited | [event.php:61](https://github.com/moderntribe/the-events-calendar/blob/e326ac333184b9ed6e577840f0e7e7d87af6e127/src/functions/advanced-functions/event.php#L61)
:x: | There must be a single space before the last parenthesis | [event.php:83](https://github.com/moderntribe/the-events-calendar/blob/e326ac333184b9ed6e577840f0e7e7d87af6e127/src/functions/advanced-functions/event.php#L83)
:x: | There must be a single space before the last parenthesis | [event.php:83](https://github.com/moderntribe/the-events-calendar/blob/e326ac333184b9ed6e577840f0e7e7d87af6e127/src/functions/advanced-functions/event.php#L83)
:x: | No space before closing parenthesis is prohibited | [event.php:83](https://github.com/moderntribe/the-events-calendar/blob/e326ac333184b9ed6e577840f0e7e7d87af6e127/src/functions/advanced-functions/event.php#L83)

src/Tribe/API.php
```diff
@@ -92,16 +103,38 @@ public static function deleteEvent( $eventId, $force_delete = false ) {
...
+				 * @param array   $raw_data The meta fields we tried to send.
+				 * @param WP_Post $event    The event itself.
+				 *
+				 * @since TBD
+				 */
+				do_action( 'tribe_events_event_save_failed_invalid_meta', $event_id, $raw_data, $event );
+
+				return false;
+			} else {
+				unset( $raw_data );
```
mitogh: Don't think this `else` block is required with the content inside as the function is completed few lines afterwards and variable is no longer available after that.

src/functions/advanced-functions/event.php
```diff
@@ -76,7 +80,11 @@ function tribe_create_event( $args ) {
 	function tribe_update_event( $postId, $args ) {
 		$postId = Tribe__Events__API::updateEvent( $postId, $args );
 
-		return $postId;
+		if ( is_wp_error( $postId ) ) {
```
mitogh: I think using ternary operator will make things more compact here.

```php
return is_wp_error( $postId ) ? false : $postId;
```

src/functions/advanced-functions/event.php
```diff
@@ -56,18 +56,22 @@
 	 */
 	function tribe_create_event( $args ) {
 		$args['post_type'] = Tribe__Events__Main::POSTTYPE;
-		$postId = Tribe__Events__API::createEvent( $args );
+		$postId            = Tribe__Events__API::createEvent( $args );
 
-		return $postId;
+		if ( is_wp_error( $postId ) ) {
```
mitogh: Same as below ⬇️

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+		 *
+		 * @param array $args The arguments sent to create/update an event post.
+		 *
+		 * @return array|WP_Error
+		 */
+		public static function sanitize_event_post_create_update_args( $args ) {
+			if (
+				empty( $args['post_type'] )
+				|| Tribe__Events__Main::POSTTYPE !== $args['post_type']
+				|| ! is_array( $args )
```
mitogh: I will move this one first instead as before accesing the key `post_type` you need to make sure `$args` is an array.

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+		 *
+		 * @param string $value The post meta value to be checked, such as '30'.
+		 *
+		 * @return string|WP_Error
+		 */
+		public static function sanitize_minute_meta_value( $value ) {
+			$value = (string) $value;
+
+			if ( '' === trim( $value ) ) {
+				return trim( $value );
```
mitogh: To avoid having to call `trim` here again you can just return `''` like 

```php
return '';
```

As `trim( $value)` is an empty string

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+					esc_html__( 'An event having a post meta value of `%s` minutes is not valid. Make sure it is from 0 to 59.', 'the-events-calendar' ),
+					$value
+				);
+
+				return new WP_Error( 'non-numeric-tribe-events-minutes-meta-value', $error_message );
+			}
+
+			$new_value = absint( $value );
+
+			if ( 59 < $new_value ) {
```
mitogh: I think by flipping the condition provides a more easier way to understand this conditional

```php
if ( $new_value > 59 )
```

- *`$new_value` is greater than 59*

Instead of 

- *`59 is greater than $new_value`*

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+				$error_message = sprintf(
+					esc_html__( 'An event having a post meta value of `%s` minutes is not valid. Make sure it is from 0 to 59.', 'the-events-calendar' ),
+					$value
+				);
+
+				return new WP_Error( 'invalid-tribe-events-minutes-meta-value', $error_message );
+			}
+
+			$new_value = str_pad( $new_value, 2, '0', STR_PAD_LEFT );
```
mitogh: To avoid resign this to a new variable you can return the result from `str_pad` instead.

```php
return str_pad( $new_value, 2, '0', STR_PAD_LEFT );
```

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+					$error_message = sprintf(
+						esc_html__( 'An event having a post meta value of `%s` hour (24-hour) is not valid. Make sure it is from 0 to 23.', 'the-events-calendar' ),
+						$value
+					);
+
+					return new WP_Error( 'invalid-tribe-events-24-hour-meta-value', $error_message );
+				}
+			}
+
+			$new_value = str_pad( $new_value, 2, '0', STR_PAD_LEFT );
```
mitogh: Return the result from `str_pad` prevents to overwrite the local variable `$new_value`

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+					$value
+				);
+
+				return new WP_Error( 'non-numeric-tribe-events-hour-meta-value', $error_message );
+			}
+
+			$new_value = absint( $value );
+
+			if ( $twelve_hour ) {
+				if (
```
mitogh: I think this conditional can fit in a single line.

mitogh: Also I think by inverting the conditional here make it more easier to follow see the details commented on `sanitize_minute_meta_value`

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+				) {
+					$error_message = sprintf(
+						esc_html__( 'An event having a post meta value of `%s` hour (12-hour) is not valid. Make sure it is from 1 to 12.', 'the-events-calendar' ),
+						$value
+					);
+
+					return new WP_Error( 'invalid-tribe-events-12-hour-meta-value', $error_message );
+				}
+			} else {
+				if ( 23 < $new_value ) {
```
mitogh: Same as above (about inverting the condition)

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+					|| 12 < $new_value
+				) {
+					$error_message = sprintf(
+						esc_html__( 'An event having a post meta value of `%s` hour (12-hour) is not valid. Make sure it is from 1 to 12.', 'the-events-calendar' ),
+						$value
+					);
+
+					return new WP_Error( 'invalid-tribe-events-12-hour-meta-value', $error_message );
+				}
+			} else {
```
mitogh: To avoid nesting you can combine else and next if in a `else if` clause

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+		 * @param string $value            The post meta value to be checked, such as 'am'.
+		 * @param bool   $empty_if_invalid If true, set an invalid value to an empty string, else generate WP_Error.
+		 *
+		 * @return string|WP_Error
+		 */
+		public static function sanitize_meridian_meta_value( $value, $empty_if_invalid = true ) {
+			// lower-case to match the `a` PHP date format used elsewhere
+			$new_value = strtolower( trim( $value ) );
+
+			if (
```
mitogh: To avoid nesting I think you can refactor this one into something like this:

```php
$is_valid = 'am' === $new_value || 'pm' === $new_value;
$invalid_value = '';

if ( ! $empty_if_invalid ) {
	$invalid_value = new WP_Error(...);
}

return $is_valid ? $new_value : $invalid_value;
```

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+				&& 12 < absint( $args['EventEndHour'] )
+				&& 24 > absint( $args['EventEndHour'] )
+			) {
+				$args['EventEndMeridian'] = '';
+			}
+
+			// Now process all but the meridians
+			foreach ( $args as $key => &$value ) {
+				if ( 'EventStartHour' === $key ) {
+					if ( empty( $args['EventStartMeridian'] ) ) {
```
mitogh: I think this one can be refactor into

```php
$twelve_hour = ! empty( $args['EventStartMeridian'] );
$value = self::sanitize_hour_meta_value( $value, $twelve_hour );
```

This will allow to get ride of the `else` block.

And the same for the next `elseif` block

src/Tribe/API.php
```diff
@@ -602,5 +642,222 @@ public static function get_and_flatten_event_meta( $event_id ) {
...
+				! empty( $args['EventEndMeridian'] )
+				&& 12 < absint( $args['EventEndHour'] )
+				&& 24 > absint( $args['EventEndHour'] )
+			) {
+				$args['EventEndMeridian'] = '';
+			}
+
+			// Now process all but the meridians
+			foreach ( $args as $key => &$value ) {
+				if ( 'EventStartHour' === $key ) {
```
mitogh: To avoid deep nesting you can return here the args instead.

```php
foreach( ... ) {
	if ( 'EventStartHour' !== $key ) {
		return $args;
	}
	
	// Remove nesting and do other operations
}
```

src/Tribe/API.php
cliffordp: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#yoda-conditions says to do it the way I did it, correct?

src/Tribe/API.php
cliffordp: `$new_value` was declared earlier. just changing its value here from integer to 2-digit, zero-left-padded string version of that integer. So OK?

cliffordp: Same as the method before this one.

src/Tribe/API.php
cliffordp: Same as the method below this one

src/Tribe/API.php
cliffordp: If `$twelve_hour` is `true` but the hour is `13` -- and the conditional was all together -- it would failover to 24-hour format conditional, which isn't correct. Please let me know if this is incorrect.

cliffordp: oh, no, incorrect... will fix

src/Tribe/API.php
cliffordp: not sure if i followed the `else`/`elseif` part so please take a fresh look upon update

src/Tribe/API.php
cliffordp: `$key` could be one of four different meta keys, not just this one, right?

mitogh: Right anything that is not `EventStartHour`

src/Tribe/API.php
mitogh: I think this has been fixed with the other changes so it should be fine

src/Tribe/API.php
mitogh: Yeah as `str_pad` returns a string value and the operation is executed before is returned so what happens here is: 

1. First executes `str_pad`
2. A string is returned from `str_pad`
3. Value from previous string is returned from this method.

As you don't need to assign back the value again to the same variable before returning.

src/Tribe/API.php
mitogh: That's correct as specified on Codex, main reason for this is to avoid confusion on assigning values like this: 

```
if ( $a = 12 ) {}
``` 

However I think on most of services sides we are moving away from this one for couple of reasons: 

1. Readability as suggested on first comment.
2. We have code reviews in place which give us an opportunity to catch for that possible scenario.
3. We are using strict comparison most of the time ( `===` `!===` ) and for `>`, `<` there's more hard to fall into the problem that "tries to solve" the problem faced by WP Yoda style

cliffordp: okie dokie

src/Tribe/API.php
cliffordp: it's clear to me now looking a 2nd time. I thought you meant quite a bit more lines to change.

src/Tribe/API.php
cliffordp: discussed via Slack and Cris said leave as-is

src/Tribe/API.php
cliffordp: Per Slack conversation, I should change all to variable on the left.

PR #2090, opened 2018-07-10
Handle fast queue processing

lucatume:
Ticket: https://central.tri.be/issues/110258

This PR fixes the issue, on the client side, reported in the ticket.
What is happening: when the queue completes on the first run, the `in progress` flag on the import record might not be unset depending on a critical path.
This will trigger re-runs of the queue (as the record is still processing) on an empty queue, the queue will never process, never complete and be stuck until health-check (every 15').


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *Support / Triage*, but this was pull requested against *release/M18.10*.

PR #2093, opened 2018-07-11
Handle the case where the queue has no items to process

lucatume:
Ticket: https://central.tri.be/issues/110311

When scheduling imports it might happen that the preview, or the runs, of the import, will not contain items to process.
This PR makes sure that case is gracefully handled.


bordoni:
Can you add a changelog??

PR #2102, opened 2018-07-24
Fix returning all posts if none assigned

cliffordp:
Per QA from https://central.tri.be/issues/105116


juanfra:
The array shouldn't have a zero as value, should be empty.

PR #2106, opened 2018-07-25
Add support for include and post_parent arguments to REST API

lucatume:
Ticket: https://central.tri.be/issues/94326

Based on: https://github.com/moderntribe/tribe-common/pull/741

This PR updates the REST API response for Events to:
* build a REST URL that uses the `include` argument to include all the events in a Series of recurring events when hitting `/all`
* support the `include` argument to fetch Events by their post IDs


src/Tribe/Validator/Base.php
```diff
@@ -201,4 +201,25 @@ public function is_venue_id_list( $venues, $sep = ',' ) {
...
+	 *
+	 * @return bool
+	 */
+	public function is_event_id_list( $events, $sep = ',' ) {
+		$sep    = is_string( $sep ) ? $sep : ',';
+		$events = Tribe__Utils__Array::list_to_array( $events, $sep );
+
+		$valid = array_filter( $events, array( $this, 'is_event_id' ) );
+
+		return ! empty( $events ) && count( $valid ) === count( $events );
```
mitogh: Same comment from: 

- https://github.com/moderntribe/tribe-common/pull/741#discussion_r205197214

src/Tribe/REST/V1/Headers/Base.php
```diff
@@ -81,7 +81,36 @@ public function get_api_origin_meta_name() {
...
+			if (
+				( $wp_query = tribe_get_global_query_object() )
+				&& ( $post_parent = $wp_query->get( 'post_parent' ) )
+				&& $post_parent == $this_post_id
+			) {
+				$all      = array( $post_parent );
+				$children = get_posts( array(
+					'post_type'      => Tribe__Events__Main::POSTTYPE,
+					'fields'         => 'ids',
+					'posts_per_page' => - 1,
```
mitogh: Maybe might be a good idea to set a limit here and paginate on this endpoint instead to avoid having a slow endpoint with ~ 1k events

src/Tribe/REST/V1/Headers/Base.php
```diff
@@ -81,7 +81,36 @@ public function get_api_origin_meta_name() {
...
+
+			/**
+			 * We are dealing with an /all kind of query.
+			 * The root URL has to be modified to include the
+			 * post parent and its children post IDs.
+			 */
+			if (
+				( $wp_query = tribe_get_global_query_object() )
+				&& ( $post_parent = $wp_query->get( 'post_parent' ) )
+				&& $post_parent == $this_post_id
```
mitogh: Wondering about the loose comparison here instead of a strict equality.

src/Tribe/Validator/Base.php
lucatume: Same answer applies here.

src/Tribe/REST/V1/Headers/Base.php
lucatume: The query might provide the `post_parent` as a string or int. A cast before would be better was it not that it would throw on bad numeric values.
So to be able to `===` here an `is_numeric` check would need to be made before and then an `int` cast; in this case loose comparison is more efficient.

src/Tribe/REST/V1/Headers/Base.php
lucatume: This is a very good idea.

PR #2109, opened 2018-07-26
Prevent users from selecting past dates in scheduled URL imports

lucatume:
Ticket: https://central.tri.be/issues/111227

This PR:
* updates the `events-admin.js` file to allow datepickers to define a min/max date using a `data` attribute
* udpates EA UI code to prevent users from selecting a past date when setting up Scheduled imports from Other URL (see reasons in ticket: https://central.tri.be/issues/108375#note-8)

What it does: http://p.tri.be/k0odtA


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *1.0.25*, but this was pull requested against *release/M18.11*.

barryhughes:
Looks good to me ... but does it need a changelog entry? If not I think it's good to go!

barryhughes:
Merging so QA can test, we can add a changelog entry later if required.

PR #2131, opened 2018-08-02
#70769 - setup future event display

jesseeproductions:
_Ref:_ [C#70769](https://central.tri.be/issues/70769)


jesseeproductions:
removed in another pr

PR #2135, opened 2018-08-06
Fix issue where Events would be duplicated in CSV import

lucatume:
Ticket: https://central.tri.be/issues/102745

This PR fixes two separate instances that would lead to event duplication during import:
1. while the query to look for events matching the criteria is setting the `tribe_remove_date_filters` query arg to `true` the query code, due to the order of checks, was ignoring it
2. since the query args did not contain a `post_status` argument it would default to `publish`; events would be duplicated any time the status of the existing event was not `publish`


mitogh:
Conflicts...

PR #2153, opened 2018-08-28
⌗112432 Remove Calls to Deprecated FB Functions; Fix Meetup API Fields

ggwicz:
**Code Review Note:** No readme edits required here, as these changes are from QA feedback on the ticket.

**Ticket:** [**⌗112432**](http://central.tri.be/issues/112432)

### DIFF
--- a/src/admin-views/tribe-options-addons-api.php
+++ b/src/admin-views/tribe-options-addons-api.php
@@ -13,7 +13,7 @@ if ( get_option( 'pue_install_key_event_aggregator' ) ) {
 	$internal = array(
 		'meetup-start' => array(
 			'type' => 'html',
-			'html' => '<h3>Meetup</h3>',
+			'html' => '<h3>' . esc_html__( 'Meetup', 'the-events-calendar' ) . '</h3>',
 		),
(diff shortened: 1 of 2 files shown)


src/admin-views/tribe-options-addons-api.php
```diff
@@ -7,6 +7,30 @@
 
 $current_url = Tribe__Settings::instance()->get_url( array( 'tab' => 'addons' ) );
 
+// if there's an Event Aggregator license key, add the Meetup.com API fields
+if ( get_option( 'pue_install_key_event_aggregator' ) ) {
+
+	$internal = array(
+		'meetup-start' => array(
+			'type' => 'html',
+			'html' => '<h3>Meetup</h3>',
```
bordoni: Why is this not translatable?

ggwicz: Because "Meetup" is a brand name/trademark that should not be altered regardless of language.

ggwicz: Nevermind, after a meeting today we have agreed this should be translatable after all. Will fix.

PR #2154, opened 2018-08-29
#109815 - move wordpress custom field metabox display setting to the …

jesseeproductions:
…events calendar pro

_Ref:_ [C#109815](https://central.tri.be/issues/109815)


jesseeproductions:
Yeah, since it is legacy I will leave it for now.

PR #2180, opened 2018-09-24
⌗108120 Alphabetize Linked Posts in Dropdowns

ggwicz:
**Ticket:** [**⌗108120**](http://central.tri.be/issues/108120)


src/Tribe/Linked_Posts.php
```diff
@@ -533,7 +533,8 @@ public function get_linked_post_info( $linked_post_type, $args = array(), $linke
 				'private',
 				'pending',
 			),
-			'orderby'              => 'post__in',
+			'order'                => 'ASC',
+			'orderby'              => 'post__in post_title',
```
Camwyn: I know this is legit, but I am personally always more comfortable with an array for more specificity when passing multiple `orderby` values. 
I just think it's clearer?
YMMV.

PR #2185, opened 2018-09-24
#111518 - fix tribe_get_events to accept 0 as an argument

jesseeproductions:
_Ref:_ [C#111518](https://central.tri.be/issues/111518)


jesseeproductions:
Looks like they are failing on the db issue with codeception.

PR #2197, opened 2018-10-03
Include only enabled views in the default view option

juanfra:
🎫 https://central.tri.be/issues/114807


ggwicz:
Tagging `enhance` because of the stuff Stephen tagged and readme conflicts. Address those and this should be okay for another review and a merge :-]

juanfra:
Hey guys, thanks for taking a look!

S, regarding the questions:

1. Definitely, I chose that one (`$default_views_options`) because the field is named "Default Views" in the front end, and that array contains the options. And we already have another `$enabled_views` right there. I think `$enabled_views_options` will make it.

2. I branched from the release branch and realized that there were some work done and the changelog entries of that (18.15 work) were included in the previous version. So, I went over central, checked the tickets target version and accommodated things over the changelog. Perhaps the MR branch is updated. I guess I'll take the time to check everything in order to resolve the conflicts consistently.

Thanks again 😃

juanfra:
Thanks! Back to you @Camwyn and @ggwicz 

I think we'll need to take a couple minutes to organize the readme entries for each MR (the hotfixes and not initializing this MR correctly messed up things)

juanfra:
Would you please take a look at this one too? https://github.com/moderntribe/events-pro/pull/1018

It's basically the same patch but for the "Default mobile view" field.

Thanks 😄

PR #2199, opened 2018-10-03
Fix single event navigation when having minical or ajax calendar scripts enqueued

juanfra:
🎫 https://central.tri.be/issues/114359


ggwicz:
Ready to merge, but tagging `enhance` because of the readme conflicts that need fixing...

PR #2208, opened 2018-10-15
Show only applicable refinement filters during Eventbrite imports.

lucatume:
Ticket: https://central.tri.be/issues/107009

This PR add a Javascript-side check on the Eventbrite current source URL to hide/show only applicable refinement filters depending on the source type (e.g. event or organizer).

Screencast: http://p.tri.be/YAlw2j


jesseeproductions:
Moving this along as it was approved.

PR #2215, opened 2018-10-19
Merge `master` into latest MR

mitogh:


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "the-events-calendar",
+  "version": "4.6.24.1",
```
barryhughes: This gives us two version properties, we only want one though right?

readme.txt
```diff
@@ -215,6 +215,12 @@ Still not happy? Shoot us an email to support@theeventscalendar.com or tweet to
 
 == Changelog ==
 
+= [4.6.24.1] 2018-10-08 =
```
barryhughes: This places the 4.6.24.1 changelog above the 4.6.25 changelog entries ... we probably want it in the other order.

package.json
barryhughes: (and for M18.15, this is the wrong one)

mitogh:
Would you mind take another look thanks @barryhughes

PR #2220, opened 2018-10-24
Linked Posts ORM work for Venues

sc0ttkclark:
https://central.tri.be/issues/116277

### DIFF
--- a/src/Tribe/Repositories/Venue.php
+++ b/src/Tribe/Repositories/Venue.php
@@ -1,0 +1,193 @@
+<?php
+/**
+ * The main ORM/Repository class for venues.
+ *
+ * @since TBD
+ */
+
+/**
+ * Class Tribe__Events__Repositories__Venue
+ *
+ *
+ * @since TBD
+ */
+class Tribe__Events__Repositories__Venue extends Tribe__Repository {
+
+	/**
+	 * The unique fragment that will be used to identify this repository filters.
+	 *
+	 * @var string
+	 */
+	protected $filter_name = 'venues';
+
+	/**
+	 * Tribe__Events__Repositories__Venue constructor.
+	 *
+	 * Sets up the repository default parameters and schema.
+	 *
+	 * @since TBD
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->default_args = array(
+			'post_type'                    => Tribe__Events__Main::POSTTYPE,
+			// We'll be handling the dates, let's mark the query as a non-filtered one.
+			'tribe_suppress_query_filters' => true,
+		);
+
+		$this->schema = array_merge( $this->schema, array(
+			'address'        => array( $this, 'filter_by_address' ),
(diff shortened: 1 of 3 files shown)


src/Tribe/Repositories/Venue.php
```diff
@@ -0,0 +1,181 @@
...
+		) );
+	}
+
+	/**
+	 * Filter by LIKE or REGEX string.
+	 *
+	 * @param string $meta_key Meta key to filter by.
+	 * @param string $value    MySQL compatible LIKE or REGEX string.
+	 */
+	public function filter_by_like_regex( $meta_key, $value ) {
```
lucatume: The same observation done for Organizers applies here: move it to the base `Tribe__Repository` class.

src/Tribe/Repositories/Venue.php
```diff
@@ -0,0 +1,181 @@
...
+
+		$this->by( 'meta_like', $meta_key, $value );
+	}
+
+	/**
+	 * Filters venues by a specific address line.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $value MySQL compatible LIKE or REGEX string.
```
lucatume: Here and in any other place where it appears: MySQL uses `REGEXP` with a final `P`. 
If we note it to refer to SQL use `REGEXP`, not `REGEX`.

src/Tribe/Repositories/Venue.php
```diff
@@ -0,0 +1,181 @@
...
+			'state_province' => array( $this, 'filter_by_state_province' ),
+			'website'        => array( $this, 'filter_by_website' ),
+		) );
+	}
+
+	/**
+	 * Filter by LIKE or REGEX string.
+	 *
+	 * @param string $meta_key Meta key to filter by.
+	 * @param string $value    MySQL compatible LIKE or REGEX string.
```
lucatume: Here and in any other place where it appears: MySQL uses REGEXP with a final P.
If we note it to refer to SQL use REGEXP, not REGEX.

lucatume:
Same backport work that should be done in organizers should be done here and fix the tests.

src/Tribe/Repositories/Venue.php
sc0ttkclark: I told Matt this is what I was going to do, glad we're on the same page :)

src/Tribe/Repositories/Venue.php
sc0ttkclark: Copy/pasted from the spec doc, will update

src/Tribe/Repositories/Venue.php
sc0ttkclark: Copy/pasted from the spec doc, will update

src/Tribe/Repositories/Venue.php
sc0ttkclark: Copy/pasted from the spec doc, will update

sc0ttkclark:
Moved into #2222

PR #2223, opened 2018-10-25
Bugfix tribe bar search if no date field

tyrann0us:
It's possible to remove the date filter from tribe bar:
```PHP
add_filter(
  'tribe-events-bar-filters',
  function (array $filters): array
  {
    unset($filters['tribe-bar-date']);
    return $filters;
  }
);
```

This will cause

https://github.com/moderntribe/the-events-calendar/blob/e483040bfca844f70bf7fa8088023ef780587f8a/src/resources/js/tribe-events-ajax-calendar.js#L374

to fail on month view: `Uncaught TypeError: Cannot read property 'length' of undefined`.

The PR fixes this.


jrserani:
_Ref:_ [C#116995](http://central.tri.be/issues/116995)

@tyrann0us - Thank you for your first PR to The Events Calendar! We appreciate the time you took to help make the plugin even better. We have this now added to our workflow and we'll let you know if we have any questions. :-)

tyrann0us:
@jrserani, do you have an ETA for this bugfix?

tyrann0us:
@jrserani, sorry to annoy you but this bug currently blocks us from doing a feature deploy for a client.

Any feedback on when to expect the fix is highly appreciated.
Thanks!

CC @barryhughes, @ggwicz, @jesseeproductions

ggwicz:
Hey @tyrann0us, thanks so much for your patience here. I'm happy to report that we've integrated your fix here over in this other Pull Request → https://github.com/moderntribe/the-events-calendar/pull/2300

The scheduled arrival for the release that includes this fix is about a week and a half from now. Is there any way you can manually implement your fix for now and then replace those files with the official update down the road?

_Note: I am only closing this PR and removing tags on it since the above PR will be used in its stead—we can certainly keep chatting here!_

tyrann0us:
@ggwicz, thanks for getting back to me.
We are going to wait for the update because implementing and testing https://github.com/cweagans/composer-patches is too expensive given the fact the update will be released soon.

PR #2249, opened 2018-11-13
Add data bundle back in. Expose datetime block

nealfennimore:
Related to https://github.com/moderntribe/events-pro/pull/1043


src/Tribe/Editor.php
```diff
@@ -425,7 +425,20 @@ public function assets() {
 				'priority' => 1
 			)
 		);
-		
+
+		tribe_asset(
```
mitogh: I don't think we wanna to load `data` on every plugin, instead I think we can create the specific selectors on externals plugins where is used. I'm sending a PR shortly for Tickets where `blocks` is used.

nealfennimore: Yeah that was my concern too, especially since you just took it out. Let's see how your PR looks and then we can decide.

The hard part is that events-pro does not use TEC as a direct dependency in the filesystem, so it needs someway of getting that selector from the datetime block, as it won't bundle it in events-pro.

mitogh: I think in the case of `PRO` we can set all the files from TEC to be a direct dependency as we can ensure that TEC is required to run PRO but in cases like Tickets or Tickets+ TEC or PRO are not requirement.

nealfennimore: The one issue is that since events-pro does not actually include TEC via JS imports, we need that selector via `window.tribe.events.blocks.datetime.selectors.[whatever]`, but we're not exposing it unless we do something with this PR.

src/modules/data/index.js
```diff
@@ -5,6 +5,8 @@ import reducer from './reducers';
 
 import { plugins } from '@moderntribe/common/data';
 import { store } from '@moderntribe/common/store';
+import * as blocks from './blocks';
+import './sagas';
```
nealfennimore: Moved saga import here, so that it runs before store can initialize

PR #2263, opened 2018-11-15
Date & Time > enlarge clickable area

juanfra:
🎫 https://central.tri.be/issues/118125

### DIFF
--- a/src/modules/blocks/event-datetime/style.pcss
+++ b/src/modules/blocks/event-datetime/style.pcss
@@ -46,6 +46,7 @@
 
 	.tribe-editor__btn--label {
 		text-align: left;
+		width: 100%;
 
 		&:disabled {
 			cursor: not-allowed;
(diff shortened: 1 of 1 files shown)


paulmskim:
Might be worth asking Leah, but I think the clickable area is quite thin. Maybe making the height of the button taller would be good?

paulmskim:
Actually did you also test this with a timezone label? If there's a timezone label in place, I think it'll be pushed to the outside of the block if you set the width to 100%?

juanfra:
Good catch @paulmskim! Didn't take that into consideration. Pushed a new fix, would you please take a look?

https://www.useloom.com/share/70f015344f994d068bb9903af299ee11

src/modules/blocks/event-datetime/style.pcss
```diff
@@ -46,6 +47,7 @@
 
 	.tribe-editor__btn--label {
 		text-align: left;
+		flex: 1 1 100%;
```
paulmskim: If you're using the `flex` approach, every flex child is going to need a `flex` property, so here would be the price and timezone (I believe). Different browsers have different defaults for the `flex` property and will not behave the same way.

PR #2268, opened 2018-11-16
Feature/112827 recurring events marker

nealfennimore:
Splits out event-datetime block into sub-blocks, so `InnerBlocks` and hooks can be used in separate contexts. Fixes:
> Note: A block can render at most a single InnerBlocks and InnerBlocks.Content element in edit and save respectively. To create distinct arrangements of nested blocks, create a separate block type which renders its own InnerBlocks and assign as the sole allowedBlocks type.

Sets up work for https://github.com/moderntribe/events-pro/pull/1050


src/modules/blocks/event-datetime/controls/template.js
```diff
@@ -0,0 +1,102 @@
...
+			separatorDate,
+			separatorTime,
+			setSeparatorDate,
+			setSeparatorTime,
+			setTimeZone,
+			showTimeZone,
+			timeZone,
+		} = this.props;
+
+		// @todo: modify so this code does not fire unless the block is selected
```
paulmskim: do we still need this todo here?

src/modules/blocks/event-datetime/subblocks/content/container.js
```diff
@@ -0,0 +1,68 @@
...
+	timeZoneLabel: dateTimeSelectors.getTimeZoneLabel( state ),
+} );
+
+const mapDispatchToProps = ( dispatch ) => ( {
+	...bindActionCreators( dateTimeActions, dispatch ),
+	...bindActionCreators( priceActions, dispatch ),
+	onDateTimeLabelClick: onDateTimeLabelClick( dispatch ),
+} );
+
+const mergeProps = ( stateProps, dispatchProps, ownProps ) => ( {
```
paulmskim: probably don't need this if all we're doing is just passing the props through.

src/modules/blocks/event-datetime/subblocks/content/template.js
```diff
@@ -0,0 +1,253 @@
...
+				);
+			case 'time-range':
+				return (
+					<span className={ classNames( 'tribe-editor__separator', className ) }>
+						{ ' '.concat( separatorTime, ' ' ) }
+					</span>
+				);
+			case 'all-day':
+				return (
+					<span className={ classNames( 'tribe-editor__separator', className ) }>{ __( 'ALL DAY', 'events-gutenberg' ) }</span>
```
paulmskim: I know this was how it was before, but from an accessibility standpoint it would be better to put in "All day" and in CSS `text-transform` it to `uppercase`.

src/modules/blocks/event-datetime/controls/template.js
nealfennimore: I actually have no idea on that, I just left it because it was already there. 

@mitogh is it save to delete this comment? I'm not sure what it was intended for

src/modules/blocks/event-datetime/subblocks/content/template.js
nealfennimore: That's a much better approach, thanks Paul - I'll put it in

src/modules/blocks/event-datetime/controls/template.js
paulmskim: Someone can correct me if I'm wrong, but I think in an old implementation of gutenberg the inspector controls would fire even if the block wasn't selected (but now it's fixed on gutenberg's side) and it may have been for that?

mitogh: I think that's right we might be able to get ride of that `TODO`

nealfennimore: Consider it :boom:

nealfennimore:
@paulmskim good to look at again

PR #2273, opened 2018-11-16
Fix/112141 date time month picker issue

paulmskim:
For ticket: https://central.tri.be/issues/112141

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -217,7 +217,7 @@ Still not happy? Shoot us an email to support@theeventscalendar.com or tweet to 
 
 = [4.7] TBD =
 
-
+* Fix - Allow for earlier dates to be selected in date time block
 
 = [4.6.26] 2018-11-13 =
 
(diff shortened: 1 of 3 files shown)


readme.txt
```diff
@@ -217,7 +217,7 @@ Still not happy? Shoot us an email to support@theeventscalendar.com or tweet to
 
 = [4.7] TBD =
 
-
+* Fix - Allow for earlier dates to be selected in date time block
```
mitogh: Same question about the changelog, are we tracking invidual changes on plugin changelog @bordoni ?

src/modules/elements/month/element.js
```diff
@@ -22,7 +22,8 @@ import './style.pcss';
 const today = new Date();
 const currentYear = today.getFullYear();
 const currentMonth = today.getMonth();
-const fromMonth = new Date( currentYear, currentMonth );
+const yearsBack = 5;
```
mitogh: might be a good idea to use a uppercase name to indicate the presence of a "constant" like on PHP side like. `YEARS_BACK`

paulmskim: I agree, though everything here has previously done in lower case, so I'm sticking with that convention.

PR #2274, opened 2018-11-16
Add filter so blocks can set the params during migration

mitogh:
add a new filter `tribe_blocks_editor_update_classic_content_params` so plugins can set the default params during the migration of blocks

Ticket: https://central.tri.be/issues/114554

### DIFF
--- a/src/Tribe/Editor.php
+++ b/src/Tribe/Editor.php
@@ -79,10 +79,11 @@ class Tribe__Events__Editor extends Tribe__Editor {
 			return false;
 		}
 
-		$on_classic_editor_page = tribe_get_request_var( 'classic-editor', false );
+		/** @var Tribe__Editor $editor */
+		$editor = tribe( 'editor' );
 
 		// Bail if in classic editor
-		if ( '' === $on_classic_editor_page || $on_classic_editor_page ) {
+		if ( $editor->is_classic_editor() ) {
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


src/Tribe/Editor.php
```diff
@@ -138,7 +139,17 @@ public function update_post_content_to_blocks( $post ) {
...
+			 * Add an opportunity to set the default params of a block when migrating from classic into
+			 * blocks editor.
+			 *
+			 * @since TBD
+			 *
+			 * @param mixed $params Either array if set to values or slug string
+			 * @param string $slug Name of the block edited
+			 * @param WP_Post $post Post that is being affected
+			 */
+			$params = apply_filters( 'tribe_blocks_editor_update_classic_content_params', end( $block_param ), $slug, $post );
```
juanfra: Why aren't we sending `$block_param` as the first param and using `end` instead?

mitogh: Great question, `end` is from the initial code here, however it has a good reason why as the `$block_params` is something as the following shape.

```php
// "regular" block
array( 'block/slug', array(
	'extra' => 'params'
	'over' 	=> 'here,
));

// no params
// "regular" block
array( 'block/slug' );
```

So `end( $params )` is either an array of params or the `slug` of the block.

juanfra: Cool, thanks for the clarification! Looking great

PR #2282, opened 2018-11-20
Fix/118381 time selector ux improvement

nealfennimore:
https://central.tri.be/issues/118381

https://www.useloom.com/share/fefc3154ab5f4c81aee42c3ac53177c7

- Updates to allow start and end time to sync together when not a multi-day event
- Moves most of logic from thunks to sagas and isolates logic flows to prevent side effects


mitogh:
Some conflicts.

src/modules/blocks/event-datetime/subblocks/dashboard/container.js
```diff
@@ -24,84 +25,31 @@ import EventDateTimeDashboard from './template';
  */
 
 const onSelectDay = ( stateProps, dispatchProps ) => ( { from, to } ) => {
```
paulmskim: we could remove `stateProps` from all these functions as we aren't using them anymore.

src/modules/data/blocks/datetime/sagas.js
```diff
@@ -11,9 +11,66 @@ import {
...
+ *
+ * @export
+ * @since 4.7.0-beta1
+ * @returns {Object} {start, end}
+ */
+export function* deriveSecondsFromDates() {
+	const moments = yield call( deriveMomentsFromDates );
+
+	return yield all( {
+		start: call( timeUtil.toSeconds, moments.start.format( 'HH:mm:ss' ) ),
```
paulmskim: There's a moment utility function `toDatabaseTime` which takes in a moment and spits out a `HH:mm:ss` format time. Not sure if you can nest call functions, but you could use that here.

src/modules/data/blocks/datetime/sagas.js
```diff
@@ -79,37 +136,350 @@ export function* resetNaturalLanguageLabel() {
...
+
+	const seconds = yield call( deriveSecondsFromDates );
+	// Prevent only date changes from using payload
+	if ( [ types.SET_END_TIME, types.SET_START_TIME ].includes( action.type ) ) {
+		// Update seconds to use payload
+		// NOTE: Mutation
+		yield call( [ Object, 'assign' ], seconds, action.payload );
+	}
+
+	// 	// If end time is earlier than start time, fix time
```
paulmskim: spacing

PR #2292, opened 2018-11-23
Update localization of variables from TEC and common

mitogh:
Move shared localized variables into `common` and update shape of TEC localized variables

Ticket: https://central.tri.be/issues/117612
Requires: https://github.com/moderntribe/tribe-common/pull/844


mitogh:
Changes has been applied as suggested.

cc @paulmskim @bordoni

PR #2294, opened 2018-11-23
Previewing and Importing CSV files with whitespace

nefeline:
Ref. https://central.tri.be/issues/117236

CSV files with whitespace at the beginning of the rows won't remove headers and skip columns on Preview

### DIFF
--- a/src/Tribe/Aggregator/Record/CSV.php
+++ b/src/Tribe/Aggregator/Record/CSV.php
@@ -92,6 +92,17 @@ class Tribe__Events__Aggregator__Record__CSV extends Tribe__Events__Aggregator__
 
 		$rows    = $importer->do_import_preview();
 
+		/*
+		 * Strip whitespace from the beginning and end of row values
+		 */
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/Record/CSV.php
```diff
@@ -92,19 +92,34 @@ public function get_csv_data() {
...
 		$rows    = $importer->do_import_preview();
 
+		/*
+		 * Strip whitespace from the beginning and end of row values
+		 */
+		$formatted_rows = array();
+
+		foreach ( $rows as $row ) {
+			$row              = array_map( 'trim', $row );
+			$formatted_rows[] = $row;
```
mitogh: I think you can take advantage that this `foreach` loop is only a single operation and avoid resigning a value to `row`.

```suggestion
			$formatted_rows[] = array_map( 'trim', $row );
```

nefeline: :+1: Done!

PR #2308, opened 2018-11-28
Add the blog_id to EA import information to make sure imports happen in the corret blog

lucatume:
Ticket: https://central.tri.be/issues/116445

This is an improvment to the EA import logic to secure imports from accidental and incorrect blog switching that might happen in the context of the handling of an async process request.
The reaction to this situation is that the item will be requeued to import it later or in another request when the blog context is correct.
Since we cannot debug the customer site this is what I *guess* is happening:
1. blog 1 has EA imports
2. blog 1 starts EA import
3. blog 1 dispatches the processing of the import to the async queue
4. during the handling of the AJAX request the blog is switched to another blog (say 2)
5. all events imported in the context of that request are created in the wrong blog (2 instead of 1)

I'm logging the situation here as an error (that will, but, appear on blog 2, not 1, from the example above and reacting to it by requeueing the item to import.  
The reason I'm not using `switch_to_blog` to change to the expected blog is to worsen what might already be a shaky situation where a "rogue" call to `switch_to_blog` was never properly handled.

### DIFF
--- a/src/Tribe/Aggregator/Processes/Import_Events.php
+++ b/src/Tribe/Aggregator/Processes/Import_Events.php
@@ -209,6 +209,35 @@ class Tribe__Events__Aggregator__Processes__Import_Events extends Tribe__Process
 		$data                  = (array) $item['data'];
 		$this->transitional_id = filter_var( $item['transitional_id'], FILTER_SANITIZE_STRING );
 
+		/*
+		 * Make sure the import is happening in the context of the same site that started it.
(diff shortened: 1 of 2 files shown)


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for **, but this was pull requested against *release/M18.17*.

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M18.18*, but this was pull requested against *release/M18.17*.

PR #2311, opened 2018-11-29
Prevent loading of an Activation hook when outdate Common

bordoni:
When we are activating The Events Calendar and we have an outdated version of Common we need to make sure we are not triggering the activation hook of The Events Calendar, because it could lead into fatals due to older version of Common been loaded first.

The implementation here delays that Activation trigger to another request, so it can have the latest Common loaded via normal common-lib version comparing.

📹 http://i.bordoni.me/b9d029f2e388

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -245,6 +245,15 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 		public static $tribeEventsMuDefaults;
 
 		/**
+		 * Key for the transient flag for a delayed activation
+		 *
+		 * @since 4.7
+		 *
+		 * @var string
+		 */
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task | [Main.php:450](https://github.com/moderntribe/the-events-calendar/blob/6cc751b392f8340a82cd217dc21d0919a07a1744/src/Tribe/Main.php#L450)

borkweb:
This looks pretty good! I'd love to see a screencast of this in action and functional with the known fail-case.

bordoni:
_Recording_

mitogh:
👍 solid!

PR #2314, opened 2018-11-29
 Fix > Price block dashboard issue

juanfra:
🎫 https://central.tri.be/issues/ 118683

### DIFF
--- a/src/modules/elements/dashboard/style.pcss
+++ b/src/modules/elements/dashboard/style.pcss
@@ -6,6 +6,7 @@
 	&--open {
 		display: block;
 		pointer-events: auto;
+		position: initial;
 		width: 100%;
 	}
 }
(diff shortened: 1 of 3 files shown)


src/modules/elements/dashboard/style.pcss
```diff
@@ -6,6 +6,7 @@
 	&--open {
 		display: block;
 		pointer-events: auto;
+		position: initial;
```
paulmskim: I believe the `position` was purposely set as `relative` to anchor the `absolute` positioned dashboard to the container. Could we just set the `top` value specifically for the price block instead, or is there another reason for setting the `position` to `initial` here?

nealfennimore: We could also just do a `transform` instead of that `top` change in order to remove this `position` change

juanfra: > I believe the position was purposely set as relative to anchor the absolute positioned dashboard to the container.

I'm not sure what was the reason behind that positioning, as I haven't worked on the dashboard. I know that was the cause of generating the gap.

I removed the positioning and adjusted the top value.

PR #2338, opened 2018-12-07
Fix the tags block delimiter

juanfra:
🎫 https://central.tri.be/issues/119393

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -216,6 +216,12 @@ Still not happy? Shoot us an email to support@theeventscalendar.com or tweet to 
 
 == Changelog ==
 
+= [4.7.4] TBD =
+
+* Fix - Display the tags block delimiter in the editor [119393]
+
+= [4.7.3] TBD =
+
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -216,6 +216,12 @@ Still not happy? Shoot us an email to support@theeventscalendar.com or tweet to
 
 == Changelog ==
 
+= [4.7.4] TBD =
+
+* Fix - Display the tags block delimiter in the editor [119393]
+
+= [4.7.3] TBD =
```
mitogh: I think this one should be 18.19 as `18.18` is `4.7.2`

juanfra: Thanks @mitogh! I've been asked to work directly on 19.01 so we can leave a clean Q for 18.19

mitogh: OMG! Didn't realize about that, so welcome to the first commit for the 2019 MR.

juanfra: 🙌

PR #2354, opened 2018-12-21
⌗119689 Twenty Nineteen Theme Compat

ggwicz:
**Ticket:** [**⌗119689**](http://central.tri.be/issues/119689)

**Code Review Note:** The typography vars here are just taken directly from the Twenty Nineteen theme's own SCSS stylesheets.


src/resources/postcss/tribe-events-full.pcss
```diff
@@ -14,6 +14,10 @@
 
 :root {
 	--featured-event-bg: #0ea0d7;
+	--twentynineteen_font__body: "NonBreakingSpaceOverride", "Hoefler Text", "Baskerville Old Face", Garamond, "Times New Roman", serif;
```
paulmskim: I don't know what the convention for CSS variable names are, but I would stick to one here. The one above uses dashes, but you're using underscores.

src/resources/postcss/tribe-events-full.pcss
```diff
@@ -2364,6 +2368,117 @@ input[name*='tribe-bar-']:-moz-placeholder {
 	}
 }
 
+/* Twenty Nineteen
+========================================================================== */
+.tribe-theme-twentynineteen {
+
+	h1.tribe-events-page-title::before {
```
paulmskim: It looks like a common recurring thing with headings that the `::before` pseudo-element has a content that needs to be set to none. Rather than setting each one with a class, why don't we do something like:

```
h1,
h2,
h3,
h4,
h5,
h6 {

  &::before {
    content: none;
  }
}
```

or only the headings of concern (if we don't use h4 - h6 in our theme). If specificity is a concern then we can add more classes above this as necessary. This would also remove the repeated `::before` selectors below.

ggwicz: great call!

ggwicz:
I've implemented all of Paul's feedback; this one is ready for another review! 🔍

PR #2357, opened 2018-12-27
Date/time block conflicts when clicking to open the block options

juanfra:
🎫 https://central.tri.be/issues/119413

There was an overlap with the divs from PRO, and it was not possible to open the options when clicking the block.

On the other hand, added a div to wrap the `tribe-editor__subtitle__headline` contents, because for example the theme twentynineteen is adding styles to the `:before` pseudo element and it was not looking good as we use flex for that block.


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *B19.01*, but this was pull requested against *release/M19.01*.

PR #2364, opened 2019-01-07
Make sure we have the correct version of wp-browser

bordoni:
Based on the documentation of `wp-browser` we need to remove the dependency for `Codeception\Command\DbSnapshot` if we go to version `wp-browser@2.2.0` 

---

If we stick to version `2.1.6` we could keep using `DbSnapshot` and that could be a faster solution.

### DIFF
--- a/codeception.dist.yml
+++ b/codeception.dist.yml
@@ -22,4 +22,6 @@ extensions:
         - 'Codeception\Command\GenerateWPAjax'
         - 'Codeception\Command\GenerateWPCanonical'
         - 'Codeception\Command\GenerateWPXMLRPC'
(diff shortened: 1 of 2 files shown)


lucatume:
Same here as per https://github.com/moderntribe/events-pro/pull/1102: why not remove DBSnapshot and setup commands and use wp-browser `2.2.0`?

lucatume:
Since tests are passing go on.

bordoni:
Should we add the lock time here too? `10s` that you added for *Events Pro*

lucatume:
> Should we add the lock time here too? `10s` that you added for _Events Pro_

Yep, it would not hurt at all.

PR #2365, opened 2019-01-07
Don't apply wpautop if it hasn't been applied already.

Camwyn:
This is re-applying `wpautop` in places we don't want it - rather than re-removing it, let's set up a check.

Discovered while debugging https://central.tri.be/issues/120562

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -218,6 +218,7 @@ Still not happy? Shoot us an email to support@theeventscalendar.com or tweet to 
 = [4.7.4] TBD =
 
 * Fix - Display the tags block delimiter in the editor [119393]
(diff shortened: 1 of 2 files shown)


bordoni:
I like that, can we check if these fails on the tests are not related to this? I will make sure to fix em if they are not.

bordoni:
I think it was failing to upload the image to WP. on :

`UrlTest::it_should_not_reimport_image_when_attached_to_different_events`

@lucatume have you heard of this problem when using `attachment->create_upload_object`?

bordoni:
Actually lets get Changelog included.

Camwyn:
I do not believe that the tests have anything to do with the added check.

I'll add a changelog post-haste

PR #2366, opened 2019-01-08
Date Time Block > Fix switching back from all day slection

juanfra:
🎫 https://central.tri.be/issues/120266

**Fixing the following problem**

In the block editor, if I select the All Day option from the start or end time dropdown, I am then not able to switch back to an event with defined start and end times. Selecting a new start time from the dropdown doesn't work (it just goes back to All Day) and changing it in the plain text entry only changes it there, but the event still shows on the frontend as All Day.

📹 http://p.tri.be/C7fhuU


src/modules/data/blocks/datetime/sagas.js
```diff
@@ -164,6 +164,8 @@ export function* onHumanReadableChange() {
 
 		const isMultiDay = ! ( yield call( momentUtil.isSameDay, result.start, result.end ) );
 
+		const isAllDay = ! isMultiDay && ( '00:00' === moments.start.format('HH:mm') && '23:59' === moments.end.format('HH:mm') );
```
bordoni: Spacing around `'HH:mm'`

PR #2368, opened 2019-01-09
Block Editor > Save "Event Options" meta data

juanfra:
🎫 https://central.tri.be/issues/120168

When creating an event with the blocks editor, the logic doesn't get to the metadata saving methods, and all the event data is being saved via the wp rest api. I ended up intercepting the event creation in the blocks editor in order to update the meta for the "Event Options". Perhaps this is not the most "tidy" way to do this. But gutenberg doesn't leave us with many options.


src/Tribe/Meta/Save.php
```diff
@@ -181,4 +186,60 @@ protected function is_event() {
...
+			delete_metadata( 'post', $event_id, '_EventHideFromUpcoming' );
+		} else {
+			update_metadata( 'post', $event_id, '_EventHideFromUpcoming', $data['EventHideFromUpcoming'] );
+		}
+
+		// Set sticky state for calendar view.
+		if ( $event instanceof WP_Post ) {
+			if (
+				isset( $data['EventShowInCalendar'] )
+				&& $data['EventShowInCalendar'] == 'yes'
```
bordoni: Lets use `tribe_is_truthy()`

src/Tribe/Meta/Save.php
```diff
@@ -181,4 +186,60 @@ protected function is_event() {
...
+		if ( empty( $data['EventHideFromUpcoming'] ) ) {
+			delete_metadata( 'post', $event_id, '_EventHideFromUpcoming' );
+		} else {
+			update_metadata( 'post', $event_id, '_EventHideFromUpcoming', $data['EventHideFromUpcoming'] );
+		}
+
+		// Set sticky state for calendar view.
+		if ( $event instanceof WP_Post ) {
+			if (
+				isset( $data['EventShowInCalendar'] )
```
bordoni: Maybe use Array utils to fetch this variable instead of doing two `isset()`.

src/Tribe/Meta/Save.php
```diff
@@ -181,4 +186,60 @@ protected function is_event() {
...
+			) {
+				$update_event = array(
+					'ID'         => $event_id,
+					'menu_order' => '-1',
+				);
+				wp_update_post( $update_event );
+			} elseif (
+				(
+					! isset( $data['EventShowInCalendar'] )
+					|| 'yes' !== $data['EventShowInCalendar']
```
bordoni: Lets use `! tribe_is_truthy()`

bordoni:
It seems that it's breaking the tests, I've tried restarting the build lets see if it's a Travis problem.

juanfra:
No, the thing is that our implementation with classic editor (old one) has a nonce in place, and the test is checking that the events can be created with that nonce (which is not there in the block editor)

bordoni:
We should still find a way to fix the tests.

juanfra:
@bordoni done ✔️

PR #2373, opened 2019-01-10
Use "Tags" without textdomain as its a WP taxonomy.

juanfra:
🎫 https://central.tri.be/issues/119393

Nico suggested to remove the textdomain as it's a WP taxonomy.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not ')' | [event-tags.php:19](https://github.com/moderntribe/the-events-calendar/blob/4d56c750e9b343cf90c6de18d2b92e9f1d998ad6/src/views/blocks/event-tags.php#L19)

PR #2380, opened 2019-01-15
Spotfix/freemius activation

barryhughes:
@zbtirrell — this removes the menu element from our call to fs_dynamic_init(). With this change, immediately after activation, the user is taken to:

screenshot_2019-01-15 events lab wordpress


zbtirrell:
where do we get redirected after the "Allow"? (Go ahead and merge this either way - it's definitely better than the branch was :) )

PR #2383, opened 2019-01-18
Spotfix: Fix Event Width in 2019 Theme's Default Page Template — ⌗119689

ggwicz:
**Ticket:** [**⌗119689**](http://central.tri.be/issues/119689)


src/resources/postcss/tribe-events-full.pcss
```diff
@@ -2426,6 +2417,18 @@ input[name*='tribe-bar-']:-moz-placeholder {
 		}
 	}
 
+	&.single-tribe_events.tribe-events-page-template {
```
juanfra: Looking good! Perhaps I would put `tribe-events-page-template` below, in case we need to add more things for `single-tribe_events` in the future.

ggwicz: I hear you. .single-tribe_events has its own separate code block above though for now, so I think we can ship with this as-is and keep anything generally related to that template in that code block.

juanfra: Cool! Yeah, it's hard to review big files! haha. If it has it's own code block you could use `&.tribe-events-page-template` inside it. Or maybe it's just me that I usually avoid that level of specificity when we have code blocks that can be inside.

PR #2386, opened 2019-01-21
Fix bad ref

Camwyn:
The product-taskmaster ref is to a deleted branch. Build fatals.


sc0ttkclark:
Will merge up via G19.02 directly

PR #2391, opened 2019-01-23
#121486 - if et is older version disable older ce to prevent fatal

jesseeproductions:
_Ref:_ [C#121486](https://central.tri.be/issues/121486)

This disables Pre PDC Version of CE if TEC shuts down due to an older version of ET.


Camwyn:
Do we need this in TEC? I hate to see the cross-referencing 😢

sc0ttkclark:
I think the only way this works is if it's here based on the problem. This looks good to go 👍

jesseeproductions:
It is needed in TEC because I cannot modify the coding in CE as that would already be on customer sites.

PR #2407, opened 2019-02-06
#122368 - prevent FB Ext from running

jesseeproductions:
prevent FB Ext from running of ET is on an older version and TEC shutsdown

_Ref:_ [C#122368](https://central.tri.be/issues/122368)


sc0ttkclark:
Are we still removing all actions instead?

jesseeproductions:
I created a new PR and going to close this one.

PR #2415, opened 2019-02-13
Ensure Only Eventbrite Imports Can Use New (do not override) Status — ⌗122648

ggwicz:
**Ticket:** [**⌗122648**](http://central.tri.be/issues/122648)


ggwicz:
I had this tagged `hold` while I just did some quick tweaks of my own. Those are done, so this is now ready for code review! 🔍

PR #2419, opened 2019-02-14
Add Image Widget Plus unload

bordoni:
_Ref:_ [C#121977](http://central.tri.be/issues/121977)
---

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -316,6 +316,8 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 			// let's initialize tec
 			add_action( 'plugins_loaded', array( $this, 'plugins_loaded' ), 0 );
 
+			// Prevents Image Widget Plus from been problematic
+			$this->compatibility_unload_iwplus_v102();
 		}
 
 		/**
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -584,6 +586,32 @@ protected function loadLibraries() {
...
+
+			if ( ! defined( 'Tribe__Image__Plus__Main::VERSION' ) ) {
+				return;
+			}
+
+			if ( ! version_compare( Tribe__Image__Plus__Main::VERSION, '1.0.2', '<=' ) ) {
+				return;
+			}
+
+			remove_action( 'plugins_loaded', array( Tribe__Image__Plus__Main::instance(), 'plugins_loaded' ), 0 );
```
jesseeproductions: Do we have time to add a notice like we do when we shutdown because of an older version of ET: https://github.com/moderntribe/the-events-calendar/blob/c68d9c6280c499503072b2d5ffdbf0f383af786a/src/Tribe/Main.php#L1449

That way at least we have some warning to people that IW+ is not running.

bordoni: I wanted to avoid any notices. Since we are going to release IW+ hotfix today I just go a sign-off from Zach that we dont need it if we can get the release on the same day.

PR #2425, opened 2019-02-26
When EA service fails, we shouldnt throw a notice with EBT

bordoni:
_Ref:_ [C#123383](http://central.tri.be/issues/123383)
---

### DIFF
--- a/src/Tribe/Aggregator/Service.php
+++ b/src/Tribe/Aggregator/Service.php
@@ -355,11 +355,13 @@ class Tribe__Events__Aggregator__Service {
 		$response = $this->get( 'eventbrite/validate', $args );
 
 		// If we have an WP_Error we return only CSV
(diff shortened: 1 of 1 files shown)


src/Tribe/Aggregator/Service.php
```diff
@@ -355,11 +355,13 @@ public function has_eventbrite_authorized() {
 		$response = $this->get( 'eventbrite/validate', $args );
 
 		// If we have an WP_Error we return only CSV
-		if ( $response instanceof WP_Error ) {
+		if ( is_wp_error( $response ) ) {
```
lucatume: Why this? `instanceof` is faster and is what the function is doing.

bordoni: I just went with what we used in other places in our code. but you are right. Moving back to `instanceof WP_Error`

PR #2427, opened 2019-02-27
 Modify behavior for the UTC notice, resets 4 times a year

bordoni:
_Ref:_ [C#123221](http://central.tri.be/issues/123221)
---

### DIFF
--- a/src/Tribe/Admin/Notice/Timezones.php
+++ b/src/Tribe/Admin/Notice/Timezones.php
@@ -17,6 +17,14 @@ class Tribe__Events__Admin__Notice__Timezones {
 	public function hook() {
 		$date = $this->get_current_reset_date();
 		$slug = $this->slug;
+		/**
+		 * Allows users to completely deactivate the resetting of the Day Light savings notice
+		 *
+		 * @since  TBD
+		 *
+		 * @param  bool
+		 */
+		$should_reset = apply_filters( 'tribe_events_admin_notice_daylight_savings_reset_notice', true );
 
 		// If we have a date append to the Slug
 		if ( $date ) {
@@ -133,6 +141,4 @@ class Tribe__Events__Admin__Notice__Timezones {
 		return sprintf( implode( '<br />', $text ), $link, $current_utc );
 
 	}
-
-
 }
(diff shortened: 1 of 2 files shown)


bordoni:
It will be added on the end of the release.

src/Tribe/Admin/Notice/Timezones.php
```diff
@@ -67,17 +138,20 @@ public function notice() {
...
 			'<a href="%1$s" target="_blank">%2$s</a>',
 			esc_url( $url ),
 			esc_html__( 'Read more', 'the-events-calendar' )
 		);
-		$text = __( 'When using The Events Calendar, we recommend that you choose a city in your timezone and avoid using a UTC timezone offset. Choosing a UTC timezone may cause problems when importing events or with Day Light Savings time. %1$s', 'the-events-calendar' );
 
-		return sprintf( $text, $link );
-
-	}
+		$text[] = __( 'When using The Events Calendar, we recommend that you geographic timezone such as "America/Los Angeles" and avoid using a UTC timezone offset such as “%2$s”.', 'the-events-calendar' );
```
ggwicz: should be `we recommend that you use a geographic timezone such as "America/Los_Angeles"`

src/Tribe/Admin/Notice/Timezones.php
```diff
@@ -1,25 +1,85 @@
...
+		$date = $this->get_current_reset_date();
+		$slug = $this->slug;
+		/**
+		 * Allows users to completely deactivate the resetting of the Day Light savings notice
+		 *
+		 * @since  TBD
+		 *
+		 * @param  bool
+		 */
+		$should_reset = apply_filters( 'tribe_events_admin_notice_daylight_savings_reset_notice', true );
```
ggwicz: seems like we're not really using this boolean to truly hide the notice; shouldn't it, for example, be down in the `should_display()` method?

if that's not the goal of this, i'd recommend naming the var better and/or making the docblock comment more clear

bordoni: That is for reseting the notice or not

src/Tribe/Admin/Notice/Timezones.php
ggwicz: hey @bordoni, you changed the time zone string in the notice which is correct, but you also still have to fix `we recommend that you geographic`

this should be `we recommend that you use a geographic`

PR #2437, opened 2019-03-07
Fix Day view AJAX issue related to featured events.

lucatume:
Ticket: https://central.tri.be/issues/123967

This PR fixes an issue where the Day view, when displaying events in AJAX context, would not show recurring events.
The issue is due to the fact that in our front-end code (e.g. the Day view) we send this information to the backend:
`featured = false` means "all events, not just the featured ones"
`featured = true` means "only featured events"

The ORM uses, instead, this logic:
* if the `featured` filter is not set "all events, not just the featured ones"
* if `featured = true` "only featured events"
* if `featured = false` only non-featured events.

At the last point, the behavior diverges and has to be adjusted.


tests/wpunit/Tribe/Events/ORM/Venues/FetchTest.php
```diff
@@ -81,10 +81,9 @@ public function should_allow_getting_venues_by_phone() {
 	 *
 	 * @test
 	 */
-	public function should_allow_getting_venues_by_postal_code() {
+	public function jhould_allow_getting_venues_by_postal_code() {
```
nealfennimore: jhould this be jhould?

juanfra: `jhould`?

lucatume: It's Danish for "it should"; I'm surprised you don't know.

lucatume: Ok, this is getting out of hand: as soon as I try to educate someone about foreign languages everyone is complaining.

PR #2438, opened 2019-03-08
Get post content from passed post, not the_post

exallium:
This was causing an error where a plugin continuously rendered a
shortcode, as it was pulling in the content for the page displaying
the given event, rather than for the event itself.


bordoni:
Resolved this and another bug we have with two new method uses on Tribe Common.
`tribe_get_the_content()` and `tribe_the_content()` which will make it cross compatible with multiple versions of PHP and WP. Released with Version `4.9.12`.

PR #2446, opened 2019-04-04
Fix(Main Events Query): Ensure UTC-based filtering of dates is applied to all query components.

lucatume:
Ticket: https://central.tri.be/issues/125196

In implementing the ORM we've modified the behaviour of event queries to use the UTC dates in place
of the local ones. Before we would incorrectly display an 8AM Tokyo event (UTC+9) as happening
before a 9AM New York event (UTC-4). This fix makes sure the change is applied across all event
queries to ensure consistency between specialized queries using UTC dates and the main query that
would use, until now, non UTC dates.

### DIFF
--- a/src/Tribe/Template/Month.php
+++ b/src/Tribe/Template/Month.php
@@ -603,7 +603,18 @@ if ( ! class_exists( 'Tribe__Events__Template__Month' ) ) {
 			 */
 			$events_in_month = apply_filters( 'tribe_events_month_get_events_in_month', null, $start_date, $end_date );
 
+
 			if ( null === $events_in_month ) {
+				/**
+				 * Documented in the `\Tribe__Events__Query::pre_get_posts` method.
+				 *
+				 * @see   \Tribe__Events__Query::pre_get_posts()
+				 *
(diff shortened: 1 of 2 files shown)


src/Tribe/Template/Month.php
```diff
@@ -603,7 +603,18 @@ protected function set_events_in_month() {
...
+
 			if ( null === $events_in_month ) {
+				/**
+				 * Documented in the `\Tribe__Events__Query::pre_get_posts` method.
+				 *
+				 * @see   \Tribe__Events__Query::pre_get_posts()
+				 *
+				 * @since TBD
+				 */
+				$use_utc = apply_filters( 'tribe_events_query_event_utc_start_date', true );
```
bordoni: Pass the query here too. As the Second param.

lucatume: There is no query to pass here and passing the global one will mislead developers in thinking that query is the one that's being filtered.
I've improved the doc block on the `pre_get_posts` filter to add `null` as a second, possible, type for the `$query` parameter.

PR #2447, opened 2019-04-04
Make the filterbar featured event filter work with ORM

juanfra:
🎫 https://central.tri.be/issues/125259

From what I talked with Luca, in ORM, if:

- the featured arg is true, then it brings only featured events.
- the featured arg is false, then it brings not featured events.
- not defined, brings all.

Adding the if statement to make this work with the filter bar. Maybe in the future we can revisit to see if we can hook to `tribe_events_month_daily_events_query_args` directly. I'm using both `featured` and `tribe_featuredevent` to get both cases working.

### DIFF
--- a/src/Tribe/Template/Month.php
+++ b/src/Tribe/Template/Month.php
@@ -896,7 +896,10 @@ if ( ! class_exists( 'Tribe__Events__Template__Month' ) ) {
 			);
 
 			// If the request is false or not set we assume the request is for all events, not just featured ones.
-			if ( tribe_is_truthy( tribe_get_request_var( 'tribe_featuredevent', false ) ) ) {
+			if (
+				tribe_is_truthy( tribe_get_request_var( 'featured', false ) )
+				|| tribe_is_truthy( tribe_get_request_var( 'tribe_featuredevent', false ) )
+			) {
 				$args['featured'] = true;
 			}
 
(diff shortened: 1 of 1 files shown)


juanfra:
ORM fix, so it shouldn't need changelog entry.

src/Tribe/Template/Month.php
```diff
@@ -895,6 +895,14 @@ private function get_daily_events( $date ) {
 				), $this->args
 			);
 
+			// If the request is false or not set we assume the request is for all events, not just featured ones.
+			if (
+				tribe_is_truthy( tribe_get_request_var( 'featured', false ) )
+				|| tribe_is_truthy( tribe_get_request_var( 'tribe_featuredevent', false ) )
+			) {
+				$args['featured'] = true;
```
Camwyn: Do you not want to set it to false? (I see that empty is the same, but wonder if there is benefit to setting false)
To set either way you could do:
`$args['featured'] = tribe_is_truthy( tribe_get_request_var( 'featured', false ) ) || tribe_is_truthy( tribe_get_request_var( 'tribe_featuredevent', false ) );`

juanfra: Yeah, the thing is the following (I left a comment regarding this in the PR description). From what I talked with Luca, in ORM, if:

- the featured arg is true, then it fetches only featured events.
- the featured arg is false, then it fetches not featured events.
- not defined, fetches all.

That's why we have it inside the if, because if the query var is not there, we shouldn't be adding the `$args['featured']` with false.

sc0ttkclark: 👍

PR #2448, opened 2019-04-05
fix some Freemius fatals

zbtirrell:
Not sure what ticket to associate this with…

### DIFF
--- a/src/Tribe/Integrations/Freemius.php
+++ b/src/Tribe/Integrations/Freemius.php
@@ -53,6 +53,11 @@ class Tribe__Events__Integrations__Freemius {
 			return;
 		}
 
+		// if the common that loaded doesn't include Freemius, let's bail
+		if ( ! class_exists( 'Tribe__Freemius' ) ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Integrations/Freemius.php
```diff
@@ -53,6 +53,11 @@ public function __construct() {
 			return;
 		}
 
+		// if the common that loaded doesn't include Freemius, let's bail
+		if ( ! class_exists( 'Tribe__Freemius' ) ) {
```
barryhughes: Don't need to debate too hard on this, because in practice I'll bet this will work just fine... but isn't the 'better' solution to wrap our call to `tribe( ... )` in a try/catch block? That way, if in a future change we alter the bound class for `tribe( 'freemius' )` from `Tribe__Fremius` to something else, we don't need to worry about also updating this test.

lucatume: The container implements the `ArrayAcces` interface.

PR #2451, opened 2019-04-09
Freemius > Maybe remove activation complete notice

juanfra:
🎫 https://central.tri.be/issues/125612

Freemius notices cannot be dismissed (at least in our implementation). This code will remove the activation complete notice.


src/Tribe/Integrations/Freemius.php
```diff
@@ -178,4 +180,34 @@ public function filter_connect_message_on_update(
...
 	}
+
+	/**
+	 * Method to remove the sticky message when the plugin is active for freemius
+	 *
+	 * @since  TBD
+	 *
+	 * @return void
+	 */
+	public function maybe_remove_activation_complete_notice() {
```
bordoni: Should we attach this to `admin_init@P1`? 
Make sure the methods and classes below exist before calling, to avoid fatals.

PR #2454, opened 2019-04-10
Only Display Freemius to 10% to New Customers

bordoni:


src/Tribe/Integrations/Freemius.php
```diff
@@ -65,7 +65,7 @@ public function __construct() {
 		 *
 		 * @param  bool  $should_load
 		 */
-		$should_load = apply_filters( 'tribe_events_integrations_should_load_freemius', true );
+		$should_load = apply_filters( 'tribe_events_integrations_should_load_freemius', $this->should_load( 10 ) );
```
Camwyn: Do we need to pass the default value here?

src/Tribe/Integrations/Freemius.php
```diff
@@ -96,6 +96,51 @@ public function __construct() {
...
+	 * When should we load Freemius to users
+	 *
+	 * @since  TBD
+	 *
+	 * @param  integer $threshold Percentage of which we will load Freemius
+	 *
+	 * @return boolean
+	 */
+	public function should_load( $threshold = 10 ) {
+		$previous_versions = Tribe__Settings_Manager::get_option( 'previous_ecp_versions', [] );
```
juanfra: is this checking for ECP or TEC?

bordoni: That was the string that was set way back when it was created. But it's right

src/Tribe/Integrations/Freemius.php
bordoni: Just for future reference, we will change that int there in the future.

PR #2455, opened 2019-04-10
Add values to avoid JSON+LD warnings on the Google console.

juanfra:
🎫 https://central.tri.be/issues/120193


src/Tribe/JSON_LD/Event.php
```diff
@@ -154,8 +165,13 @@ public function get_data( $posts = null, $args = array() ) {
 	 * @return string
 	 */
 	protected function normalize_price( $price ) {
+
+		// Make it work with different languages
+		$regex_free  = '/^\\s*' . __( 'Free', 'the-events-calendar' ) . '\\s*$/i';
```
bordoni: We should make sure to keep free in English, as well as the translation.

juanfra: @bordoni Json LD price doesn't take `free` as a value in `offers.price`, it's required to be a number. And 0 for the "Free" cases.

As we're using `tribe_get_cost` to get the price there, we receive the string when it's free, and that's the purpose of the regex. It was working for english, replacing `Free` with `0`. But that's only english. What happens with sites in other languages?

This improvement will make it work when the site is in other languages (As it uses the catalog translations. For example: `Gratis` for german, `Gratuito` in spanish, and so on, and so forth).

This is something the customers were complaining of and wasn't part of the scope of the task, but this code fix it. And we should keep the code, if we want to solve the problem.

PR #2458, opened 2019-04-12
Respect the correct post Status for List Widget

bordoni:
_Ref:_ [C#125787](http://central.tri.be/issues/125787)
---
On this particular change here we are basically adding the `post_status` default, should we have this on the `tribe_get_events` level? for backwards compatible, it current is using `any` which causes problems.


lucatume:
Merging this.

PR #2460, opened 2019-04-13
Resolve SQL errors around ORM and direct modifications to SQL

bordoni:
_Ref:_ [C#125795](http://central.tri.be/issues/125795)
---

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -52,6 +52,13 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 			}
 
 			/**
+			 * When using the ORM we cannot use EventStartDate injection
+			 */
+			if ( isset( $query->builder ) && $query->builder instanceof Tribe__Repository ) {
+				$can_inject = false;
+			}
+
+			/**
 			 * Determine whether a date field can be injected into various parts of a query.
 			 *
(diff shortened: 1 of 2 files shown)


src/Tribe/Query.php
```diff
@@ -51,6 +51,13 @@ public static function can_inject_date_field( $query ) {
 				$can_inject = false;
 			}
 
+			/**
+			 * When using the ORM we cannot use EventStartDate injection
+			 */
+			if ( isset( $query->builder ) && $query->builder instanceof Tribe__Repository ) {
```
mitogh: If `$query` is an object the `isset` is not required as `instanceof` will return `false` if is not defined.

bordoni: On this case the `builder` param might now always be set, since we are relying on maybe been a normal WP_Query.

mitogh: I see but `$query` is defined right, which in this case `$query->builder` is `null` and in that case `instanceof` return `false`

bordoni: What I meant, is that `$query` is based on `WP_Query` which doesn't by definition have a `builder` param, which would lead to a notice of trying to fetch an un-existent param on the object.

mitogh: sounds good

PR #2461, opened 2019-04-15
fix(AJAX Pagination): Fix a number of issues with AJAX pagination in local timezone.

lucatume:
These fixes makes sure AJAX pagination keeps working correctly when using the "Use manual timezones
for each event" Timezone Setting. In the List view the issues were caused on a the "now" date not
being "recycled" when generating the hash, in the other views the issue was caused by duplicate
event date meta queries caused by the way the next and previous links queries work by "recycling"
the global query arguments.

re https://central.tri.be/issues/123950

screencast: https://github.com/moderntribe/the-events-calendar/pull/2461


src/Tribe/Service_Providers/ORM.php
```diff
@@ -23,5 +23,42 @@ public function register() {
...
+			return $args;
+		}
+
+		if ( empty( $query->builder ) || ! $query->builder instanceof Tribe__Repository__Interface ) {
+			return $args;
+		}
+
+		$args['meta_query'] = tribe_filter_meta_query(
+			$args['meta_query'],
+			array( 'key' => '/_Event(Start|End)Date(UTC)*/' )
```
bordoni: `[]` Array.

PR #2468, opened 2019-04-23
Remove featured image if the value is empty

mitogh:
If the value of `FeaturedImage` on the request is present but is set to
null or empty value it means we want to remove the image or update the
resource with an empty image, as right now the check was in place to
review if the request has a valid image to update the image value.

Updates: 

- Applies the same logic for the featured image of an Event.

Ticket: https://central.tri.be/issues/120913

### DIFF
--- a/src/Tribe/API.php
+++ b/src/Tribe/API.php
@@ -156,8 +156,12 @@ if ( ! class_exists( 'Tribe__Events__API' ) ) {
 				$data['EventCost'] = reset( $data['EventCost'] );
 			}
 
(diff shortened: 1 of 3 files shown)


mitogh:
BTW do you think we can have enable the merge without travis to be pass as the tests are already failing on this bucket branch.

I think the plan is to move to `master` once RBE is completed so this change needs to be ported back to `master`.

Thanks.

nealfennimore:
@bordoni would you be able to merge this in for us?

PR #2469, opened 2019-04-23
Fix slug translation with WPML

dgwatkins:
The `wpml_st_load_slug_translation` function was removed from String Translation and so was `WPML_Slug_Translation::get_translations`.

I rewrote it in a way that it works again and added support for translating archive slugs.

🎫 https://central.tri.be/issues/127434


juanfra:
Thanks for the contribution @dgwatkins! I'd highly appreciate if you could send me the dev version via email and maybe we can have a short conversation over there about backwards compat.

PR #2471, opened 2019-04-24
check if variable exists before using it

zbtirrell:
This was reported in a support thread. I can't reproduce the issue - however, I can see where it could occur in the code, so, fixing.


bordoni:
Reopening, Github likes to close PRs when you delete base branch.

PR #2479, opened 2019-04-30
Kitchen Sink Templating + Template Bootstraping

bordoni:
_Ref:_ [C#126213](http://central.tri.be/issues/126213)
---


src/functions/views/provider.php
```diff
@@ -0,0 +1,35 @@
...
+	if ( defined( 'TRIBE_EVENTS_V2_VIEWS' ) ) {
+		return (bool) TRIBE_EVENTS_V2_VIEWS;
+	}
+
+	$env_var = getenv( 'TRIBE_EVENTS_V2_VIEWS' );
+	if ( false !== $env_var ) {
+		return (bool) $env_var;
+	}
+
+	$enabled = (bool) tribe_get_option( View::$option_enabled, false );
```
paulmskim: might want to `use` the `View` class here, getting a fatal if I don't define a constant.

src/Tribe/Views/V2/Hooks.php
```diff
@@ -0,0 +1,130 @@
...
+ * To remove an action:
+ * remove_action( 'some_action', [ tribe( Tribe\Events\Views\V2\Hooks::class ), 'some_method' ] );
+ * remove_action( 'some_action', [ tribe( 'views-v2.hooks' ), 'some_method' ] );
+ *
+ * @package Tribe\Events\Views\V2
+ * @since TBD
+ */
+
+namespace Tribe\Events\Views\V2;
+
```
juanfra: I think we should include Tribe__Events__Rewrite here. `use Tribe__Events__Rewrite as Rewrite;`

PR #2484, opened 2019-05-01
Prevent notices around multiple Queries with `start_date`

bordoni:
_Ref:_ [C#126790](http://central.tri.be/issues/126790)
---

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -1278,6 +1278,11 @@ if ( ! class_exists( 'Tribe__Events__Query' ) ) {
 					Arr::get( $args, 'start_date' ),
 					Arr::get( $args, 'startDate' ),
 					Arr::get( $args, 'starts_after' ),
+					Arr::get( $args, 'starts_before' ),
+					Arr::get( $args, 'end_date' ),
+					Arr::get( $args, 'endDate' ),
+					Arr::get( $args, 'ends_after' ),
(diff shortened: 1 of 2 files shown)


src/Tribe/Query.php
```diff
@@ -1274,11 +1274,20 @@ public static function getEvents( $args = array(), $full = false ) {
 				}
 
 				$display = Arr::get( $args, 'eventDisplay' );
-				$possible_start_date = Arr::get( $args, 'start_date', Arr::get( $args, 'startDate' ) );
+				$has_date_args = array_filter( [
```
juanfra: why are you using array filter here?

bordoni: To prevent having empty results. Without a callback `array_filter` will filter by empty values.

juanfra: Then it would just work with the conditional you're using next? `! $has_date_args` or you should be checking if empty?

bordoni: No, because if we need to remove the non-existent entries before checking if there is something on `$has_date_args`because otherwise, it would always be true. Since the array would have a few indexes and wouldn't validate that conditional.

juanfra: Yeah, what I mean is that you're using a bool comparison to check an array. After it's filtered. But I assume you've checked already the different cases.

PR #2488, opened 2019-05-06
Connect main query to View v2 render system, add tests

lucatume:
Based on: https://github.com/moderntribe/tribe-common/pull/1003

Tickets:
* https://central.tri.be/issues/126078
* https://central.tri.be/issues/126077
* https://central.tri.be/issues/126076

This PR adds an important building block to the View v2 system: main query (PHP initial state) rendering.
On an high level all the Views will work like this:
1. from a Context get some ORM/Repository args
2. from the Repository fetch the matching posts
3. from the matching posts the View will use a Template to render

This series of commits adds the step `0`, the one where the Context, instead of being built from a REST request, a widget, a shortcode or some other entry point is built as a consequence of the main query.
The Views v2 service provider(s) take care of removing the filters used from the v1 implementation of the views (see integration tests to "detect" them), to then hook on the main query and inject matching posts in it.
The flow above then follows.

I've avoided speculating about what we might, or might not, need and kept the code lean.
Since the structure is there reading more, or less, from the context will be just a question of adding entries to it.


juanfra:
This is awesome. What do you think, in order to keep the compat hooks independent, if we add some hooks on the methods of `/V2/Hooks.php` and then we create a file specifically for v1 compatibility? That way we could keep the v2 implementation a bit cleaner and we trace a line between the implementation of v2 and the compatibility with v1. Perhaps it doesn't make much sense right now, but as soon as we start finding more things we can end up having a lot of ifs and special code in the new files just to maintain compatibility.

lucatume:
> This is awesome. What do you think, in order to keep the compat hooks independent, if we add some hooks on the methods of `/V2/Hooks.php` and then we create a file specifically for v1 compatibility? That way we could keep the v2 implementation a bit cleaner and we trace a line between the implementation of v2 and the compatibility with v1. Perhaps it doesn't make much sense right now, but as soon as we start finding more compatibility things/issues we might end up having a lot of ifs and special code in the new files just to maintain compatibility.

This is a great idea, I've merged this PR and opened a new PR implementing this proposal in here: https://github.com/moderntribe/the-events-calendar/pull/2490.

PR #2491, opened 2019-05-07
removing po files to save space

zbtirrell:
See: https://central.tri.be/issues/126690


zbtirrell:
I'd love to have Victor test this branch before it is merged. It should save about 15MB

PR #2496, opened 2019-05-10
PHP and JS for Views Manager and template injection

bordoni:
_Ref:_ [C#126086](http://central.tri.be/issues/126086)
---

PR #2499, opened 2019-05-10
VR > Add list view v2 templates

juanfra:
🎫 https://central.tri.be/issues/126102

With this PR we're basically covering the base HTML via templates for the new events bar, the top bar and the different events on the list view. Ideally there's some more things that require some more time (responsiveness, I missed some SVGs for the featured and recurring, and a couple more things).

I also wanted to talk with Paul next week about how is that we're gonna handle spacing and we are going to use the grid system for all views.

📹 https://www.loom.com/share/ae91702f8a91443986f4d81076d8b4d0

🔗 [The HTML of the Events bar](https://gist.github.com/juanfra/d75122c7347fd8d870ef6d91fdf00567)
🔗 [The HTML of the Top Bar](https://gist.github.com/juanfra/a09d70d8dfa9fef54bee6bc6dc4f2395)
🔗 [The HTML of the list view](https://gist.github.com/juanfra/17e397efaac97f2ee03736fe64cf09fd) - _for the featured event we're only adding a featured class to the event container_


juanfra:
🗒 **NOTE** - I've added the variables and mixins here, ideally they'll live in their own repo afterwards.

src/views/v2/list/nav/prev.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @version TBD
+ *
+ */
+if ( ! tribe_has_previous_event() ) {
+	return;
+}
+?>
+<a
+	href="<?php echo esc_url( tribe_get_listview_prev_link() ); ?>"
```
bordoni: These links need to Passed in, to allow for tesing.

src/views/v2/list/nav/prev.php
```diff
@@ -0,0 +1,25 @@
...
+	rel="prev"
+	class="tribe-common-c-nav--prev"
+>
+	<?php echo esc_html( sprintf( __( 'Previous %s', 'the-events-calendar' ), tribe_get_event_label_plural() ) ); ?>
```
bordoni: As well as the label.

src/views/v2/list/month-separator.php
```diff
@@ -0,0 +1,18 @@
...
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+?>
+<div class="tribe-events-calendar-list__separator--month">
+	<span>May</span>
```
bordoni: Pass this to the template.

src/views/v2/list/event/title.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+
+$event_id = $event->ID;
```
bordoni: Why are we using globals? We should bring it from context.

src/views/v2/list/event/featured-image.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+
+$event_id = $event->ID;
+
+if ( ! has_post_thumbnail( $event_id ) ) {
```
bordoni: Same here, we need to stop using Methods like this, prevents testing.

src/views/v2/list/event/featured-image.php
```diff
@@ -0,0 +1,25 @@
...
+	return;
+}
+
+?>
+<div class="tribe-events-calendar-list__event--featured-image">
+	<?php echo get_the_post_thumbnail( $event_id, 'large' ); ?>
```
bordoni: Same here.

src/views/v2/list/event/date.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @version TBD
+ *
+ */
+
+$event_id = $event->ID;
+?>
+<div class="tribe-events-calendar-list__event--datetime">
+	<time datetime="1970-01-01T00:00:00+00:00">
+		<?php echo tribe_events_event_schedule_details(); ?>
```
bordoni: Let's build a static version of this, for now, a template that replaces this function.

src/views/v2/list.php
```diff
@@ -1,17 +1,36 @@
...
+ *
+ * @version TBD
+ *
+ */
+
+$events_label_plural = tribe_get_event_label_plural();
+
+$this->template( 'events-bar' );
+
+$this->template( 'top-bar' );
```
bordoni: maybe call it header? To avoid having `filter-bar`, `events-bar` and `top-bar`

src/views/v2/events-bar/views.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ */
+?>
+<div class="tribe-events-calendar-events-bar--views">
+	<h3 class="tribe-common-a11y-visual-hide"><?php printf( esc_html__( '%s Views Navigation', 'the-events-calendar' ), tribe_get_event_label_singular() ); ?></h3>
+	<div class="tribe-common-form-control-tabs">
+		<button id="tabButton" aria-haspopup="listbox" aria-labelledby="tabButton" aria-expanded="true"><?php esc_html_e( 'Views', 'the-events-calendar' ); ?></button>
+		<ul tabindex="-1" role="listbox" aria-activedescendant="tabOneLabel">
+			<li role="presentation">
+				<input id="tabList" name="tabGroup" type="radio" value="tabList" checked="checked" />
```
bordoni: Same stuff with the ID camelCased.

src/views/v2/events-bar/form/keyword.php
```diff
@@ -0,0 +1,26 @@
...
+ * @version TBD
+ *
+ */
+?>
+<div class="tribe-common-form-control-text">
+	<label for="keyword"><?php esc_html_e( 'Keyword', 'the-events-calendar' ); ?></label>
+	<input
+		type="text"
+		id="keyword"
+		name="keyword"
```
bordoni: Let's move towards using `tribe-views[search]`?

src/views/v2/events-bar/form.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @version TBD
+ *
+ */
+
+$current_url = tribe_events_get_current_filter_url();
+?>
+<div class="tribe-events-calendar-events-bar--form">
+
+	<form action="<?php echo esc_attr( $current_url ); ?>" method="post" class="tribe-common-c-search">
```
bordoni: The form will move out. The whole View will be the form.
Check here: https://github.com/moderntribe/the-events-calendar/pull/2496/files#diff-a94dec2f05fa33e5969767bcecff4943R34

src/views/v2/events-bar/form/keyword.php
paulmskim: Let's talk about this, as we never really set on how we're passing information to the BE via inputs (what format BE needs, what format works for FE, etc.)

src/views/v2/events-bar/form/submit.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+?>
+<input
```
paulmskim: Let's use `<button type="submit">` here instead. Also no need for `aria-label`, `name`, or `value`.

src/views/v2/events-bar/form.php
paulmskim: If I'm understanding you correctly, Gustavo, you're saying that this should be all one view and not multiple pieces (using `$this->template()` below)? This is my feel, that this search (form) should be one view, as we even have a search component defined in `tribe-common`. That should have all the styles for the form.

src/resources/postcss/components/_events-bar.pcss
```diff
@@ -0,0 +1,28 @@
...
+
+.tribe-events-calendar-events-bar {
+	display: flex;
+	width: 100%;
+	flex-direction: row;
+	flex-wrap: wrap;
+	align-items: center;
+}
+
+.tribe-events-calendar-events-bar--form {
```
paulmskim: let's make this `.tribe-events__events-bar-form`

src/resources/postcss/components/_top-bar.pcss
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+
+	.tribe-events-calendar-top-bar--today {
+		display: flex;
+		flex: 1 1 60%;
+		align-items: center;
+	}
+
+	.tribe-events-calendar-top-bar--today--button {
```
paulmskim: These styles should come from the class `.tribe-events-c-btn-border` I believe.

src/resources/postcss/components/_top-bar.pcss
```diff
@@ -0,0 +1,61 @@
...
+		flex-wrap: wrap;
+		align-items: center;
+	}
+
+	.tribe-events-calendar-top-bar--nav {
+		display: flex;
+		padding-right: 15px;
+		align-items: center;
+
+		nav a {
```
paulmskim: We should target all elements with classes rather than nesting like so. Every element should have a unique class name using BEM format so that we should really only be using one class.

For example, we may have `.tribe-events__top-bar-nav-wrapper`, `.tribe-events__top-bar-nav`, and `.tribe-events__top-bar-nav-link`, for example.

src/views/v2/top-bar/nav.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+?>
+<div class="tribe-events-calendar-top-bar--nav">
+	<nav>
+		<a href="#" class="tribe-common-c-btn-icon tribe-common-c-btn-icon--caret-left tribe-common-b3"></a>
```
paulmskim: All our navs should be a `<ul>` with each link wrapped in `<li>`. So for example:

```
<nav class="">
    <ul class="">
        <li class="">
            <a href="#" class=""></a>
        </li>
        ...
    </ul>
</nav>
```

src/views/v2/list/nav.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+?>
+<nav
+	class="tribe-common-c-nav"
+	aria-label="<?php echo esc_html( sprintf( esc_html__( '%s List Navigation', 'the-events-calendar' ), tribe_get_event_label_plural() ) ); ?>"
```
paulmskim: Let's omit this. The `<nav>` element already communicates that this is a navigation element.

src/resources/postcss/views/_list.pcss
```diff
@@ -0,0 +1,113 @@
+/**
+ * The Events Calendar Tribe List View V2 Stylesheet
+ *
+ * @package TribeEventsCalendar
+ *
+ */
+.tribe-events {
```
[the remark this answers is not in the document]
paulmskim: Is there a reason why we shouldn't use `tribe-events` here @bordoni ? We defined this in the CSS planning document and it would be good if we didn't have to switch it up here. It also helps us follow BEM format.

paulmskim:
Sweet stuff here @juanfra , and love the screencast. Most things here are things on HTML and CSS semantics, but visually things are looking great. Let's get together on Monday to chat.

src/views/v2/events-bar/form/submit.php
juanfra: Any reason in particular to go for `button` instead of `input`?

src/views/v2/list.php
juanfra: I followed the terminology used by the designers -> https://moderntribe.invisionapp.com/d/main/#/console/17163687/359236873/preview

Maybe we can discuss a bit about it?

src/views/v2/list/event/featured-image.php
juanfra: Do you mean to pass the post thumbnail to the template? I don't get it. The info should be retrieved in the template part. We can pass the event variable and then test according to that.

src/resources/postcss/views/_list.pcss
```diff
@@ -0,0 +1,113 @@
...
+
+	.tribe-events-calendar-list {
```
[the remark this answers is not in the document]
juanfra: @paulmskim thoughts about naming the main views wrappers?

src/views/v2/events-bar/form/submit.php
paulmskim: Only in that buttons allow for a wider range of rendering possibilities, input is only limited to text (whatever is defined in the value attr).

PR #2506, opened 2019-05-15
VR > Add body classes for themes with compatibility

juanfra:
🎫 https://central.tri.be/issues/127066

Add body classes for the themes we'll include compatibility.

### DIFF
--- a/src/Tribe/Views/V2/Hooks.php
+++ b/src/Tribe/Views/V2/Hooks.php
@@ -182,8 +182,7 @@ class Hooks  extends \tad_DI52_ServiceProvider {
 	}
 
 	/**
-	 * Filters the body classes.
-	 * Add theme compatibility classes.
+	 * Filters the body classes to add theme compatibility ones.
 	 *
 	 * @since TBD
 	 */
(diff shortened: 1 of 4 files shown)


lucatume:
I've restarted the build as it seem it did run into some hiccups.

src/Tribe/Views/V2/Hooks.php
```diff
@@ -179,4 +180,14 @@ public function filter_query_vars( array $query_vars = [] ) {
 
 		return $query_vars;
 	}
+
+	/**
+	 * Filters the body classes.
+	 * Add theme compatibility classes.
```
lucatume: Add a blank line here or, better, join to read "Filters the body classes to add..."

src/Tribe/Views/V2/Theme_Compatibility.php
```diff
@@ -0,0 +1,100 @@
...
+		$child_theme  = get_option( 'stylesheet' );
+		$parent_theme = get_option( 'template' );
+
+		// if the 2 options are the same, then there is no child theme.
+		if ( $child_theme == $parent_theme ) {
+			$child_theme = false;
+		}
+
+		if ( $child_theme ) {
+			$theme_classes = "tribe-theme-parent-$parent_theme tribe-theme-child-$child_theme";
```
paulmskim: From a CSS point of view, is there a way we can target only one class, rather than looking for `.tribe-theme-parent-twentyseventeen`, `.tribe-theme-child-twentyseventeen`, and `.tribe-theme-twentyseventeen`, for example? This will make the theme overrides 3x as large.

juanfra: Sure! I was replicating what we currently have right now (classes). Maybe instead of parent we set the `tribe-theme-twentyseventeen` there and add the child theme class just in case. Agree?

paulmskim: Sure yeah, I wonder what the reason was for the decision previously. If it was a valid reason we may need to keep it or think of something to cover that case.

PR #2507, opened 2019-05-16
VR > Theme compatibility

juanfra:
🎫 https://central.tri.be/127066

### DIFF
--- a/src/resources/postcss/views/_list.pcss
+++ b/src/resources/postcss/views/_list.pcss
@@ -49,11 +49,11 @@
 			font-size: var(--font-size-7);
 			margin-bottom: var(--spacer-2);
 		}
+	}
 
-		a {
-			color: var(--color-text-primary);
-			text-decoration: none;
-		}
+	.tribe-events-calendar-list__event--title-link {
+		color: var(--color-text-primary);
+		text-decoration: none;
 	}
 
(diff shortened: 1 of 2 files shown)


src/resources/postcss/views/_list.pcss
```diff
@@ -38,15 +38,22 @@
...
+		line-height: var(--line-height-1);
 		margin-bottom: var(--spacer-1);
 		padding: 0;
 
 		@media (--viewport-medium) {
 			font-size: var(--font-size-7);
 			margin-bottom: var(--spacer-2);
 		}
+
+		a {
```
paulmskim: could we add a class to this link and target the class only rather than increasing the specificity with nested selectors?

src/resources/postcss/views/_list.pcss
```diff
@@ -49,6 +51,11 @@
 		}
 	}
 
+	.tribe-events-calendar-list__event--title-link {
```
paulmskim: So we need to follow BEM styles with this. It should probably be something like `.tribe-events-calendar-list__event-title-link`. The `--` is reserved for the M (modifier) of BEM, where if we have different variations of the same element. If we had a red link in addition to the regular link, for example, we might have a class `.tribe-events-calendar-list__event-title-link--red`.

PR #2519, opened 2019-05-22
Feature/128104 month day block event

paulmskim:
For ticket: https://central.tri.be/issues/128104


lucatume:
Refresh before merging.

bordoni:
Should we have tests added here?

paulmskim:
@lucatume @bordoni @juanfra added snapshot test for this, have another pass at this.

paulmskim:
@lucatume I'm not sure why `EndpointsCest: Should_provide_an_html_endpoint_on_the_rest_api` is failing here on travis.

lucatume:
> @lucatume I'm not sure why `EndpointsCest: Should_provide_an_html_endpoint_on_the_rest_api` is failing here on travis.

Probably this: https://github.com/moderntribe/the-events-calendar/pull/2520/commits/ead1538042b751aa63e7d855c39f12dd267ecec6

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,62 @@
...
+
+	.tribe-events-calendar-month__calendar-event-title {
+		font-weight: var(--font-weight-regular);
+	}
+
+	.tribe-events-calendar-month__calendar-event-tooltip {
+		background-color: var(--color-background);
+		border: 1px solid var(--color-border-default);
+		box-shadow: var(--box-shadow-tooltip);
+		display: none;
```
ryanurban: If we are going with a simple CSS implementation for the tooltip, for simple a11y, I'd recommend one of two things:

1) Don't do display none, instead go with the height 0/opacity 0 approach where the content is still accessible to say a screen reader

OR

2) Keep this approach and add a `aria-hidden="true"` to the tooltip container. Technically we'll still link off to the event itself where a user can get all that event's info, so I think it would be fine to not worry about making that particular functionality and content be more functionally accessible.

I'd rec the 2nd approach ; )

juanfra: We're tied to what tooltipster brings here. I'm not sure if it'll work if we do something different from `display: none`. I'm implementing this and will go with the 2nd approach

juanfra: FYI: Tooltipster is creating a `div` _on-the-fly_ outside the wrapper (at a body level), so I'll need to take this as a special case for some classes.

ryanurban: Oh ok, didn't' realize that, there is this fwiw: https://iamceege.github.io/tooltipster/#accessibility

paulmskim: hey @juanfra do you think you could handle this? I can merge this and you can handle the CSS as you need for the tooltip.

juanfra: Sure thing, the wrapper will be some sort of "tooltipster theme", so probably that will live outside, but I can make the rest play nicely with your work. Thanks @paulmskim !

lucatume:
I've pushed changes into this PR to make sure tests are passing.

PR #2520, opened 2019-05-23
Update Browser History on navigation

bordoni:
_Ref:_ [C#126089](http://central.tri.be/issues/126089)
---


tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
 namespace Tribe\Events\Views\V2;
 
 class TemplateBootstrapTest extends \Codeception\TestCase\WPTestCase {
+	public function setUp() {
```
lucatume: Remove this if it's just calling the parent method.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+		// before
+		parent::setUp();
+	}
+
+	public function tearDown() {
```
lucatume: Remove this if it's just calling the parent method.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+
+	/**
+	 * It should only allow permitted values on base template option
+	 *
+	 * @test
+	 * @dataProvider base_template_options
+	 */
+	public function should_only_allow_permitted_values_on_base_template_option( $input, $expected ) {
+		tribe_update_option( 'tribeEventsTemplate', $input );
+		$option_value = $this->make_instance()->get_template_setting();
```
lucatume: l'd add a blank line before this, to keep with the "visual" structure of: Arrange, Act, Assert.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+	}
+
+	/**
+	 * It should return template event instance
+	 *
+	 * @test
+	 */
+	public function it_should_return_template_event_instance() {
+		tribe_update_option( 'tribeEventsTemplate', 'event' );
+		$instance = $this->make_instance()->get_template_object();
```
lucatume: l'd add a blank line before this, to keep with the "visual" structure of: Arrange, Act, Assert.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+	}
+
+	/**
+	 * It should return template page instance
+	 *
+	 * @test
+	 */
+	public function it_should_return_template_page_instance() {
+		tribe_update_option( 'tribeEventsTemplate', 'default' );
+		$instance = $this->make_instance()->get_template_object();
```
lucatume: l'd add a blank line before this, to keep with the "visual" structure of: Arrange, Act, Assert.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+				false,
+			],
+			'numeric' => [
+				[
+					'post_type' => 'tribe_events',
+				],
+				true,
+			],
+			'stdObject' => [
+				[],
```
lucatume: This is not an object.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+
+	public function invalid_queries() {
+		return [
+			'string' => [
+				'foo',
+				false,
+			],
+			'numeric' => [
+				[
+					'post_type' => 'tribe_events',
```
lucatume: Either the value, or the test data name, are wrong.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+		$should_load = $this->make_instance()->should_load( $query );
+
+		$this->assertEquals( $should_load, $expected );
+	}
+
+	public function invalid_queries() {
+		return [
+			'string' => [
+				'foo',
+				false,
```
lucatume: Here and below: what is this second value for? It does not seem to be used in the test case.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+			],
+			'post_type_eq_tribe_events' => [
+				[
+					'post_type' => 'tribe_events',
+				],
+				true,
+			],
+			'post_type_contains_tribe_events' => [
+				[
+					'post_type' => [ 'tribe_events', 'foo' ],
```
lucatume: Split this to have `tribe_events` with a valid test case (e.g. `post`) and an invalid one (e.g. `invalid_post_type`).
If we're not checking if the post types are valid let's make it explicit and clear.

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+			],
+			'boolean_false' => [
+				false,
+				'event',
+			],
+			'boolean_true' => [
+				false,
+				'event',
+			],
+			'event' => [
```
lucatume: What about `slug_event` to make it clearer?

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+				'event',
+				'event',
+			],
+			'page_isnt_a_value' => [
+				'page',
+				'event',
+			],
```
lucatume: Where is `slug_page`?

tests/views_integration/Tribe/Events/Views/V2/TemplateBootstrapTest.php
```diff
@@ -2,5 +2,186 @@
...
+			'numeric' => [
+				2,
+				'event',
+			],
+			'default' => [
+				'default',
+				'page',
+			],
+			'empty' => [
```
lucatume: What about `empty_string`?
And maybe add a `0` (zero) case?

src/Tribe/Views/V2/Assets.php
```diff
@@ -42,7 +42,7 @@ public function register() {
 			$plugin,
 			'tribe-events-views-v2-manager',
 			'views/manager.js',
-			[ 'jquery', 'tribe-common', 'tribe-query-string' ],
+			[ 'jquery', 'tribe-common', 'tribe-query-string', 'underscore' ],
```
paulmskim: Does WP ship with underscore? If not, can we pull in modules as required from our code instead to not download the full underscore library?

src/resources/postcss/views/_loader.pcss
```diff
@@ -0,0 +1,24 @@
+.tribe-events-container {
```
paulmskim: this should just be `.tribe-events`

src/resources/postcss/views/_loader.pcss
```diff
@@ -0,0 +1,24 @@
...
+		bottom: 0;
+		left: 0;
+		z-index: 25000;
+		background: rgba( 255, 255, 255, 0.6 );
+
+		&.tribe-hidden {
+			@mixin hidden;
+		}
+
+		.tribe-events-view-loader-spinner {
```
paulmskim: this should be `.tribe-events-view-loader__spinner` and not wrapped in `.tribe-events-view-loader`.

src/resources/postcss/views/_loader.pcss
```diff
@@ -0,0 +1,24 @@
...
+	.tribe-events-view-loader {
+		align-content: center;
+		justify-content: center;
+		display: flex;
+		position: absolute;
+		top: 0;
+		right: 0;
+		bottom: 0;
+		left: 0;
+		z-index: 25000;
```
paulmskim: Let's use a sensible `z-index` value here, rather than a random high number. We can create a `z-index` variables file in `tribe-common-styles` if needed so we're being consistent.

src/resources/postcss/views/_loader.pcss
```diff
@@ -0,0 +1,24 @@
+.tribe-events-container {
+	position: relative;
+
+	.tribe-events-view-loader {
+		align-content: center;
```
paulmskim: `align-items`

src/resources/postcss/views/_loader.pcss
```diff
@@ -0,0 +1,24 @@
...
+		justify-content: center;
+		display: flex;
+		position: absolute;
+		top: 0;
+		right: 0;
+		bottom: 0;
+		left: 0;
+		z-index: 25000;
+		background: rgba( 255, 255, 255, 0.6 );
```
paulmskim: Let's define this value in the colors variables file in `tribe-common-styles`.

src/resources/postcss/views/_loader.pcss
```diff
@@ -0,0 +1,24 @@
...
+
+		&.tribe-hidden {
+			@mixin hidden;
+		}
+
+		.tribe-events-view-loader-spinner {
+			align-self: center;
```
paulmskim: this shouldn't be needed if you use `align-items` above.

src/resources/postcss/views/_loader.pcss
```diff
@@ -0,0 +1,24 @@
...
+		display: flex;
+		position: absolute;
+		top: 0;
+		right: 0;
+		bottom: 0;
+		left: 0;
+		z-index: 25000;
+		background: rgba( 255, 255, 255, 0.6 );
+
+		&.tribe-hidden {
```
paulmskim: This is OK for now, but can we not wrap this in `.tribe-events-view-loader`? I think that if this is a common occurrence, we should use a common class, like `.tribe-common-a11y-hidden` or something similar.

src/views/v2/list.php
```diff
@@ -12,25 +12,55 @@
...
+ */
+tribe_asset_enqueue( 'tribe-events-views-v2-manager' );
+
+$events_label_plural = tribe_get_event_label_plural();
 $events = $this->get( 'events' );
 
-$this->template( 'events-bar' );
+?>
+<form
+	class="tribe-common tribe-events tribe-events-container"
```
paulmskim: reason why we're using `tribe-events-container` here?

src/views/v2/list.php
```diff
@@ -12,25 +12,55 @@
...
+ * @todo  remove this and properly enqueue assets
+ */
+tribe_asset_enqueue( 'tribe-events-views-v2-manager' );
+
+$events_label_plural = tribe_get_event_label_plural();
 $events = $this->get( 'events' );
 
-$this->template( 'events-bar' );
+?>
+<form
```
paulmskim: I don't think semantically wrapping the entire list view in a `<form>` is going to work. Is there a specific reason why we're doing this?

src/views/v2/list/nav/next.php
```diff
@@ -22,7 +22,7 @@
 	<a
 		href="<?php echo esc_url( $link ); ?>"
 		rel="next"
-		class="tribe-common-c-nav__next"
+		class="tribe-common-c-nav__next tribe-events-navigation-link"
```
paulmskim: any reason why we're not following BEM format with this?

src/views/v2/loader.php
```diff
@@ -0,0 +1,3 @@
+<div class="tribe-events-view-loader tribe-hidden">
+	<div class="tribe-events-view-loader-spinner">Loading...</div>
```
paulmskim: `.tribe-events-view-loader__spinner` class

src/views/v2/list.php
bordoni: My goal is to have a views manager JS that will be initialized on any element that uses that class. Should we use another naming?

src/views/v2/list.php
bordoni: I wanted customers to be able to include a field in any place inside of the view manager container, which the whole view been a form allows us to handle both with JS active and inactive since when the JS is deactivated it would just be a normal request to the backend.

src/resources/postcss/views/_loader.pcss
bordoni: Perfecto, using that.

src/views/v2/list/nav/next.php
bordoni: Not I was looking to figure out with you guys what would be the best formatting for this. I wasnt sure.

src/views/v2/list.php
paulmskim: I'd be partial to naming this `tribe-events-js` or something similar. However, I also don't have a good idea of the larger picture of the JS of the new views (not just views manager) and it would be tough for me to say something about it without understanding that bigger picture.

src/resources/postcss/views/_loader.pcss
paulmskim: Sorry, we do have a class from common that does exactly this: `.tribe-common-a11y-hidden`. However, if we're planning to target this class in the JS, we should probably use something like `.tribe-events-view-loader--hidden` and use `@mixin hidden`.

paulmskim: actually scratch that, using `tribe-common-a11y-hidden` is fine with how the JS is set up.

src/Tribe/Views/V2/Assets.php
bordoni: WordPress ship with Underscore.

src/resources/postcss/views/_loader.pcss
bordoni: On this particular case might be both. We wanted in particular to the Views Manager container to be relative so the loader would cover all of it.

src/views/v2/list.php
bordoni: I really don't know how I feel about these classes with `js` included. Would love to talk about the particular naming of the container better.

src/views/v2/list.php
paulmskim: I don't see a submit button anywhere in the list view markup, the navigation is done by links. But also I'd like to chat with you about this and the need for allowing customers to include a field anywhere.

src/resources/postcss/views/_loader.pcss
bordoni: What would be a sensible `z-index` here? Since we want this to be on top of everything and customers might have other stuff running around?

src/views/v2/list.php
bordoni: Lets do that on monday? There are some good reasons for this, it might seem weird I know.

src/resources/postcss/views/_loader.pcss
bordoni: Added those variables; https://github.com/moderntribe/tribe-common-styles/pull/7

src/views/v2/list.php
ryanurban: Would it make sense in this upcoming JS work to utilize the `data-js` attribute approach we do on agency? It's performant and very nicely provides a clear notice to our users who may hack on things as they customize that a given element should be treated a bit more carefully as it is utilized for JS driven functionality.

ryanurban: ^ @paulmskim @faction23 @bordoni

src/views/v2/list.php
ryanurban: Agree with Paul on this, this is really funky and not conforming to best practices.

bordoni: Added a `@todo` there and we can chat afterward about this one.

src/resources/postcss/views/_loader.pcss
```diff
@@ -0,0 +1,18 @@
...
+		display: flex;
+		position: absolute;
+		top: 0;
+		right: 0;
+		bottom: 0;
+		left: 0;
+		z-index: var(--layer-spinner-container);
+		background: var(--color-background-transparent);
+
+		&__spinner {
```
paulmskim: this works, but I'm gonna be picky and say let's move this out to it's own fully written out class `.tribe-events-view-loader__spinner`. It just makes it easier to follow and `ctrl+f`.

src/views/v2/loader.php
```diff
@@ -0,0 +1,3 @@
+<div class="tribe-events-view-loader tribe-common-a11y-hidden">
```
paulmskim: A note on this. It'd be nice to use `role="alert"` and `aria-live="assertive"` for accessibility here, but we need custom JS to interact with this (add Loading text, add Loaded text). Not high priority to add this, but something to consider down the road.

tests/views_integration/Tribe/Events/Views/V2/HTML/LoaderTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+	/**
+	 * @test
+	 */
+	public function it_should_contain_correct_html_classes() {
+		$template = $this->template->template( 'loader' );
+		$html = $this->document->html( $template );
+
+		$this->assertEquals(
+			$html->find( '.tribe-events-view-loader' )->count(),
```
Camwyn: You use$html->find( '.tribe-events-view-loader' ) three times, I'd move it out to a variable?

bordoni: Not really because it breaks due to how it modifies the original object in some odd way. Still need to see why.

PR #2522, opened 2019-05-23
VR > Initial month calendar markup and styles

juanfra:
🎫 https://central.tri.be/issues/128100

https://www.loom.com/share/bc7deaba90b14b8b8325ac21f3acc8a6


src/views/v2/month/grid-header.php
```diff
@@ -0,0 +1,27 @@
...
+$days_of_week = tribe_events_get_days_of_week();
+global $wp_locale;
+?>
+<header class="tribe-events-calendar-month__header" role="rowgroup">
+	<?php foreach ( $days_of_week as $day ) : ?>
+			<div
+				class="tribe-events-calendar-month__header-column"
+				role="columnheader"
+				aria-label="<?php echo esc_attr( $day ); ?>"
+				><?php echo esc_html( $wp_locale->get_weekday_abbrev( $day ) ); ?></div>
```
bordoni:
```suggestion
			>
				<?php echo esc_html( $wp_locale->get_weekday_abbrev( $day ) ); ?>
			</div>
```

src/views/v2/month.php
```diff
@@ -18,9 +18,34 @@
...
+	<div class="tribe-events-calendar-month__body" role="rowgroup">
+
+		<?php // @todo: replace this with the actual month days. Using these for(s) for presentation purposes. ?>
+		<?php for ( $week = 0; $week < 4; $week++ ) : ?>
+
+			<div class="tribe-events-calendar-month__week" role="row">
+
+				<?php for ( $day = 0; $day < 7; $day++ ) : ?>
+
+					<div class="tribe-events-calendar-month__day" role="gridcell" tabindex="-1">
```
ryanurban: I know we are following the pattern that gcal has established, which is great, however I'm curious if we have any grand designs in mind as of now that would result in needing to implement these `tabindex="-1"` attributes? If not, we can likely nix them for now and add them later and as we would need them for our implementation ; )

We should follow along with gcal and relate the gridcell to the day heading, so `aria-labelledby` on this and then a matching/corresponding `id` on the day heading. This will add additional context to this wrapper.

One other thought, I would recommend including this day wrapper as part of the day template partial itself, as that will nicely organize that as a completely self-contained unit, especially in the case where we may need to do various conditional logic for days. Just my two cents!

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,66 @@
...
+		&:hover {
+			@media (--viewport-medium) {
+				border-bottom: 1px solid var(--color-border-active);
+				box-shadow: inset 0 -1px 0 0 var(--color-border-active);
+			}
+		}
+
+		&:first-of-type {
+			@media (--viewport-medium) {
+				border-left: 1px solid var(--color-border-default);
```
paulmskim: Let's put this style on the row instead. If we do this, the first cell will be 1px smaller than the others (due to the border).

src/views/v2/month/day.php
```diff
@@ -0,0 +1,28 @@
...
+
+$day_classes = [ 'tribe-events-calendar-month__day-date', 'tribe-common-h4' ];
+
+// @todo: check if we use classes here or if we wrap the day block directly with the classes (we also have `.tribe-events-calendar-month__day-date--current`).
+if ( $day_number == date( 'd', time() ) ) {
+	$day_classes[] = 'tribe-events-calendar-month__day-date--current';
+}
+?>
+<div class="<?php echo esc_attr( implode( ' ', $day_classes ) ); ?>" aria-label="Weekday, Month Daynumber">
+	<time><?php echo esc_html( $day_number ); ?></time>
```
paulmskim: We should include the `datetime` attribute.

ryanurban: Let's be sure to add `datetime` attribute ; ) Also wonder if it would be appropriate to wrap this time tag in a heading (whatever is appropriate based on your current setup, maybe an `h4`?). Thoughts?

I also like how as part of the heading for a day, gcal also states how many events there are for the day: http://p.tri.be/WqsYUj. This would be a nice thing to add here as well (not as part of the time tag) and wrap that text in an a11y hidden span.

Last thought, if I remember right, the date for each cell will link off to the day view, correct? If so, we should make sure to add a link too : )

src/views/v2/month/day.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+// @todo: This is just for presentation purposes, while building the view.
+$day_number  = ( $week < 1 ) ? ( $day + 1 ) * ( $week + 1 ) : ( $day + 1 ) + $week * 7;
+
+$day_classes = [ 'tribe-events-calendar-month__day-date', 'tribe-common-h4' ];
```
paulmskim: Desktop seems to use `desktop-h4` size, but mobile uses `mobile-h6` size. This would be a good case to define these in the `tribe-events-calendar-month__day-date` CSS instead. The headings are not yet mixins, but we should probably turn them into mixins for cases like this.

src/views/v2/month/grid-header.php
```diff
@@ -0,0 +1,27 @@
...
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+// @todo: Check if we're keeping these template tags or how is that we'll be populating data here.
+$days_of_week = tribe_events_get_days_of_week();
+global $wp_locale;
+?>
+<header class="tribe-events-calendar-month__header" role="rowgroup">
```
ryanurban: Think this `role` can just be `row`?

src/views/v2/month/day.php
```diff
@@ -0,0 +1,28 @@
...
+
+// @todo: check if we use classes here or if we wrap the day block directly with the classes (we also have `.tribe-events-calendar-month__day-date--current`).
+if ( $day_number == date( 'd', time() ) ) {
+	$day_classes[] = 'tribe-events-calendar-month__day-date--current';
+}
+?>
+<div class="<?php echo esc_attr( implode( ' ', $day_classes ) ); ?>" aria-label="Weekday, Month Daynumber">
```
ryanurban: Given comment in the month template, you can likely nix the `aria-label` here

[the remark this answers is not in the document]
juanfra: I've based this on what I saw google did on their calendars.

src/views/v2/month/grid-header.php
```diff
@@ -0,0 +1,27 @@
...
+	<?php foreach ( $days_of_week as $day ) : ?>
+			<div
```
ryanurban: The text within should probably be wrapped in an appropriate heading tag and a time tag with datetime attribute (all while still being wrapped in a div as I'm sure that'll be helpful for layout)?

You can nix the `aria-label` if you make the updates above I believe.

src/views/v2/month/day.php
paulmskim: I've added heading mixins to the latest `tribe-common-styles`, check out https://github.com/moderntribe/tribe-common/pull/1029 on how they're being used.

src/views/v2/month/day.php
juanfra: > I also like how as part of the heading for a day, gcal also states how many events there are for the day: http://p.tri.be/WqsYUj. This would be a nice thing to add here as well (not as part of the time tag) and wrap that text in an a11y hidden span.

I'm not sure if I'm getting you on this one. What attribute is being used for that on the example you shared?

src/views/v2/month/grid-header.php
juanfra: I've now added a new one as a wrapper for that, that its role is row

src/views/v2/month/grid-header.php
juanfra: This is just for the day of week - Like "Monday", "Tuesday" and so on. So I'm not sure if the time tag makes sense. Thoughts?

src/views/v2/month.php
```diff
@@ -18,9 +18,34 @@
...
 
 $this->template( 'top-bar' );
-
 ?>
 
-<div class="tribe-events-calendar-month">
-	<h1>Welcome to the month view</h1>
-</div>
\ No newline at end of file
+<h2 class="tribe-common-a11y-hidden" id="tribe-calendar-header"><?php printf( esc_html__( 'Calendar of %s', 'the-events-calendar' ), tribe_get_event_label_plural() ); ?></h2>
```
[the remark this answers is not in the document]
juanfra: This is describing the general calendar. As it is right now. Do you think it should be removed?

bordoni:
Tests are not passing and lets make sure it's update with `release/B19.07`

src/views/v2/month.php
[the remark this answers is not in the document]
ryanurban: I _think_ so as I think we can pair up the purpose of this with the heading for the calendar itself (add `<span class="a11y visual hide">Calendar of Events For </span> Date`). Let's get one more vote one way or another. @paulmskim, what do you think?

src/views/v2/month/day.php
ryanurban: Sorry about that, there is a lot here :P No attribute is being used, but as part of the heading, they nicely inform the user how many events are occurring on a given day. So we could do that as well `<span class="a11y visually hide">X events, </span><time datetime="">time</time>`

src/views/v2/month/grid-header.php
ryanurban: Ahh, you are right on the time tag, good call!

src/views/v2/month.php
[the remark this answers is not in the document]
paulmskim: I don't think it's necessary to have this, though we will probably need a `<h1>` for the page (just saying events for "month", no need to specify calendar), much like wpshindig.com for their month view. I do like the idea of putting a span in the header labeling the grid as a calendar, so let's do that too.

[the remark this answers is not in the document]
paulmskim: Also if we use a `<h1>`, follow the headings down as you include files. The next heading I see is `<h4>`. We should only be stepping down by 1 each heading, so the next level should be `<h2>`.

juanfra:
Tests were not passing because it's getting out of resources. We're not adding anything that could break the tests here. I've re-started them and the green checkmark is there.

src/views/v2/month.php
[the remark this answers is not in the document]
juanfra: >  though we will probably need a <h1> for the page (just saying events for "month", no need to specify calendar), much like wpshindig.com for their month view. I do like the idea of putting a span in the header labeling the grid as a calendar, so let's do that too.

Haven't included that as it's not part of the design. So, this H2 will be the one that will be saying "Hey, I'm the calendar page for Events on ... Month"

It's the only place where this is happening. Do you want to move it?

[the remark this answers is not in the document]
paulmskim: Yeah, let's move it to just within the header. Also the class to visually hide things is `tribe-common-a11y-visual-hide`. The class `tribe-common-a11y-hidden` hides from the the page and screen readers.

src/views/v2/month/day.php
```diff
@@ -0,0 +1,39 @@
...
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+// @todo: This is just for presentation purposes, while building the view.
+$day_number  = ( $week < 1 ) ? ( $day + 1 ) * ( $week + 1 ) : ( $day + 1 ) + $week * 7;
+
+$day_title_classes = [ 'tribe-events-calendar-month__day-date' ];
+$day_id = 'tribe-events-calendar-day-' . $day_number;
```
paulmskim: We should maybe add the month number to this too, so it's like `'tribe-events-calendar-day-' . $month_number . '-' . $day_number` (or any other way to create a unique id). This will result in gridcells with the same id at edge cases like the start and end of the month.

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,99 @@
...
+		@mixin heading;
+		@mixin heading-6;
+
+		@media (--viewport-medium) {
+			@mixin desktop-heading-4;
+		}
+
+		padding: var(--spacer-1);
+
+		a {
```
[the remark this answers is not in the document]
juanfra: The problem here is that the link will be shown only for days where there are events. If I follow that then we'll have to make logic on the template, or set the same class for both elements (with the reset, links don't have the cursor: pointer).

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,99 @@
...
+			&:active {
+				color: var(--color-accent-secondary-active);
+			}
+		}
+	}
+
+	.tribe-events-calendar-month__day-date--current {
+		color: var(--color-accent-primary);
+
+		a {
```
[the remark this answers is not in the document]
juanfra: 👉 https://github.com/moderntribe/the-events-calendar/pull/2522#discussion_r287864745

src/views/v2/month/day.php
juanfra: Yes, the thing is that we don't have the month number, and it's just data with testing purposes. We want to hardcode this stuff just for the sake of doing it?

src/views/v2/month/day.php
```diff
@@ -0,0 +1,39 @@
...
+
+// @todo: check if we use classes here or if we wrap the day block directly with the classes (we also have `.tribe-events-calendar-month__day-date--current`).
+if ( $day_number == date( 'd', time() ) ) {
+	$day_title_classes[] = 'tribe-events-calendar-month__day-date--current';
+}
+?>
+
+<div class="tribe-events-calendar-month__day" role="gridcell" aria-labelledby="<?php echo esc_attr( $day_id ); ?>">
```
bordoni: Would like to see tests checking for the presence of these A11y attributes.

src/views/v2/month/day.php
```diff
@@ -0,0 +1,39 @@
...
+	<div id="<?php echo esc_attr( $day_id ); ?>">
+		<h3 class="<?php echo esc_attr( implode( ' ', $day_title_classes ) ); ?>">
```
bordoni: Same here would love to check on these classes with tests.

src/views/v2/month/grid-header.php
```diff
@@ -0,0 +1,33 @@
...
+// @todo: Check if we're keeping these template tags or how is that we'll be populating data here.
+$days_of_week = tribe_events_get_days_of_week();
+global $wp_locale;
+?>
+<header role="rowgroup">
+	<div role="row" class="tribe-events-calendar-month__header">
+		<?php foreach ( $days_of_week as $day ) : ?>
+				<div
+					class="tribe-events-calendar-month__header-column"
+					role="columnheader"
```
bordoni: Would like to see tests checking for the presence of these A11y attributes.

src/views/v2/month.php
```diff
@@ -18,9 +18,28 @@
...
+<div class="tribe-events-calendar-month" role="grid" aria-labelledby="tribe-calendar-header" aria-readonly="true">
+
+	<?php $this->template( 'month/grid-header' ); ?>
+
+	<div class="tribe-events-calendar-month__body" role="rowgroup">
+
+		<?php // @todo: replace this with the actual month days. Using these for(s) for presentation purposes. ?>
+		<?php for ( $week = 0; $week < 4; $week++ ) : ?>
+
+			<div class="tribe-events-calendar-month__week" role="row">
```
bordoni: Would like to see tests checking for the presence of these A11y attributes.

src/views/v2/month.php
```diff
@@ -18,9 +18,28 @@
...
 
 $this->template( 'top-bar' );
-
 ?>
 
-<div class="tribe-events-calendar-month">
-	<h1>Welcome to the month view</h1>
-</div>
\ No newline at end of file
+<div class="tribe-events-calendar-month" role="grid" aria-labelledby="tribe-calendar-header" aria-readonly="true">
```
bordoni: Would like to see tests checking for the presence of these A11y attributes.

src/views/v2/list/event/date.php
```diff
@@ -25,8 +25,8 @@
 	<?php if ( $is_featured ) : ?>
 		<span
 			class="tribe-events-calendar-list__event-datetime-featured-icon tribe-common-svgicon tribe-common-svgicon--featured"
-			aria-label="<?php esc_html_e( 'Featured', 'the-events-calendar' ); ?>"
-			title="<?php esc_html_e( 'Featured', 'the-events-calendar' ); ?>"
+			aria-label="<?php esc_attr_e( 'Featured', 'the-events-calendar' ); ?>"
```
bordoni: Would like to see tests checking for the presence of these A11y attributes.

src/resources/postcss/views/_month.pcss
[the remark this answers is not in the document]
paulmskim: Even if we only show links for those days that have events, we should still target them with classes. We are already doing `if` and `for` logic on templates, and there is always going to be this sort of logic.

Let's add the `cursor: pointer` to the reset then so we don't have to repeat this.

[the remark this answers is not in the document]
juanfra: I still don't see the point of this. Having the same styles for different classes, duplicating just for the sake of having a particular class for a hyperlink that is inside many elements. ¯\_(ツ)_/¯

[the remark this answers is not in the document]
ryanurban: It's more about consistency, future maintainability and also provides a nice set of css "hooks" for our users as well and makes their customizations cleaner : ) I'd rec having the same class on the <a> for the shared styles regardless of state and then add a modifier class to either the day with or without the event.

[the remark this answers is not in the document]
juanfra: If I put in the customer shoes and I have to change two classes with the same values to set the month day number, I won't understand why. Or if I'm a client I changed the way the "current" day looked by inspecting the element and then when the current day has events my styles don't take effect, because there are two classes.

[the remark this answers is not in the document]
ryanurban: Should be like that or a frustrating experience depending on how we set this up. Let's chat about this 😉

PR #2525, opened 2019-05-24
VR > Multiday styles & markup

juanfra:
🎫 https://central.tri.be/issues/128105

Markup and styles for the multi-day event. It actually looks like:

`@todo-next:` Implement this in a way that works for having multiday events across different cells, etc.


src/views/v2/month/event-multiday.php
```diff
@@ -0,0 +1,32 @@
...
+$event_id = $event->ID;
+
+$classes = [ 'tribe-events-calendar-month__event-multiday' ];
+
+if ( $is_featured = tribe( 'tec.featured_events' )->is_featured( $event_id ) ) {
+	$classes[] = 'tribe-events-calendar-month__event-multiday--featured';
+}
+?>
+<div class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>">
+	<?php if ( $is_featured ) : ?>
```
ryanurban: Same comment I made to Paul around our icon usage, especially for cases like this where there is importance to what this icon represents. Let's be sure to either add a11y hidden text noting this is a featured event or add a `title` & `aria-label` to communicate that.

Last thing, I'd recommend going with an `<em>` tag for this particular implementation as it will give it a bit more semantic juice and importance.

cc @paulmskim, good idea to do where you have implemented as well ; )

src/views/v2/month/event-multiday.php
```diff
@@ -0,0 +1,36 @@
+<?php
```
ryanurban: It'd be good to go ahead and keep a `<time datetime="...">...</time>` element here and a11y visually hide it. Just be consistent in your html implementation of it between this and the events for a given day.

src/views/v2/month/event-multiday.php
paulmskim: @ryanurban Yup, haven't forgotten ;)

src/views/v2/month/event-multiday.php
```diff
@@ -0,0 +1,36 @@
...
+$event    = $this->get( 'event' );
+$event_id = $event->ID;
+
+$classes = [ 'tribe-events-calendar-month__event-multiday' ];
+
+if ( $is_featured = tribe( 'tec.featured_events' )->is_featured( $event_id ) ) {
+	$classes[] = 'tribe-events-calendar-month__event-multiday--featured';
+}
+?>
+<div class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>">
```
paulmskim: Let's add another inner wrapper to this, for a couple reasons. So our markup will look something like:

```html
<div class="tribe-events-calendar-month__event-multiday">
    <div class="tribe-events-calendar-month__event-multiday-inner">
        <em ...></em>
        <h3></h3>
    </div>
</div>
```

1. When we had our call with Matt about the month view, there was discussion about the 0.1 opacity background and crossing over day blocks and having the grid lines show through the background. We're going to want to add a white background to the outer wrapper and have the blue 0.1 opacity background on the inner wrapper.
2. Matt is going to pick this out, but the border radius implementation isn't quite the same as what the designs are. Using an outer and inner wrapper will help fix this, I'll explain further in the CSS.

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,38 @@
...
+		flex-wrap: wrap;
+		align-items: center;
+		margin-bottom: var(--spacer-0);
+
+		&:hover {
+			background-color: rgba(var(--color-accent-primary), .2);
+		}
+	}
+
+	.tribe-events-calendar-month__event-multiday-featured-icon {
```
paulmskim: We've used this style a few times now, and its the exact same style to get the featured icon. I've added svg icon classes in this PR: https://github.com/moderntribe/tribe-common/pull/1029. Let's use these going forward (apply these classes to the `<em>` element).

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,38 @@
...
+		border-style: solid;
+		border-width: 0 2px 0px 2px;
+		border-color: var(--color-accent-primary);
+		padding: 1px var(--spacer-1);
+		display: flex;
+		flex-wrap: wrap;
+		align-items: center;
+		margin-bottom: var(--spacer-0);
+
+		&:hover {
```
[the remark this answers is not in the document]
paulmskim: So there's a bit of a tricky situation here: the wrapper `<div>` cannot receive a focus. We can tackle this a couple ways:

1. Make the inner wrapper a `<a>` to the event itself so we can apply `:hover` and `:focus` to it.
2. Use JS to apply a `--focused` class to the inner wrapper when the link is focused, with the `--focused` class styles having the background style update.

@ryanurban @juanfra thoughts? My preference would be to go the `<a>` route, but I wonder if there are semantic markup concerns with this (with the `<h3>` wrapped inside the `<a>` and the featured icon as well).

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,38 @@
  * @package TribeEventsCalendar
  *
  */
-.tribe-events {}
\ No newline at end of file
+.tribe-events {
+
+	.tribe-events-calendar-month__event-multiday {
+		background-color: rgba(var(--color-accent-primary), .1);
+		border-radius: 6px;
```
paulmskim: Let's apply the border radius to the outer wrapper and all the border styles to the inner wrapper. This prevents the blue border from curving, as we're seeing in the screenshot.

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,38 @@
  * @package TribeEventsCalendar
  *
  */
-.tribe-events {}
\ No newline at end of file
+.tribe-events {
+
+	.tribe-events-calendar-month__event-multiday {
+		background-color: rgba(var(--color-accent-primary), .1);
```
paulmskim: let's move this to the inner wrapper, and apply a white background to the outer wrapper.

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,38 @@
...
+		border-radius: 6px;
+		border-style: solid;
+		border-width: 0 2px 0px 2px;
```
paulmskim: We're going to need to allow for various cases, such as when a multi-day event spans multiple weeks. In this case, the end of the first week will not have a border on the end, the middle weeks will not have a border on either end, and the last week will not have a border at the start. I would suggest having class modifiers to apply each border separately.

src/resources/postcss/views/_month.pcss
```diff
@@ -4,4 +4,38 @@
  * @package TribeEventsCalendar
  *
  */
-.tribe-events {}
\ No newline at end of file
+.tribe-events {
+
+	.tribe-events-calendar-month__event-multiday {
```
paulmskim: We're going to need various widths of this multiday component (1 day - 7 days). Let's add modifier classes to account for this.

juanfra: Yes, I was leaving that for the time we actually worked on that. The idea of this PR was to set a basic markup and styles for the multiday marker.

src/views/v2/month/event-multiday.php
```diff
@@ -0,0 +1,39 @@
...
+	<time datetime="the-date-and-or-duration" class="tribe-common-a11y-hidden">The date and duration</time>
```
paulmskim: we should use the `tribe-common-a11y-visual-hide` class here.

tests/views_integration/Tribe/Events/Views/V2/HTML/Month/MonthEventMultidayTest.php
```diff
@@ -0,0 +1,48 @@
...
+	/**
+	 * @test
+	 */
+	public function it_should_contain_a11y_attributes() {
+		$template = $this->template->template( 'month/event-multiday', [ 'event' => (object) [ 'ID' => 0 ] ] );
+		$html = $this->document->html( $template );
+		$html = $html->find( '.tribe-events-calendar-month__event-multiday' );
+
+		/*
+			@todo: If the event is featured we should check the a11y classes
```
bordoni: Create an Assertion and mark it as skipped. And we are good for merging.

PR #2526, opened 2019-05-24
Avoid fiddling with the start date of query if is a query into the past

borkweb:
The start_date query param was being overridden on the list view for all queries that didn't have the start date set. However, on the base list view, a query into the past is needed to determine if the previous link is necessary, so overriding the start_date breaks it. This change prevents the override from occuring on queries that are looking into the past.

Additionally, this PR adds an acceptance test suite with tests to verify that the previous and next links are present at the right moments on initial page load.

Here's a screencast of the fix in action: http://p.tri.be/tl3g2q

See: https://central.tri.be/issues/127477


borkweb:
Hrm. There are a lot test errors that seem to have come from B19.07. We'll need to hunt those down in another PR and get them fixed before merging this PR.

bordoni:
Those errors seem to be related to a method missing in common.

bordoni:
@borkweb can you solve that common conflict, it prevents Travis from running.

.travis.yml
```diff
@@ -138,4 +138,5 @@ script:
   - codecept run views_rest
   - codecept run views_integration
   - codecept run views_ui
+  #- codecept run acceptance
```
juanfra: Looks good! And with tests in place! Are we adding this commented for any reason in particular?

borkweb: Yeah, adding as commented because I need to figure out some TravisCI issues that I'm having with the tests. I'll be figuring that out soon-ish!

juanfra: Perfect, thanks!

PR #2528, opened 2019-05-28
add gherkin scenarios for frontend list view layout

vicskf:
https://central.tri.be/issues/128333

First attempt to write list view scenarios. Comments and suggestions are welcome

### DIFF
--- a/tests/features/front-end/list-view/view-layout.feature
+++ b/tests/features/front-end/list-view/view-layout.feature
@@ -2,33 +2,61 @@ Feature: layout and visual elements of an event in List View
 
 Background:
     Given I am a Visitor on a site with TEC
-    Given List View is enabled under Events --> Settings --> Display
-    Given the site timezone is America/Los_Angeles
-    Given "Number of events to show per page" is set to 10
+    And List View is enabled under Events --> Settings --> Display
+    And the site timezone is America/Los_Angeles
+    And "Number of events to show per page" is set to 10
 
 Scenario: display of single-day timed event in List View
+    Given I have a single-day timed event
+    Then I should see the event date as ["start date" "Date time separator" "start time" "Time range separator" "end time"]
 
 Scenario: display of single-day event with the same start and end time in List View
-#This is a thing in the current views, but maybe not supported after VR?
+# This is a thing in the current views, but maybe not supported after VR?
+    Given I have a single-day event with "start time" the same as "end time"
+    Then I should see the event date as ["start date" "Date time separator" "start time"]
 
 Scenario: display of multi-day timed event in List View
+    Given I have a multi-day timed event
+    Then I should see the event date as ["start date" "Date time separator" "start time" "Time range separator" "end date" "Date time separator" "end time"]
 
 Scenario: display of single-day all day event in List View
+    Given I have a single-day all day event
+    Then I should see the event date as ["start date"]
 
 Scenario: display of multi-day all day event in List View
+    Given I have a multi-day all day event
+    Then I should see the event date as ["start date" "Time range separator" "end date"]
 
 Scenario: display of featured event in List View
+    Given I have a featured event with a featured image
+    Then I should see the event with a background colour
+    And I should see the featured image in a bigger size if it has one
 
 Scenario: display of an event with a venue in List View
+    Given I have an event with a venue
+    And The venue has an address
+    Then I should see the event showing the venue and address
 
-Scenario: display of an event with a price in List View
+Scenario: display of an event with a cost in List View
+    Given I have an event with a cost
+    Then I should see the event showing the cost
+    And The cost should show the correct currency symbol in the correct currency position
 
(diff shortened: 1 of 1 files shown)


tests/features/front-end/list-view/view-layout.feature
```diff
@@ -2,33 +2,61 @@ Feature: layout and visual elements of an event in List View
 
 Background:
     Given I am a Visitor on a site with TEC
-    Given List View is enabled under Events --> Settings --> Display
-    Given the site timezone is America/Los_Angeles
-    Given "Number of events to show per page" is set to 10
+    And List View is enabled under Events --> Settings --> Display
```
lucatume: Quotes around "List View"; here and everywhere else you might think of "variables" use quotes.
e.g. 
`And "List" view is enabled`
`And "Month" view is enabled`
`And "Another" view is enabled`

The tests are not going to consume the information "under Events --> Settings --> Display" as the tests will just set an option in the database; if we really want to embed the setting current location in the tests then I'd use a shorter notation like `Events > Settings > Display` or `Some > Other > Menu in the plugin`.

tests/features/front-end/list-view/view-layout.feature
```diff
@@ -2,33 +2,61 @@ Feature: layout and visual elements of an event in List View
...
+    And the site timezone is America/Los_Angeles
```
lucatume: `And the site timezone is "America/Los_Angeles"`

tests/features/front-end/list-view/view-layout.feature
```diff
@@ -2,33 +2,61 @@ Feature: layout and visual elements of an event in List View
...
+    And "Number of events to show per page" is set to 10
```
lucatume: I feel this could either be a more direct instruction like `And "posts_per_page" calendar option is set to "10"` or it should be more colloquial: `And "Events > Settings > Display > Number of Events to show per page` is set to "10".
The second would increase the test-as-documentation value.

tests/features/front-end/list-view/view-layout.feature
```diff
@@ -2,33 +2,61 @@ Feature: layout and visual elements of an event in List View
...
 
 Scenario: display of single-day timed event in List View
+    Given I have a single-day timed event
```
lucatume: Always define events in context of current time: 
```
Given I have the following event:
| title  | start_date         | timezone                     |
| Test | tomorrow 9am | America/Los_Angeles |
```

This is still an automated test that requires translation into code, the structure above is still readable and clear but lends itself to be reused:
```
Given I have the following events:
| title     | start_date           | timezone                      | duration |
| Test 1 | tomorrow 9am    | America/Los_Angeles | 2H          |
| Test 2 | tomorrow 12pm | Europe/Paris                 | 5H          |
```

The same step, once standardized, can cover a lot of mileage. The columns would be the ones supported by the ORM syntax.

tests/features/front-end/list-view/view-layout.feature
```diff
@@ -2,33 +2,61 @@ Feature: layout and visual elements of an event in List View
...
+    Then I should see the event date as ["start date" "Date time separator" "start time" "Time range separator" "end time"]
```
lucatume: It' missing a `When` step:
`When I visit the event single`

lucatume: The `Then` step is flaky. There are a lot of unspoken assumptions in there and I'm not sure what you're trying to achieve here.
If you're relying on default values then hard-code those in else the test would pass even if those are empty.
`Then I should see the "event date" has the format "Y:m:d H:i:s"` is clearer.

tests/features/front-end/list-view/view-layout.feature
leahkoerper: @lucatume I've been using List View, etc. without quotation marks all over the place in other scenarios. Is it worth changing those, or should I just keep this in mind going forward? I'll try to remember to do setting paths like you suggest above in future scenarios.

tests/features/front-end/list-view/view-layout.feature
leahkoerper: Same question as above- is this worth me changing in all the existing scenarios? I'd like to err on the non-technical version, e.g. `And "Events > Settings > Display > Number of Events to show per page is set to "10".`

tests/features/front-end/list-view/view-layout.feature
leahkoerper: I've been using something like this:
```
    Given there are 12 events on this site
        | event name       | start date-time        | duration       |
        | strawberry S-T-P | today -3 days -2 hours | 4 hours        |
        | banana M-A-P     | today -2 days all day  | 1 day          |
        | eggplant M-A-N   | today -2 days all day  | 4 days         |
```

bordoni:
Closing this since we are no longer going this route.

PR #2532, opened 2019-05-29
add tribe_suppress_query_filters arg

codingmusician:
_Ref:_ [C#127822](http://central.tri.be/issues/127822)
---

For AJAX requests, we want tribe_suppress_query_filters to be false.


bordoni:
It also seems like the code above is not allowing Filters to be applied to the ORM query args. Which was the intention of the PR right? 

We also want to make sure what ever change we apply to month gets applied to all other views.
- [ ] Day
- [ ] Day

codingmusician:
> It also seems like the code above is not allowing Filters to be applied to the ORM query args. Which was the intention of the PR right?
> 
> We also want to make sure whatever change we apply to month gets applied to all other views.
> 
> * [ ]  List
> * [x]  Month
> * [ ]  Day
> * [ ]  Week
> * [ ]  Photo
> * [ ]  Map

The filter in question does get triggered in all other views other than month. I've tested this out as noted here: https://central.tri.be/issues/127822#note-6

src/Tribe/Template/Month.php
```diff
@@ -401,6 +401,7 @@ protected function set_args( $args = array() ) {
 					'eventDisplay' => 'month',
 					'eventDate'    => is_array( $_POST['eventDate'] ) ? Tribe__Utils__Array::get( $_POST, array( 'eventDate', 0 ) ) : $_POST['eventDate'],
 					'post_status'  => $post_status,
+					'tribe_suppress_query_filters' => false
```
tr1b0t: :no_entry_sign: Each line in an array declaration must end in a comma

<sub>hash: 36ff7e4f8717c6acbb5730b77364179c</sub>

PR #2536, opened 2019-05-31
Rework tests.md

cliffordp:


lucatume:
@cliffordp wow! I've made some smaller changes, mainly the bullet list numbering and some re-arrangements of the text but nothing major.
Worth noticing: I've changed references to the the `.env.local` file to `env.testing.local` file to:
1. keep with an established standard of prefixing any environment file related to tests with `.env.testing`
2. PHPStorm, and other IDEs, will pick up `.env.testing.local` as an environment file and will provide syntax support for it; the same is not true of `.env.local`.

cliffordp:
Great. Thanks, Luca.

Why manually type each ordered list item instead of letting Markdown assign the numbers automatically, especially in case of rearranging or other potential future edits?

lucatume:
> Great. Thanks, Luca.
> 
> Why manually type each ordered list item instead of letting Markdown assign the numbers automatically, especially in case of rearranging or other potential future edits?

The initial version of the code had all list bullets prefixed with `1.` no matter the position; unless I'm losing something a plain Markdown viewer will show a list of bullets all with `1.`.

PR #2539, opened 2019-06-04
Spotfix/rbe recurrence 2

nealfennimore:


src/Tribe/REST/V1/Endpoints/Archive_Event.php
```diff
@@ -161,8 +161,17 @@ public function get( WP_REST_Request $request ) {
 
 			$data['rest_url'] = $this->get_current_rest_url( $args, $extra_rest_args );
 
+			$events = tribe_events();
 
-			$events = tribe_events()->by_args( $args )->pluck( 'ID' );
+			if ( $request['runs_between'] && $args['start_date'] && $args['end_date'] ){
```
mitogh: I think using: 

```suggestion
			if ( ! empty( $request['runs_between'] ) && ! empty( $args['start_date'] ) && ! empty( $args['end_date'] ) ){
```

Allow to check if the variable is set on `$request` and also make sure the value is not empty as right now if the value is not set on the `$request` the line above will trigger a warning.

nealfennimore: I put in the `! empty( $request['runs_between'] )`. We should be good with not having the `empty` around the start and end date arguments, as they're always defined beforehand

mitogh: Perfect! 👍

PR #2542, opened 2019-06-04
Create tests for the Theme Compatibility code

bordoni:
_Ref:_ [C#127999](http://central.tri.be/issues/127999)
---


src/Tribe/Views/V2/Theme_Compatibility.php
```diff
@@ -25,74 +30,94 @@ class Theme_Compatibility {
...
 		}
 
-		/**
-		 * Allows to filter the theme list with compatibility fixes.
-		 *
-		 * @since  4.9.3
-		 *
-		 * @param  array  $themes_with_compatibility_fixes A list of themes we provide compatibility for.
-		 */
-		$theme_compatibility_list = apply_filters( 'tribe_events_views_v2_themes_compatibility_fixes', $this->themes_with_compatibility_fixes );
```
juanfra: We would keep this filter. That way we could add an extension to give compatibility with whatever theme we want.

bordoni: We moved to use `tribe_events_views_v2_theme_compatibility_registered` which happens on a get method.

juanfra: awesome 🎉

PR #2543, opened 2019-06-05
update nav markup to show disabled button if no link

paulmskim:


src/views/v2/list/nav/next.php
```diff
@@ -14,17 +14,22 @@
...
-}
 ?>
 <li class="tribe-common-c-nav__list-item">
-	<a
-		href="<?php echo esc_url( $link ); ?>"
-		rel="next"
-		class="tribe-common-c-nav__next"
-		data-js="tribe-events-view-link"
-	>
+	<?php if ( empty( $link ) ) : ?>
```
juanfra: Great stuff! I was thinking we could have a partial for the disabled ones and this could look a bit more clean. What do you think? I'm not a big fan of large if/else within templates, I have to say though.

paulmskim: I'm OK either way, though I do lean a bit more toward the `if/else` in this case as there is a lot of repeated pieces. @bordoni @ryanurban thoughts?

bordoni: I feel like a case for each creates an easier overall template structure.

paulmskim: sounds good, will separate

ryanurban: I could also see either way being "better". @bordoni, given your institution knowledge here and with our customers and their feedback, have you noticed any sort of pattern around their thoughts and requests for ease of templateage overrides? Better to keep it all inside one template with conditionals to cut down on needing to nav around files vs. having things be cleaner in two templates but now needing to track two if doing customizations around the pagination?

bordoni: Normally customers tend to appreciate when the is less duplicated logic for the sake of HTML.

PR #2537, opened 2019-06-01
Feature/128097 mobile month day

paulmskim:
For ticket: https://central.tri.be/issues/128097

PR #2544, opened 2019-06-05
[VR] > Month View data and refinements

juanfra:
🎫 https://central.tri.be/issues/127870

After a couple long days we ended up putting all the pieces together. The month view is working with multiday and regular events. This required a new way of structuring the data. We have defined this and ideally we have some temporary helper functions to hydrate the template with the required arrays and the different functions to loop on that. (can be found on `month-view-demo-template-tags.php`).

Added some more tests to some of the new templates and been fixing different glitches that appeared once we put everything together.

Ideally, this is covering things looking to the goal of presenting the demo. But, we need to continue working on:

- Improving the partials folders/structure to something more friendly. A lot of parts for this view and ideally it should be pretty straightforward to understand what you are supposed to find on each one of them. That's not ready.
- Make a pass on the class naming and see how everything plays together.
- See if we're gonna use this tooltip format everywhere and define a pattern.
- Treatment of multi-day events on multiple weeks.
- A11y refinements for the grid.

I'll be addressing the reviews that we can find in the following hours, but we'll need to merge this at some point today (June 5th) for the demo tomorrow. We can always keep a list of actionable items/improvements for the upcoming days.

📹 https://www.loom.com/share/83c11a7d7f7b4264aad575c449728c2d

**Thank you** 🎉


src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -5,10 +5,34 @@
...
+			overflow: hidden;
+			position: absolute;
+			z-index: calc( var(--z-index-spinner-container) - 1 );
+		}
+	}
+
+	.tribe-events-calendar-month__event-multiday--start {
+		@media (--viewport-medium) {
+			border-radius: 6px 0 0 6px;
+			&.tribe-events-calendar-month__event-multiday--end {
```
paulmskim: instead of repeating this style here and below with `&.tribe-events-calendar-month__event-multiday--start`, we can do one of two things:

1. we can target `border-top-left-radius` and `border-bottom-left-radius` for `.tribe-events-calendar-month__event-multiday--start`, and `border-top-right-radius` and `border-bottom-right-radius` for `.tribe-events-calendar-month__event-multiday--end`
2. add a style for `.tribe-events-calendar-month__event-multiday--start.tribe-events-calendar-month__event-multiday--end` below both these styles with a `border-radius: 6px`.

I'd go with approach 1 for modularity.

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -5,10 +5,34 @@
...
+			border-radius: 6px;
+			background-color: white;
+			overflow: hidden;
+			position: absolute;
+			z-index: calc( var(--z-index-spinner-container) - 1 );
+		}
+	}
+
+	.tribe-events-calendar-month__event-multiday--start {
+		@media (--viewport-medium) {
```
paulmskim: don't need this as we don't show multiday events on mobile.

src/resources/postcss/views/month/_day.pcss
```diff
@@ -73,4 +73,13 @@
 	.tribe-events-calendar-month__day-date--disabled {
 		color: var(--color-text-disabled);
 	}
+
+	.tribe-events-calendar-month__day-events {
```
paulmskim: we should put these display styles on the entire calendar day rather than here, since we have a button for mobile.

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -5,10 +5,34 @@
 	 * ----------------------------------------------------------------------------- */
 
 	.tribe-events-calendar-month__event-multiday {
-		border-radius: 6px;
-		margin-bottom: var(--spacer-0);
-		background-color: white;
-		overflow: hidden;
+		display: none;
```
paulmskim: same here for putting display styles on entire calendar day

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -5,10 +5,34 @@
...
+
+		@media (--viewport-medium) {
+			display: block;
+			border-radius: 6px;
```
paulmskim: this shouldn't be a default style, only for `--start` or `--end`

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -26,6 +50,15 @@
 		}
 	}
 
+	.tribe-events-calendar-month__multiday-wrapper {
+
+		@media (--viewport-medium) {
```
paulmskim: shouldn't need this

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -5,10 +5,34 @@
 	 * ----------------------------------------------------------------------------- */
 
 	.tribe-events-calendar-month__event-multiday {
-		border-radius: 6px;
-		margin-bottom: var(--spacer-0);
-		background-color: white;
-		overflow: hidden;
+		display: none;
+
+		@media (--viewport-medium) {
```
paulmskim: shouldn't need this

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -42,4 +75,42 @@
...
+		width: calc( 600% + 6px );
+	}
+
+	.tribe-events-calendar-month__event-multiday-width-7 {
+		width: calc( 700% + 7px );
+	}
+
+	.tribe-events-calendar-month__event-multiday--empty,
+	.tribe-events-calendar-month__event-multiday--hidden {
+		display: none;
```
paulmskim: same thing with display styles on entire calendar day rather than each component

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -42,4 +75,42 @@
...
+
+		@media (--viewport-medium) {
+			visibility: hidden;
```
paulmskim: This style is fine for `--empty`, but `--hidden` should use an `opacity: 0` style.

src/views/v2/month/calendar-event/title.php
```diff
@@ -12,14 +12,17 @@
  * @version TBD
  *
  */
+$event    = $this->get( 'event' );
+$event_id = $event->ID;
 ?>
-<h3 class="tribe-events-calendar-month__calendar-event-title tribe-common-h8">
+<h3 class="tribe-events-calendar-month__calendar-event-title tribe-common-h8 tribe-events-tooltip"
```
paulmskim: the tooltip stuff should be on the anchor rather than the heading, since that's what the user is interacting with to reveal the tooltip

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * Selectors used for configuration and setup
+	 *
+	 * @since TBD
+	 *
+	 * @type {PlainObject}
+	 */
+	obj.selectors = {
+		tooltip: '.tribe-events-tooltip',
```
paulmskim: can we use `data-js` attributes here rather than classes? this helps us separate what is used for styles and what we're targeting for js.

src/views/v2/month/calendar-event/tooltip.php
```diff
@@ -12,11 +12,17 @@
...
-// $event    = $this->get( 'event' );
-// $event_id = $event->ID;
+$event    = $this->get( 'event' );
+$event_id = $event->ID;
 ?>
-<div class="tribe-events-calendar-month__calendar-event-tooltip">
-	<?php $this->template( 'month/calendar-event/tooltip/featured-image', [ 'event' => $event ] ); ?>
-	<?php $this->template( 'month/calendar-event/tooltip/description', [ 'event' => $event ] ); ?>
-	<?php /* RSVP Ticket CTA goes here */ ?>
+<div class="tribe-events-tooltip__content">
```
paulmskim: no need for wrapper class, add `data-js` attribute to div below.

src/views/v2/month/calendar-event/title.php
paulmskim: let's also use `data-js` attributes rather than the `.tribe-events-tooltip` class

src/views/v2/month/event-multiday-spacer.php
```diff
@@ -0,0 +1,19 @@
...
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+
+?>
+<div class="tribe-events-calendar-month__multiday-wrapper">
```
paulmskim: we don't need 2 divs here, we just need 1 with the right padding/height and margin

src/views/v2/month/day-events-multiday.php
```diff
@@ -0,0 +1,39 @@
...
+// Get the multiday events for that day
+// @todo: This is a function with demo purposes.
+$events_multiday = tribe_events_views_v2_month_demo_day_get_events_multiday( $month, $day_number );
+
+if ( ! $events_multiday ) {
+	return;
+}
+?>
+
+<?php foreach ( $events_multiday as $event ) : ?>
```
paulmskim: If you're putting a wrapper around `day-events`, then we should put a wrapper around `day-events-multiday` as well to keep consistent. But we also need to consider the different spacing cases here and think about whether having a wrapper for both makes sense or not.

src/views/v2/month/event-multiday.php
```diff
@@ -15,25 +15,55 @@
...
+// Check if it's featured.
+if ( $is_featured = isset( $event->featured ) && $event->featured ) { // @todo: later use tribe( 'tec.featured_events' )->is_featured( $event_id ) or similar
 	$classes[] = 'tribe-events-calendar-month__event-multiday--featured';
 }
+
+// If it starts today and this week, let's add the left border and set the width
+if ( $starts_today = $event->start_date == $day_number ) { // @todo:later we can check mm/dd or even year
+
+	// @todo: check if it ends this week or not, in order to split the duration
+	$classes[] = 'tribe-events-calendar-month__event-multiday-width-' . $event->duration;
```
paulmskim: should be a `--width-2` modifier

src/views/v2/month/event-multiday.php
```diff
@@ -15,25 +15,55 @@
...
+$classes_inner = [ 'tribe-events-calendar-month__event-multiday-inner' ];
 
-if ( $is_featured = tribe( 'tec.featured_events' )->is_featured( $event_id ) ) {
+// Check if it's featured.
+if ( $is_featured = isset( $event->featured ) && $event->featured ) { // @todo: later use tribe( 'tec.featured_events' )->is_featured( $event_id ) or similar
 	$classes[] = 'tribe-events-calendar-month__event-multiday--featured';
 }
+
+// If it starts today and this week, let's add the left border and set the width
+if ( $starts_today = $event->start_date == $day_number ) { // @todo:later we can check mm/dd or even year
```
paulmskim: The logic here, at least from description, seems a bit confusing. Maybe I'm not understanding, but the labels we use should also be descriptive enough for users to understand what's happening. To me, we should have one `T/F` flag that is `show`. If `show === true`, we do everything (`start_today` flag, `end_this_week` flag). If `show === false`, we only add `--hidden`.

src/views/v2/month/event-multiday.php
```diff
@@ -15,25 +15,55 @@
...
+		$classes[] = 'tribe-events-calendar-month__event-multiday--start';
+	}
+
+} else {
+	$classes[] = 'tribe-events-calendar-month__event-multiday--hidden';
+}
+
+// if it ends this week, let's add the right border
+$end_this_week = isset( $event->end_this_week ) && $event->end_this_week;
+if ( $end_this_week ) {
```
paulmskim: If the event is hidden, but still ends this week, we add unnecessary classes and apply unneeded styles.

src/views/v2/month/event-multiday.php
```diff
@@ -15,25 +15,55 @@
...
+	$classes_inner[] = 'tribe-events-calendar-month__event-multiday-inner--border-right';
+	$classes[] = 'tribe-events-calendar-month__event-multiday--end';
```
paulmskim: We should discuss this, as this was done elsewhere, but if we're adding a `--end` modifier, we shouldn't need the `--border-right` modifier on `tribe-events-calendar-month__event-multiday-inner` and just work off the `--end` modifier. Yes, we may be creating more another class to override this way, but the benefits are twofold: semantics and maintenance.

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -42,4 +75,42 @@
 	.tribe-events-calendar-month__event-multiday-title {
 		font-weight: var(--font-weight-regular);
 	}
+
+	.tribe-events-calendar-month__event-multiday-width-1 {
+		width: 100%;
```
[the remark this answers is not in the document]
paulmskim: this should also be the default for `__event-multiday`, the modifiers should be `--width-2`, `--width-3`, and so on.

juanfra:
Thanks! Yes, the purpose of this PR was to have everything for the demo. There was no time, unfortunately, to split everything into different PRs and treat each one. I'll go through the code reviews, thanks.

src/views/v2/month/event-multiday.php
juanfra: `--end` is for the wrapper, `--border-right` is for the inner.

The wrapper needs to be curved in order to match the design, except if it's an event that continues next week. In that case the multiday edge on the right side should be just straight, and for that we need to modify the wrapper.

src/Tribe/Views/V2/Assets.php
```diff
@@ -33,7 +33,7 @@ public function register() {
 			$plugin,
 			'tribe-events-calendar-views-v2',
 			'views/tribe-events-v2.css',
-			[ 'tribe-common-style' ],
+			[ 'tribe-common-style', 'tribe-tooltipster-css' ],
```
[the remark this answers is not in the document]
juanfra: That was one of my doubts. If we're going to have a toggle for the "Hide Recurring Events" as we been told, then I think the tooltip should be included everywhere. I'll add the comment but we need to talk about the tooltip implementation project-wide.

src/Tribe/Views/V2/Assets.php
```diff
@@ -45,5 +45,14 @@ public function register() {
 			[ 'jquery', 'tribe-common', 'tribe-query-string', 'underscore' ],
 			null // prevent it from loading
 		);
+
+		tribe_asset(
+			$plugin,
+			'tribe-events-views-scripts',
+			'views/scripts.js',
```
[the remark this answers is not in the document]
juanfra: The idea was to include the general views scripts here. Do you feel we should be creating a file for each? (tooltip, a11y, events-bar, etc)

src/resources/postcss/base/_layouts.pcss
```diff
@@ -8,6 +8,7 @@
 	.tribe-events-l-container {
 		margin-bottom: var(--spacer-11);
 		margin-top: var(--spacer-10);
+		position: relative; /* making this relative so the loader can take its dimension */
```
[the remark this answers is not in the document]
juanfra: Gustavo put that over `.tribe-events` and it was bothering the tooltips (as the tooltips are on a div at the body level, we needed to use `.tribe-events` on them so the styles inside of them are actually OK).

So, I ended up moving it to the layout container so the loader can take the space that he wanted.

src/resources/postcss/components/_tooltip.pcss
```diff
@@ -0,0 +1,40 @@
...
+	max-width: 254px;
+
+	.tooltipster-box {
+		background: transparent;
+		border: none;
+		border-radius: 0;
+		box-shadow: none;
+		margin: 0;
+
+		.tooltipster-content {
```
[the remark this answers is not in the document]
juanfra: Yes, it's their theme, their way of doing things and this is our way to override and "reset" their styles.

src/resources/postcss/views/month/_multiday-event.pcss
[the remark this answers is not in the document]
juanfra: We can get one of duration 1 if it starts in a week and it ends on the first one of the next week.

src/Tribe/Views/V2/Assets.php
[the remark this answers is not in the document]
paulmskim: I think so, there are scripts (multiday) that should only run on month view, so if we can split them up that's best.

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+	/**
+	 * Handles the initialization of the scripts when Document is ready
+	 *
+	 * @since  TBD
+	 *
+	 * @return {void}
+	 */
+	obj.ready = function() {
+
+		obj.initTooltips();
```
bordoni: The initalization of the tooltips should be hooked up to the Views Manager, on the `afterSetup.tribeEvents` action from the container.

Something similar to:
`tribe.events.views.manager.$containers.on( 'afterSetup.tribeEvents', obj.initTooltips );`

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+		tooltipContent: '.tribe-events-tooltip__content',
+	};
+
+	/**
+	 * Initialize accessible tooltips via tooltipster
+	 *
+	 * @since TBD
+	 *
+	 */
+	obj.initTooltips = function() {
```
bordoni: Should we use the same structure for the method names like on `manager.js` using `setup`.
I feel like there is no need for a `Tooltips`. We should move each piece of "Script" into it's own file, so it can be removed without issues.

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+tribe.events.views = tribe.events.views || {};
+
+/**
+ * Configures Views Object in the Global Tribe variable
+ *
+ * @since  TBD
+ *
+ * @type   {PlainObject}
+ */
+tribe.events.views.scripts = {};
```
bordoni: Let's create one file for each script, e.g.: `tooltip`, `datepicker`, `month-day-picker`

src/resources/js/views/scripts.js
bordoni: I agree with the above.

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+	 *
+	 * @since TBD
+	 *
+	 */
+	obj.initTooltips = function() {
+
+		$( obj.selectors.tooltip ).tooltipster( {
+			interactive: true,
+			theme: [ 'tribe-common', 'tribe-events', 'tribe-events-tooltip-theme' ],
+			functionInit: function( instance, helper ) {
```
bordoni: Move this method to the `obj`, so it makes it easier to extend this code.

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+				instance.content( content );
+				$( helper.origin )
+					.focus( function() {
+						$( this ).tooltipster( 'open' );
+					})
+					.blur( function() {
+						$( this ).tooltipster( 'close' );
+				});
+			},
+			functionReady: function( instance, helper ) {
```
bordoni: Same here as the `functionInit.`

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+				$( helper.origin ).find( obj.selectors.tooltipContent ).attr( 'aria-hidden', false );
+			},
+			functionAfter: function( instance, helper ) {
```
bordoni: Same here as the `functionInit.`

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+				$( helper.origin ).find( obj.selectors.tooltipContent ).attr( 'aria-hidden', true );
```
bordoni: Should we use `prop`?

src/resources/js/views/scripts.js
```diff
@@ -0,0 +1,93 @@
...
+				$( helper.origin )
+					.focus( function() {
+						$( this ).tooltipster( 'open' );
+					})
+					.blur( function() {
+						$( this ).tooltipster( 'close' );
+				});
+			},
+			functionReady: function( instance, helper ) {
+				$( helper.origin ).find( obj.selectors.tooltipContent ).attr( 'aria-hidden', false );
```
bordoni: Should we use `prop`?

src/views/v2/month/event-multiday.php
paulmskim: Right, I mean to do something like this:

```
.tribe-events-calendar-month__event-multiday--end {

    .tribe-events-calendar-month__event-multiday-inner {
        // styles for --border-right here
    }
}
```

src/resources/postcss/views/month/_multiday-event.pcss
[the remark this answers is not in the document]
paulmskim: ah you're right, my bad. either way, I think we should make this the default rather than adding a class for it.

src/resources/postcss/components/_tooltip.pcss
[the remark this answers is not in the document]
paulmskim: sounds good

src/Tribe/Views/V2/Assets.php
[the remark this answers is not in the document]
juanfra: Maybe a general one (it was my intention for this one) for the ones that should be part of all views and then one for each view? (month-view.js with multiday, grid a11y). It'd be great if we can define if we group by view or by feature.

src/resources/js/views/scripts.js
juanfra: this is a method provided by `tooltipster`, and it needs to be declared there.

src/resources/js/views/scripts.js
juanfra: this is a method provided by `tooltipster`, and it needs to be declared there.

src/resources/js/views/scripts.js
juanfra: this is a method provided by `tooltipster`, and it needs to be declared there.

src/resources/js/views/scripts.js
juanfra: This is an a11y recommendation by tooltipster to make them accessible > https://iamceege.github.io/tooltipster/#accessibility

src/resources/js/views/scripts.js
juanfra: This is an a11y recommendation by tooltipster to make them accessible > https://iamceege.github.io/tooltipster/#accessibility

src/resources/postcss/views/month/_multiday-event.pcss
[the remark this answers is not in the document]
juanfra: Yeah can do that, but ideally if we have the logic set in place to add the number dynamically to the class, we'll be adding the `-1`, unless we specifically state the conditional to be `$duration > 1`

src/resources/postcss/views/month/_day.pcss
juanfra: The idea of this one was also to wrap the regular events and add the required side padding (no padding or anything for the multiday events).

juanfra: And also, wrapping the entire day will make us re-work on the whole multiday structure as it won't work the way it is.

src/resources/postcss/views/month/_multiday-event.pcss
juanfra: Yes, it is required, because if not when switching to mobile, it continue taking the height and it looks weird: https://cloudup.com/cT2HSGJN2nL

src/resources/js/views/scripts.js
bordoni: Normally for Prop is the recommended for booleans like that one. But TBH either way is fine by me.

src/resources/js/views/scripts.js
bordoni: I meant to say, that it would be good to declare it on the object like:
```
{
...
    functionReady: obj.onFunctionReady,
    functionAfter: obj.onFunctionAfter,
    functionInit: obj.onFunctionInit
}
```

src/resources/postcss/views/month/_multiday-event.pcss
juanfra: `--empty` can have an `opacity:0;` but for hidden need to be have `visibility: hidden;` because if not, the element comes next in the dom and the hover for multiday doesn't work.

src/views/v2/month/event-multiday.php
juanfra: I'm not sure if I'm following you on this one. We shouldn't be hardcoding here, it comes from the data structure.

src/views/v2/month/event-multiday.php
juanfra: If it's starting today, it means it should be shown. `$starts_today` is a boolean (it can be renamed to `$should_show` or something else) that is defined if the event starting date equals `$day_number` (the day number of the cell)

src/views/v2/month/event-multiday.php
juanfra: I thought we were trying to avoid nesting. Are we making an exception on this one?

src/resources/postcss/views/month/_calendar-event.pcss
```diff
@@ -7,6 +7,10 @@
 	.tribe-events-calendar-month__calendar-event {
 		margin: var(--spacer-1) 0;
 		position: relative;
+
+		&:first-of-type {
```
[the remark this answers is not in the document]
juanfra: https://www.loom.com/share/89aba7b05b52476c9b61804cc0a4d5cc

src/views/v2/month/event-multiday.php
paulmskim: I understand, I mean the double `--` before the `width` as it is a modifier, I used `--width-2` as an example.

src/views/v2/month/event-multiday.php
paulmskim: Yes, I think using a `$should_show` variable name would be better and easier to understand.

src/views/v2/month/event-multiday.php
paulmskim: Yes you're right, we are trying to avoid nesting. But in cases like this, the semantics surrounding this is specifically around the multiday event. We are saying that this is a case where we have an `--end` modifier to the multiday, that the multiday event component has an end cap. So given this case, we need to add all the styles that make sense for an `--end` modifier.

The difference between the above and having both `.tribe-events-calendar-month__event-multiday--end` and `tribe-events-calendar-month__event-multiday-inner--border-right` styles is around semantics and maintenance. For maintenance, now we have 2 places to make changes if we decide to change anything. For semantics, we could technically now have `tribe-events-calendar-month__event-multiday-inner--border-right` styles without `.tribe-events-calendar-month__event-multiday--end` styles, as these are 2 different modifiers. However, in the first example above, we can't have `tribe-events-calendar-month__event-multiday-inner--border-right` styles without `.tribe-events-calendar-month__event-multiday--end` as it is all controlled by 1 class, the `--end` modifier.

Hopefully these examples make sense. If not, let me know and we can hop on a call for it.

src/views/v2/month/event-multiday.php
```diff
@@ -15,25 +15,55 @@
...
 	$classes[] = 'tribe-events-calendar-month__event-multiday--featured';
 }
+
+// If it starts today and this week, let's add the left border and set the width
+if ( $starts_today = $event->start_date == $day_number ) { // @todo:later we can check mm/dd or even year
+
+	// @todo: check if it ends this week or not, in order to split the duration
+	$classes[] = 'tribe-events-calendar-month__event-multiday-width-' . $event->duration;
+
+	if ( isset( $event->start_this_week ) && $event->start_this_week ) {
```
paulmskim: `start_this_week` for the `--start` modifier is misleading, shouldn't this be where we have the `start_today` flag?

src/views/v2/month/event-multiday.php
```diff
@@ -15,25 +15,55 @@
...
-				title="<?php esc_attr_e( 'Featured', 'the-events-calendar' ); ?>"
-			></em>
-		<?php endif; ?>
-		<h3 class="tribe-events-calendar-month__event-multiday-title tribe-common-h8">
-			Lorem ipsum
-		</h3>
-	</a>
+<div class="tribe-events-calendar-month__event-multiday-wrapper">
+
+	<div class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>">
```
paulmskim: maybe we should use `article` here too, much like the single day event

paulmskim: also could we add a `data-id` attribute here with the event id? this is for the JS.

src/Tribe/Views/V2/Assets.php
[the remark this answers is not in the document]
bordoni: Lets go with what we talked earlier today @juanfra we want something very similar to what @paulmskim suggested, as modular as possible. No ties to a particular View, it should be dependent on been included and on the selectors.

src/resources/postcss/views/month/_calendar-event.pcss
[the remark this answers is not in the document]
paulmskim: So you're right in that case, but that's only one case. Other cases include: if we have a featured event (it has `--spacer-1` spacing above and below it), this could be below a multiday event too, if we have a regular event below a multiday event (it has a `--spacer-0`). What I mean by the wrapper around these single day events is that the `:first-of-type` selector will be relative to the `div` as a child of the `.tribe-events-calendar-month__day-events` wrapper, where the first single day event will always be the first of type.

So given all these cases, we need to think about whether we want the `.tribe-events-calendar-month__day-events` wrapper (and same for multiday) or not, and what the specific spacing cases will be.

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -5,10 +5,34 @@
...
-		overflow: hidden;
+		display: none;
+
+		@media (--viewport-medium) {
+			display: block;
+			border-radius: 6px;
+			background-color: white;
+			overflow: hidden;
+			position: absolute;
+			z-index: calc( var(--z-index-spinner-container) - 1 );
```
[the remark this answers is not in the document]
paulmskim: question: do we need a positive `z-index` here, or can we get away with just `position: absolute`? are we trying to place it above something else?

src/resources/postcss/views/month/_multiday-event.pcss
```diff
@@ -5,10 +5,26 @@
 	 * ----------------------------------------------------------------------------- */
 
 	.tribe-events-calendar-month__event-multiday {
-		border-radius: 6px;
-		margin-bottom: var(--spacer-0);
-		background-color: white;
-		overflow: hidden;
+		display: none;
+		min-width: calc(100% + 1px);
```
paulmskim: let's set this as `width`, as `min-width` will have unintended `width` consequences if we have long content.

src/resources/postcss/views/month/_multiday-event.pcss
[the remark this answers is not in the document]
paulmskim: This will also affect the hidden ones, which have an effect on the hover. The hidden ones will not get the `--width-1` class, which is why we want this as default.

src/resources/postcss/views/month/_multiday-event.pcss
paulmskim: So in https://github.com/moderntribe/the-events-calendar/pull/2537/files#diff-a7442de4a8d47551d117e298780f4c19R30 I'll be adding separate markup for mobile and hiding the mobile on desktop (just the way mobile works, we can't use the same markup). `.tribe-events-calendar-month__day-cell` should be hidden on mobile and display on desktop.

If you want, you can merge as is and I'll fix it up on my branch when I merge the release branch into the mobile work.

src/resources/postcss/views/month/_day.pcss
paulmskim: the padding is fine, i mean the `display: none` and `display: block`. see https://github.com/moderntribe/the-events-calendar/pull/2544#discussion_r291478904

src/resources/postcss/views/month/_multiday-event.pcss
paulmskim: not sure i follow you here. isn't `--empty` supposed to be a blank spacer? that should work with no styles for opacity or visibility, just height/padding and margin. but for `--hidden`, yes the point is for that to lay above the visible multiday. That is what the JS handles to allow for those on hover and focus to display the correct styles on the visible multiday event. That is what's happening below:

https://www.loom.com/share/7146f385e65f4911b09aa6b052815476

src/Tribe/Views/V2/Assets.php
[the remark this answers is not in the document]
lucatume: Neat!

src/views/v2/month/day-events-multiday.php
```diff
@@ -0,0 +1,39 @@
...
+ * [your-theme]/tribe/events/views/v2/month/day-events-multiday.php
+ *
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+$day_number = $this->get( 'day' );
```
lucatume: Should we not use the `$day` and `$month` extracted vars here?

src/views/v2/month/day-events.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+$day_number = $this->get( 'day' );
+$month      = $this->get( 'month' );
```
lucatume: Same question as above here: should we not use the `$day` and `$month` extracted vars?

src/views/v2/month/day-events.php
```diff
@@ -0,0 +1,35 @@
...
+
+// Get the multiday events for that day
+// @todo: This is a function with demo purposes.
+$events_regular = tribe_events_views_v2_month_demo_day_get_events_regular( $month, $day_number );
```
lucatume: I think this beats me at any game of long names.

src/views/v2/month/day.php
```diff
@@ -14,7 +14,8 @@
  */
 // @todo: This is just for presentation purposes, while building the view.
 $day_number   = ( $week < 1 ) ? ( $day + 1 ) * ( $week + 1 ) : ( $day + 1 ) + $week * 7;
-$month_number = 5;
+$month_number = 6;
+$month = $this->get( 'month' );
```
lucatume: Use extracted `$month` var here?

tests/views_integration/Tribe/Events/Views/V2/HTML/Month/MonthEventMultidayTest.php
```diff
@@ -9,7 +9,20 @@ class MonthEventMultidayTest extends TestHtmlCase {
 	 * @test
 	 */
 	public function it_should_contain_correct_html_classes() {
-		$template = $this->template->template( 'month/event-multiday', [ 'event' => (object) [ 'ID' => 0 ] ] );
+
+		$event = [
```
lucatume: Why not use the Event factory here? See https://github.com/moderntribe/the-events-calendar/blob/9ef75801268a195c91f0525d13f1693f9c305288/tests/_support/Factories/Event.php#L7 and https://github.com/moderntribe/the-events-calendar/blob/7b736aadc603b10bbb8a46a6ac12514882edd8e3/tests/wpunit/Tribe/Events/REST/V1/Endpoints/Single_Organizer_SlugTest.php#L30

tests/views_integration/Tribe/Events/Views/V2/HTML/Month/MonthEventMultidayTest.php
```diff
@@ -37,7 +50,20 @@ public function it_should_contain_correct_html_classes() {
 	 * @test
 	 */
 	public function it_should_contain_a11y_attributes() {
-		$template = $this->template->template( 'month/event-multiday', [ 'event' => (object) [ 'ID' => 0 ] ] );
+
+		$event = [
```
lucatume: Use the event factory here.

src/views/v2/month/event-multiday.php
juanfra: oh, I see what you mean now.

src/views/v2/month/event-multiday.php
juanfra: Yes, I didn't put too much focus on that because I feel these will be replaced with template tags, and they will be self descriptive.

src/resources/js/views/scripts.js
bordoni: As we discussed on hangout, we should still move it to the exposed `obj` as a method, so there is an easier way to extend this.

src/resources/js/views/scripts.js
bordoni: As we discussed on hangout, we should still move it to the exposed `obj` as a method, so there is an easier way to extend this.

src/views/v2/month/event-multiday.php
juanfra: Yes, I get it - But at some point I feel we're having cases where we can do the same for `--featured` for example, but still, and we're not doing it. Trying to find some clarity on where is that we should be doing the exceptions.

src/resources/postcss/views/month/_multiday-event.pcss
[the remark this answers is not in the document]
juanfra: https://www.loom.com/share/e9b86afcfc6643328a6afd0fde3be2ab

src/views/v2/month/day.php
juanfra: Maybe it's a bit confusing, but `$month` here, is the array containing the month data structure (with events on each day). It's hooked from `month-view-demo-template-tags.php` where the function with long names live.

tests/views_integration/Tribe/Events/Views/V2/HTML/Month/MonthEventMultidayTest.php
juanfra: Yep, I was just setting the smallest piece of data that the template will require so the test doesn't die. And the idea was coming back once the templates actually have real data an adapt the tests. (Basically, my idea of adding the tests at this point is that will make us come back and fix them later, and we are sure we'll have tests for that template, but at this point there's no data or anything to test with).

src/resources/postcss/views/month/_multiday-event.pcss
juanfra: The thing when I set that to width is that it was defaulting the rest (even with modifiers) to 100%

src/views/v2/month/day-events-multiday.php
juanfra: We need the day number to calculate if it's starting today (`$day_number` is the cell number) and `$month` contains the month data (with events, to calculate how long the event last, etc)

src/views/v2/month/day-events.php
juanfra: We need the day number to retrieve the events from `$month` (`$day_number` is the cell number).

src/resources/postcss/views/month/_multiday-event.pcss
juanfra: Okies.

src/resources/postcss/views/month/_multiday-event.pcss
juanfra: Ok, didn't know that you were taking that for the JS, thanks for the clarification.

src/views/v2/month/event-multiday.php
juanfra: The case in which we have a multiday event spanning across two different weeks forces us to do it that way. Let's say the week start on monday, and you have a multiday event that comes from the past week, then that event won't be starting that week: sharp edge on the left and no left border.

src/views/v2/month/event-multiday.php
paulmskim: We should be doing it for the `--featured` cases too. There were a couple cases earlier where it was done like this, but we should be sticking to only one class that affects the modified case.

src/resources/postcss/views/month/_multiday-event.pcss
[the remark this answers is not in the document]
paulmskim: Ah that's because of the `position: relative` on the next cell's wrapper. Not having a `z-index` should not be a problem once we put in the JS, I can handle that when we merge everything with the JS.

src/views/v2/month/event-multiday.php
paulmskim: No worries, I had something else in mind but this works too.

src/resources/postcss/views/month/_multiday-event.pcss
paulmskim: it sounds like you've got competing styles somewhere. given the order of this CSS document, the `--width-#` modifiers come after this class, so it should override this style if using `width: calc(100% + 1px);`.

src/resources/postcss/views/month/_multiday-event.pcss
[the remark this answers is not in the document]
juanfra: Awesome 🎉

src/views/v2/month/calendar-event/tooltip.php
juanfra: Yes, unfortunately for the a11y implementation we need a wrapper.

src/views/v2/month/event-multiday.php
juanfra: Cool, as long as we follow a pattern, I'm in, will change that.

src/resources/postcss/views/month/_calendar-event.pcss
```diff
@@ -37,16 +41,8 @@
...
-		border: 1px solid var(--color-border-default);
-		box-shadow: var(--box-shadow-tooltip);
-		display: none;
-		left: calc(var(--spacer-7) * -1);
-		padding: var(--spacer-4);
-		position: absolute;
-		top: calc(100% + var(--spacer-1));
-		width: calc(100% + (2 * var(--spacer-7)));
+	.tribe-events-calendar-month__calendar-event-title-link {
+		color: var(--color-text-primary);
```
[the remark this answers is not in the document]
juanfra: I'm not sure if I follow you on this one. I'd be all in for setting the color on `.tribe-common a`, otherwise on classes like this one. If not, any theme will end up having their own `a` color style. And the idea would be to be in control of the link color for the calendar event title, right?

tests/views_integration/Tribe/Events/Views/V2/HTML/Month/MonthEventMultidayTest.php
juanfra: And also, at this point we're not actually using an `$event` per se, but a simple small implementation of an object with a few pieces of info to display the demo. Will leave a `@todo`

src/resources/postcss/views/month/_multiday-event.pcss
juanfra: Perfect, looking at that 👍

src/resources/postcss/views/month/_calendar-event.pcss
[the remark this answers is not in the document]
paulmskim: Yup, `.tribe-common a would be the way to go. We want to override user agent and theme styles.

src/Tribe/Views/V2/Assets.php
```diff
@@ -33,7 +33,7 @@ public function register() {
 			$plugin,
 			'tribe-events-calendar-views-v2',
 			'views/tribe-events-v2.css',
-			[ 'tribe-common-style' ],
+			[ 'tribe-common-style', 'tribe-tooltipster-css' ], // @todo: check if we're including tooltips only in month view.
 			'wp_enqueue_scripts',
```
bordoni: We should have logic here to only load this in the correct pages.

src/Tribe/Views/V2/Assets.php
```diff
@@ -45,5 +45,14 @@ public function register() {
 			[ 'jquery', 'tribe-common', 'tribe-query-string', 'underscore' ],
 			null // prevent it from loading
 		);
+
+		tribe_asset(
+			$plugin,
+			'tribe-events-views-v2-tooltip',
+			'views/tooltip.js',
+			[ 'jquery', 'tribe-common', 'tribe-events-views-v2-manager', 'tribe-tooltipster' ],
+			'wp_enqueue_scripts',
```
bordoni: We should have logic here to only load this in the correct pages. I mean on all of the new V2 views.

src/Tribe/Views/V2/Service_Provider.php
```diff
@@ -46,6 +46,9 @@ public function register() {
 
 		// Since the View main class will act as a DI container itself let's provide it with the global container.
 		View::set_container( $this->container );
+
+		// @todo: remove this when we hydrate the month view with data and we use the correct template tags.
+		require_once tribe( 'tec.main' )->plugin_path . 'src/Tribe/Views/V2/month-view-demo-template-tags.php';
```
bordoni: Can we move this into the template files for month view? I makes it easier to not forget.

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+	 * @since TBD
+	 *
+	 */
+	obj.onFunctionInit = function( instance, helper ) {
+
+		var content = $( helper.origin ).find( obj.selectors.tooltipContent ).html();
+		instance.content( content );
+		$( helper.origin )
+			.focus( function() {
+				$( this ).tooltipster( 'open' );
```
bordoni: Move this into a `onOriginFocus`

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+			})
+			.blur( function() {
+				$( this ).tooltipster( 'close' );
```
bordoni: Move this into an `onOriginBlur`

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+	 *
+	 * @since TBD
+	 *
+	 */
+	obj.initTooltips = function() {
+
+		$( obj.selectors.tooltip ).tooltipster( {
+			interactive: true,
+			theme: [ 'tribe-common', 'tribe-events', 'tribe-events-tooltip-theme' ],
+			functionInit: function( instance, helper ) {
```
bordoni:
```suggestion
			functionInit: obj.onFunctionInit,
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+				obj.onFunctionInit( instance, helper );
```
bordoni:
```suggestion
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+			},
```
bordoni:
```suggestion
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+			functionReady: function( instance, helper ) {
```
bordoni:
```suggestion
			functionReady: obj.onFunctionReady,
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+				obj.onFunctionReady( instance, helper );
```
bordoni:
```suggestion
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+			},
```
bordoni:
```suggestion
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+			functionAfter: function( instance, helper ) {
```
bordoni:
```suggestion
			functionAfter: obj.onFunctionAfter,
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+				obj.onFunctionAfter( instance, helper );
```
bordoni:
```suggestion
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+			}
```
bordoni:
```suggestion
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * Handles the initialization of the scripts when Document is ready
+	 *
+	 * @since  TBD
+	 *
+	 * @return {void}
+	 */
+	obj.ready = function() {
+		tribe.events.views.manager.$containers.on( 'afterSetup.tribeEvents', obj.initTooltips() );
```
bordoni: No need to call the method, just pass it as a variable.
```suggestion
		tribe.events.views.manager.$containers.on( 'afterSetup.tribeEvents', obj.initTooltips );
```

src/resources/js/views/tooltip.js
```diff
@@ -0,0 +1,132 @@
...
+	 *
+	 * @since TBD
+	 *
+	 */
+	obj.onFunctionInit = function( instance, helper ) {
+
+		var content = $( helper.origin ).find( obj.selectors.tooltipContent ).html();
+		instance.content( content );
+		$( helper.origin )
+			.focus( function() {
```
bordoni: would love to see us using `.on( 'focus', obj. onOriginFocus )`

src/views/v2/index.php
```diff
@@ -1,8 +1,9 @@
...
-
-get_header();
-
-echo tribe( Template_Bootstrap::class )->get_view_html();
-
-get_footer();
\ No newline at end of file
+// @TODO: We're setting the div wrapper over here, but we'll need to revise this.
+get_header();?>
+<div class="tribe-common tribe-events">
```
bordoni: These need to be included inside of each view, not here.

juanfra: Is there any specific reason for this? This wrapper must be part of all views, because it's the one that "give sense" to all of our styles. Wouldn't it make sense not to repeat ourselves on each view and have it outside?

src/Tribe/Views/V2/Assets.php
```diff
@@ -33,9 +33,12 @@ public function register() {
 			$plugin,
 			'tribe-events-calendar-views-v2',
 			'views/tribe-events-v2.css',
-			[ 'tribe-common-style' ],
+			[ 'tribe-common-style', 'tribe-tooltipster-css' ], // @todo: check if we're including tooltips only in month view.
 			'wp_enqueue_scripts',
-			[ 'priority' => 10 ]
+			[
+				'priority'     => 10,
+				'conditionals' => array( $this, 'should_enqueue_frontend' ),
```
bordoni: use `[]`

src/Tribe/Views/V2/Assets.php
```diff
@@ -45,5 +48,40 @@ public function register() {
...
+
+		tribe_asset(
+			$plugin,
+			'tribe-events-views-v2-tooltip',
+			'views/tooltip.js',
+			[ 'jquery', 'tribe-common', 'tribe-events-views-v2-manager', 'tribe-tooltipster' ],
+			'wp_enqueue_scripts',
+			[
+				'priority'     => 10,
+				'conditionals' => array( $this, 'should_enqueue_frontend' ),
```
bordoni: use `[]`

src/Tribe/Views/V2/Assets.php
```diff
@@ -45,5 +48,40 @@ public function register() {
...
+	/**
+	 * Checks if we should enqueue frontend assets for the V2 views
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function should_enqueue_frontend() {
+
+		$display_mode = tribe_context()->get( 'event_display_mode' );
```
bordoni: We should use the `Template Bootstrap` for checking

src/resources/postcss/views/month/_calendar-event.pcss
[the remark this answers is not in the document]
ryanurban: ^ Agreed

PR #2547, opened 2019-06-06
fix(Url.php): remove imports and usages of `parse_query` function from Guzzle

lucatume:


juanfra:
The same comment as the one I left in common. Let's be sure these changes are not lost in the middle in case we start with B19.08

PR #2551, opened 2019-06-07
Newsletter signup changes

barryhughes:
Updates the newsletter signup forms (which are embedded in the welcome and update screens) so that they submit to the Support API platform. Functionality is essentially as it used to be, but the data will no longer be sent directly to createsend.com.

replacement-newsletter-signups

:warning: Important! This change depends on [Support API PR 28](https://github.com/moderntribe/support-api.tri.be/pull/28), which introduces the endpoint we need for this to work. That change must be approved and be deployed before this work ships.

:ticket: [♯129034](https://central.tri.be/issues/129034)

### DIFF
--- a/dev/docker/ci-compose.yml
+++ b/dev/docker/ci-compose.yml
@@ -49,7 +49,7 @@ services:
 
   chromedriver:
     container_name: ci_chromedriver
-    image: selenium/standalone-chrome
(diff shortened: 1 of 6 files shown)


barryhughes:
:memo: Tests are failing, but it is in relation to rewrite, theme compat and other tests unrelated to this change.

bordoni:
Changelog.

PR #2557, opened 2019-06-10
fix(Context): move TEC specific locations in dedicated service provider

lucatume:


bordoni:
Why are the tests failing here?

lucatume:
> Why are the tests failing here?

Common is not aligned.

PR #2559, opened 2019-06-12
List View TEC filters support

lucatume:
Ticket: https://central.tri.be/issues/127992

Based on: https://github.com/moderntribe/the-events-calendar/pull/2541


src/resources/js/views/manager.js
```diff
@@ -200,11 +200,23 @@ tribe.events.views.manager = {};
...
 	obj.onSubmit = function( event ) {
 		event.preventDefault();
-		var $container = $( this );
-		var formData = Qs.parse( $container.serialize() );
 
-		// pass the data to the request using `tribe-events-views`
-		obj.request( formData['tribe-events-views'], $container );
+		// The submit event is triggered on the form, not the container.
+		var $form = $( this );
+		var $container = $form.closest('.tribe-events');
```
bordoni: Spacing on this JS file seems to be missing.

src/resources/js/views/manager.js
```diff
@@ -200,11 +200,23 @@ tribe.events.views.manager = {};
...
+		var nonce = $container.data('view-rest-nonce');
+
+		var formData = Qs.parse($form.serialize());
+		var viewData = formData['tribe-events-views'] || {};
```
bordoni: why are we doing this relaying of information instead of allowing the form data to do its job?
This would make subviews more complicated, would love to see if we could keep it using the form fields, and if URL field is not available uses the `window.location.href`.

src/Tribe/Views/V2/View.php
```diff
@@ -830,4 +859,34 @@ protected function filter_view_url( $canonical, string $url ) {
...
+	 * @return array An array of Template variables for the View Template.
+	 */
+	protected function setup_template_vars(  ) {
+		$template_vars = [
+			'title'       => wp_title( null, false ),
+			'events'      => $this->repository->all(),
+			'url'         => $this->get_url( true ),
+			'prev_url'    => $this->prev_url( true ),
+			'next_url'    => $this->next_url( true ),
+			'bar_keyword' => $this->context->get( 'keyword', '' ),
```
bordoni: Why not pass these two Bar stuff as an array?

src/resources/js/views/manager.js
lucatume: I'm not sure I understand the question here.

src/Tribe/Views/V2/View.php
lucatume: I've modified the code to do that.

PR #2560, opened 2019-06-13
[VR] List view refinements

juanfra:
🎫  https://central.tri.be/issues/129307

Updating markup and styles to match new design.


src/views/v2/list.php
```diff
@@ -40,7 +40,7 @@ class="tribe-common tribe-events"
 
 		<div class="tribe-events-calendar-list">
 
-			<?php $this->template( 'list/month-separator', [ 'month' => date( 'M' ) ] ); ?>
+			<?php $this->template( 'list/month-separator', [ 'month' => date( 'M' ), 'year' => date( 'Y' ) ] ); ?>
```
bordoni: Lets prevent problems, and pass a single date variable.

src/views/v2/list/month-separator.php
```diff
@@ -13,7 +13,8 @@
  *
  */
 $month = $this->get( 'month' );
+$year  = $this->get( 'year' )
 ?>
 <div class="tribe-events-calendar-list__separator-month">
-	<time class="tribe-events-calendar-list__separator-month-text tribe-common-b1 tribe-common-b1--bold" datetime="1970-01-01T00:00:00+00:00"><?php echo esc_html( $month ); ?></time>
+	<time class="tribe-events-calendar-list__separator-month-text tribe-common-b1" datetime="1970-01-01T00:00:00+00:00"><?php echo esc_html( $month ); ?> <?php echo esc_html( $year ); ?></time>
```
bordoni: Break into multiple lines
And lets make sure the `datetime=""` is generated dynamically.

src/resources/postcss/views/_list.pcss
```diff
@@ -93,9 +93,12 @@
 	}
 
 	.tribe-events-calendar-list__event-title {
+		@mixin heading;
+		@mixin heading-6;
```
paulmskim: I'm seeing only `h5` in the designs, we should add the `h5` class to the markup.

src/views/v2/list/event/description.php
```diff
@@ -15,6 +15,6 @@
 $event    = $this->get( 'event' );
 $event_id = $event->ID;
 ?>
-<div class="tribe-events-calendar-list__event-description tribe-common-b1">
+<div class="tribe-events-calendar-list__event-description tribe-common-b2">
```
paulmskim: I'm still seeing `b1` here, there is one event in list view that has `b2` styles it seems, but all other events have `b1` styles. Check with matt/rachel?

src/views/v2/list/event/title.php
```diff
@@ -15,7 +15,7 @@
 $event    = $this->get( 'event' );
 $event_id = $event->ID;
 ?>
-<h3 class="tribe-events-calendar-list__event-title tribe-common-h4">
```
paulmskim: keep this but change to `h5`

src/views/v2/list/month-separator.php
```diff
@@ -12,8 +12,15 @@
...
  */
-$month = $this->get( 'month' );
+$date  = $this->get( 'date' );
+$month = date( 'M', $date );
+$year  = date( 'Y', $date );
 ?>
 <div class="tribe-events-calendar-list__separator-month">
-	<time class="tribe-events-calendar-list__separator-month-text tribe-common-b1 tribe-common-b1--bold" datetime="1970-01-01T00:00:00+00:00"><?php echo esc_html( $month ); ?></time>
+	<time
+		class="tribe-events-calendar-list__separator-month-text tribe-common-b1"
```
paulmskim: This should be `h7` on mobile and `h6` on desktop. Let's add the `h7` class and override with `@mixin heading-6` in the separator month text class.

src/resources/postcss/views/_list.pcss
juanfra: If you inspect the designs, you will see that for mobile is 6 (16px) and for desktop, 5.

src/views/v2/list/event/description.php
juanfra: Unfortunately Matt is not here, and he was in charge of this refresh. I recall he mentioned that updating the size for this was one of the changes, but maybe he referred to the featured only. Asked Rachel and she's not sure.

src/views/v2/list/event/title.php
juanfra: We are updating the styles. I don't understand why we should change the semantics. Can you please expand a bit more on this request?

src/views/v2/list/month-separator.php
juanfra: Why should it be that? The size is for the b1 match the design and they all have the same line height. Would you please expand a bit more on why you think it should be like that?

src/resources/postcss/views/_list.pcss
paulmskim: Ah I assumed we had different mobile and desktop heading sizes for heading 5, by looking at the class names mobile says `mobile-heading-5` and desktop says `desktop-heading-5`. This is something we should check with matt/rachel on, if they named these classes this way by accident or it's intended. The typography sheet hasn't been updated either, so we should reconcile this.

src/views/v2/list/event/description.php
paulmskim: I'm going off the latest designs on invision. Confirming with Matt/Rachel is best.

src/views/v2/list/event/title.php
paulmskim: I don't mean the tag, i meant change to `tribe-common-h5` class. This is connected to https://github.com/moderntribe/the-events-calendar/pull/2560#discussion_r294078672

src/views/v2/list/month-separator.php
paulmskim: I'm going based off the class names when you inspect the elements. My takeaway from the class names is it gives insight into Matt and Rachel's thoughts when they were designing this. So should any styles regarding `h6` or `h7` change, everything that uses the mixins and classes are updated easily. You're right, the `b1` styles match the design, but if their intention was not for the text to be a body text but a heading with heading styles, then we're not building this in an easily maintainable way. I would check with Matt and Rachel on what they intended here.

paulmskim: Also I'm iffy about their heading sizes from `h5` to `h8`, I'm seeing conflicting class names and font sizes. I'm going to poke Matt a bit about this.

src/resources/postcss/views/_list.pcss
juanfra: I think the problem is that, this time we didn't have a description of the changes, so I did a pass on the updated designs and updated accordingly.

juanfra:
On hold until MattCR is back and he clarifies.

src/resources/postcss/views/_list.pcss
paulmskim: from the call we'll add `.tribe-common-h5--min-medium` and use that for the title.

src/views/v2/list/event/description.php
paulmskim: confirmed with design, this is good.

src/views/v2/list/event/title.php
paulmskim: from the call with design, we'll use `tribe-common-h6 tribe-common-h5--min-medium` classes here.

src/views/v2/list/month-separator.php
paulmskim: from the call with design, we'll use heading classes here. Let's define an alt class as needed.

juanfra: I'm adding the `--alt` modifier for now. I see that's a repeated pattern. Maybe we should think about having a generic modifier for `font-weight: normal` that can be used everywhere. `the-prefix--font-normal` or similar.

paulmskim: sounds good. I think it's OK for now, we could say the same for the same 3 styles that are applied to every heading (color, font family, and font weight), but we'll also need to balance how easily someone can follow the code. not the highest priority for now either.

src/views/v2/month/day-marker.php
```diff
@@ -14,5 +14,5 @@
  */
 ?>
 <div class="tribe-events-c-day-marker">
-	<time class="tribe-events-c-day-marker__date tribe-common-b1 tribe-common-b1--bold" datetime="2019-05-12">May 12</time>
+	<time class="tribe-events-c-day-marker__date tribe-common-h6 tribe-common-h6--alt" datetime="2019-05-12">May 12</time>
```
paulmskim: isn't this h7 on mobile and h6 on desktop?

src/resources/postcss/views/_list.pcss
```diff
@@ -101,13 +101,7 @@
...
 
 	.tribe-events-calendar-list__event-title-link {
-		border-bottom: 2px solid transparent;
-		transition: var(--transition-border-color);
-
-		&:focus,
-		&:hover {
-			border-bottom: 2px solid var(--color-text-primary);
-		}
+		@mixin anchor-default;
```
paulmskim: same as https://github.com/moderntribe/the-events-calendar/pull/2564#discussion_r295138902

src/views/v2/month/day-marker.php
juanfra: Yep, typo, thanks! good catch.

PR #2561, opened 2019-06-14
Ensure common version is declared early to make decisions on loading

borkweb:
This moves the tribe-common registration back to `__construct()` so that it happens _before_ `plugins_loaded` priority `0` as per the Plugin Dependency spec.

See: https://central.tri.be/issues/129478
Related: https://github.com/moderntribe/event-tickets/pull/1285


borkweb:
Here's a video demo of the fix: http://p.tri.be/wf8WPp

PR #2565, opened 2019-06-15
Views V2 shortcode support

bordoni:
_Ref:_ [C#127995](http://central.tri.be/issues/127995)
---

Depends #2556 https://github.com/moderntribe/events-pro/pull/1161

### DIFF
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -207,7 +207,7 @@ class View implements View_Interface {
 
 		$view->url = $url_object;
 
-		// Setup wether this view should manage url or not, based on the Rest Request Sent.
+		// Setup whether this view should manage URL or not, based on the Rest Request Sent.
 		$view->get_template()->set( 'should_manage_url', tribe_is_truthy( Arr::get( $params, 'should_manage_url', true ) ) );
 
 		return $view;
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -203,9 +203,12 @@ class View implements View_Interface {
 				)
 			);
 
-		$view =  static::make( $slug, $context );
+		$view = static::make( $slug, $context );
 
 		$view->url = $url_object;
+
+		// Setup wether this view should manage url or not, based on the Rest Request Sent.
+		$view->get_template()->set( 'should_manage_url', Arr::get( $params, 'should_manage_url', true ) );
 
 		return $view;
 	}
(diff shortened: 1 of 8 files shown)


src/Tribe/Views/V2/View.php
```diff
@@ -201,10 +203,13 @@ public static function make_for_rest( \WP_REST_Request $request ) {
 				)
 			);
 
-		$view =  static::make( $slug, $context );
+		$view = static::make( $slug, $context );
 
 		$view->url = $url_object;
 
+		// Setup wether this view should manage url or not, based on the Rest Request Sent.
```
borkweb:
```suggestion
		// Setup whether this view should manage URL or not, based on the Rest Request Sent.
```

src/resources/js/views/manager.js
```diff
@@ -72,8 +72,15 @@ tribe.events.views.manager = {};
 	obj.setup = function( index, container ) {
 		var $container = $( container );
 		var $form = $container.find( obj.selectors.form );
+		var $data = $container.find( obj.selectors.dataScript );
+		var data = {};
```
borkweb:
```suggestion
		var data  = {};
```

src/resources/js/views/manager.js
```diff
@@ -107,6 +114,30 @@ tribe.events.views.manager = {};
...
+	 *
+	 * @since TBD
+	 *
+	 * @param  {Element|jQuery} element Which element we are using as the container.
+	 *
+	 * @return {Boolean}
+	 */
+	obj.shouldManageUrl = function( $container ) {
+		var shouldManageUrl = $container.data( 'view-manage-url' );
+		var tribeIsTruthy = /^(true|1|on|yes)$/;
```
borkweb:
```suggestion
		var tribeIsTruthy   = /^(true|1|on|yes)$/;
```

src/views/v2/events-bar/form.php
```diff
@@ -12,13 +12,18 @@
...
+
+use Tribe\Events\Views\V2\Rest_Endpoint;
 ?>
 <div class="tribe-events-c-events-bar__form">
 	<form
 		class="tribe-common-c-search"
 		method="get"
 		data-js="tribe-events-view-form"
 	>
+		<?php wp_nonce_field( 'wp_rest', 'tribe-events-views[_wpnonce]' ); ?>
```
borkweb: I'd suggest the action be something other than `wp_rest` as the view won't always hit a REST endpoint.

src/views/v2/top-bar/today.php
```diff
@@ -12,9 +12,20 @@
  * @version 4.9.3
  *
  */
+
+// If we didnt have a view setup we cannot print what today's link
```
borkweb:
```suggestion
// If we didn't have a view setup we cannot print today's link
```

src/views/v2/events-bar/form.php
bordoni: Agreed, that is something we will need to address by having this be a method inside of the of another class later, or template.

PR #2566, opened 2019-06-15
Ensure the deactivation feedback dialog for Freemius is present

borkweb:
This fix resolves an issue where Freemius' hook to register the deactivation dialog was hooked to an action that has already come and gone within our plugins (`plugins_loaded`).

Additionally, the `should_load()` method had a chance of loading if the seed threshold was too high. (or not load if someone cleared `previous_ecp_versions` while having a freemius seed)

Here's the screencast of this in action: http://p.tri.be/h7U3f9

See: https://central.tri.be/issues/128341


bordoni:
Good catches.

PR #2568, opened 2019-06-16
Feature/128099 multiday event js

paulmskim:
https://central.tri.be/issues/128099


src/resources/js/views/multiday-events.js
```diff
@@ -0,0 +1,179 @@
...
+		$hiddenMultidayEvents.each( function( index, hiddenMultidayEvent ) {
+			var $hiddenMultidayEvent = $( hiddenMultidayEvent );
+			var $visibleMultidayEvents = obj.findVisibleMultidayEvents( $hiddenMultidayEvent );
+
+			$visibleMultidayEvents.each( function( index, visibleMultidayEvent ) {
+				var $visibleMultidayEvent = $( visibleMultidayEvent );
+				var $visibleMultidayEventInner = $visibleMultidayEvent.find( obj.selectors.multidayEventInner );
+				var $hiddenMultidayEventInner = $hiddenMultidayEvent.find( obj.selectors.multidayEventInner );
+
+				$hiddenMultidayEventInner.hover( obj.onHoverIn( $visibleMultidayEventInner ), obj.onHoverOut( $visibleMultidayEventInner ) );
```
bordoni: Since these are all on one element I would love to have all of those on a single `on()` call. What do you think?

src/resources/js/views/multiday-events.js
```diff
@@ -0,0 +1,179 @@
...
+	obj.ready = function() {
+		/**
+		 * @todo: this isn't working... need to figure out why and fix.
+		 */
+		tribe.events.views.manager.$containers.on( 'afterSetup.tribeEvents', obj.bindEvents );
+		/**
+		 * @todo: do below for ajax events
+		 */
+		// on 'beforeAjaxBeforeSend.tribeEvents' event, remove all listeners
+		// on 'afterAjaxSuccess.tribeEvents' or 'afterAjaxError.tribeEvents', add all listeners
```
bordoni: With the container changes to `bindEvents` we wouldn't need to do those.

src/resources/js/views/tooltip.js
```diff
@@ -145,9 +137,9 @@ tribe.events.views.tooltip = {};
 	 */
 	obj.ready = function() {
 		// @todo: make it work with variable instead of function, so it's triggered how's supposed to be
-		tribe.events.views.manager.$containers.on( 'afterSetup.tribeEvents', obj.initTooltips() );
+		tribe.events.views.manager.$containers.on( 'afterSetup.tribeEvents', obj.initTooltips );
```
bordoni:
```suggestion
		$document.on( 'afterSetup.tribeEvents', tribe.events.views.manager.selectors.container , obj.initTooltips );
```

src/resources/js/views/multiday-events.js
paulmskim: probably need to keep `afterAjaxError`, right? if we're removing before send, then reinit on success or fail? since success runs `setup`, we don't need that one, but on fail, if we don't reinit, then functionality fails.

src/resources/js/views/multiday-events.js
paulmskim: not sure how we would do that? the events call different functions.

paulmskim: i had to use 2 `on()` calls since they called different functions.

src/resources/js/views/multiday-events.js
```diff
@@ -0,0 +1,179 @@
...
+	 * @param {jQuery} $hiddenMultidayEvent jQuery object of hidden multiday event
+	 *
+	 * @return {jQuery} jQuery object of visible multiday event or false if none found
+	 */
+	obj.findVisibleMultidayEvents = function( $hiddenMultidayEvent ) {
+		var $calendar = $hiddenMultidayEvent.closest( obj.selectors.calendar );
+		var eventId = $hiddenMultidayEvent.attr( 'data-id' );
+
+		return $calendar
+			.find( obj.selectors.multidayEvent + '[data-id=' + eventId + ']' )
```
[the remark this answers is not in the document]
paulmskim: any reason why we want to do this rather than do it in one pass? doing this we're getting all multiday events, no matter whether they match the id or not, then running through them again. seems like we could avoid the second pass.

src/Tribe/Views/V2/Assets.php
```diff
@@ -66,6 +66,19 @@ public function register() {
...
 
+		tribe_asset(
+			$plugin,
+			'tribe-events-views-v2-multiday-events',
+			'views/multiday-events.js',
+			[ 'jquery', 'tribe-common', 'tribe-events-views-v2-manager' ],
+			'wp_enqueue_scripts',
+			[
+				'priority'     => 10,
+				'conditionals' => [ $this, 'should_enqueue_frontend' ],
```
juanfra: So far this should only be included in month view. Let's see if it's worth it to do it now or leave a to-do to revisit this in the future (probably when week view is implemented).

src/resources/js/views/tooltip.js
```diff
@@ -117,21 +116,21 @@ tribe.events.views.tooltip = {};
...
-			},
-			functionAfter: function( instance, helper ) {
-				obj.onFunctionAfter( instance, helper );
-			}
+	obj.initTooltips = function( index, $container, data ) {
+		$container.find( obj.selectors.tooltip ).each( function( index, tooltip ) {
+			$( tooltip ).tooltipster( {
+				interactive: true,
+				theme: [ 'tribe-common', 'tribe-events', 'tribe-events-tooltip-theme' ],
+				functionInit: obj.onFunctionInit,
```
juanfra: Did you test this and see if it is working? Or just updated the code?

bordoni: What do you mean @juanfra?

src/Tribe/Views/V2/Assets.php
bordoni: It needs to be included on all views.

src/resources/js/views/multiday-events.js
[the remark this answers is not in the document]
bordoni: Lets keep the initial way, like we talked on the Slack.

src/resources/js/views/tooltip.js
juanfra: That you requested me to change that previously, to have the same format it has here. But when I made those changes it wasn't triggering the tooltip nor the different things we implemented for the accessible tooltip. So, I was wondering if we did some dev QA on these changes or if we'll end up finding out that tooltips are not working next time we want to use them.

PR #2569, opened 2019-06-17
Fix/128033 multiday end date

paulmskim:
https://central.tri.be/issues/128033


paulmskim:
failed tests are from `restv1` suite

PR #2570, opened 2019-06-17
Include a place for the content of the Actions container in top-bar

bordoni:
_Ref:_ [C#129165](http://central.tri.be/issues/129165)
---
https://github.com/moderntribe/events-filterbar/pull/177
https://github.com/moderntribe/events-pro/pull/1163


src/views/v2/events-bar/filters.php
```diff
@@ -13,21 +13,3 @@
  *
  */
 ?>
-<div class="tribe-events-c-events-bar__filters">
```
paulmskim: should we remove this file completely? or are you keeping this to replace with filter bar when activated?

src/views/v2/top-bar/actions.php
```diff
@@ -14,16 +14,5 @@
  */
 ?>
 <div class="tribe-events-c-top-bar__actions">
```
paulmskim: are we thinking of having multiple actions? or is this to future-proof in the case we add functionality? my questions are around why we didn't move this entire template to pro.

src/views/v2/top-bar/actions/content.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
paulmskim: same question as above

paulmskim:
also @bordoni where are we placing the styles for filter bar and pro stuff? I mean pro doesn't need styles as its all from common, but if we decide to expand pro features, we'll need styles and JS. should we have a discussion around the FE asset structure for these plugins?

src/Tribe/Views/V2/View.php
```diff
@@ -732,17 +690,30 @@ protected function map_args_to_query_args( array $args = null ) {
...
+		 *                                      template.
+		 * @param View_Interface $view          The current view whose template variables are being set.
+		 */
+		$template_vars = apply_filters( 'tribe_events_views_v2_view_template_vars', $template_vars, $this );
+
 		/**
 		 * Filters the variables that will be set on the View template.
 		 *
 		 * @since 4.9.3
+		 * @since TBD Renamed the filter to be alined with other fitlers on this class.
```
borkweb:
```suggestion
		 * @since TBD Renamed the filter to be aligned with other filters on this class.
```

src/Tribe/Views/V2/Views/Day_View.php
```diff
@@ -22,6 +22,14 @@ class Day_View extends View {
...
 	protected $slug = 'day';
 
+	/**
+	 * Visibility for this view.
+	 *
+	 * @since TBD
+	 *
+	 * @var bool
+	 */
+	protected $publicly_visible = true;
```
borkweb: Should this variable be moved to the parent `View` class with a default of being set to `true`? Then we don't need to declare it in all of the child classes. It feels like a default of `true` is better than a default of `false` for folks that are trying to create views.

Perhaps your intent was to have it in all of the child classes so that folks who copy these View classes know that is a setting. Either way, we should probably still declare it in the parent so that `View::is_publicly_visible()` doesn't throw a warning if someone extends and doesn't declare this variable.

src/Tribe/Views/V2/Views/Reflector_View.php
```diff
@@ -18,6 +18,15 @@
...
 class Reflector_View extends View {
 
+	/**
+	 * Slug for this view.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected $slug = 'reflector';
```
borkweb: If you add `publicly_visible` to the parent class as `true`, you'll need to declare it as `false` here.

src/views/v2/events-bar/views/item.php
```diff
@@ -0,0 +1,44 @@
...
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ */
+use Tribe\Events\Views\V2\View;
+
+
```
borkweb: Superfluous line break

src/views/v2/top-bar/actions.php
borkweb: I imagine themers/plugin authors may wish to add additional actions.

tests/views_integration/Tribe/Events/Views/V2/ViewTest.php
```diff
@@ -110,28 +110,6 @@ public function should_print_custom_html_when_specifying_it_() {
 		$this->expectOutputString( 'Alice in Wonderland' );
 	}
 
-	/**
-	 * It should allow getting the slug currently associated to a View
-	 *
-	 * @test
-	 */
-	public function should_allow_getting_the_slug_currently_associated_to_a_view() {
```
borkweb: I'm curious why this test was removed.

tests/wpunit/Tribe/Events/Views/V2/TemplateTest.php
```diff
@@ -1,41 +0,0 @@
-<?php
-
-namespace Tribe\Events\Views\V2;
-
-use Codeception\Exception\Fail;
-
-class TemplateTest extends \Codeception\TestCase\WPTestCase {
```
borkweb: Was this test class removed on purpose?

tests/views_integration/Tribe/Events/Views/V2/ViewTest.php
bordoni: It was moved to Manage class.

src/views/v2/events-bar/filters.php
bordoni: Was moved to filter-bar.

src/Tribe/Views/V2/Views/Day_View.php
bordoni: Resolved.

bordoni:
Ready for another look.

PR #2572, opened 2019-06-17
Bail out of bootstrapping non-destructively if older ET is present

borkweb:
**Background**

The `maybe_set_common_lib_info()` method was in `plugins_loaded()` inappropriately. Once relocated _out_ of `plugins_loaded()` and back to `__construct()` where it belongs, a problem reared its head with regards to plugin validation and bailing. 

Before the Plugin Dependency work, we never checked for out-of-date versions from _both_ TEC and ET. When Plugin Dependency was completed, we added a new check to support this desired validation at the expense of moving the common bootstrap to the wrong place.

**The fix**

In a previous PR for this hotfix, `maybe_set_common_lib_info()` was relocated back to `__construct()`. This PR follows along with some more changes of note:

1. There is now a class property that holds a boolean that allows plugins_loaded() to bail if we determine the plugin should not initialize & bootstrap.
2. I've added two hooks to the `plugins_loaded` action at `-1` priority to ensure it runs _before_ we attempt to initialize and bootstrap.
    1. The first hook validates that the other core plugin–TEC if executing from within ET, or ET if executing from within TEC–is of the appropriate version. If not, it _re-points_ common _back_ to the older plugin.
    2. The second hook validates that the appropriate WP or PHP version is running. If not, it _re-points_ common _back_ to the older plugin.
3. If the above hooks determine that the plugin should not init/bootstrap, it doesn't.

See: https://central.tri.be/issues/129478 and https://central.tri.be/issues/129479


src/Tribe/Main.php
```diff
@@ -277,6 +277,9 @@ class Tribe__Events__Main {
 		/** @var Tribe__Events__Default_Values */
 		private $default_values = null;
 
+		/** @var bool Prevent autoload intialization */
```
barryhughes: Minor, but, typo: `intialization` should be `initialization`.

src/Tribe/Main.php
```diff
@@ -364,38 +369,80 @@ public function maybe_set_common_lib_info() {
...
-		 * better compatibility with the autoloader methods.
+		 * Resets the global common info back to ET's common path
 		 *
-		 * @return void
+		 * @since 4.9.3.2
 		 */
-		public function plugins_loaded() {
+		private function reset_common_lib_info_back_to_et() {
+			$et = Tribe__Tickets__Main::instance();
+			$et_common_version = file_get_contents( $et->plugin_path . 'common/src/Tribe/Main.php' );
```
barryhughes: (1/2) apologies if I'm misreading but it looks like you're loading the Tribe__Main.php _source_ itself ...

src/Tribe/Main.php
```diff
@@ -364,38 +369,80 @@ public function maybe_set_common_lib_info() {
...
+			$GLOBALS['tribe-common-info'] = [
+				'dir'     => "{$et->plugin_path}common/src/Tribe",
+				'version' => $et_common_version,
```
barryhughes: (2/2) ...and are then setting the version to be the source. Shouldn't we be parsing out the version number?

src/Tribe/Main.php
```diff
@@ -364,38 +369,80 @@ public function maybe_set_common_lib_info() {
...
+			if ( version_compare( Tribe__Tickets__Main::VERSION, $this->min_et_version, '>=' ) ) {
+				return;
+			}
+
+			$this->should_prevent_autoload_init = true;
+
+			add_action( 'admin_notices', [ $this, 'compatibility_notice' ] );
+			add_action( 'network_admin_notices', [ $this, 'compatibility_notice' ] );
+			add_filter( 'tribe_ecp_to_run_or_not_to_run', [ $this, 'disable_pro' ] );
+			add_action( 'tribe_plugins_loaded', [ $this, 'remove_exts' ], 0 );
```
barryhughes: Not directly related to _this_ PR and possibly not a big concern: but going to note/ask anyway...won't this do more than just disable extensions? For instance, won't it stop Eventbrite from integrating with Aggregator (because that also happens at `tribe_plugins_loaded:10`) and so possibly other weirdness?

src/Tribe/Main.php
borkweb: Ha! Well, at least we know that `version_compare()` identifies the entire source code as a greater version than that which is running. Great catch.

src/Tribe/Main.php
borkweb: A great question and I believe you are absolutely correct. I'll add this to my doc where I'm tracking what needs to be addressed when we revisit the Plugin Dependency work.

src/Tribe/Main.php
```diff
@@ -364,38 +376,86 @@ public function maybe_set_common_lib_info() {
...
+		 */
+		public function maybe_bail_if_invalid_wp_or_php() {
+			if ( self::supportedVersion( 'wordpress' ) && self::supportedVersion( 'php' ) ) {
+				return;
+			}
+
+			add_action( 'admin_notices', array( $this, 'notSupportedError' ) );
+
+			// if we get in here, we need to reset the global common to ET's version so that we don't cause a fatal
+			$this->reset_common_lib_info_back_to_et();
```
barryhughes: I don't think I can comment on it because it's not something changed in _this_ PR, but I believe this method (`reset_common_lib_info_back_to_et()`, I mean) needs further attention; similar to a previous comment it looks to be populating `$GLOBALS['tribe-common-info']['version']` with PHP source rather than the parsed out version number.

borkweb: This should now be resolved in https://github.com/moderntribe/the-events-calendar/pull/2572/files#diff-e5ad0809a1399fb28c50f3775e36ba7eR394

barryhughes: Ah yes, ignore me!

PR #2574, opened 2019-06-18
VR > Add new icons

juanfra:
🎫 https://central.tri.be/issues/129296

Add new icons that will be used on the new events bar.


juanfra:
We're moving icons to their own repo. Closing this one.

PR #2575, opened 2019-06-18
Fix/129298 tab select update

paulmskim:


src/views/v2/events-bar/views/item.php
```diff
@@ -23,24 +23,22 @@
...
-		value="<?php echo esc_attr( $view_slug ); ?>"
-		<?php checked( $is_current_view ); ?>
-	/>
-	<label
-		class="tribe-common-form-control-tabs__label"
-		id="<?php echo sanitize_html_class( "tribe-views-{$view_slug}-label" ); ?>"
-		for="<?php echo sanitize_html_class( "tribe-views-{$view_slug}" ); ?>"
-		role="option"
-		aria-selected="true"
+<li class="tribe-events-c-view-selector__list-item">
```
juanfra: I think it'd also be cool to know if the item is the one active. Thinking about customizations or future amendments.

src/resources/images/icons/list.svg
```diff
@@ -0,0 +1 @@
+<svg width="21" height="18" xmlns="http://www.w3.org/2000/svg"><g fill="#141827" fill-rule="nonzero"><path d="M19.883 3.272c.342 0 .622-.21.622-.467v-.464c0-.257-.28-.467-.622-.467H7.127c-.342 0-.622.21-.622.467v.464c0 .257.28.467.622.467h12.756zM1.838 4.207a1.75 1.75 0 1 0 0-3.5 1.75 1.75 0 0 0 0 3.5zM19.883 9.98c.342 0 .622-.21.622-.466v-.465c0-.257-.28-.467-.622-.467H7.127c-.342 0-.622.21-.622.467v.465c0 .256.28.466.622.466h12.756zM1.838 10.916a1.75 1.75 0 1 0 0-3.5 1.75 1.75 0 0 0 0 3.5z"/><g><path d="M19.883 16.689c.342 0 .622-.21.622-.467v-.465c0-.256-.28-.466-.622-.466H7.127c-.342 0-.622.21-.622.466v.465c0 .257.28.467.622.467h12.756zM1.838 17.624a1.75 1.75 0 1 0 0-3.5 1.75 1.75 0 0 0 0 3.5z"/></g></g></svg>
```
juanfra: It's a pity we didn't coordinated on the iconography update.

src/resources/js/views/accordion.js
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @return {void}
+	 */
+	obj.toggleAccordion = function( event ) {
+		var $container = event.data.container;
+		var $header = $( event.data.target );
+		var contentId = $header.attr( 'aria-controls' );
+		var $content = $container.find( '#' + contentId );
+
+		if ( $header.attr( 'aria-expanded' ) === 'true' ) {
```
juanfra: Let's use yoda conditionals.

src/resources/js/views/accordion.js
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @param {Event} event event object for 'afterSetup.tribeEvents' event
+	 * @param {integer} index jQuery.each index param from 'afterSetup.tribeEvents' event
+	 * @param {jQuery} $container jQuery object of view container
+	 * @param {object} data data object passed from 'afterSetup.tribeEvents' event
+	 *
+	 * @return {void}
+	 */
+	obj.bindEvents = function( event, index, $container, data ) {
+		$container.find( obj.selectors.accordionTrigger ).each( function( index, header ) {
```
juanfra: Maybe split into multiple lines.

src/resources/postcss/components/_view-selector.pcss
```diff
@@ -0,0 +1,273 @@
...
+	 * </div>
+	 *
+	 * ----------------------------------------------------------------------------- */
+
+	.tribe-events-c-view-selector {
+		position: relative;
+		width: 29px;
+
+		@media (--viewport-medium) {
+			width: 114px;
```
juanfra: Are we setting the selector to a specific view? Have we tried with texts that are not only the ones included in the designs? Thinking about translations here and how this can end up looking bad if those words change.

src/resources/postcss/components/_view-selector.pcss
```diff
@@ -0,0 +1,273 @@
...
+
+			flex: auto;
+		}
+	}
+
+	.tribe-events-c-view-selector__content {
+		display: none;
+		background-color: var(--color-background);
+		border: 1px solid var(--color-border-secondary);
+		border-radius: 4px;
```
juanfra: We're repeating the same border radius on buttons on some other places. Maybe it's a good time to define this as a variable and use it on other places, as it looks like it'll be part of the design system.

src/resources/postcss/components/_view-selector.pcss
```diff
@@ -0,0 +1,273 @@
...
+		box-shadow: var(--box-shadow-tooltip);
+		padding: calc(var(--spacer-4) / 2) var(--spacer-1);
+		position: absolute;
+		right: 0;
+		top: 100%;
+		transform: translateY(var(--spacer-2));
+		width: 168px;
```
juanfra: The same concern here with the width and different content from what we have in English.

src/resources/postcss/components/_view-selector.pcss
```diff
@@ -0,0 +1,273 @@
...
+		}
+	}
+
+	/* .tribe-events-c-view-selector__list {} */
+
+	/* .tribe-events-c-view-selector__list-item {} */
+
+	.tribe-events-c-view-selector__list-item-link {
+		align-items: center;
+		border-radius: 4px;
```
juanfra: The same thing about having the `border-raduis` value coming from a var.

src/resources/postcss/components/_view-selector.pcss
```diff
@@ -0,0 +1,273 @@
...
+		margin-right: 12px;
+		width: 18px;
+
+		@media (--viewport-medium) {
+			@mixin hidden;
+		}
+	}
+
+	.tribe-events-c-view-selector__list-item-text {
+		@mixin desktop-body-2;
```
juanfra: I feel like in other places we're using the common class, instead of the mixin on the newly defined class. Why do we have it like that here, instead of including the b2 class from common in the markup?

src/views/v2/events-bar/views.php
```diff
@@ -13,22 +13,29 @@
  *
  */
 use Tribe\Events\Views\V2\Manager;
+use Tribe\Events\Views\V2\View;
 
 $public_views = tribe( Manager::class )->get_publicly_visible_views();
+$view_instance = View::make( $this->get( 'view_class_name' ) );
```
juanfra: I feel like at some point if these vars can come from the backend and we document them on the docblock, the template will be way cleaner. Maybe we should discuss this with the BE devs.

src/views/v2/events-bar/views.php
```diff
@@ -13,22 +13,29 @@
...
-			aria-labelledby="tribe-views-button"
-			aria-expanded="true"
+			class="tribe-events-c-view-selector__button"
+			aria-controls="tribe-events-view-selector-content"
+			aria-expanded="false"
+			aria-selected="false"
+			data-js="tribe-events-accordion-trigger"
 		>
-			<?php esc_html_e( 'Views', 'the-events-calendar' ); ?>
+			<span class="tribe-events-c-view-selector__button-icon tribe-common-svgicon tribe-common-svgicon--<?php echo esc_attr( $view_slug ); ?>"></span>
```
bordoni:
```suggestion
			<span class="tribe-events-c-view-selector__button-icon tribe-common-svgicon <?php echo sanitize_html_class( "tribe-common-svgicon--{$view_slug}" ); ?>"></span>
```

src/views/v2/events-bar/views.php
bordoni:
```suggestion
```

src/views/v2/events-bar/views.php
```diff
@@ -13,22 +13,29 @@
  *
  */
 use Tribe\Events\Views\V2\Manager;
+use Tribe\Events\Views\V2\View;
 
 $public_views = tribe( Manager::class )->get_publicly_visible_views();
+$view_instance = View::make( $this->get( 'view_class_name' ) );
+$view_slug = $view_instance->get_slug();
```
bordoni:
```suggestion
$view_slug = $this->get( 'view' )->get_slug();
```

src/views/v2/events-bar/views.php
```diff
@@ -13,22 +13,29 @@
...
+$view_label = $view_instance->get_label();
```
bordoni:
```suggestion
$view_label = $this->get( 'view' )->get_label();
```

src/resources/postcss/components/_view-selector.pcss
paulmskim: The `tribe-common-b2` class has a mobile font size and a desktop font size. The designs here call for the same font size on both mobile and desktop.

src/views/v2/events-bar/views/item.php
paulmskim: The link below this has an active class. Are you talking about the list item getting an active class?

juanfra: Yes. As it's the item the one that will be active.

paulmskim: Got it. My preference would be to not, as the `<li>` elements here are basically semantic wrappers. They don't do much of anything at all, the main actors here are the links.

paulmskim: If customers want to add an active wrapper class for styling, it's not difficult to extend what's already here.

src/resources/postcss/components/_view-selector.pcss
paulmskim: Good question. No, this wouldn't allow for longer words. The designs have some odd spacing that I'd like to clear up with Matt/Rachel.

src/resources/postcss/components/_view-selector.pcss
bordoni: We can come back to this, would you mind noting this in a ticket @juanfra  we need to review some of those design points.

src/resources/postcss/components/_view-selector.pcss
bordoni: We can come back to this, would you mind noting this in a ticket @juanfra  we need to review some of those design points.

paulmskim: I have a ticket for tabs/select work: https://central.tri.be/issues/129298 I can have a look later, this shouldn't hold up the merge for QA.

src/views/v2/events-bar/views/item.php
juanfra: Yes, my idea was basically that if they want to do it, it would be a small effort for us now and they will have to add some CSS instead of having to touch the template.

paulmskim: I understand your point, I don't see the benefit to us to do so other than an active modifier class that doesn't do anything style-wise or functionality-wise. I believe we should provide customers the tools to do what they want, but I don't believe we should try to provide them everything, whether they will use them or not.

paulmskim: moved active class to list item and removed from anchor.

PR #2576, opened 2019-06-19
Feature/128097 mobile month day updated

paulmskim:
https://central.tri.be/issues/128097

Also reorganized file structure of templates as that was not done in the initial pass (https://central.tri.be/issues/129312). @juanfra would love your thoughts here.

On hold until https://github.com/moderntribe/the-events-calendar/pull/2570 is merged in, do not CR as the diff is too big.


src/resources/js/views/accordion.js
```diff
@@ -55,7 +55,7 @@ tribe.events.views.accordion = {};
 		$container.find( obj.selectors.accordionTrigger ).each( function( index, header ) {
 			var $header = $( header );
 			var contentId = $header.attr( 'aria-controls' );
```
bordoni: Could it be just a selector? Not ID? Making it available to use classes too.

src/resources/js/views/month-mobile-events.js
```diff
@@ -0,0 +1,193 @@
...
+	 * @since TBD
+	 *
+	 * @param {jQuery} $container jQuery object of view container
+	 *
+	 * @return {void}
+	 */
+	obj.closeAllEvents = function( $container ) {
+		$container.find( obj.selectors.calendarDay ).each( function( index, header ) {
+			var $header = $( header );
+			var contentId = $header.attr( 'aria-controls' );
```
bordoni: Same as the above, would love to see this being selector not a ID only kinda thing.

src/resources/js/views/month-mobile-events.js
```diff
@@ -0,0 +1,193 @@
...
+	 */
+	obj.openMobileEvents = function( $header, $content ) {
+		// only perform accordion actions if $header has aria-controls attribute.
+		var contentId = $header.attr( 'aria-controls' );
+		if ( contentId ) {
+			tribe.events.views.accordion.openAccordion( $header, $content );
+		}
+
+		$header.addClass( obj.selectors.calendarDaySelected.className() );
+		$content.parent().css( 'display', 'block' );
```
bordoni: Should we control this with a class? Makes it more flexible.
Also, should we target the parent always?

src/resources/js/views/month-mobile-events.js
```diff
@@ -0,0 +1,193 @@
...
+	 * @since TBD
+	 *
+	 * @param {Event} event event object of click event
+	 *
+	 * @return {void}
+	 */
+	obj.toggleMobileEvents = function( event ) {
+		var $container = event.data.container;
+		var $header = $( event.data.target );
+		var contentId = $header.attr( 'aria-controls' );
```
bordoni: Same as the above, would love to see this being selector, not an ID only kinda thing.

src/resources/js/views/accordion.js
paulmskim: An accordion header button controls the accordion contents. They are linked by an id (`aria-controls` attribute is the `id` of the contents that the header controls). So there's no other good way of getting the contents that are linked to the header.

src/resources/js/views/month-mobile-events.js
paulmskim: When you ask if we should always target the parent, do you mean in `accordion.js`? I'm targeting the parent here because it's specific to mobile events js, the accordion doesn't always have this structure.

Can do to toggle a class here.

src/resources/js/views/accordion.js
bordoni: Could we link them via a selector instead of an ID?

paulmskim: not likely. if a page has multiple accordions with the same markup, we need to know which header controls which content. the accordion.js is meant to be extendable for any element.

if you're thinking of something else, would love to chat it out to see if we can balance both.

bordoni: Sounds good.

src/resources/js/views/multiday-events.js
```diff
@@ -9,7 +9,7 @@ tribe.events = tribe.events || {};
 tribe.events.views = tribe.events.views || {};
 
 /**
- * Configures Views Object in the Global Tribe variable
+ * Configures Multiday Events Object in the Global Tribe variable
```
juanfra: Are we planning to handle multi-day events for the week view here as well? If not it'd be ideal to specify the file name (obj, etc) for the month -> e.g.: `views/month-multiday-events.js`

src/views/v2/month/calendar-body.php
```diff
@@ -0,0 +1,40 @@
...
+ * View: Month View - Calendar Body
+ *
+ * Override this template in your own theme by creating a file at:
+ * [your-theme]/tribe/events/views/v2/month/calendar-body.php
+ *
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version 4.9.3
```
juanfra: This shouldn't stop us from merging, but it'd be good if we all start setting the TBD version every time we touch a template. Then it will be replaced during the release tasks

src/views/v2/month/calendar-body/day/multiday-events/multiday-event.php
```diff
@@ -17,48 +17,48 @@
 $event_id = $event->ID;
 $day_number = $this->get( 'day' );
 
-$classes = [ 'tribe-events-calendar-month__event-multiday' ];
+$classes = [ 'tribe-events-calendar-month__multiday-event' ];
 
 // Check if it's featured.
 if ( $is_featured = isset( $event->featured ) && $event->featured ) { // @todo: later use tribe( 'tec.featured_events' )->is_featured( $event_id ) or similar
-	$classes[] = 'tribe-events-calendar-month__event-multiday--featured';
+	$classes[] = 'tribe-events-calendar-month__multiday-event--featured';
```
juanfra: Maybe we can leave a `@todo` for BE devs to be sure that we later move the classes configuration to a template tag, something like what `post_class` is for WP.

src/resources/js/views/multiday-events.js
paulmskim: good call, will make that change. if we find that week view can basically reuse most/all of the js, we can change this back then.

bordoni:
Once we fix that test that is broken we are good. We have the same output but it seems to be in a weird order. So we can pass a `sort` method and we would be ok.

PR #2577, opened 2019-06-19
Fix TImezone definition on .ICS file

mitogh:
Ticket: https://central.tri.be/issues/89999


src/Tribe/iCal.php
```diff
@@ -183,6 +198,70 @@ public function do_ical_template() {
...
+	 * @param int|null $post If you want the ical file for a single event
+	 * @param boolean  $echo Whether the content should be echoed or returned
+	 *
+	 * @return string
+	 */
+	public function generate_ical_feed( $post = null, $echo = true ) {
+		$this->post = $post;
+		$this->events = $this->get_event_posts();
+
+		$this->set_headers();
```
bordoni: Should we send headers when we are not echoing?

src/Tribe/iCal.php
```diff
@@ -237,77 +316,282 @@ private function get_month_view_events() {
...
+	/**
+	 * Get the full content of the *.ics file.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	protected function get_content() {
+		return sprintf(
+			"%s%s%s%s",
```
bordoni: If we are not doing anything specific with the `sprintf` why not just implode an array?

src/Tribe/iCal.php
```diff
@@ -237,77 +316,282 @@ private function get_month_view_events() {
...
+		$content .= 'PRODID:-//' . $blog_name . ' - ECPv' . Tribe__Events__Main::VERSION . "//NONSGML v1.0//EN\r\n";
+		$content .= "CALSCALE:GREGORIAN\r\n";
+		$content .= "METHOD:PUBLISH\r\n";
+
+		/**
+		 * Allows for customizing the value of the generated iCal file's "X-WR-CALNAME:" property.
+		 *
+		 * @param string $blog_name The value to use for "X-WR-CALNAME"; defaults to value of get_bloginfo( 'name' ).
+		 */
+		$x_wr_calname = apply_filters( 'tribe_ical_feed_calname', $blog_name );
```
bordoni: Docblock with a since

src/Tribe/iCal.php
```diff
@@ -237,77 +316,282 @@ private function get_month_view_events() {
...
+
+		if ( ! empty( $x_wr_calname ) ) {
+			$content .= 'X-WR-CALNAME:' . $x_wr_calname . "\r\n";
+		}
+
+		$content .= 'X-ORIGINAL-URL:' . $blog_home . "\r\n";
+		$content .= "X-WR-CALDESC:" . sprintf( esc_html_x( 'Events for %s', 'iCal feed description', 'the-events-calendar' ), $blog_name ) . "\r\n";
+
+		return apply_filters( 'tribe_ical_properties', $content );;
```
bordoni: Docblock with a since required.

src/Tribe/iCal.php
```diff
@@ -237,77 +316,282 @@ private function get_month_view_events() {
...
+			if ( count( $transitions ) === 1 ) {
+				$transitions[] = array_values( $transitions )[ 0 ];
+			}
+
+			$id = $timezone->getName();
+			$item[] = 'TZID:"' .  $id . '"';
+
+			$last_transition = null;
+			foreach ( $transitions as $i => $transition ) {
+				if ( $i === 0 ) {
```
bordoni: yoda.

src/Tribe/iCal.php
```diff
@@ -322,51 +606,47 @@ public function generate_ical_feed( $post = null, $echo = true ) {
...
 			$item[] = 'DTSTAMP:' . date( $full_format, time() );
 			$item[] = 'CREATED:' . $tzoned->created;
 			$item[] = 'LAST-MODIFIED:' . $tzoned->modified;
 			$item[] = 'UID:' . $event_post->ID . '-' . $time->start . '-' . $time->end . '@' . parse_url( home_url( '/' ), PHP_URL_HOST );
-			$item[] = 'SUMMARY:' . str_replace( array( ',', "\n", "\r" ), array( '\,', '\n', '' ), html_entity_decode( strip_tags( $event_post->post_title ), ENT_QUOTES ) );
-			$item[] = 'DESCRIPTION:' . str_replace( array( ',', "\n", "\r" ), array( '\,', '\n', '' ), html_entity_decode( strip_tags( str_replace( '</p>', '</p> ', apply_filters( 'the_content', tribe( 'editor.utils' )->exclude_tribe_blocks( $event_post->post_content ) ) ) ), ENT_QUOTES ) );
+			$item[] = 'SUMMARY:' . str_replace(
+					[ ',', "\n", "\r" ],
+					[ '\,', '\n', '' ],
+					html_entity_decode( strip_tags( $event_post->post_title ), ENT_QUOTES )
```
bordoni: Move this into another method.

src/Tribe/iCal.php
```diff
@@ -322,51 +606,47 @@ public function generate_ical_feed( $post = null, $echo = true ) {
...
+				);
+			$item[] = 'DESCRIPTION:' . str_replace(
+					[ ',', "\n", "\r" ],
+					[ '\,', '\n', '' ],
+					html_entity_decode( strip_tags( str_replace( '</p>', '</p> ', apply_filters( 'the_content', tribe( 'editor.utils' )->exclude_tribe_blocks( $event_post->post_content ) ) ) ), ENT_QUOTES )
```
bordoni: Move this into another method.

tests/wpunit/Tribe/Events/iCalTest.php
```diff
@@ -238,4 +242,55 @@ public function test_get_ical_link_single() {
...
+				],
+				"BEGIN:VTIMEZONE\r\nTZID:\"UTC\"\r\nBEGIN:STANDARD\r\nTZOFFSETFROM:+0000\r\nTZOFFSETTO:+0000\r\nTZNAME:UTC\r\nDTSTART:20190101T000000\r\nEND:STANDARD\r\nEND:VTIMEZONE\r\n"
+			],
+			[
+				[
+					'when' => '2019-01-01',
+					'duration' => DAY_IN_SECONDS,
+					'meta_input' => [ '_EventTimezone' => 'UTC+6' ]
+				],
+				"EGIN:VTIMEZONE\r\nTZID:\"UTC\"\r\nBEGIN:STANDARD\r\nTZOFFSETFROM:+0000\r\nTZOFFSETTO:+0000\r\nTZNAME:UTC\r\nDTSTART:20190101T000000\r\nEND:STANDARD\r\nEND:VTIMEZONE\r\n",
```
bordoni: Missing `B`

src/Tribe/iCal.php
mitogh: This one existed before this release so the `@since` will not be accurate.

src/Tribe/iCal.php
mitogh: I think we are missing the `dockblock` however same as above this one existed already before this release so will not be accurate to add it on this release.

src/Tribe/iCal.php
mitogh: Great question I think we shouldn't, this was on the original implementation but let me add a test for this one as it seems we shouldn't deliver those headers when we only need the content.

src/Tribe/iCal.php
mitogh: Not really a specific reason so I think an implode will work great here 👍

src/Tribe/iCal.php
```diff
@@ -183,6 +204,69 @@ public function do_ical_template() {
...
+
+	/**
+	 * Get an array with all the Events to be used to process the *.ics file
+	 *
+	 * @since TBD
+	 *
+	 * @return array|null
+	 */
+	protected function get_event_posts() {
+		if ( $this->post ) {
```
juanfra: I'm curious why we are using the  elseif instead of just using the ifs, as each one has a return.

src/Tribe/iCal.php
```diff
@@ -237,77 +321,287 @@ private function get_month_view_events() {
...
+		$filename = sprintf( '%s-%s.ics', $site, $hash );
 
-			if ( ! $wp_query = tribe_get_global_query_object() ) {
-				return;
+		/**
+		 * Modifies the filename provided in the Content-Disposition header for iCal feeds.
+		 *
+		 * @var string       $filename
+		 * @var WP_Post|null $post
+		 */
```
juanfra: Let's add a `@since TBD` here.

src/Tribe/iCal.php
```diff
@@ -237,77 +321,287 @@ private function get_month_view_events() {
...
+		}
+
+		$content .= 'X-ORIGINAL-URL:' . $blog_home . "\r\n";
+		$content .= "X-WR-CALDESC:" . sprintf( esc_html_x( 'Events for %s', 'iCal feed description', 'the-events-calendar' ), $blog_name ) . "\r\n";
+
+		/**
+		 * Allows for customization of the various properties at the top of the generated iCal file.
+		 *
+		 * @param string $content Existing properties atop the file; starts at "BEGIN:VCALENDAR", ends at "X-WR-CALDESC".
+		 */
```
juanfra: Let's add a `@since TBD` here.

src/Tribe/iCal.php
```diff
@@ -237,77 +321,287 @@ private function get_month_view_events() {
...
+				$item[] = 'END:' . $type;
+				$last_transition = $transition;
 			}
+		}
+
+		/**
+		 * Allow for customization of an individual "VTIMEZONE" item to be rendered inside an iCal export file.
+		 *
+		 * @param array $item The various iCal file format components of this specific event item.
+		 * @param array $timezones The various iCal timzone components of this specific event item.
```
juanfra: Let's add a `@since TBD` here.

src/Tribe/iCal.php
```diff
@@ -237,77 +321,287 @@ private function get_month_view_events() {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param $events array An array with all the events to parse the timezones.
+	 *
+	 * @return array
+	 */
+	protected function parse_timezones( $events ) {
+		$data = [];
+		foreach ( $events as $event ){
```
juanfra: spacing ``
```suggestion
		foreach ( $events as $event ) {
```

src/Tribe/iCal.php
mitogh: No real reason, but primarily as this one was the original code this change just moved this piece into a function. But I think is a great point which will make this function so much easier to understand.

src/Tribe/iCal.php
mitogh: Oh this one is an existing filter that was moved around due the changes on the code, do you think is still relevant to add a `@since` even was not added on this PR?

src/Tribe/iCal.php
mitogh: Oh this one is an existing filter that was moved around due the changes on the code, do you think is still relevant to add a `@since` even was not added on this PR?

juanfra: Perfect!

src/Tribe/iCal.php
juanfra: Perfect!

mitogh:
@bordoni would you mind take another look.

Thanks.

PR #2581, opened 2019-06-20
Feature/128095 mobile month js

paulmskim:
https://central.tri.be/issues/128095

On hold until https://github.com/moderntribe/the-events-calendar/pull/2576 is merged.

Screencast: https://www.loom.com/share/292e0779e5e646839cf7b93f02a782e0


paulmskim:
@bordoni @juanfra this is good for CR.

bordoni:
As I noted to you on that PR, we need to make sure all of the Modules that interact with Views need to be by container, so a global state here will create problems. 

I wonder if instead of managing the state in a variable, it would be "simpler" to keep the state tied to the DOM, so when a new view is pulled it it resets.

paulmskim:
That's an interesting solution, will think about how else we could tackle it, but there is always that option. Thanks for the input!

src/resources/js/views/month-grid.js
```diff
@@ -0,0 +1,353 @@
...
+	/**
+	 * Handle keydown event to move focused grid cell
+	 *
+	 * @since TBD
+	 *
+	 * @param {Event} event event object
+	 *
+	 * @return {void}
+	 */
+	obj.handleKeydown = function( event ) {
```
ryanurban: Nice

PR #2583, opened 2019-06-21
Feature/129310 event date tag

paulmskim:
https://central.tri.be/issues/129310


src/resources/postcss/views/list/_event.pcss
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	.tribe-events-calendar-list__event-date-tag-datetime {
+		display: flex;
+		flex-direction: column;
+		text-align: center;
+
+		@media (--viewport-full) {
+			margin-left: 28%;
```
juanfra: Is this calculated in any form? Does it look good when sliding the viewport? (I'm starting to wander about the views on sites using the page template)

src/views/v2/list/event/date-tag.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @version TBD
+ *
+ */
+
+$event = $this->get( 'event' );
+?>
+<div class="tribe-events-calendar-list__event-date-tag tribe-common-g-col">
+	<time class="tribe-events-calendar-list__event-date-tag-datetime" datetime="1970-01-01T00:00:00+00:00">
+		<span class="tribe-events-calendar-list__event-date-tag-weekday">Wed</span>
```
juanfra: Maybe if we have some time, given that we already have data for the events, we could use the real date here.

src/resources/postcss/views/list/_event.pcss
paulmskim: I based this roughly off the designs, and picked the `--viewport-full` breakpoint as it didn't break the text into two lines. Using page template, it may not work so well. But I think that's a question for when we do a theme pass.

src/views/v2/list/event/date-tag.php
paulmskim: Yeah, I don't know the template tags for this data. If you or @bordoni want to go in and fix those up, feel free. I just didn't want to spend a bunch of time looking.

src/resources/postcss/views/list/_event.pcss
juanfra: Okies. Yes, my question had more to do with using the page template and not the events template.

src/views/v2/list/event/date-tag.php
juanfra: Ok, right now we don't have template tags for those specifically. But, yeah, we can take a look at these later.

src/resources/postcss/views/list/_event.pcss
```diff
@@ -56,6 +56,7 @@
 
 		color: var(--color-text-secondary);
 		margin-bottom: calc(var(--spacer-0) * -1);
+		text-transform: uppercase;
```
juanfra: Do you feel it would be a good idea to add a mod class for uppercase and use it site wide?

juanfra:
Seems like you have to update some snapshots. Other than that, it's ready to go for me!

src/resources/postcss/views/list/_event.pcss
paulmskim: I'm not seeing a consistent enough pattern to say that it warrants a modifier class for uppercase. If you can point out a clear pattern, I wouldn't be opposed.

juanfra: I don't see a pattern either (I saw one for the month, day separators and the font weight), but I was just wondering if from a "framework" perspective it would be nice to have those.

ryanurban: Worth asking the q, but sounds like based on the lack of a pattern, meh :)

PR #2588, opened 2019-06-21
VR > Events Bar > Form Inputs JS

juanfra:
🎫 https://central.tri.be/issues/129300


juanfra:
This one **can be code reviewed**. Setting on focus until we define where we are setting the focus class names on https://github.com/moderntribe/tribe-common/pull/1061

Thanks 🎉

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,116 @@
...
+	 *
+	 * @return {void}
+	 */
+	obj.bindEvents = function( event, index, $container, data ) {
+
+		// Bind event for the keyword input.
+		$container
+			.find( obj.selectors.inputKeyword )
+			.each( function( index, input ) {
+				console.log( $( input ).val() );
```
paulmskim: probably don't want this right?

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,116 @@
...
+	/**
+	 * Toggles accordion on header click
+	 *
+	 * @since TBD
+	 *
+	 * @param {Event} event event object of click event
+	 *
+	 * @return {void}
+	 */
+	obj.inputSetFocusMaybe = function( event ) {
```
paulmskim: Let's call this `setInputFocusClass` or something similar. In PHP you may use `maybe` in function names, but in JS we don't.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,116 @@
...
+		var $input = $( event.target );
+
+		// Set the focus class if it has content.
+		$input.toggleClass( event.data.inputClassFocus, '' !== $input.val().trim() );
+
```
paulmskim: remove extra space

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,116 @@
...
+	 *
+	 * @param {Event} event event object of click event
+	 *
+	 * @return {void}
+	 */
+	obj.inputSetFocusMaybe = function( event ) {
+		var $input = $( event.target );
+
+		// Set the focus class if it has content.
+		$input.toggleClass( event.data.inputClassFocus, '' !== $input.val().trim() );
```
paulmskim: either use `event.data.target` here (`event.data.target` is a jQuery object already) or don't pass `target` in the `on()` function

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,116 @@
...
+	 */
+	obj.selectors = {
+		inputKeyword: '.tribe-common-c-search__input--keyword',
+		inputKeywordFocus: '.tribe-common-c-search__input--keyword-focus',
+		inputLocation: '.tribe-common-c-search__input--location',
+		inputLocationFocus: '.tribe-common-c-search__input--location-focus',
+	};
+
+	/**
+	 * Toggles accordion on header click
```
paulmskim: Description should match what function is doing.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,116 @@
...
+	 */
+	obj.bindEvents = function( event, index, $container, data ) {
+
+		// Bind event for the keyword input.
+		$container
+			.find( obj.selectors.inputKeyword )
+			.each( function( index, input ) {
+				console.log( $( input ).val() );
+				$( input ).toggleClass( obj.selectors.inputKeywordFocus.className(), '' !== $( input ).val().trim() );
+				$( input ).on( 'change', { target: $( this ), inputClassFocus: obj.selectors.inputKeywordFocus.className() }, obj.inputSetFocusMaybe );
```
paulmskim: love the use of passing the class name as data, nice work :+1:

src/resources/js/views/events-bar.js
juanfra: haha you're right 🤦‍♂️

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,114 @@
...
+	 * @param {integer} index jQuery.each index param from 'afterSetup.tribeEvents' event
+	 * @param {jQuery} $container jQuery object of view container
+	 * @param {object} data data object passed from 'afterSetup.tribeEvents' event
+	 *
+	 * @return {void}
+	 */
+	obj.bindEvents = function( event, index, $container, data ) {
+
+		// Bind event for the keyword input.
+		$container
```
bordoni: Each one of these binds should live into a separate method.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,114 @@
...
+			.find( obj.selectors.inputKeyword )
+			.each( function( index, input ) {
+				$( input ).toggleClass( obj.selectors.inputKeywordFocus.className(), '' !== $( input ).val().trim() );
+				$( input ).on( 'change', { target: $( this ), inputClassFocus: obj.selectors.inputKeywordFocus.className() }, obj.setInputFocusClass );
+			} );
+
+		// Bind event for the location input.
+		$container
+			.find( obj.selectors.inputLocation )
+			.each( function( index, input ) {
```
bordoni: Each one of these binds should live into a separate method.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,158 @@
...
+
+	/**
+	 * Selectors used for configuration and setup
+	 *
+	 * @since TBD
+	 *
+	 * @type {PlainObject}
+	 */
+	obj.selectors = {
+		input: '.tribe-common-form-control-text__input',
```
paulmskim: we should try and use `[data-js]` as much as possible, except where we're actually adding classes.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,158 @@
...
+	 * @param {jQuery} $container jQuery object of container.
+	 *
+	 * @return {void}
+	 */
+	obj.bindEventsInputKeyword = function( $container ) {
+
+		// Bind event for the keyword input.
+		$container
+			.find( obj.selectors.inputKeywordWrapper )
+			.find( obj.selectors.input )
```
paulmskim: Instead of finding the input wrapper, then finding the input, then doing the `closest()` to get the wrapper again, why don't we `.find( obj.selectors.inputKeywordWrapper )`, toggle class, then find the input and add the `on( 'change' )`? Saves us from running multiple dom queries.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,158 @@
...
+	 * @param {jQuery} $container jQuery object of container.
+	 *
+	 * @return {void}
+	 */
+	obj.bindEventsInputLocation = function( $container ) {
+
+		// Bind event for the location input.
+		$container
+			.find( obj.selectors.inputLocationdWrapper )
+			.find( obj.selectors.input )
```
paulmskim: same here as the note above.

src/resources/js/views/events-bar.js
paulmskim: @bordoni was this resolved? did you mean that the `each()` should be calling something like `obj.bindInputChangeListener` rather than an anonymous `function()`?

src/resources/js/views/events-bar.js
juanfra: Because, as the modifier class is in the wrapper now, we need to go and check the input value. Even if we do it that way, we would have to go through the DOM in a different way, but I'm not sure if we could "economize" anything here.

The class changes are tied to what's in the input.

src/resources/js/views/events-bar.js
juanfra: Where would you suggest it?

src/resources/js/views/events-bar.js
ryanurban: Agreed, we should able to more efficiently target for what you need to do here. Whether that's being more extensive in your use of `data-js` and/or adding additional hooks with which to reduce the amount of dom queries needed. Optimizations on how you target and select elements add up : )

src/resources/js/views/events-bar.js
paulmskim: on the element, we can add `data-js` attributes and selectors used in JS can be `[data-js="the-selector"]`. if you're not sure, let's hop on a call and walk through it.

juanfra: Yep. I was asking in this particular case if you were suggesting to do that in a particular thing.

juanfra:
@paulmskim ready for new 👀 - I've used the data-js naming from what I saw you used on `feature/129425-search-filter-tab-js` for consistency.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,161 @@
...
+	/**
+	 * Bind events for the keyword input of the events bar
+	 *
+	 * @since TBD
+	 *
+	 * @param {jQuery} $container jQuery object of container.
+	 *
+	 * @return {void}
+	 */
+	obj.bindEventsInputKeyword = function( $container ) {
```
bordoni: Instead of creating a method for each Input that we might have there we should make a generic `[data-js]` selector and just apply it to any element that has it. We need to make sure we are lowering the amount of JS logic that needs to be made custom for every piece of HTML we do.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,161 @@
...
+	/**
+	 * Bind events for the location input of the events bar
+	 *
+	 * @since TBD
+	 *
+	 * @param {jQuery} $container jQuery object of container.
+	 *
+	 * @return {void}
+	 */
+	obj.bindEventsInputLocation = function( $container ) {
```
bordoni: Instead of creating a method for each Input that we might have there we should make a generic `[data-js]` selector and just apply it to any element that has it. We need to make sure we are lowering the amount of JS logic that needs to be made custom for every piece of HTML we do.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,161 @@
...
+	 * @param {object} data data object passed from 'afterSetup.tribeEvents' event
+	 *
+	 * @return {void}
+	 */
+	obj.bindEvents = function( event, index, $container, data ) {
+		// Bind event for the keyword input.
+		obj.bindEventsInputKeyword( $container );
+
+		// Bind event for the location input.
+		obj.bindEventsInputLocation( $container );
```
bordoni: The two comments above are especially true because location input is not on by default so this logic would need to move to Pro. Let's make a generic one.

src/resources/js/views/events-bar.js
paulmskim: Definitely, let's make this generic. We can do so even with the `toggleClass()` that we're doing.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,161 @@
...
+	 *
+	 * @return {void}
+	 */
+	obj.bindEventsInputLocation = function( $container ) {
+
+		// Bind event for the location input.
+		$container
+			.find( obj.selectors.inputLocationdWrapper )
+			.each( function( index, wrapper ) {
+				$( wrapper )
```
paulmskim: If we find ourselves calling `$()` more than once, like here with `$( wrapper )` and `$( input )`, it's a good idea to store them in variables so you can easily access them again. DOM queries are expensive, and jQuery is particularly more expensive than native JS. The less we can rely on calling `$()`, `.find()`, `.closest()`, and related DOM methods the lighter of an application we'll get.

paulmskim: also do we need this and below wrapped in an `.each()` function?

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,138 @@
...
+	 * @param {object} data data object passed from 'afterSetup.tribeEvents' event
+	 *
+	 * @return {void}
+	 */
+	obj.bindEvents = function( event, index, $container, data ) {
+		// Bind event for the keyword input.
+		obj.bindEventsInputFocus( $container, obj.selectors.inputKeywordWrapper, obj.selectors.inputKeywordWrapperFocus );
+
+		// Bind event for the location input.
+		obj.bindEventsInputFocus( $container, obj.selectors.inputLocationWrapper, obj.selectors.inputLocationWrapperFocus );
```
bordoni: We want to make it generic and controlled by a `[data-js]` attribute, which would allow creating more fields without any JS added.

src/resources/js/views/events-bar.js
```diff
@@ -0,0 +1,138 @@
...
+	 * @since TBD
+	 *
+	 * @type {PlainObject}
+	 */
+	obj.selectors = {
+		input: '[data-js="tribe-events-events-bar-input-control-input"]',
+		inputWrapper: '[data-js="tribe-events-events-bar-input-control"]',
+		inputKeywordWrapper: '.tribe-common-c-search__input-control--keyword',
+		inputKeywordWrapperFocus: '.tribe-common-c-search__input-control--keyword-focus',
+		inputLocationWrapper: '.tribe-common-c-search__input-control--location',
```
bordoni: We want to make these generic, no specificity to Location or Keyword should be present on the JS. Including classes.

src/resources/js/views/events-bar.js
juanfra: We actually need to pass what classes we need to set. If it was on the input we could use something generic, but as we're targeting via wrapper, we'll need to know what wrapper are we on.. location, keyword or whatever, to add the `--focus` modifier class.

PR #2591, opened 2019-06-25
List view dynamic dates output

bordoni:
_Ref:_ [C#129119](http://central.tri.be/issues/129119)
---


src/Tribe/Views/V2/Utils/Separators.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @since  TBD
+	 *
+	 * @param  array       $events WP_Post or numeric ID for events
+	 * @param  WP_Post|int $event  Event we want to determine
+	 *
+	 * @return boolean
+	 */
+	public static function should_have_month( $events, $event ) {
+		if ( ! is_numeric( reset( $events ) ) ) {
+			$ids = wp_list_pluck( $events, 'ID' );
```
lucatume: This approach will break with a mix of IDs and events; it should be easy to refactor this code into an `array_map` like:
```php
$ids = array_map(static function($event){
    return is_numeric($event) ? $event : $event->ID;
}, $events);
```

src/Tribe/Views/V2/Utils/Separators.php
```diff
@@ -0,0 +1,47 @@
...
+		} else {
+			$ids = array_map( 'absint', $events );
+		}
+
+		$event_id = is_numeric( $event ) ? $event : $event->ID;
+
+		$start_dates = array_map( function( $id ) {
```
lucatume: use `static function` since we're not using `$this` inside the closure.

lucatume:
Make sure tests pass, then merge

PR #2592, opened 2019-06-26
Add request argument into the filter

mitogh:
Ticket: https://central.tri.be/issues/129858


src/Tribe/REST/V1/Endpoints/Archive_Event.php
```diff
@@ -164,12 +164,13 @@ public function get( WP_REST_Request $request ) {
 			/**
 			 * Filter the arguments used to get the events on the archive page via REST API.
 			 *
```
bordoni: Add a missing `@since` of when this param was added.

mitogh: This filter was added during this MR so `since TBD` is part of the dock block of this PR Let me move it a bit.

PR #2594, opened 2019-06-26
Views V2: Single Event template loading behavior fixed

bordoni:
Our single event pages were broken due to View V2 deactivation of the V1 actions without correctly pulling the old single views to be included.

### DIFF
--- a/src/Tribe/Views/V2/Template_Bootstrap.php
+++ b/src/Tribe/Views/V2/Template_Bootstrap.php
@@ -71,7 +71,7 @@ class Template_Bootstrap {
 	 *
 	 * @return string
 	 */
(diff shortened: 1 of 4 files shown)


src/Tribe/Views/V2/Template_Bootstrap.php
```diff
@@ -64,6 +64,21 @@ public function get_template_object() {
...
 	}
 
+	/**
+	 * Fetches the HTML for the Single Event page using the legacy view system
+	 *
+	 * @since  TBD
+	 *
+	 * @return string
+	 */
+	protected function get_legacy_single_event_html() {
```
lucatume: Maybe `get_v1...`?

PR #2595, opened 2019-06-26
VR > Events Bar > Filters Button Styles

juanfra:
🎫 https://central.tri.be/issues/129301

### DIFF
--- a/src/resources/postcss/components/_events-bar.pcss
+++ b/src/resources/postcss/components/_events-bar.pcss
@@ -192,9 +192,9 @@
 	}
 
 	.tribe-events-c-events-bar__tab-icon {
-		height: var(--spacer-3);
+		height: 16px;
 		margin-bottom: var(--spacer-0);
-		width: var(--spacer-3);
+		width: 16px;
 	}
 
 	.tribe-events-c-events-bar__tab--active {
--- a/src/resources/postcss/components/_events-bar.pcss
+++ b/src/resources/postcss/components/_events-bar.pcss
@@ -159,6 +159,40 @@
 		width: 21px;
 	}
 
+	.tribe-events-c-events-bar__tabs {
+		display: flex;
+	}
+
+	.tribe-events-c-events-bar__tab {
+		border-radius: var(--border-radius-default);
+		display: flex;
+		flex: 1 1 50%;
+		flex-direction: column;
+		align-items: center;
+		padding: var(--spacer-3) 0;
+
+		&:hover {
+			background-color: var(--color-background-secondary);
+			.tribe-events-c-events-bar__tab-text {
+				font-weight: var(--font-weight-bold);
+			}
+
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


src/resources/postcss/components/_events-bar.pcss
```diff
@@ -159,6 +159,40 @@
...
+
+	.tribe-events-c-events-bar__tab {
+		border-radius: var(--border-radius-default);
+		display: flex;
+		flex: 1 1 50%;
+		flex-direction: column;
+		align-items: center;
+		padding: var(--spacer-3) 0;
+
+		&:hover {
```
paulmskim: also `&:focus`

src/resources/postcss/components/_events-bar.pcss
```diff
@@ -159,6 +159,40 @@
 		width: 21px;
 	}
 
+	.tribe-events-c-events-bar__tabs {
+		display: flex;
+	}
+
+	.tribe-events-c-events-bar__tab {
+		border-radius: var(--border-radius-default);
```
paulmskim: let's alphabetize all the styles, we don't have a linter yet but the linter will flag these.

src/resources/postcss/components/_events-bar.pcss
```diff
@@ -159,6 +159,40 @@
 		width: 21px;
 	}
 
+	.tribe-events-c-events-bar__tabs {
+		display: flex;
+	}
+
+	.tribe-events-c-events-bar__tab {
```
paulmskim: there should be a 8px margin between the two tabs (it's a bit hard to see, but if you focus on the text you'll get the width of the tabs)

src/resources/postcss/components/_events-bar.pcss
```diff
@@ -159,6 +159,40 @@
...
+			background-color: var(--color-background-secondary);
+			.tribe-events-c-events-bar__tab-text {
+				font-weight: var(--font-weight-bold);
+			}
+
+		}
+	}
+
+	.tribe-events-c-events-bar__tab-icon {
+		width: var(--spacer-3);
```
paulmskim: i don't think we should use spacers here for width and height. they match the px size, but that's not what we're using spacers for. if we decide to double all the spacer sizes but keep the font size, icon size, etc. the same, this would be unintentionally changed. extremely unlikely, but thinking in terms of semantics and maintenance.

src/resources/postcss/components/_events-bar.pcss
paulmskim: see note in fe-de-views chat on hover and active state

paulmskim: see note in fe-de-views chat on hover and active state

juanfra: Ok, let's wait until they chime in for this one. My interpretation was that the hover had also the bg color.

PR #2599, opened 2019-06-28
VR > Events Bar > Datepicker styles

juanfra:
🎫 https://central.tri.be/issues/129302


src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+		margin-top: var(--spacer-1);
+		padding: 0 var(--spacer-3) var(--spacer-3);
+
+		@media (--viewport-medium) {
+			margin-top: var(--spacer-3);
+		}
+
+		&:before,
+		&:after {
+			display: none;
```
paulmskim: let's use `content: none` here to completely remove it rather than hiding it.

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+
+		.datepicker-switch {
+			color: var(--color-text-primary);
+			font-size: var(--font-size-3);
+			font-weight: var(--font-weight-regular);
+		}
+
+		.datepicker-switch,
+		.prev, .next {
+			line-height: initial;
```
paulmskim: we should set a line height here, initial for line height can vary from font to font and we don't have control over that.

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+			padding: initial;
```
paulmskim: don't need this if we have padding set just below

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+		th {
+			height: initial;
+			width: initial;
+		}
+
+		.dow {
+			color: var(--color-text-primary);
+			font-size: var(--font-size-0);
+			font-weight: var(--font-weight-regular);
+			padding: initial;
```
paulmskim: let's set the padding rather than letting the browser decide.

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+
+		td,
+		th {
+			height: initial;
+			width: initial;
+		}
+
+		.dow {
+			color: var(--color-text-primary);
+			font-size: var(--font-size-0);
```
paulmskim: we should always set a line height if we're setting a font size so we have control over the height of the element. could this possibly match one of the mixins that we have already defined? if not, could you check with matt/rachel to see what they were thinking?

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+			font-weight: var(--font-weight-regular);
+			padding: initial;
+			padding-bottom: var(--spacer-2);
+		}
+
+		.datepicker-switch {
+			color: var(--color-text-primary);
+			font-size: var(--font-size-3);
```
paulmskim: same comment as above on line height and using an already-defined mixin.

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+			&:hover,
+			&:focus,
+			&:active {
+				background-color: var(--color-background);
+			}
+		}
+
+		.prev,
+		.next {
+			text-indent: -999em;
```
paulmskim: we won't need to do this, as we can define what markup we use for the prev and next arrows. but i can take care of that on my end, so no need to do anything here.

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+			&:hover,
+			&:focus,
+			&:active {
+				&:before {
+					background-image: svg-inline(caret-right-active);
+				}
+			}
+		}
+
+		.datepicker-switch {
```
paulmskim: this style is already defined above.

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		.day,
+		.month,
+		.year {
+			&:hover,
+			&:focus,
+			&:active,
+			&.focused {
+				color: #fff;
```
paulmskim: we have a variable for this.

src/resources/postcss/components/_datepicker.pcss
```diff
@@ -0,0 +1,135 @@
...
+
+		.month,
+		.year {
+			font-size: var(--font-size-1);
+		}
+
+		.day,
+		.month,
+		.year {
+			&:hover,
```
paulmskim: ask matt/rachel about the hover/focus state for these, I can't imagine that it would be the same as the active state, but i could be wrong.

src/resources/postcss/components/_datepicker.pcss
ryanurban: Do we have a color var for white?

juanfra: Yes, but it has `background` in the name so I thought that would be confusing.

src/resources/postcss/components/_datepicker.pcss
juanfra: The initial padding is zero.

src/resources/postcss/components/_datepicker.pcss
juanfra: With the current implementation we don't have control over the markup. Is your implementation ready?

src/resources/postcss/components/_datepicker.pcss
juanfra: Yes, for a background, not related. If we change the background on the design system it'll affect this.

src/resources/postcss/components/_datepicker.pcss
juanfra: From the designs I assume that when they hover, focus or click they want to have the blue background. They haven't specified.

src/resources/postcss/components/_datepicker.pcss
paulmskim: Yup, you're right. What I mean is that we should explicitly set it rather than relying on the browser. Also, there is no IE support at all, so the width and height `initial` above will also not work.

src/resources/postcss/components/_datepicker.pcss
paulmskim: I've used the variable background color for these cases too. The background color of the datepicker is also the background color, and if that changes to, say, light grey, I would expect the color of the text to be such too in this case.

But if you want to create a new variable for this too I wouldn't be opposed. I do, however, think they are linked.

src/resources/postcss/components/_datepicker.pcss
paulmskim: not yet, but don't worry about the arrows, i can tackle them in the JS.

src/resources/postcss/components/_datepicker.pcss
paulmskim: still need line height here.

juanfra: I set it here, with the others: https://github.com/moderntribe/the-events-calendar/pull/2599/files#diff-572931b824026e2e53c9d15dc50c838dR46

src/resources/postcss/components/_datepicker.pcss
juanfra: I feel it's confusing, but if you already used it for something similar I don't mind using it here.

juanfra:
Thank you, Paul!

src/Tribe/Views/V2/Assets.php
```diff
@@ -166,6 +180,24 @@ public function register() {
...
+			'views/datepicker.js',
+			[ 'jquery', 'tribe-common', 'tribe-events-views-v2-bootstrap-datepicker' ],
+			null,
+			[
+				'priority' => 10,
+			]
+		);
+
+		/**
+		 * @todo: remove once we can not load v1 scripts in v2
```
tr1b0t: :warning: Comment refers to a TODO task

<sub>hash: bd0daab6fccfd12904b60d29985a07d2</sub>

paulmskim:
Pending one note left in views redesign tech

juanfra:
Closing in favor of https://github.com/moderntribe/the-events-calendar/pull/2600

PR #2604, opened 2019-07-01
fix(composer): Change case of package name.

Camwyn:
Change the `mikey179/vfsStream` package to all lowercase to squash warnings about composer 2.0.
Screen Shot 2019-07-01 at 7 34 21 PM

See also: https://github.com/moderntribe/events-pro/pull/1168


borkweb:
Well...tests need to pass first

lucatume:
I've restarted the build on this to make sure tests are not failing due to a fluke.

PR #2608, opened 2019-07-05
Feature/130235 theme overrides twenty nineteen

paulmskim:
https://central.tri.be/issues/130235


juanfra:
Let's update the snapshot and it's good to go 🚀

PR #2623, opened 2019-07-19
Add snapshot testing for day view data.

bordoni:
_Ref:_ [C#130580](http://central.tri.be/issues/130580)
---


tests/views_integration/Tribe/Events/Views/V2/Views/__snapshots__/List_ViewTest__test_render_empty__1.php
```diff
@@ -188,8 +188,8 @@ class="tribe-common-h2 tribe-common-h3--min-medium tribe-common-h--alt tribe-eve
 		data-js="tribe-events-top-bar-datepicker-button"
 	>
 					Now				&mdash;
-		<time datetime="2019-07-18">
-			July 18, 2019		</time>
+		<time datetime="2019-07-19">
+			July 19, 2019		</time>
```
paulmskim: we should look into mocking the date rather than having it update every time

bordoni: Yeah, we need to make sure that is the case.

PR #2625, opened 2019-07-19
Fixed database storage of pipe separated additional fields

jwoolsch:
When importing pipe separated additional fields via CSV, they were being stored as a single row pipe separated string in the meta data table instead of multiple searchable rows.


jwoolsch:
Reference ticket: GRT-EOVTP-386

PR #2629, opened 2019-07-24
REST failover to AJAX

bordoni:
_Ref:_ [C#130051](http://central.tri.be/issues/130051)
---


src/Tribe/Views/V2/Rest_Endpoint.php
```diff
@@ -16,6 +38,106 @@ class Rest_Endpoint {
...
+
+	/**
+	 * AJAX fallback for when REST endpoint is disabled. We try to mock a WP_REST_Request
+	 * and use the same method behind the scenes to make sure we have consistency.
+	 *
+	 * @since  TBD
+	 *
+	 * @return void
+	 */
+	public function ajax_fallback() {
```
lucatume: Nitpicking here: I'd prefer a method name like <verb>_<noun> that will keep making sense in time.
Here I'd use `handle_ajax_request`.

src/Tribe/Views/V2/Rest_Endpoint.php
```diff
@@ -16,6 +38,106 @@ class Rest_Endpoint {
...
+		$request = new Request( 'GET', static::NAMESPACE . '/html' );
+
+		$body_params = (array) $_GET;
```
lucatume: Sanitize here; as first step drop any argument we're not using, as a second step sanitize each by type.

src/Tribe/Views/V2/Rest_Endpoint.php
```diff
@@ -16,6 +38,106 @@ class Rest_Endpoint {
...
+
+	/**
+	 * Check if the REST endpoint is available.
+	 *
+	 * @since  TBD
+	 *
+	 * @return boolean If the REST API endpoint is available.
+	 */
+	public function is_available() {
+		$is_available = tribe( 'tec.rest-v1.system' )->tec_rest_api_is_enabled();
```
lucatume: Should this check be for the WordPress REST API itself?
One could have deactivated TEC REST API but kept the WordPress REST API active.
The REST API we're using here is another matter entirely from the TEC REST API and we're filtering its availability below.

tests/views_integration/Tribe/Events/Views/V2/Rest_EndpointTest.php
```diff
@@ -0,0 +1,38 @@
+<?php
+namespace Tribe\Events\Views\V2;
+
+class Rest_EndpointTest extends \Codeception\TestCase\WPTestCase {
```
lucatume: Add a test to check for sanitization.
Add a test for the discrimination between WordPress REST API and TEC REST API (a test matrix).
Add some tests for the AJAX handling method to make sure it returns the same stuff the REST-based approach returns.

src/Tribe/Views/V2/Rest_Endpoint.php
bordoni: Yeah I made a change to check what I could, but since 4.7 you cannot disable the rest API anymore, directly on WP. But we might need to have a internal HTTP request to do a real test of availability of the API.

src/Tribe/Views/V2/Rest_Endpoint.php
bordoni: I am not sure if we want to add a layer of Sanitization here, because it should just pass the params directly without filtering since it would be a second place to check for more params to be used. I would like that to be done on the `make_for_rest`. Or do you mean something else?

tests/views_integration/Tribe/Events/Views/V2/Rest_EndpointTest.php
bordoni: What do you mean a test between TEC and WP rest API?

src/Tribe/Views/V2/Rest_Endpoint.php
bordoni: Had a convo with @lucatume will abstract that info and make sure we are running the same checks for both.

PR #2631, opened 2019-07-25
Mobile view redirecting to list view indefinitely

drclaw:
When using the month view as the default, but the list view for mobile, the month view always gets redirected back to the list view. Looks like it might just be a matter of using an ID selector instead of class?


bordoni:
CTL/137877

PR #2634, opened 2019-08-01
VR: fix some Month View test issues

lucatume:
Ticket: n/a

While working on new code I've found out and fixed some issues with the month view
and tests related to the global `today` and `today_date` variables.
I've regenerated the snapshots to reflect that change.


src/views/v2/base.php
```diff
@@ -1,7 +1,28 @@
...
+ * View: Base View
+ *
+ * Override this template in your own theme by creating a file at:
+ * [your-theme]/tribe/events/views/v2/base.php
+ *
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version 4.9.4
```
juanfra: Let's set the version TBD

lucatume: This is there from version 1, I just forgot to set it before.

juanfra: Yeah, I know. But just mentioning because the point of the `@version` is to highlight if there have been changes every time the plugin updates.

PR #2636, opened 2019-08-01
Fix/131554 events bar dqa

paulmskim:
first pass on https://central.tri.be/issues/131554


bordoni:
As soon as the tests are passing we can merge.

PR #2637, opened 2019-08-02
VR: fix orderby repository issue

lucatume:
Ticket: https://central.tri.be/issues/130598

Depends on https://github.com/moderntribe/tribe-common/pull/1103

This PR fixes two issues:
1. the Event ORM/Repository handling of multiple `order_by` calls.
2. the duplicated call to the `View::setup_repository_args` method in the `View::get_html` method.

The first, especially, would lead to "stacking" of order-by clauses and would lead to SQL errors.


src/Tribe/Repositories/Event.php
```diff
@@ -1526,4 +1444,205 @@ protected function format_item( $id ) {
...
+			$wpdb->prepare(
+				"
+				LEFT JOIN {$wpdb->postmeta} AS {$postmeta_table}
+					ON (
+						{$postmeta_table}.post_id = {$wpdb->posts}.ID
+						AND {$postmeta_table}.meta_key = %s
+					)
+				LEFT JOIN {$wpdb->posts} AS {$posts_table}
+					ON {$wpdb->posts}.ID = {$postmeta_table}.meta_value
+			",
```
bordoni: Wrong indentation.

src/Tribe/Repositories/Event.php
```diff
@@ -1526,4 +1444,205 @@ protected function format_item( $id ) {
...
```
bordoni: Wrong indentation.

src/Tribe/Repositories/Event.php
```diff
@@ -1526,4 +1444,205 @@ protected function format_item( $id ) {
...
+
+		$this->filter_query->join(
+			$wpdb->prepare(
+				"
+				LEFT JOIN {$wpdb->postmeta} AS {$postmeta_table}
+					ON (
+						{$postmeta_table}.post_id = {$wpdb->posts}.ID
+						AND {$postmeta_table}.meta_key = %s
+					)
+			",
```
bordoni: Wrong indentation.

src/Tribe/Repositories/Event.php
```diff
@@ -1526,4 +1444,205 @@ protected function format_item( $id ) {
...
+
+		$check_orderby = $order_by;
+
+		if ( ! is_array( $check_orderby ) ) {
+			$check_orderby = explode( ' ', $check_orderby );
+		}
+
+		$timestamp_key = 'TIMESTAMP(mt1.meta_value)';
+
+		$by_event_start_date     = isset( $check_orderby['event_date'] )
```
bordoni: For the date related ones do we need to check for all the other possible ways `event_date` is represented? Like when you use `$orm->by( 'date_overlaps', $start, $end );`. I am not sure if it's handled but wanted to make sure.

lucatume: It's handled; filtering and ordering are different matters but the code that this PR shows as changed is really just moved around, not changed in functionality.

PR #2639, opened 2019-08-06
Feature/130237 theme overrides divi

paulmskim:
https://central.tri.be/issues/130237


src/Tribe/Views/V2/Theme_Compatibility.php
```diff
@@ -46,7 +46,7 @@ public function is_compatibility_required() {
 			return false;
 		}
 
-		if ( in_array( $template, $this->get_registered_themes() ) ) {
+		if ( in_array( strtolower( $template ), $this->get_registered_themes() ) ) {
```
paulmskim: @bordoni would love your feedback on this. When I extracted the theme into the themes folder, the theme folder was named `Divi`. Because we get the theme name from the folder, `Divi` did not match `divi` in https://github.com/moderntribe/the-events-calendar/pull/2639/files#diff-0b6d28a0d302f614926ff090305c2222L26 and did not apply the proper body class. I'm also open to any other approach here that achieves the same thing.

juanfra: Nice catch! One question then: Shouldn't we apply `strtolower` here as well? https://github.com/moderntribe/the-events-calendar/blob/ca21a831e4415d1286b9b495485c19ff05db3051/src/Tribe/Views/V2/Theme_Compatibility.php#L94

Or the body class is in lowercase?

paulmskim: yes, I think you're right. I changed it on my local and didn't change it back, so it was still lowercase.

paulmskim:
@bordoni @juanfra this is good for another pass.

PR #2646, opened 2019-08-13
Feature/130236 theme overrides avada

paulmskim:
https://central.tri.be/issues/130236


src/resources/postcss/components/_search.pcss
```diff
@@ -199,4 +199,17 @@
 			}
 		}
 	}
+
+	/* -----------------------------------------------------------------------------
+	 * Theme Overrides - Enfold
```
bordoni: Copy and Pasta, should be avada right?

paulmskim: You're right, yeah should be avada

PR #2648, opened 2019-08-15
Views v2 - dynamic Week View support

lucatume:
Tickets:

* https://central.tri.be/issues/130599
* https://central.tri.be/issues/130598

Updates base Views V2 code to support dynamic Week view step 2:


src/views/v2/day/top-bar/today.php
```diff
@@ -11,14 +11,9 @@
  *
  * @version 4.9.3
```
juanfra: Not a stopper, but every time we update a template we need to set this number as TBD so it's replaced with the new version (and eventually when this is live, users can know which templates have been updated).

lucatume: Ok, thanks, will update.

juanfra: Thanks 👍  Not a stopper now, but something we should get used to do every time we touch a template, to create the habit.

PR #2654, opened 2019-08-20
Display title properly multiday and on mobile

bordoni:
_Ref:_ [C#132859](http://central.tri.be/issues/132859)
---


bordoni:
@lucatume I am not sure what is going on with the `restv1` or the `wpunit` they are both working locally, and nothing has changed here that would break those, do you think they are false alerts? Meaning, I see the `restv1` fail every now and again and when restarted it works.

Wondering if we should dig in further here, currently focused on the Alpha 1 for the VR, would you mind checking this on your AM?

lucatume:
> @lucatume I am not sure what is going on with the `restv1` or the `wpunit` they are both working locally, and nothing has changed here that would break those, do you think they are false alerts? Meaning, I see the `restv1` fail every now and again and when restarted it works.
> 
> Wondering if we should dig in further here, currently focused on the Alpha 1 for the VR, would you mind checking this on your AM?
> 
> --
> _Edit_
> It feels like https://travis-ci.org/moderntribe/the-events-calendar/builds/574587573#L2002 is being caused by something to do with timezones on Travis.

The build was pulling an older version of common. Merging latest code fixed the issue.

src/views/v2/month/calendar-body/day/multiday-events/multiday-event.php
```diff
@@ -77,7 +77,7 @@ class="tribe-events-calendar-month__multiday-event-featured-icon tribe-common-sv
 				></em>
 			<?php endif; ?>
 			<h3 class="tribe-events-calendar-month__multiday-event-title tribe-common-h8">
-				<?php echo wp_kses_post( $event->title ) ?>
+				<?php echo get_the_title( $event ); ?>
```
juanfra: I'm curious if we need to define if we're going to go with the template tags or echoing the getters on the templates. Maybe we could define this. Like, using `the_title` vs `echo get_the_title`

bordoni: Problem with `the_title` is that it doesnt accept the event ID, which is a problem for testing.

PR #2655, opened 2019-08-20
Makes sure we take year into consideration on List View header date

bordoni:
_Ref:_ [C#129829](http://central.tri.be/issues/129829)
---


bordoni:
Same as: https://github.com/moderntribe/the-events-calendar/pull/2654#issuecomment-523281508

src/views/v2/list/top-bar/datepicker.php
```diff
@@ -14,25 +14,39 @@
...
+
+/**
+ * @todo @bordoni @lucatume Abstract the handling of these dates into a method/function
+ */
 $default_start_date = 'now';
 $selected_start_date_value = $this->get( [ 'bar', 'date' ], $default_start_date );
 if ( empty( $selected_start_date_value ) ) {
-	$selected_start_date_value = $default_start_date;
+	$first_event = $this->get( 'view' )->get_repository()->first();
+
```
juanfra: Maybe assign `$default_start_date` to `$selected_start_date_value` and have only the `if` to check if it's intance of and replace, to avoid the else?

juanfra: Just to be consistent with what we use elsewhere.

bordoni: On this case we need that else, because of the way initial `$selected_start_date_value` is set.
I wanted to avoid that else too.

PR #2652, opened 2019-08-16
Scenarios > List View > event excerpt

leahkoerper:
changes excerpt to 30 words instead of 35 per Central #131088

PR #2659, opened 2019-08-20
Scenarios > specifying for V2

leahkoerper:
Clarifying when scenarios specifically require V2 Views


leahkoerper:
Depends on #2652

PR #2661, opened 2019-08-21
VR > Featured Events > List

juanfra:
🎫 https://central.tri.be/issues/131750

🎥 https://www.loom.com/share/47ed388c066b4eca84ba0539f0e54b94


src/resources/postcss/views/list/_event.pcss
```diff
@@ -118,7 +141,19 @@
 	}
 
 	.tribe-events-calendar-list__event-datetime-featured-icon {
-		margin-left: 10px;
+		margin-right: var(--spacer-1);
+	}
+
+	.tribe-events-calendar-list__event-datetime-featured-text {
+		@mixin hidden;
```
paulmskim: let's use the `visually-hide` mixin here instead, and `visually-show` below.

src/resources/postcss/views/list/_event.pcss
```diff
@@ -30,20 +30,42 @@
...
+				box-shadow: 2px 0 0 0 var(--color-accent-primary);
+			}
+		}
+	}
+
 	.tribe-events-calendar-list__event-date-tag {
 		flex: none;
 		margin-right: var(--grid-gutter-small-half-negative);
-		width: var(--grid-width-1-of-4);
+		width: calc(var(--spacer-9) + var(--grid-gutter-small-half));
```
paulmskim: I thought we were going to use no gutters here (removing the `tribe-common-g-row--gutters` class). There's also an overlap that does not allow a user to select half the text, which could be fixed with removing the gutters.
list-view

src/resources/postcss/views/list/_event.pcss
```diff
@@ -30,20 +30,42 @@
...
 
+	.tribe-events-calendar-list__event-row--featured {
+
+		.tribe-events-calendar-list__event-date-tag-datetime {
+			border-left: 3px solid var(--color-accent-primary);
+
+			@media (--viewport-medium) {
+				border-left: none;
+				border-right: 2px solid var(--color-accent-primary);
+				box-shadow: 2px 0 0 0 var(--color-accent-primary);
```
paulmskim: curious, why are we using a box shadow here rather than using a border-right of 4px (or even an `&:after` pseudoelement)?

juanfra: I made that split, because in the design the border is is positioned in two different parts: https://cloudup.com/cLNgfzxICpA

src/resources/postcss/views/list/_event.pcss
juanfra: Yes, the thing is that we need the gutters on desktop. Maybe it's worth to have a partially thing for no gutters on mobile? Like we have with headings?

paulmskim: I'd recommend having no gutters and putting in the gutters on desktop.

src/resources/postcss/views/list/_event.pcss
```diff
@@ -11,6 +11,18 @@
...
 			margin-bottom: var(--spacer-8);
 			margin-top: var(--spacer-9);
+			margin-left: var(--grid-gutter-small-half-negative);
+			margin-right: var(--grid-gutter-small-half-negative);
+
+			& > .tribe-common-g-col {
+				padding-left: var(--grid-gutter-small-half);
+				padding-right: var(--grid-gutter-small-half);
+
+				@media (--viewport-medium) {
```
paulmskim: A few things here:
1. You've got a media query wrapped in a media query here.
2. I don't think we're using padding on mobile right?
3. For desktop, the negative margin should match the padding.

juanfra: 1. Nice catch!
2. Yes, no padding on mobile.
3. 👍

PR #2663, opened 2019-08-22
Add logic to the Month view Previous and Next URL

bordoni:
_Ref:_ [C#133006](http://central.tri.be/issues/133006)
---
http://i.bordoni.me/77c25b6ed6a2

### DIFF
--- a/src/Tribe/Views/V2/Views/Month_View.php
+++ b/src/Tribe/Views/V2/Views/Month_View.php
@@ -72,31 +72,18 @@ class Month_View extends By_Day_View {
 		$date           = $this->context->get( 'event_date', $default_date );
 		$event_date_var = $default_date === $date ? '' : $date;
 
-		$previous_date = date( 'Y-m-t', strtotime( $date . ' -1 month' ) );
+		// Start of the month passed
+		$previous_date = date( 'Y-m', strtotime( $date . '-01 00:00' ) );
 
-		$past = tribe_events()->by_args( [
-			'on_date'  => $previous_date,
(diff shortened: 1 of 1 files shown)


src/Tribe/Views/V2/Views/Month_View.php
```diff
@@ -64,6 +64,101 @@ public function __construct( Stack $stack) {
...
+			return '';
+		}
+
+		$next_date = tribe_get_start_date( $next_event, false, 'Y-m' );
+
+		$query_args = [ 'eventDate' => $next_date ];
+		$url = remove_query_arg( [ 'tribe-bar-date' ], $this->get_url() );
+		$url = add_query_arg( $query_args, $url );
+
+		if ( ! empty( $url ) && $canonical ) {
```
mitogh: It seems like this chunk: https://github.com/moderntribe/the-events-calendar/pull/2663/files#diff-88603a33bd81edac461866548086b791R139-R156

Can be reused as a private method as is the same logic as above something along the lines of: 

But I don't think is a blocker, as it looks okay to me.

bordoni: We might need to think about putting this on the View parent class, since this basically repeats on all methods of `prev` and `next`. Adding a `@todo` for that.

bordoni:
```suggestion
		/**
		 * @todo @be move this repeating piece of code to the parent `View` class.
		 */
		if ( ! empty( $url ) && $canonical ) {
```

PR #2666, opened 2019-08-22
VR > Featured Events > Day View

juanfra:
🎫 https://central.tri.be/issues/131752

🎥 https://www.loom.com/share/1f40d4dc5cb242ed939c5ad1f71c32af


src/resources/postcss/views/day/_event.pcss
```diff
@@ -28,6 +28,21 @@
...
 
+	.tribe-events-calendar-day__event--featured {
+
+		.tribe-events-calendar-day__event-details {
+
+			&:before {
+				background: var(--color-accent-primary);
+				content: '';
+				height: 100%;
+				left: var(--grid-gutter-small-half-negative);
```
paulmskim: shouldn't this have different values for mobile and desktop (since the gutter changes)?

paulmskim:
@juanfra what does this look like if the event has an image? I can see it looking the same on desktop, but on mobile, I don't think the line will run from the image down to the event details.

juanfra:
You're right. Given the margins that we have I think we'll need to have a wrapper and apply the featured styles there for mobile.

src/resources/postcss/views/day/_event.pcss
```diff
@@ -28,6 +28,40 @@
...
+			content: '';
+			height: 100%;
+			left: var(--grid-gutter-small-half);
+			position: absolute;
+			width: 3px;
+		}
+
+		@media (--viewport-medium) {
+			&:before {
+				width: 0;
```
paulmskim: We should be removing this by using `content: none`. But also, can we reuse this for the desktop instead of creating another pseudo element below? We're copying most of the styles from the `&:before` above.

src/views/v2/day/event/date.php
```diff
@@ -9,25 +9,28 @@
...
 		<em
 			class="tribe-events-calendar-day__event-datetime-featured-icon tribe-common-svgicon tribe-common-svgicon--featured"
 			aria-label="<?php esc_attr_e( 'Featured', 'the-events-calendar' ); ?>"
 			title="<?php esc_attr_e( 'Featured', 'the-events-calendar' ); ?>"
 		>
 		</em>
+		<span class="tribe-events-calendar-day__event-datetime-featured-text tribe-common-b2"><?php esc_html_e( 'Featured', 'the-events-calendar' ); ?></span>
 	<?php endif; ?>
+	<time class="tribe-events-calendar-day__event-datetime tribe-common-b2" datetime="<?php echo esc_attr( $event_date_attr ); ?>">
+		<?php echo tribe_events_event_schedule_details( $event ); ?>
```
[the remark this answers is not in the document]
juanfra: @bordoni I think if we're moving that way then we should make that a separate task and the update it on all the different views where it's currently being used.

src/resources/postcss/views/day/_event.pcss
juanfra: @paulmskim thanks, updated the `content`. With regards to the pseudo, we're defining them for two different elements to match the design.

paulmskim: Great stuff. For the pseudo-element, I understand why you did this, what I mean is we can do it with one pseudo-element and position it to match the design rather than with two pseudo-elements and copied styles.

juanfra: https://www.loom.com/share/856372c4fd5e49fd9bd78cab1b3c4a32

paulmskim: So if you set `left: calc(var(--grid-width-1-of-9) - var(--grid-gutter-half));` on the article pseudo-selector for `--viewport-medium` and remove the `__event-details` pseudo-selector, you should get the same effect as the design but using the same pseudo-selector. Let's hop on a call if it doesn't work out.

juanfra: Ok, cool, thanks!

src/views/v2/day/event/date.php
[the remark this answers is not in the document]
paulmskim: Can you create a ticket for this @juanfra and make sure it's assigned to someone?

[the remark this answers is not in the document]
juanfra: Sure, thanks!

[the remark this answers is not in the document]
juanfra: https://central.tri.be/issues/133516

src/views/v2/day/event/date.php
```diff
@@ -9,25 +9,30 @@
...
 			aria-label="<?php esc_attr_e( 'Featured', 'the-events-calendar' ); ?>"
 			title="<?php esc_attr_e( 'Featured', 'the-events-calendar' ); ?>"
 		>
 		</em>
+		<span class="tribe-events-calendar-day__event-datetime-featured-text tribe-common-b2">
+			<?php esc_html_e( 'Featured', 'the-events-calendar' ); ?>
+		</span>
 	<?php endif; ?>
+	<time class="tribe-events-calendar-day__event-datetime tribe-common-b2" datetime="<?php echo esc_attr( $event_date_attr ); ?>">
+		<?php echo tribe_events_event_schedule_details( $event ); ?>
```
lucatume: Use the `schedule_details` property here.
```suggestion
		<?php echo esc_html( $event->schedule_details->value() ); ?>
```

PR #2669, opened 2019-08-23
Use tribe_is_truthy to verify Meta values to allow other false/truthy

bordoni:
_Ref:_ [C#133098](http://central.tri.be/issues/133098)
---


bordoni:
@lucatume tests are now passing. Just needed to restart travis.

PR #2670, opened 2019-08-23
VR > Month View > Past events

juanfra:
🎫 https://central.tri.be/issues/133035

🎥 https://www.loom.com/share/5af2f4c2710e4079ac649a44876a746e

### DIFF
--- a/src/resources/postcss/views/month/_calendar-event.pcss
+++ b/src/resources/postcss/views/month/_calendar-event.pcss
@@ -63,6 +63,28 @@
 		}
 	}
 
+	.tribe-events-calendar-month__day--past {
+
+		.tribe-events-calendar-month__calendar-event--featured {
+
+			&:before {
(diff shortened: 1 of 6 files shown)


juanfra:
Using vars from https://github.com/moderntribe/tribe-common-styles/pull/37 - Let's confirm we are using those before merging this.

src/resources/postcss/views/month/_calendar-event.pcss
```diff
@@ -63,6 +63,28 @@
...
+
+		.tribe-events-calendar-month__calendar-event--featured {
+
+			&:before {
+				background: var(--color-past-events);
+			}
+		}
+
+		.tribe-common-c-image__bg {
+			filter: grayscale(100%);
```
paulmskim: This property won't work on IE11. We need to think of something that'll support IE11.

PR #2671, opened 2019-08-23
Update to Freemius integration

barryhughes:
Increasing target ratio from 10% to 50%.

:ticket: [♯133148](https://central.tri.be/issues/133148)


barryhughes:
Unfortunately, didn't merge this into B19.11 and we're now targeting B19.12. The code is basically unchanged but I did move the changelog entry up into the 4.9.9 block. For safety, removing the merge label so we get a quick check by a 2nd pair of eyes.

PR #2676, opened 2019-08-27
fix(tribe_get_event) use Lazy_Events

lucatume:
Ticket: n/a

Based on https://github.com/moderntribe/tribe-common/pull/1128 and https://github.com/moderntribe/tribe-testing-facilities/pull/9

This PR leverages the `Tribe\Utils\Lazy_Events` trait to make sure the decorated post produced by the `tribe_get_event` function will be cached on `shutdown` and only if one of the "lazy" properties was accessed.


lucatume:
Currently failiing as the `tribe-testing-facilities` PR is required.

PR #2685, opened 2019-09-02
Doc/vr tests readme 2

paulmskim:


tests/views_integration/README.md
```diff
@@ -316,10 +316,155 @@ This second test example shows what is, probably, the main feature of data-drive
...
+			[
+				'today'      => $this->mock_date_value,
+				'now'        => $this->mock_date_value,
+			]
+		);
+
+		$list_view = View::make( List_View::class, $context );
+		$html      = $list_view->get_html();
+
+		// Let's make sure the View is displaying what events we expect it to display.
```
lucatume: @juanfra  This part of the code, down to line `363` should be removed from this example: this is **data** testing, not snapshot testing.
Trusting the fact that data is tested in its own tests, this test should not re-test it.

tests/views_integration/README.md
```diff
@@ -316,10 +316,155 @@ This second test example shows what is, probably, the main feature of data-drive
...
+			'today'      => $this->mock_date_value,
+			'now'        => $this->mock_date_value,
+			'events_per_page' => 2,
+		] ) );
+		$html = $list_view->get_html();
+
+		// Let's make sure the View is displaying what events we expect it to display.
+		$expected_post_ids = wp_list_pluck( array_slice( $events, 0, 2 ), 'ID' );
+		$this->assertEquals(
+			$expected_post_ids,
```
lucatume: @juanfra  As above, this part of the code, lines `407` to `412`, should be removed as this is **data** testing, not snapshot testing.

tests/views_integration/README.md
```diff
@@ -316,10 +316,155 @@ This second test example shows what is, probably, the main feature of data-drive
...
+		$this->assertEquals( 3, tribe_events()->where( 'ends_after', 'now' )->count() );
+
+		$this->remap_posts( $events, [
+			'events/featured/1.json',
+			'events/single/1.json',
+			'events/single/2.json'
+		] );
+
+		$list_view = View::make( List_View::class );
+		$list_view->set_context( tribe_context()->alter( [
```
lucatume: @juanfra  I think we could add some comments here, in the PHP comment format, to make clear what we're doing at each step.

tests/views_integration/README.md
```diff
@@ -316,10 +316,155 @@ This second test example shows what is, probably, the main feature of data-drive
...
+				'timezone'   => 'Europe/Paris',
+				'duration'   => 3 * HOUR_IN_SECONDS,
+				'title'      => 'Test Event - ' . $start_date,
+				'status'     => 'publish',
+			] )->create();
+		}
+		// Sanity check
+		$this->assertEquals( 3, tribe_events()->where( 'ends_after', 'now' )->count() );
+
+		$this->remap_posts( $events, [
```
lucatume: What are we doing here? Why?
@juanfra  Please add inline comments here.

tests/views_integration/README.md
```diff
@@ -316,10 +316,155 @@ This second test example shows what is, probably, the main feature of data-drive
...
+		);
+
+		$this->assertMatchesSnapshot( $html );
+	}
+}
+```
+
+By using the `MatchesSnapshots` trait and calling the `assertMatchesSnapshot` method, we can set an initial snapshot and compare the markup each time the test is run.
+
+When a markup change occurs, the test will fail as the html markup will not match the snapshot. In this case, review the differences. if they are what you expect, then delete the snapshot file and run the test again to generate a new snapshot. Commit this snapshot to the repo so that all others running tests will have the latest snapshot to compare to.
```
lucatume: @juanfra  Here add a mention about the `--debug` flag.
```suggestion
When a markup change occurs, the test will fail as the HTML markup will not match the snapshot. In this case, review the differences. if they are what you expect, then delete the snapshot file and run the test again to generate a new snapshot. Commit this snapshot to the repo so that all others running tests will have the latest snapshot to compare to.
```

tests/views_integration/README.md
```diff
@@ -316,10 +316,155 @@ This second test example shows what is, probably, the main feature of data-drive
 * we manipulate data as the full View, in production, would do
 
 ### Snapshot testing
-@todo @juanfra
+
+This type of testing answers the following question:
+
+> Is this View, or View partial, rendering the correct markup (HTML structure, attributes, and data output) given a specific set of template variables?
```
lucatume: @juanfra  @paulmskim  If we use snapshot testing in the following section, then we need to remove the mention of partials from here.
This sections should deal with entire Views only and should be renamed "View snapshot testing".

tests/views_integration/README.md
```diff
@@ -316,10 +316,155 @@ This second test example shows what is, probably, the main feature of data-drive
 * we manipulate data as the full View, in production, would do
 
 ### Snapshot testing
```
lucatume: See comment about possible title change below.

tests/views_integration/README.md
```diff
@@ -316,10 +316,155 @@ This second test example shows what is, probably, the main feature of data-drive
...
+
+This type of testing deals with the following question:
+
+> Is this View partial rendering the correct markup (HTML structure, attributes, and data output) given a specific set of template variables?
+
+This testing is often run using snapshot testing as we are most interested in the View partial markup. See **Snapshot Testing** above for details.
+
+The View partials folder structure is organized in the same structure as the view partials. When creating a new test for a partial, place the test within the appropriate folder matching that of the actual markup.
+
+Below is an example of List View Nav:
```
lucatume: @paulmskim Above is said that this kind of testing "is often run using snapshot testing" and that's the only available example of testing. Please add more examples not using snapshot testing.

tests/views_integration/README.md
lucatume: @juanfra @paulmskim I think we're mixing **what we test** (View data, View HTML, Partial HTML) with **how** we test it (db queries, snapshot testing, HTML structure lookup).
This section should be renamed to "View (HTML) testing".

Snapshot testing is the testing technique used for testing entire Views, so it makes sense to explain it here.
Please add a sub-section dedicated to snapshot testing and move the presentation of the technique there; this should serve as an high level introduction for it, that could be referenced in the "View (HTML) Testing" and "Component (HTML) Testing" sections.

PR #2689, opened 2019-09-03
fix(timezone notice): The timezone notice responds incorrectly.

Camwyn:
The timezone notice appears to respond incorrectly when on an event edit page (admin). The function checks the event timezone properly, but when the notice is set, it uses the site timezone. This is all well and good until we're using "manual timezones for each event" - in which case the site timezone is fine but the event one is not (defaults to UTC+0).

Then we get messages like "...and avoid using a UTC timezone offset such as “America/Los_Angeles”. This just confuses folks.

Proposed fix: store the timezone string when we check it and reuse it in the notice.

📷 (good): http://p.tri.be/2lwzwW
📷 (good): http://p.tri.be/Vk0q4A
📷 (bad): http://p.tri.be/SoGWi9

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -215,6 +215,9 @@ Still not happy? Shoot us an email to support@theeventscalendar.com or tweet to 
 
 == Changelog ==
 
+= [TBD] TBD =
+* Tweak - Use the correct timezone in notice when on an event edit page.
+
 = [4.9.7] 2019-08-19 =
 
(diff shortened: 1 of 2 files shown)


sc0ttkclark:
I like this solution

Camwyn:
@bordoni what happened here (closed with no explanation)?

sc0ttkclark:
This probably auto-closed because B19.12 branch was deleted.

bordoni:
Sorry branch deleted, can we move it to B19.13?

PR #2692, opened 2019-09-04
remove quotes from TZID

zbtirrell:
See: https://central.tri.be/issues/133626


zbtirrell:
On hold because this needs to be repointed to release/B19.12 after B19.11 gets merged in.

PR #2705, opened 2019-09-11
Fix the use of `get_the_content` with a post ID

rossluebe:
Referencing the [get_the_content](https://developer.wordpress.org/reference/functions/get_the_content/) docs, the post ID needs to be the third parameter. I've added in the default values for the first two parameters for this function.


rossluebe:
Just noticed this is related to #2438, either resolution would fix this issue.

bordoni:
Resolved this and another bug we have with two new method uses on Tribe Common.
`tribe_get_the_content()` and `tribe_the_content()` which will make it cross compatible with multiple versions of PHP and WP. Released with Version `4.9.12`.

PR #2716, opened 2019-09-18
Fix/133898 datepicker pro code

paulmskim:
https://central.tri.be/issues/133898

https://www.loom.com/share/351b882414ec45e886bea0b08dacf42b

### DIFF
--- a/src/resources/js/views/datepicker.js
+++ b/src/resources/js/views/datepicker.js
@@ -40,18 +40,10 @@ tribe.events.views.datepicker = {};
 	 */
 	obj.selectors = {
 		datepickerContainer: '[data-js="tribe-events-top-bar-datepicker-container"]',
-		datepickerDays: '.datepicker-days',
 		datepickerDaysBody: '.datepicker-days tbody',
-		datepickerDaysRow: '.datepicker-days tbody tr',
-		datepickerDay: '.day',
-		datepickerDayNotDisabled: '.day:not(.disabled)',
 		input: '[data-js="tribe-events-top-bar-date"]',
 		button: '[data-js="tribe-events-top-bar-datepicker-button"]',
 		buttonOpenClass: '.tribe-events-c-top-bar__datepicker-button--open',
-		activeClass: '.active',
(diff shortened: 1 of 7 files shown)


src/resources/js/views/datepicker.js
```diff
@@ -65,6 +57,39 @@ tribe.events.views.datepicker = {};
...
+	 *
+	 * @since TBD
+	 *
+	 * @type {PlainObject}
+	 */
+	obj.options = {
+		container: null,
+		daysOfWeekDisabled: [],
+		/**
+		 * @todo: use format from BE. Paul.
```
bordoni: Lets start using the `@` from github on these.

src/resources/js/views/datepicker.js
```diff
@@ -363,61 +314,49 @@ tribe.events.views.datepicker = {};
...
+		// set up mutation observer
+		obj.observer = new MutationObserver( obj.handleMutation( { container: $container } ) );
+
+		// set options for datepicker
+		obj.options.container = $container.find( obj.selectors.datepickerContainer );
+		obj.options.minViewMode = isMonthView ? 'year' : 'month';
+		var tribeL10nDatatables = window.tribe_l10n_datatables || {};
+		var datepickerI18n = tribeL10nDatatables.datepicker || {};
+		var nextText = datepickerI18n.nextText || 'Next';
+		var prevText = datepickerI18n.prevText || 'Prev';
```
bordoni: No need for this extra string here. It should always come from the localize var.

paulmskim: You're right, we should always have this available, but the 2 lines above I've set empty objects if in the off case `tribe_l10n_datatables` does not exist. So `nextText` and `prevText` from `datepickerI18n` object will be undefined. `Next` and `Prev` are fallbacks.

src/resources/js/views/datepicker.js
paulmskim: copied these from other places, but will add.

PR #2725, opened 2019-09-23
Revert change to tribe_get_venue_details() since we are no longer using it

bordoni:
_Ref:_ [C#134532](http://central.tri.be/issues/134532)
---


bordoni:
Going to merge, since the tests are a false positive we have from _I think_ `tribe-common` not syncing correctly.

PR #2741, opened 2019-10-07
VR > Filter excerpt length for list and day views.

juanfra:
🎫 https://central.tri.be/issues/131088

🎥 https://www.loom.com/share/a81a41a0447247638decdd51a7ccc574

### DIFF
--- a/src/Tribe/Views/V2/Template/Excerpt.php
+++ b/src/Tribe/Views/V2/Template/Excerpt.php
@@ -19,13 +19,6 @@ namespace Tribe\Events\Views\V2\Template;
 class Excerpt {
 
 	/**
-	 * Excerpt constructor.
-	 *
-	 * @since TBD
-	 */
-	public function __construct() {}
-
-	/**
 	 * Filters the excerpt length.
 	 *
 	 * Set the excerpt length for list and day view.
@@ -40,11 +33,7 @@ class Excerpt {
 		$context = tribe_context();
 		$view = $context->get( 'event_display_mode', 'list' );
 
-		if ( 'list' === $view || 'day' === $view ) {
-			$length = 30;
-		}
-
-		return $length;
+		return in_array( $view, [ 'list', 'day' ] ) ? 30 : $length;
 	}
 
 	/**
--- a/src/Tribe/Views/V2/Template/Excerpt.php
+++ b/src/Tribe/Views/V2/Template/Excerpt.php
@@ -8,6 +8,8 @@
  */
 
 namespace Tribe\Events\Views\V2\Template;
+use Tribe__Template as Base_Template;
+use Tribe__Events__Main as Plugin;
 
 /**
  * Class Excerpt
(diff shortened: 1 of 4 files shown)


src/Tribe/Views/V2/Template/Excerpt.php
```diff
@@ -0,0 +1,79 @@
...
+ * @package Tribe\Events\Views\V2\Template
+ */
+class Excerpt {
+
+	/**
+	 * Excerpt constructor.
+	 *
+	 * @since TBD
+	 */
+	public function __construct() {}
```
bordoni: Why have a construct here?

src/Tribe/Views/V2/Template/Excerpt.php
```diff
@@ -0,0 +1,79 @@
...
+	 *
+	 * @param int $length The excerpt length.
+	 *
+	 * @return int The excerpt length modified, if necessary.
+	 */
+	public function maybe_filter_excerpt_length( $length ) {
+		$context = tribe_context();
+		$view = $context->get( 'event_display_mode', 'list' );
+
+		if ( 'list' === $view || 'day' === $view ) {
```
bordoni: use `in_array()` and return 30 directly, avoids the person having to read the whole method to get to the return.

src/Tribe/Views/V2/Template/Excerpt.php
```diff
@@ -0,0 +1,79 @@
...
+
+		$template = strtolower( get_template() );
+
+		// Check if theme is twentyseventeen.
+		if ( ! $template || 'twentyseventeen' !== $template ) {
+			return $link;
+		}
+
+		$link  = '<div class="tribe-events-c-small-cta tribe-common-b3 tribe-events-c-small-cta--readmore">';
+		$link .= '<a href="' . esc_url( get_permalink( get_the_ID() ) ) . '" class="tribe-events-c-small-cta__link tribe-common-cta tribe-common-cta--thin-alt">';
```
bordoni: This should be loaded via a view file.

src/Tribe/Views/V2/Template/Excerpt.php
juanfra: I thought it would be better for readability if it was this way.

src/Tribe/Views/V2/Template/Excerpt.php
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		$template = strtolower( get_template() );
+
+		// Check if theme is twentyseventeen.
+		if ( ! $template || 'twentyseventeen' !== $template ) {
+			return $link;
+		}
+
+		return $this->template( 'components/read-more', [], false );
```
bordoni: We should pass down to the template `'link' => $link` and instead of using `get_the_ID()` on the actual template we should be passing `'event' => tribe_get_event( get_the_ID() )`, so it more closely matches the other templates.

Don't forget to add these two on the docblock of the new view.

juanfra: Thanks! I don't understand why is that we should be sending the $link variable. That variable contains what twentyseventeen set as a string for the whole "continue reading". And what we're trying to do here is basically override that with our template. We don't need any info from that variable in the template: https://github.com/WordPress/twentyseventeen/blob/master/functions.php#L222

juanfra: And actually, if we use tribe_get_event there it end up in a fatal.

bordoni: What I meant is that we need to pass into the template the variables in case someone wants to use that in the template. Why do we end up with a fatal? Does it trigger an infinite loop? if yes we need to make sure it doesn't happen, our goal is to make the templates simpler.

PR #2739, opened 2019-10-04
VR > BE > Date header format in mobile views

juanfra:
🎫 https://central.tri.be/issues/134525
🎥 https://www.loom.com/share/57d0b5e7f4db48c3a562b11e586e507f

PR #2744, opened 2019-10-08
Feature/v2 filter bar

paulmskim:
https://central.tri.be/issues/133523
https://central.tri.be/issues/133588
https://central.tri.be/issues/134133
https://central.tri.be/issues/134292
https://central.tri.be/issues/134295
https://central.tri.be/issues/134296


src/Tribe/Views/V2/Views/Month_View.php
```diff
@@ -309,17 +309,18 @@ static function ( $event ) use ( $date_object ) {
...
-				'multiday_events'  => $day_stack,
-				'found_events'     => $day_found_events,
-				'more_events'      => $more_events,
-				'day_url'          => $day_url,
+				'date'              => $day_date,
+				'is_start_of_week'  => $start_of_week === $date_object->format( 'N' ),
+				'year_number'       => $date_object->format( 'Y' ),
+				'month_number'      => $date_object->format( 'm' ),
+				'day_number'        => $date_object->format( 'd' ),
+				'day_number_no_pad' => $date_object->format( 'j' ),
```
juanfra: I'm not sure about having these two, I'd maybe keep just the old one, but with the `j` format. One question: Is there any reason why you defined the new one instead of modifying the old? I think the leading 0 wouldn't affect any calculation or anything that could be taking place.

paulmskim: I'll need to check again, but `'day_number'` is used in the `<time>` element for the `datetime` attribute. If I can get rid of the month and day number and just go with a `yyyy-mm-dd` date format, then I'll do that.

juanfra: Awesome, I'm adding a new var in all views for `date_formats` - Maybe if you need any format in particular (to generate the meta) you could use that -> https://github.com/moderntribe/the-events-calendar/pull/2739/files#diff-828c168861c5b8f4cb8331c20c65d55aR1001

PR #2750, opened 2019-10-14
Fixed PHP 7 compatibility issue: Use of func_get_args() in Tribe__Eve…

JacobMC:
Fixed PHP 7 compatibility issue: Use of func_get_args() in Tribe__Events__Templates::getTemplateHierarchy

Original Error:
> Since PHP 7.0, functions inspecting arguments, like func_get_args(), no longer report the original value as
>                 passed to a parameter, but will instead provide the current value. The parameter "$args" was changed on line
>                 546. (PHPCompatibility.FunctionUse.ArgumentFunctionsReportCurrentValue.Changed)

### DIFF
--- a/src/Tribe/Templates.php
+++ b/src/Tribe/Templates.php
@@ -543,8 +543,8 @@ if ( ! class_exists( 'Tribe__Events__Templates' ) ) {
 		 **/
 		public static function getTemplateHierarchy( $template, $args = array() ) {
 			if ( ! is_array( $args ) ) {
+				$passed        = array( $template, $args );
 				$args          = array();
-				$passed        = func_get_args();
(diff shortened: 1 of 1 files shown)


bordoni:
CTL/137878

src/Tribe/Templates.php
```diff
@@ -543,8 +543,8 @@ public static function showInLoops( $query ) {
 		 **/
 		public static function getTemplateHierarchy( $template, $args = array() ) {
 			if ( ! is_array( $args ) ) {
+				$passed        = array( $template, $args );
```
sc0ttkclark, line 546: This does not seem to be backwards compatible with the `$backwards_map` usage below where it still supports other arguments coming from `func_get_args()`

PR #2755, opened 2019-10-16
Feature/135512 tec base

paulmskim:
https://central.tri.be/issues/135512


src/Tribe/Views/V2/Assets.php
```diff
@@ -40,13 +40,29 @@ class Assets extends \tad_DI52_ServiceProvider {
...
+		);
+
 		tribe_asset(
 			$plugin,
 			'tribe-events-views-v2-full',
 			'views-full.css',
 			[
-				'tribe-common-style',
-				'tribe-tooltipster-css',
+				'tribe-common-full-style',
```
juanfra: I assume we're working on the option for skeleton vs full in another ticket?

paulmskim: yeah, that's a BE thing I'll have to connect with G on

PR #2761, opened 2019-10-17
VR > Messages > Markup & Styles

juanfra:
🎫 https://central.tri.be/issues/133311
🎥 https://www.loom.com/share/247cbf75511e427f80ae72d4e7dc5acd


src/views/v2/components/messages.php
```diff
@@ -20,16 +20,17 @@
...
-<div class="tribe-events-notices" style="text-align: center; width: 100%; margin: 0 0 1em auto;">
-	<?php
-	foreach ( $messages as $message_type => $message_group ) {
-		foreach ( $message_group as $message ) {
-			echo wp_kses_post( "{$message_type}: $message" );
-		}
-	}
-	?>
+<div <?php tribe_classes( $classes ); ?>>
+	<?php foreach ( $messages as $message_type => $message_group ) : ?>
```
ryanurban: Semantically maybe we should updated this to be a `<ul>` & set of `<li>`s, thoughts?

src/views/v2/components/messages.php
```diff
@@ -20,16 +20,17 @@
...
 
-<div class="tribe-events-notices" style="text-align: center; width: 100%; margin: 0 0 1em auto;">
-	<?php
-	foreach ( $messages as $message_type => $message_group ) {
-		foreach ( $message_group as $message ) {
-			echo wp_kses_post( "{$message_type}: $message" );
-		}
-	}
-	?>
+<div <?php tribe_classes( $classes ); ?>>
```
ryanurban: Let's add `role="alert"` to this outer wrapper ⚡️

tests/views_integration/Tribe/Events/Views/V2/Partials/Components/MessagesTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+class MessagesTest extends HtmlPartialTestCase
+{
+
+	protected $partial_path = 'components/messages';
+
+	/**
+	 * Test render with context
+	 */
+	public function test_render_with_context() {
```
paulmskim: let's test the loop by writing a test for multiple notices as well as multiple message groups.

src/resources/postcss/components/_messages.pcss
```diff
@@ -0,0 +1,52 @@
...
+		order: -1;
+
+		@media (--viewport-medium) {
+			order: 3;
+			padding: var(--spacer-3);
+			margin: 0 0 var(--spacer-7) auto;
+		}
+	}
+
+	.tribe-events-c-messages__inner {
```
paulmskim: do we know what this will look like with multiple messages? do we need spacing below each message?

tests/views_integration/Tribe/Events/Views/V2/Partials/Components/MessagesTest.php
bordoni: I like that idea.

tests/views_integration/Tribe/Events/Views/V2/Partials/Components/MessagesTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+		$messages = [
+			'notice' => [
+				'There were no results found',
+			],
+		];
+
+		$this->assertMatchesSnapshot( $this->get_partial_html( [ 'messages' => $messages ] ) );
+	}
+
```
bordoni: We should also test that the Notices accept HTML.

src/resources/postcss/components/_messages.pcss
juanfra: No, so far the designs are for one message only. But the data structure allows more than one. That's why I did it this way.

paulmskim: If we're going to be using lists and foreach loops, we should be accounting for this in the styles as well.

juanfra:
👋 hey y'all, worked on the suggestions. This is ready for a new CR.

https://www.loom.com/share/ab7f123e327342a8a1834ffc9e0cb7a6

src/resources/postcss/components/_messages.pcss
```diff
@@ -0,0 +1,53 @@
...
+	 *
+	 * ----------------------------------------------------------------------------- */
+
+	.tribe-events-c-messages {
+		align-items: center;
+		background-color: var(--color-background-messages);
+		border-radius: var(--border-radius-default);
+		display: flex;
+		margin: 0 0 var(--spacer-3) auto;
+		order: -1;
```
paulmskim: Let's move the message HTML above the events bar so we don't have to set order twice, once on mobile and again on desktop. It's a good idea to try to mess with order as little as possible.

src/views/v2/components/messages.php
```diff
@@ -20,16 +20,16 @@
...
-	<?php
-	foreach ( $messages as $message_type => $message_group ) {
-		foreach ( $message_group as $message ) {
-			echo wp_kses_post( "{$message_type}: $message" );
-		}
-	}
-	?>
-</div>
-
+<?php foreach ( $messages as $message_type => $message_group ) : ?>
```
paulmskim: I feel we should have a wrapper div for all these messages, so we can control all the messages with one div, what do you think?

src/resources/postcss/components/_messages.pcss
```diff
@@ -0,0 +1,53 @@
...
+	 * </div>
+	 *
+	 * ----------------------------------------------------------------------------- */
+
+	.tribe-events-c-messages {
+		align-items: center;
+		background-color: var(--color-background-messages);
+		border-radius: var(--border-radius-default);
+		display: flex;
+		margin: 0 0 var(--spacer-3) auto;
```
paulmskim: curious why we have margin auto on left but 0 on the right?

src/resources/postcss/components/_messages.pcss
```diff
@@ -0,0 +1,53 @@
...
+		order: -1;
+		padding: var(--spacer-2);
+		width: 100%;
+
+		@media (--viewport-medium) {
+			justify-content: center;
+			margin: 0 0 var(--spacer-7) auto;
```
paulmskim: same question on margin

PR #2768, opened 2019-10-19
Fix event REST endpoint which prevented events from appearing

maxm123:
As I described [here](https://wordpress.org/support/topic/events-created-via-rest-api-do-not-show-up-in-list-of-upcoming-events/#post-12045755), there is a difference in handling events 'EventHideFromUpcoming' field between the REST API and the front end. While the latter only checks for its existence, the former works with 'yes' and an empty string. This results in events being created with 'EventHideFromUpcoming' set to an empty string by the REST API which is the reason why it doesn't appear on the web page.

Now, it only sets 'EventHideFromUpcoming' if the value is meant to represent 'yes'.

### DIFF
--- a/src/Tribe/REST/V1/Endpoints/Single_Event.php
+++ b/src/Tribe/REST/V1/Endpoints/Single_Event.php
@@ -622,7 +622,9 @@ class Tribe__Events__REST__V1__Endpoints__Single_Event
 		// Event presentation data
 		$postarr['EventShowMap']          = tribe_is_truthy( $request['show_map'] );
 		$postarr['EventShowMapLink']      = tribe_is_truthy( $request['show_map_link'] );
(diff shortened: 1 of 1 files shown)


lucatume:
Rebasing on 19.16

bordoni:
CTL/137880

maxm123:
What happened to my PR? I only see that it was closed.

lucatume:
@maxm123 Uhn, it looks like this was closed by mistake, thanks for pointing it out.

I've re-based the PR on the current working code and wrapped it to add the required accessory code: https://github.com/moderntribe/the-events-calendar/pull/3066

This PR will be closed, but the code "will live on", thanks!

maxm123:
Ok, thank you for taking care of it.

PR #2783, opened 2019-10-24
VR > Messages > Skeleton & full

juanfra:
🎫 https://central.tri.be/issues/133311
🎥 https://www.loom.com/share/247cbf75511e427f80ae72d4e7dc5acd

### DIFF
--- a/src/resources/postcss/components/skeleton/_messages.pcss
+++ b/src/resources/postcss/components/skeleton/_messages.pcss
@@ -1,0 +1,45 @@
+.tribe-events {
+	/* -----------------------------------------------------------------------------
+	 *
+	 * Messages
+	 *
+	 * Example:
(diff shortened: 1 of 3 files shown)


src/resources/postcss/components/skeleton/_messages.pcss
```diff
@@ -0,0 +1,44 @@
...
+
+		@media (--viewport-medium) {
+			justify-content: center;
+			margin: 0 0 var(--spacer-7) 0;
+			padding: var(--spacer-3);
+		}
+	}
+
+	.tribe-events-c-messages__message-list {
+		flex-direction: column;
```
paulmskim: are we setting this as `display: flex` somewhere? if not, let's remove this.

PR #2791, opened 2019-10-29
VR > Read more QA

juanfra:
🎫 https://central.tri.be/issues/131089
🎥 https://www.loom.com/share/f94677f418b048a8b0ed731b0c825177

- Making sure we have the hellip.
- Fixing size and spacing.


src/Tribe/Views/V2/Template/Excerpt.php
```diff
@@ -74,7 +74,7 @@ public function maybe_filter_excerpt_more( $link ) {
 			return $link;
 		}
 
-		return $this->template( 'components/read-more', [], false );
+		return ' &hellip; ' . $this->template( 'components/read-more', [], false );
```
paulmskim: Can we wrap this in a span and put it in the template rather than here?

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
  *
  * @link {INSERT_ARTCILE_LINK_HERE}
  *
- * @version 4.9.10
+ * @version TBD
  *
  */
```
tr1b0t: * :no_entry_sign: There must be exactly one blank line after the file comment
* :no_entry_sign: Missing @package tag in file comment
* :no_entry_sign: Additional blank lines found at end of doc comment

<sub>hash: 8c2446a4f511b5c6f79df7515a6d10db</sub>

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
...
 ?>
+<span class="tribe-events-c-small-cta-hellip"> &hellip; </span>
 <div class="tribe-events-c-small-cta tribe-common-b3 tribe-events-c-small-cta--readmore">
 	<a
 		href="<?php echo esc_url( get_permalink( get_the_ID() ) ) ?>"
```
tr1b0t: :no_entry_sign: Inline PHP statement must end with a semicolon

<sub>hash: 84e29842ddb71c7d549fd8cde9aa8b31</sub>

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
...
 		class="tribe-events-c-small-cta__link tribe-common-cta tribe-common-cta--thin-alt"
-	>
-		<?php esc_html_e( 'Continue Reading' , 'the-events-calendar' ); ?>
-	</a>
+	><?php esc_html_e( 'Continue Reading' , 'the-events-calendar' ); ?></a>
```
tr1b0t: :no_entry_sign: Space found before comma in function call

<sub>hash: 47f74ca234a723b96dfbb2599797a970</sub>

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
...
 </div>
```
tr1b0t: :no_entry_sign: File must end with a newline character

<sub>hash: b434c0fcb472f42bc5d829c74860e42c</sub>

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
  *
  * @link {INSERT_ARTCILE_LINK_HERE}
  *
- * @version 4.9.10
- *
+ * @version TBD
  */
```
tr1b0t: :no_entry_sign: Missing @package tag in file comment

<sub>hash: 9ab9f078f4103409c69a479dc2c4a966</sub>

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
...
+
 ?>
+<span class="tribe-events-c-small-cta-hellip"> &hellip; </span>
```
paulmskim: since this doesn't really have anything to do with the small cta component, I think the class name should be something related to the read more component.

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
...
 <div class="tribe-events-c-small-cta tribe-common-b3 tribe-events-c-small-cta--readmore">
```
paulmskim: 2 notes here:
1. Do we have styles for `tribe-events-c-small-cta--readmore`? If not, we probably shouldn't have that class.
2. Since this is a read more component, it should have a read more component class (like `tribe-events-c-read-more` or something similar).

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
...
 	<a
-		href="<?php echo esc_url( get_permalink( get_the_ID() ) ) ?>"
+		href="<?php echo esc_url( get_permalink( get_the_ID() ) ); ?>"
```
paulmskim: Shouldn't this be getting the ID from the `$event` object rather than the `get_the_ID()` template tag?

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,14 @@
  *
  * @link {INSERT_ARTCILE_LINK_HERE}
  *
- * @version 4.9.10
- *
+ * @version TBD
```
paulmskim: Do we have tests for this template? If not, can we add tests?

src/views/v2/components/read-more.php
juanfra: Yes we have https://github.com/moderntribe/the-events-calendar/blob/release/B19.14/src/resources/postcss/components/skeleton/_small-cta.pcss#L20

src/views/v2/components/read-more.php
juanfra: No, because we are not sending the `$event` there, given that if we get it, it would give a fatal: https://github.com/moderntribe/the-events-calendar/pull/2741#discussion_r333066555

src/views/v2/components/read-more.php
juanfra: I can add a snapshot test for this template in particular. But it'll be only loaded for twentyseventeen, given that we're hooking on top of their filter, to make the read more link look like what they want on the design: https://github.com/WordPress/twentyseventeen/blob/master/functions.php#L222

src/views/v2/components/read-more.php
juanfra: 2. I treated this as a cta component (has the same styles) but with a modifier given that requires different spacing as a variation of the cta (it's a read more).

src/views/v2/components/read-more.php
paulmskim: We should be adding snapshots for every template we have, regardless of whether it's used in one theme or all the themes we support.

src/views/v2/components/read-more.php
paulmskim: So I never got a chance to review https://github.com/moderntribe/the-events-calendar/pull/2741, but if we're calling this template a read more component, we should be naming it as such. Not all of our CSS components have a corresponding template, just a template structure we use. Some are just styles we reuse throughout the code, much like `tribe-common-h1`, for example. So if we're reusing the `--read-more` modifier for the small CTA in other places, I'd agree. But since it's specific to the read more component, I'd be more inclined to put the style on the read more component class.

src/views/v2/components/read-more.php
paulmskim: I see what we're hooking into from https://github.com/moderntribe/the-events-calendar/pull/2741. Instead, can we do the `get_permalink( get_the_ID() )` in the hook and pass down a `$link` variable? We want to maintain consistency with our other templates. Ideally we would want to pass down `$event` too, but seems like we can't. If users want to modify the read more template, they still have access to the template tags.

src/views/v2/components/read-more.php
bordoni: We should be adding tests for all templates for sure.

src/views/v2/components/read-more.php
bordoni: Yes I agree on that, we should be passing down either the Event Object or at least a link.

src/views/v2/components/read-more.php
juanfra: Well, I see it differently. Since we're using this template only to override the twentyseventeen read more button and make it look like the CTA but with the difference of the margin, I though it would be pointless to define a new component, and it would be way more efficient to define a modifier, as it is a really really special case.

src/views/v2/components/read-more.php
juanfra: As we discussed before, it not possible to pass the event object here.

src/Tribe/Views/V2/Template/Excerpt.php
```diff
@@ -74,7 +74,7 @@ public function maybe_filter_excerpt_more( $link ) {
 			return $link;
 		}
 
-		return $this->template( 'components/read-more', [], false );
+		return $this->template( 'components/read-more', [ 'permalink' => get_permalink( get_the_ID() ) ], false );
```
bordoni:
```suggestion
		return $this->template( 'components/read-more', [ 'permalink' => get_permalink( get_the_ID() ) ], false );
```

src/views/v2/components/read-more.php
```diff
@@ -9,15 +9,16 @@
...
+ * @var string $permalink The permalink.
  */
+
 ?>
-<div class="tribe-events-c-small-cta tribe-common-b3 tribe-events-c-small-cta--readmore">
+<span class="tribe-events-c-read-more-hellip"> &hellip; </span>
+<div class="tribe-events-c-small-cta tribe-common-b3 tribe-events-c-read-more">
 	<a
-		href="<?php echo esc_url( get_permalink( get_the_ID() ) ) ?>"
+		href="<?php echo esc_url( $permalink ); ?>"
```
bordoni:
```suggestion
		href="<?php echo esc_url( $event->permalink ); ?>"
```

src/Tribe/Views/V2/Template/Excerpt.php
```diff
@@ -74,7 +74,7 @@ public function maybe_filter_excerpt_more( $link ) {
 			return $link;
 		}
```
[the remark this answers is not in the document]
juanfra: @bordoni we already discussed this here https://github.com/moderntribe/the-events-calendar/pull/2741#discussion_r333855348 and it's not possible to do that.

src/views/v2/components/read-more.php
juanfra: @bordoni we already discussed this here https://github.com/moderntribe/the-events-calendar/pull/2741#discussion_r333855348 and it's not possible to do that.

src/Tribe/Views/V2/Template/Excerpt.php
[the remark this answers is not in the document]
bordoni: Oh yeah that is true, I forgot about that, I feel like what we could do to prevent that Fatal/Infinite loop by removing the filter right after including it.

Something along the lines of:

```
		remove_filter( 'excerpt_more', [ $this, 'filter_excerpt_more' ], 999 );
```

Note, also we should move away from `999` as the priority and use something like `100`.

PR #2793, opened 2019-10-30
Add get directions data

bordoni:
_Ref:_ [C#135288](http://central.tri.be/issues/135288)
---
We were just missing the Directions link, and we are using the old method via `tribe_get_event()`
📹 http://i.bordoni.me/bbe7b0165421


src/Tribe/Models/Post_Types/Venue.php
```diff
@@ -41,7 +43,8 @@ protected function build_properties( $filter ) {
 				'state'                 => $state,
 				'province'              => $province,
 				'zip'                   => $zip,
-				'permalink'             => get_permalink( $this->post->ID ),
+				'permalink'             => $permalink,
+				'directions_link'       => esc_url( $directions_link ),
```
borkweb: Is this escaping too early? Should we do `esc_url_raw()` instead? Or nothing here and escape late.

bordoni: Agreed.

juanfra: I think the method is already escaping the return.

PR #2794, opened 2019-10-30
updating contributor instructions

zbtirrell:


CONTRIBUTING.md
```diff
@@ -38,9 +46,9 @@ that with the npm `-g` flag.
...
 ```
 
-#### Install all the things
+#### Install all the NPM packages
 
-With all of those in place, simply run:
+With all of those in place, run:
 
 ```
 npm install
```
lucatume: It's worth mentioning we use version `8.9.4`.
If one does not have that (that's old) then `nodenv` should be used.

CONTRIBUTING.md
```diff
@@ -53,6 +61,27 @@ rebuild and try again.
...
+composer install
+```
+
+#### Install npm and composer packages in common
+
+One of our submodules is Tribe Common, which is a shared set of libraries in use by our plugins (see also [Event Tickets](https://github.com/moderntribe/event-tickets)). It uses npm and composer as well, so, let's get those as well.
+
+```
+cd common
+npm install
```
lucatume: Same note about `node` versions here.

CONTRIBUTING.md
```diff
@@ -53,6 +61,27 @@ rebuild and try again.
...
+composer install
```
lucatume: `common` has sub-modules too:
```suggestion
composer install
git submodule update --init --recursive
```

zbtirrell: the --recursive should cover that though, right?

borkweb: Yup, it will. We don't need the `git submodule update --init --recursive` here.

PR #2798, opened 2019-10-31
Fix/136456 datepicker issue

paulmskim:
https://central.tri.be/issues/136456


bordoni:
Wait for tests and we golden.

PR #2799, opened 2019-10-31
Fix allowing admin event list to be ordered by post date

wullaski:
This is a fix for the admin event list which allows for explicitly ordering by post date. Default ordering will still be ordered by event start date as will ordering explicitly by any other column.


bordoni:
CTL/137882

PR #2814, opened 2019-11-06
Dont disable V1 on single event views

bordoni:
_Ref:_ [C#136503](http://central.tri.be/issues/136503)
---
Super simple change that was causing regressions, due to how agressive the V1 disabling was happening.

📹 http://i.bordoni.me/2cc7af256998

### DIFF
--- a/src/Tribe/Views/V2/Template_Bootstrap.php
+++ b/src/Tribe/Views/V2/Template_Bootstrap.php
@@ -93,6 +93,17 @@ class Template_Bootstrap {
 			: tribe( Template\Event::class );
 	}
 
+	public function is_single_event() {
+		$conditions = [
+			is_singular( TEC::POSTTYPE ),
+			'single-event' === tribe_context()->get( 'view' ),
+		];
+
+		$is_single_event = in_array( true, $conditions );
+
+		return $is_single_event;
+	}
+
 	/**
 	 * Fetches the HTML for the Single Event page using the legacy view system
 	 *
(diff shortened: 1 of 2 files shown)


src/Tribe/Views/V2/Template_Bootstrap.php
```diff
@@ -93,6 +93,25 @@ public function get_template_object() {
 			: tribe( Template\Event::class );
 	}
 
+	/**
+	 * Detemines wether we are in a Single event page or not,
+	 * base only on global context.
+	 *
+	 * @since  TBD
+	 *
+	 * @return bool
```
lucatume: Add return value description:
```suggestion
	 * @return bool Whether the current request is for the single event template or not.
```

src/Tribe/Views/V2/Template_Bootstrap.php
```diff
@@ -93,6 +93,25 @@ public function get_template_object() {
...
+	 */
+	public function is_single_event() {
+		$conditions = [
+			is_singular( TEC::POSTTYPE ),
+			'single-event' === tribe_context()->get( 'view' ),
+		];
+
+		$is_single_event = in_array( true, $conditions );
```
lucatume: Just use an `||` here and above? It's just 2 conditions.
Do you foresee more conditions being added?

bordoni: Yeah, my intention is to allow filtering + adding more conditionals. 
Something I am logging for after beta, with a couple other Single/Bootstrap fixes.

PR #2818, opened 2019-11-06
Move filtering of which views from registered to publicly available

bordoni:
_Ref:_ [C#136962](http://central.tri.be/issues/136962)
---
[Screencap](https://drive.google.com/open?id=136rDSulQyggBNbdImZhntodFS_9m_xqn&authuser=luca@tri.be&usp=drive_fs)


lucatume:
I've refactored this PR to remove the need, while checking for publicly visible Views, to instantiate Views (and all the filtering involved) to... not show the View.
I've made the `View::publicly_visible` property static and updated all its usages and its accessor method.

PR #2828, opened 2019-11-09
Add new components for the After/Before settings

bordoni:
_Ref:_ [C#136982](http://central.tri.be/issues/136982)
---

Added components for the After and Before settings.
📹 http://i.bordoni.me/5b2cfdde1945


paulmskim:
can we add tests for this component, testing both cases of with and without content?

PR #2831, opened 2019-11-11
Respect Featured params + Remove tax archive notice

bordoni:
_Ref:_ [C#135490](http://central.tri.be/issues/135490)
---
I found a notice when visiting the Category archives while I was fixing Featured param on shortcodes. So these changes are to support those fixes.

📹 https://i.bordoni.me/4gu78LWj

### DIFF
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -1483,16 +1483,18 @@ class View implements View_Interface {
 			// Set up breadcrumbs for category
 			if ( ! empty( $term_slug ) ) {
 				$term  = get_term_by( 'slug', $term_slug, $taxonomy );
-				$label = $term->name;
+				if ( ! empty( $term->name ) ) {
+					$label = $term->name;
 
-				$breadcrumbs[] = [
-					'link'  => $this->get_today_url( true ),
-					'label' => tribe_get_event_label_plural(),
-				];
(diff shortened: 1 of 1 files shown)


src/Tribe/Views/V2/View.php
```diff
@@ -603,6 +603,10 @@ public function get_url( $canonical = false ) {
 			'tribe-bar-search' => $this->context->get( 'keyword', '' ),
 		];
 
+		if ( tribe_is_truthy( $this->context->get( 'featured', false ) ) ) {
```
borkweb: Should we assign the results of this `tribe_is_truthy()` so we don't need to execute that and the `context->get()` twice? Like:

```php
if ( $is_featured = tribe_is_truthy( $this->context->get( 'featured', false ) ) ) {
```

src/Tribe/Views/V2/View.php
```diff
@@ -983,6 +987,11 @@ protected function setup_repository_args( \Tribe__Context $context = null ) {
 			$args[ TEC::TAXONOMY ] = $context_arr[ TEC::TAXONOMY ];
 		}
 
+		// Setup featured only when set to true.
+		if ( tribe_is_truthy( $context->get( 'featured', false ) ) ) {
```
borkweb: Same comment as the one above on reducing duplicate calls.

PR #2836, opened 2019-11-12
VR > Swap `&mdash;` for hyphen

juanfra:
🎫 https://central.tri.be/issues/135003
🎥 https://www.loom.com/share/cf5d0a3f56bd420a8191ece91f800ac9


src/Tribe/Views/V2/View.php
```diff
@@ -1081,8 +1081,10 @@ protected function setup_template_vars() {
 			'prev_label'           => $this->get_link_label( $this->prev_url( false ) ),
 			'next_label'           => $this->get_link_label( $this->next_url( false ) ),
 			'date_formats'         => (object) [
-				'compact'        => Dates::datepicker_formats( tribe_get_option( 'datepickerFormat' ) ),
-				'month_and_year' => tribe_get_date_option( 'monthAndYearFormat', 'F Y' ),
+				'compact'              => Dates::datepicker_formats( tribe_get_option( 'datepickerFormat' ) ),
+				'month_and_year'       => tribe_get_date_option( 'monthAndYearFormat', 'F Y' ),
+				'time_range_separator' => tribe_get_date_option( 'timeRangeSeparator', ' - ' ),
+				'date_time_separator'  => tribe_get_date_option( 'dateTimeSeparator', ' @ ' ),
```
paulmskim: question because I don't know, do the separators always have the spaces around them, or are they just the `-` or `@` characters?

paulmskim: I'm also not sure if this is the right place to put this, as this technically isn't a date format. @bordoni thoughts?

src/views/v2/list/top-bar/datepicker.php
```diff
@@ -46,7 +46,7 @@ class="tribe-events-c-top-bar__datepicker-time"
 			</time>
 		<?php endif; ?>
 		<?php if ( $show_end ) : ?>
-			<span class="tribe-events-c-top-bar__datepicker-separator">&mdash;</span>
+			<span class="tribe-events-c-top-bar__datepicker-separator"> - </span>
```
paulmskim: this should be the separator var

src/Tribe/Views/V2/View.php
juanfra: Default values are with spaces: https://github.com/moderntribe/the-events-calendar/search?q=dateTimeSeparator&unscoped_q=dateTimeSeparator

src/views/v2/list/top-bar/datepicker.php
juanfra: I didn't change that and wanted to ask design/strategy, because technically the option for the separator is for time range and not the datepicker.

juanfra: https://tribe.slack.com/archives/CEUHNUF3R/p1573755087278900?thread_ts=1573753404.276600&cid=CEUHNUF3R - We should keep the hyphen for the datepicker and use the option for dates.

src/Tribe/Views/V2/View.php
bordoni: I feel like either way is fine here. In the placement of the actual formatting. What do you think @lucatume ??
Any particular opinions?

lucatume: I think it's fine to have it here, while a bit stretched, the `date_formats` array could either contain only date formats, or loosely contain anything required to format dates.
I this case I think this is a good solution.

PR #2837, opened 2019-11-12
Prevent browser URL from been used as the base for Form

bordoni:
_Ref:_ [C#135491](http://central.tri.be/issues/135491)
---
https://i.bordoni.me/E0uPknrG


borkweb:
Code looks good. Needs an artifact.

PR #2844, opened 2019-11-14
VR - fix multi-day display on Month and Week view

lucatume:
Ticket: https://central.tri.be/issues/137222

[Screencap](https://drive.google.com/open?id=13VKbUFksath2jrf6cQDVA8DtztDk5U4a&authuser=luca@tri.be&usp=drive_fs)
[Screencap 2](https://drive.google.com/open?id=13YzA3ZsbUUJ5pc8u14-W4oQRgLv_DqyV&authuser=luca@tri.be&usp=drive_fs)

This PR fixes an issue that would prevent multi-day events from displaying correctly, or from showing at all.
See the screencast to have a before and after idea of what the issue is.

The issue cause is that the LIMIT to the number of fetched events per day applied to each day query in the Month and Week Views (really in the `By_Day_View` class) would "push" multi-day, non all-day events out of the results on the first day.

To make an example:

* on 2019-10-01 we have 3 events starting
	* at 6 AM (ID `23`)
	* at 8 AM (ID `89`)
	* at 10 AM - this is a multi-day event, ending 2 days later (ID `66`)
* the limit to events per day is 2

With the above situation we would get, as results for the day query, the events with IDs `23` and `89`, but not `66` as, in `ASC` date and time order, it starts after the others.

This PR fixes this by leveraging the fact we're in a View, Month or Week, that queries more than one day and a multi-day event might not be first on the day it starts, but **will be first on the following** day.

With this information we can "go back" and "append" the multi-day event ID, that will be then picked up by the Stack, to the previous day(s) results.


lucatume:
The "file noise" comes from snapshots I've updated as a result of an added class.

lucatume:
I've updated the code to fix a further issue dealing with multi-day events **starting on the last day of the period**.

lucatume:
Rebased on `bucket/views-redesign-beta-1`.

PR #2847, opened 2019-11-14
Month View performance tweak work

lucatume:
Ticket: https://central.tri.be/issues/136624

Based on: https://github.com/moderntribe/tribe-common/pull/1209

This PR **is step 1 of a work in progress**; it should be reviewed as such.

What this PR does: 

* add a `Query` class used by Month and Week View (by the parent `By_Day_View` class, really) to run 2 queries to fetch all events in a period.
* leverage the work done in the `common` PR (https://github.com/moderntribe/tribe-common/pull/1209) to improve timings and avoid useless computations.

The performance improvement is obtained by both reducing the number of queries and by using some of those queries a bit better, step by step:

1. Fetch all the events in a period (say a month grid); to find them we use start date and end date, since those values are already used by MySQL to filter the result, we can just return them.
2. The results of the query are save in an `Event_Result` value object. This is little than a wrapper around the array to make the object immutable and allow type-hinting.
3. The results are grouped by day from PHP, using the information of each result.
4. Each day result is saved in cache (mind that the implementation will benefit from persistent caching, but is optimized as if the only cachine available is per-request).
5. When required, each day results are read from the cache into an `Event_Result_Set`; this is a collection with *some* repository-like capabilities (more on this later) that operates on the results for the day *kinda like a query*.
6. Results can now be sorted, ordered and filtered by the View (currently Month or Week).
7. Templates expect `WP_Post` objects (decorated by the `tribe_get_event` function) for each event to show; so we pre-fill the cache for each post in one single query.
8. Views go on as normal.

## TODO:

I've left plenty of them in the code to mark where I see a missing feature or something we should discuss.

a. re: 1 and 7  the queries are currently unbound (just fetch all of them); these need to have filterable limits.
b. in this implementation the "period query" code (fetch all the events in a period) lives in a `Query` class. This should really live in the repository, see next paragraph.

## Repository integration

Re: TODO.b - I would like to move from the current implementation to this:

```php
<?php
$start = '2019-09-30 00:00:00';
$end = '2019-01-04 23:59:59';
// Just the vanilla event repository.
$events = tribe_events()
    // Fetch all events in the period.
    ->where( 'in_period', $start, $end )
    // Return the "midddle-ground" Event_Result_Set, not a full event, not just an ID.
    // Sets are grouped by day.
    ->get_sets();
```

I would like, then, to create a caching decorator and use it seamlessly.

```php
<?php
$start = '2019-09-30 00:00:00';
$end = '2019-01-04 23:59:59';
// The caching version of the event repository.
$events = tribe_events( 'caching' )
    // Fetch all events in the period.
    ->where( 'in_period', $start, $end )
    // Return the "midddle-ground" Event_Result_Set, not a full event, not just an ID.
    // Sets are grouped by day.
    ->get_sets();
```

The difference is that the repository will check on the cache before running any query and will hydrate it if required.

One could then operate on the sets as those allow:

```php
<?php
// $repository comes from the second snippet above.
$day_sets = $repository->get_sets();
foreach( $day_sets as $day_set ){
    $day_set->order_by( 'start_date', 'DESC');
    echo implode( PHP_EOL, $day_set->pluck( 'ID' );
}
```

I would like some feedback on the proposed API, please.


lucatume:
Tests are failing on a cache-related issue. I'm aware of that and looking into it.

src/Tribe/Views/V2/Query/Query.php
```diff
@@ -0,0 +1,225 @@
...
+	 * @since TBD
+	 *
+	 * @param \DateTimeInterface $start The start date of the period, correct time should be set already.
+	 * @param \DateTimeInterface $end The end date of the period, correct time should be set already.
+	 */
+	public static function update_period_cache( \DateTimeInterface $start, \DateTimeInterface $end ) {
+		global $wpdb;
+
+		$use_site_timezone = Timezones::is_mode( Timezones::SITE_TIMEZONE );
+
```
bordoni:
```suggestion
$meta_key_timezone = '_EventTimezone';
$meta_key_all_day = '_EventAllDay'; 
$post_type = TEC::POSTTYPE;
```

src/Tribe/Views/V2/Query/Query.php
```diff
@@ -0,0 +1,225 @@
...
+					INNER JOIN {$wpdb->postmeta} start_date 
+						ON (p.ID = start_date.post_id AND start_date.meta_key = %s)
+					WHERE p.post_type = 'tribe_events'
+					-- Starts before the period end.
+					AND start_date.meta_value <= %s
+				) start_date ON p.ID = start_date.ID
+				 INNER JOIN {$wpdb->postmeta} end_date 
+				 	ON (p.ID = end_date.post_id AND end_date.meta_key = %s)
+				 INNER JOIN {$wpdb->postmeta} timezone 
+				 	ON (p.ID = timezone.post_id AND timezone.meta_key = '_EventTimezone')
```
bordoni:
```suggestion
				 	ON (p.ID = timezone.post_id AND timezone.meta_key = '{$meta_key_timezone}')
```

src/Tribe/Views/V2/Query/Query.php
```diff
@@ -0,0 +1,225 @@
...
+				 -- LEFT JOIN to allow NULL post_id if meta key not set.
+				 LEFT JOIN {$wpdb->postmeta} all_day 
+				 	ON (p.ID = all_day.post_id AND all_day.meta_key = '_EventAllDay')
```
bordoni:
```suggestion
				 	ON (p.ID = all_day.post_id AND all_day.meta_key = '{$meta_key_all_day}')
```

src/Tribe/Views/V2/Query/Query.php
```diff
@@ -0,0 +1,225 @@
...
+
+		WHERE p.post_type = 'tribe_events'
```
bordoni:
```suggestion
		WHERE p.post_type = '{$post_type}'
```

src/Tribe/Views/V2/Query/Query.php
```diff
@@ -0,0 +1,225 @@
...
+							// Sanity check: break when the current day is equal to the event end date.
+							$reached_end = $d->format( Dates::DBDATEFORMAT ) === $end_date->format( Dates::DBDATEFORMAT );
+							if ( $reached_end ) {
+								break;
+							}
+						}
+					}
+
+					// Normalize the timezone to the site one.
+					$result['start_date'] = $start_date->setTimezone( $site_timezone )->format( 'Y-m-d H:i:s' );
```
bordoni:
```suggestion
					$result['start_date'] = $start_date->setTimezone( $site_timezone )->format( Dates::DBDATEFORMAT );
```

src/Tribe/Views/V2/Query/Query.php
```diff
@@ -0,0 +1,225 @@
...
+					$result['end_date']   = $end_date->setTimezone( $site_timezone )->format( 'Y-m-d H:i:s' );
```
bordoni:
```suggestion
					$result['end_date']   = $end_date->setTimezone( $site_timezone )->format( DBDATEFORMAT );
```

src/Tribe/Views/V2/Query/Query.php
```diff
@@ -0,0 +1,225 @@
...
+	 * `Event_Result` class.
+	 *
+	 * @since TBD
+	 *
+	 * @param \DateTimeInterface $start The start date of the period, correct time should be set already.
+	 * @param \DateTimeInterface $end The end date of the period, correct time should be set already.
+	 */
+	public static function update_period_cache( \DateTimeInterface $start, \DateTimeInterface $end ) {
+		global $wpdb;
+
```
bordoni: We should have a filter here to be able to bypass this whole method.

src/Tribe/Views/V2/Query/Query.php
```diff
@@ -0,0 +1,225 @@
...
+							$buffer[ $overlap_day ][] = new Event_Result( $result );
+						} else {
+							$buffer[ $overlap_day ] = [ new Event_Result( $result ) ];
+						}
+					}
+
+					return $buffer;
+				}, [] );
+		}
+
```
bordoni: Filter before setting to cache.

src/Tribe/Views/V2/Query/Events_Result_Set.php
```diff
@@ -0,0 +1,129 @@
...
+	 * {@inheritDoc}
+	 */
+	public function all() {
+		return $this->event_results;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function jsonSerialize() {
```
bordoni: Does it need to be `jsonSerialize` upper case is that because of the parent class?

src/Tribe/Views/V2/Query/Query.php
borkweb:
```suggestion
					$result['end_date']   = $end_date->setTimezone( $site_timezone )->format( Dates::DBDATEFORMAT );
```

src/Tribe/Views/V2/Views/By_Day_View.php
```diff
@@ -133,44 +135,45 @@ public function get_grid_days( $date = null, $force = false ) {
...
 			$day_string = $day->format( 'Y-m-d' );
 
-			if ( isset( $this->grid_days_cache[ $day_string ] ) ) {
-				continue;
+			$cache_key   = Query::get_cache_key( $day_string );
+			$day_results = Events_Result_Set::from_value(
+				(array) $cache->get( $cache_key, Cache_Listener::TRIGGER_SAVE_POST )
+			);
+
+			if ( ! $day_results->count() ) {
```
mitogh: Probably personal preference here but I think moving empty state as "default" state of the ids removes the need of the else block¿ and negation operation

``` php
$event_ids = [];

if ( $day_results->count() ) {
	$day_results->order_by( $order_by, $order );
	$event_ids = array_map( 'absint', $day_results->pluck( 'ID' ) );
}
```

src/Tribe/Views/V2/Query/Events_Result_Set.php
lucatume: Yes, it's the specification of the `jsonSerializable` interface.

src/Tribe/Views/V2/Query/Query.php
lucatume: This will be moved into the repository and that will be taken care of using the default extension system of the repository.
This would need to be filtered if it stayed like this, I agree.

src/Tribe/Views/V2/Query/Query.php
lucatume: Once this goes into a `caching` repository implementation, the default repository extension and filtering system will apply.

src/Tribe/Views/V2/Views/By_Day_View.php
lucatume: Good suggestion, thanks.

lucatume:
I've updated this PR after [updates on `common`](https://github.com/moderntribe/tribe-common/pull/1224).

The changes:

* kill the `Query` class
* move the period fetching code to the `Event_Period` repository; the post cache warmup part is in `common`, see [the PR](https://github.com/moderntribe/tribe-common/pull/1224)
* draft the minimum req. code to make the `Event_Period` repository releasable; thus the high number of TODOs in the code
* add tests for the `Event_Period` repository.

PR #2849, opened 2019-11-14
Broswer Navigation management

bordoni:
_Ref:_ [C#137312](http://central.tri.be/issues/137312)
---

📹 https://i.bordoni.me/7KuxllNn


bordoni:
I think this is another false positive on the tests.

src/resources/js/views/manager.js
```diff
@@ -291,6 +306,42 @@ tribe.events.views.manager = {};
...
+		var shouldManageUrl = obj.shouldManageUrl( $container );
+
+		var data = {
+			url: url,
+			_wpnonce: nonce
+		};
+
+		obj.request( data, $container );
+
+		$container.trigger( 'afterPopState.tribeEvents', event );
```
paulmskim: All of this should be in an `each()` in the case we have multiple containers right?

src/resources/js/views/manager.js
```diff
@@ -291,6 +306,42 @@ tribe.events.views.manager = {};
...
+	obj.popstate = function( event ) {
+		var $container = $( obj.selectors.container );
+		var containerData = obj.getContainerData( $container );
+
+		// Flag that we are doing popstate globally.
+		obj.doingPopstate = true;
+
+		$container.trigger( 'beforePopState.tribeEvents', event );
+
+		var url = event.originalEvent.originalTarget.location.href;
```
paulmskim: We shouldn't use `originalTarget`, only FF supports it and it's experimental: https://developer.mozilla.org/en-US/docs/Web/API/Event/originalTarget

Instead, we'd probably want to use the history API to be compatible with the browsers we support.

PR #2850, opened 2019-11-15
Support all browsers + better container selection for Navigation

bordoni:
_Ref:_ [C#137312](http://central.tri.be/issues/137312)
---

📹 https://i.bordoni.me/yAu4RwmD


bordoni:
Every night, the Travis CI API tests give these false positives...

PR #2853, opened 2019-11-15
VR > Fix iOS tapping for datepicker.

juanfra:
🎫 https://central.tri.be/issues/137306
🎥 iOS: https://drive.google.com/open?id=13S-bCUmVoWsQJi6ViQ9aotK34lXEW_wz
🎥 Desktop: https://www.loom.com/share/dc2fc161196b4217981a427e3eb3bf97


src/resources/js/views/datepicker.js
```diff
@@ -282,6 +282,24 @@ tribe.events.views.datepicker = {};
 	obj.handleMousedown = function( event ) {
 		var $datepickerButton = event.data.target;
 		var state = $datepickerButton.data( 'tribeEventsState' );
+		var tapHide = false;
+
+		if ( 'touchstart' === event.type ) {
```
paulmskim: 2 notes here:
1. it would be good to separate the `touchstart` event handler from the `mousedown` handler
2. it would also be good to have a flag for this so that all this code below here doesn't run on every single tap, but only the first. especially for mobile, since their processors are smaller than computers, saving computing power is even more important.

PR #2855, opened 2019-11-16
Prevent butchering of UTC dates on submission in the block editor

borkweb:
There are two problems that this PR fixes. One is silly. One is a result of a WP 5.3 change.

* Silly fix: The timestamps were H:I:s rather than H:i:s. The `I` is a boolean indicator of whether or not the date is on DST.

* WP 5.3 fix: With the block editor, WP 5.3 now submits _all_ fields during the REST call regardless of whether we wish to update those fields or not. This butchers our UTC fields that we had historically updated manually upon submission. This fix moves our datetime handling to a point _after_ WP does its thing with the data. Then we fix it.  # Please enter the commit message for your changes. Lines starting

:ticket: https://central.tri.be/issues/137421
:movie_camera: http://p.tri.be/K3Xnbw (9 minutes)

### DIFF
--- a/src/Tribe/Editor/Meta.php
+++ b/src/Tribe/Editor/Meta.php
@@ -1,5 +1,7 @@
 <?php
 
+use Tribe__Date_Utils as Date;
+
 /**
  * Initialize Gutenberg Event Meta fields
  *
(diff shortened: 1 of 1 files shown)


borkweb:
This test is busted: https://travis-ci.org/moderntribe/the-events-calendar/builds/612645168#L1033

Gotta debug that.

sc0ttkclark:
Looks good so far!

PR #2860, opened 2019-11-19
VR > iCal links in new views.

juanfra:
🎫 https://central.tri.be/issues/137446
🎥 https://www.loom.com/share/d4a17d8da72a4ded8014f8c69f4b23ab


src/Tribe/Views/V2/View.php
```diff
@@ -1568,4 +1569,44 @@ protected function filter_display_events_bar( $display ) {
...
+		/**
+		 * Allow for customization of the iCal export link "Export Events" text.
+		 *
+		 * @since unknown
+		 *
+		 * @param string $text The default link text, which is "Export Events".
+		 */
+		$link_text  = apply_filters( 'tribe_events_ical_export_text', __( 'Export Events', 'the-events-calendar' ) );
+
+		$link_title = __( 'Use this to share calendar data with Google Calendar, Apple iCal and other compatible apps', 'the-events-calendar' );
```
paulmskim: I would assume we would want this to be filtered too. I'm not sure of the extent of how much we want to filter the other data (url, display_link), @bordoni ?

src/Tribe/Views/V2/View.php
```diff
@@ -1568,4 +1569,44 @@ protected function filter_display_events_bar( $display ) {
...
+		$ical_data = [
+			'display_link' => $display_ical,
+			'link'         => (object) [
+				'url'    => esc_url( tribe_get_ical_link() ),
+				'anchor' => $link_text,
+				'title'  => $link_title,
+			],
+		];
+
+		return (object) $ical_data;
```
paulmskim: any specific reason why you're typecasting the data to objects here? not sure of the pros and cons here from a BE perspective.

src/Tribe/Views/V2/View.php
```diff
@@ -1568,4 +1569,44 @@ protected function filter_display_events_bar( $display ) {
 
 		return $display;
 	}
+
+	/**
+	 * Returns the iCal data we're sending to the view.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
```
paulmskim: you're returning an object in the function, but in the docs you're returning an array.

src/views/v2/components/ical-link.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+if ( empty( $ical->display_link ) ) {
+	return;
+}
+
+?>
+<div class="tribe-events-c-ical tribe-common-b1">
+	<a
+		class="tribe-events-c-ical__link tribe-events-c-small-cta__link tribe-common-cta--thin-alt"
```
paulmskim: This shouldn't have the `tribe-events-c-small-cta__link` class right?

Also, if we're using the `--thin-alt` modifier for `tribe-common-cta`, we should be using the `tribe-common-cta` class with it. We also have `tribe-common-anchor`, `tribe-common-anchor-alt`, and `tribe-common-anchor-thin` classes if those are more what you want to match. The `-alt` and `-thin` classes do not need the `tribe-common-anchor` to be present to work, as they are not modifiers.

src/Tribe/Views/V2/View.php
```diff
@@ -1568,4 +1569,44 @@ protected function filter_display_events_bar( $display ) {
...
+		 */
+		$link_text  = apply_filters( 'tribe_events_ical_export_text', __( 'Export Events', 'the-events-calendar' ) );
+
+		$link_title = __( 'Use this to share calendar data with Google Calendar, Apple iCal and other compatible apps', 'the-events-calendar' );
+
+		$ical_data = [
+			'display_link' => $display_ical,
+			'link'         => (object) [
+				'url'    => esc_url( tribe_get_ical_link() ),
+				'anchor' => $link_text,
```
paulmskim: can we change this to `'label'` or something that's a bit clearer? was there a specific reason for naming it `'anchor'`? to me, `'anchor'` means the link or url and had me confused.

tests/views_integration/Tribe/Events/Views/V2/Partials/Components/iCalTest.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Tribe\Events\Views\V2\Partials\Components;
+
+use Tribe\Test\Products\WPBrowser\Views\V2\HtmlPartialTestCase;
+
+class iCalTest extends HtmlPartialTestCase
```
paulmskim: small thing, but in staying consistent with all the other tests, if our template is `ical-link.php`, our test class and file name should be `Ical_LinkTest`

src/Tribe/Views/V2/View.php
```diff
@@ -1568,4 +1569,44 @@ protected function filter_display_events_bar( $display ) {
...
+	 */
+	protected function get_ical_data() {
+		/**
+		 * A filter to control whether the "iCal Import" link shows up or not.
+		 *
+		 * @since unknown
+		 *
+		 * @param boolean $show Whether to show the "iCal Import" link; defaults to true.
+		 */
+		$display_ical = apply_filters( 'tribe_events_list_show_ical_link', true );
```
bordoni: These are here for backward compatibility, perfecto. But I would love to see one big filter at the end of the method.

src/Tribe/Views/V2/View.php
bordoni: I would preferer it to be an array, makes it easier to work with `Arr::get()`

src/Tribe/Views/V2/View.php
bordoni: I would go with `link_url`, `link_text` and `link_title`

src/Tribe/Views/V2/View.php
paulmskim: one for all views and one for view specific (following other filters)?

src/Tribe/Views/V2/View.php
juanfra: These filters are already present in the current views, so I left them as they are.

src/Tribe/Views/V2/View.php
juanfra: Made it this way so we can be consistent with the rest (thumbnails, tickets) and use, for example: `$ical->link->title` in the template.

src/views/v2/components/ical-link.php
juanfra: We're using those classes to quickly style up this as a link, but with a font size equal to the previous element (In most views, the navigation). There's no design requirement for this and after testing out different things, this style and size made sense.

src/Tribe/Views/V2/View.php
juanfra: @bordoni this info goes to the template and ideally we could use it like we do with the post thumbs and the rest -> `$ical->link->title`. Do you want to use the array utils functions in the templates?

src/views/v2/components/ical-link.php
juanfra: @paulmskim would you suggest to redefine styles for this new component?

src/Tribe/Views/V2/View.php
bordoni: Ohh I see, this information goes into the `Event_Model`? For `tribe_get_event()`.

juanfra: It's view based, not related to the event: https://github.com/moderntribe/the-events-calendar/pull/2860/files#diff-828c168861c5b8f4cb8331c20c65d55aR1102

src/views/v2/components/ical-link.php
paulmskim: The class `tribe-events-c-small-cta__link` only applies `cursor: pointer` to the element. Also by semantics, you should be applying all the `tribe-events-c-small-cta` classes if you're applying the `tribe-events-c-small-cta__link` class to this element. Further, the `<a>` should by default have `cursor: pointer`. So I'm not entirely sure what you're asking here.

src/views/v2/components/ical-link.php
```diff
@@ -0,0 +1,29 @@
...
+ * Override this template in your own theme by creating a file at:
+ * [your-theme]/tribe/events/views/v2/components/ical-link.php
+ *
+ * See more documentation about our views templating system.
+ *
+ * @link {INSERT_ARTCILE_LINK_HERE}
+ *
+ * @version TBD
+ *
+ * @var obj $ical Object containing iCal data
```
paulmskim: object not obj.

src/views/v2/components/ical-link.php
paulmskim: I understand. So a few things you want to do here:
1. Remove `tribe-events-c-small-cta__link`. As mentioned above, this class doesn't belong here.
2. Remove `tribe-common-cta--thin-alt`. As mentioned above, if you're using this class, you want to use it in conjunction with `tribe-common-cta` as the `--thin-alt` is a modifier. That doesn't belong here either.
3. Add a class `tribe-common-anchor-thin-alt` to common under `base/full/typography/_anchors.pcss`. Use the mixin `anchor-thin-alt`. Add proper overrides. Add `tribe-common-anchor-thin-alt` to this element. As `tribe-common-cta--thin-alt` only applies the `anchor-thin-alt` mixin, rather than using a modifier class meant for a CTA component, using an anchor class meant for applying anchor styles (and not messing with the font size) is what you want to do.

juanfra: Whoops! I just pushed some fixes for this. https://github.com/moderntribe/the-events-calendar/pull/2860/commits/4176b2c7c24d686693f39e88ce869310d80377d3 - I think this way is more clean, I didn't want to re-write a lot of stuff. I think using the mixing for the link can make it directly. Can you check these changes?

paulmskim: Just add `tribe-common-anchor-alt` class to this element and nix the styles in https://github.com/moderntribe/the-events-calendar/commit/4176b2c7c24d686693f39e88ce869310d80377d3 then, we're just duplicating CSS by doing this when we already have classes created for this.

juanfra: Right, that's what I originally intended to do, but with the cta.

src/views/v2/components/ical-link.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+if ( empty( $ical->display_link ) ) {
+	return;
+}
+
+?>
+<div class="tribe-events-c-ical tribe-common-b1">
+	<a
+		class="tribe-common-anchor-alt"
```
paulmskim: can you add back the `tribe-events-c-ical__link` class, to allow users to hook into it if they want to make changes? so we have both `tribe-common-anchor-alt` and `tribe-events-c-ical__link` classes.

PR #2866, opened 2019-11-21
Improve View base class and Interface for better HTML class handling

bordoni:
_Ref:_ [C#124741](http://central.tri.be/issues/124741)
---


src/views/v2/day.php
```diff
@@ -21,9 +21,10 @@
 if ( empty( $disable_event_search ) ) {
 	$header_classes[] = 'tribe-events-header--has-event-search';
 }
+
 ?>
 <div
-	class="tribe-common tribe-events tribe-events-view tribe-events-view--day"
+	<?php tribe_classes( $this->get_view()->get_html_classes() ); ?>
```
paulmskim: do you think it would make more sense to pass this through context rather than in the template?

PR #2874, opened 2019-11-25
feat(Views V2) update the date marker and month sep. to use req. date

lucatume:
Ticket: https://central.tri.be/issues/135365 

[Screencap](https://drive.google.com/open?id=14S0VC2pvpRKB3TcUu2WyzP6xFDVg_h_j&authuser=luca@tri.be&usp=drive_fs)

This PR modifies the behavior of list-like views (List, Phooto and Map) and the month separator on List view to handle the concept of "request date".  

The "request date" is the one that the user has, implicitly by visiting a View or explicitly by using the date picker to select a date, requested.  

If today is 11/25, then the requested date would be 11/25; if, instead, I selected 12/3 using the datepicker, then the request date would be 12/3.

Multi-day events might be among the results for a request date and, yet have started before.

Say we have a 10/1 to 11/30 event: visitign the List View on 11/25 would (correctly) show that event, yet the date marker would say "10/1".

To alleviate this confusion concerning events started "in the past" in relation to the request date, we "move" the event date marker to the request date.

In so doing we update the List View month separator date too.


bordoni:
I love the implementation 🥇

PR #2876, opened 2019-11-27
Featured Archive Views V2

bordoni:
_Ref:_ [C#135422](http://central.tri.be/issues/135422)
---
I've added the WP Matched Query since it was being reset on the `setup_the_loop`.
📹 https://i.bordoni.me/P8ulnX1D

### DIFF
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -1585,6 +1585,20 @@ class View implements View_Interface {
 			}
 		}
 
+		// Setup breadcrumbs for when it's featured.
+		if ( $is_featured = tribe_is_truthy( $this->context->get( 'featured', false ) ) ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Views/V2/View.php
```diff
@@ -1585,6 +1585,20 @@ protected function get_breadcrumbs() {
 			}
 		}
 
+		// Setup breadcrumbs for when it's featured.
+		if ( $is_featured = tribe_is_truthy( $this->context->get( 'featured', false ) ) ) {
+			$non_featured_link = tribe_events_get_url( [ 'featured' => 0 ] );
+
+			$breadcrumbs[] = [
+				'link'  => $non_featured_link,
```
juanfra: Why aren't we using `tribe_get_events_link()` here?

bordoni: That is the old method. We need to figure out how to deprecated a lot of those or route them correctly.

PR #2878, opened 2019-11-28
Create the required methods for proper Smart Activation for new Installs

bordoni:
_Ref:_ [C#137510](http://central.tri.be/issues/137510)
---
📹 https://i.bordoni.me/WnuERqGP


bordoni:
Merging as these are false positives.

PR #2879, opened 2019-11-28
Feature/135367 events per page 12 default

bordoni:
Depends on #2878 
📹 https://i.bordoni.me/YEudqBwv


bordoni:
TravisCI likes to throw false alerts at night.

PR #2888, opened 2019-12-03
Overwrite the global Post  using WP_Query::setup_postdata

bordoni:
_Ref:_ [C#137977](http://central.tri.be/issues/137977)
---


src/views/v2/list.php
```diff
@@ -23,6 +23,7 @@
 if ( empty( $disable_event_search ) ) {
 	$header_classes[] = 'tribe-events-header--has-event-search';
 }
+global $wp_query;
```
juanfra: we don't need this one, right?

bordoni: No, we dont. my Bad.

PR #2891, opened 2019-12-03
VR > Avada override

juanfra:
🎫 https://central.tri.be/issues/137527
🎥 https://www.loom.com/share/1696c19d44b048e4b308e3503f519f1c


src/resources/postcss/views/full/month/_multiday-event.pcss
```diff
@@ -61,4 +61,10 @@
 		font-weight: var(--font-weight-regular);
 		text-overflow: ellipsis;
 	}
+
+	.tribe-theme-avada #main & {
```
paulmskim: can we add a comment block above this like we do for the other theme overrides to clearly indicate where the specific theme override begins?

juanfra: Sure!

paulmskim:
looks like snapshots are failing

PR #2896, opened 2019-12-05
Pagination on Venue Organizer 

bordoni:
ref/137384
---
📹 https://i.bordoni.me/L1uk5l8o


src/Tribe/Repositories/Organizer.php
```diff
@@ -48,26 +48,11 @@ public function __construct() {
...
-		$this->schema = array_merge( $this->schema, array(
-			'name' => array( $this, 'filter_by_name' ),
-		) );
-
 		$this->add_simple_meta_schema_entry( 'email', '_OrganizerEmail' );
 		$this->add_simple_meta_schema_entry( 'phone', '_OrganizerPhone' );
 		$this->add_simple_meta_schema_entry( 'website', '_OrganizerWebsite' );
 	}
 
-	/**
```
lucatume: Why remove this?

src/Tribe/Views/V2/Template/Page.php
```diff
@@ -270,7 +270,7 @@ public function should_hijack_page_template( \WP_Query $query ) {
 		}
 
 		// We wont hijack in case we are not dealing with a Post Type query
-		if ( ! in_array( TEC::POSTTYPE, (array) $query->get( 'post_type' ) ) ) {
+		if ( empty( $query->tribe_is_event_query ) ) {
```
lucatume: The flag property is set [in the `Tribe__Events__Query` class](https://github.com/moderntribe/the-events-calendar/blob/e95c1d91b4091e6865c6e7f9fd9cf179ee16b435/src/Tribe/Query.php#L186-L191), that will, eventually, be cycled out of v2.
I would re-implement this check in a Context location.

bordoni: We should revisit that and the method inside of `Bootstrap::should_load`. For now I think we can keep it.

src/Tribe/Repositories/Organizer.php
bordoni: Because that was breaking `tribe_organizers()->where( 'name' )` which uses `post_name` which should be an exact method instead of search.

src/Tribe/Views/V2/View.php
```diff
@@ -726,7 +726,7 @@ public function get_url( $canonical = false, $force = false ) {
 		}
 
 		$event_display_mode = $this->context->get( 'event_display_mode', false );
-		if ( false !== $event_display_mode && $event_display_mode !== $this->context->get( 'eventDisplay' ) ) {
+		if ( 'past' === $event_display_mode && $event_display_mode !== $this->context->get( 'eventDisplay' ) ) {
```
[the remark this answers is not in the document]
bordoni: No need, because if `=== 'past'` it's for sure diff than `false`.

PR #2898, opened 2019-12-05
Fix/137760 datepicker fixes

paulmskim:
https://central.tri.be/issues/137760
https://central.tri.be/issues/137766

datepicker-styles


src/resources/js/views/datepicker.js
```diff
@@ -256,6 +256,19 @@ tribe.events.views.datepicker = {};
 		obj.submitRequest( $container, dateValue );
 	};
 
+	/**
+	 * Handle datepicker show event
+	 *
+	 * @since 4.9.8
```
juanfra: Are we introducing this? If yes, then we should set the `@since` `TBD`

juanfra:
Was this tested on iOs devices?

paulmskim:
I don't have an iOS device to test on. However the change you made regarding the touchstart should be the same. If we find that the datepicker breaks on iOS we can revert this back.

PR #2903, opened 2019-12-09
VR > Tooltip: break words and wrap overflow for really really really really really long words.

juanfra:
🎫 https://central.tri.be/issues/137400
🎥 https://www.loom.com/share/4cd6ca71a0224ff682e84ec6d7560b73

### DIFF
--- a/src/resources/postcss/components/skeleton/_tooltip.pcss
+++ b/src/resources/postcss/components/skeleton/_tooltip.pcss
@@ -15,8 +15,7 @@
 
 		.tooltipster-content {
 			padding: 0;
-			overflow-wrap: break-word;
-			word-break: break-all;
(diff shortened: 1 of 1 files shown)


src/resources/postcss/components/skeleton/_tooltip.pcss
```diff
@@ -15,6 +15,8 @@
 
 		.tooltipster-content {
 			padding: 0;
+			overflow-wrap: break-word;
+			word-break: break-all;
```
paulmskim: We should almost never use `break-all` for `word-break`, otherwise even just regular sentences that have words that are meant to wrap won't wrap. Use `break-word` instead.

Also if we use `break-word`, do we need `overflow-wrap`?

PR #2907, opened 2019-12-10
Feature/138282 add container media query bps

paulmskim:
https://central.tri.be/issues/138282


src/Tribe/Views/V2/View.php
```diff
@@ -1755,11 +1760,33 @@ protected function get_show_datepicker_submit() {
 		$live_refresh       = tribe_is_truthy( tribe_get_option( 'liveFiltersUpdate', true ) );
 		$disable_events_bar = tribe_is_truthy( tribe_get_option( 'tribeDisableTribeBar', false ) );
 
+		$show_datepicker_submit = false;
```
lucatume: Simplify the setting:
```php
$show_datepicker_submit = empty( $live_refresh) && ! empty( $disable_events_bar ) );
```

The following lines would need to be removed.

src/Tribe/Views/V2/Behaviors/Breakpoint_Behavior.php
```diff
@@ -0,0 +1,89 @@
+<?php
+/**
+ * Provides methods for Views for breakpoint behavior.
+ *
+ * @since   TBD
+ *
+ * @package Tribe\Events\Views\V2\Behaviors
+ */
+
+namespace Tribe\Events\Views\V2\Behaviors;
```
lucatume: Question: could you explain why you've created a `Behaviors` directory for this? Do you think we'll add more?

bordoni: Yeah, I had a convo with Paul around possibly moving some things out of the View Class, since currently, it creates a lot of bloat. Tomorrow lets talk about that, I think it could be a possible solution. Mostly around we could use for Breadcrumbs, Messages, and move the currently existing `List_Behavior` into that folder.

PR #2908, opened 2019-12-10
fix(Views V2) avoid inf. loop on object cache

lucatume:
Ticket: https://central.tri.be/issues/138249  

This PR is a slight variation on the fix @bordoni had originally implemented to avoid infinite loops on excerpt.
This PR re-introduces the fix and takes into account the additional scenario of object-cache presence.

[Screencap](https://drive.google.com/open?id=1-R3zdQKmdE_Z2Sz7vv5VZwl6t_a-Cc4l&authuser=luca@tri.be&usp=drive_fs)

To test the fix I'v put together a really incomplete, yet fitting for this specific issue testing, pluging [in this gist](https://gist.github.com/e44c5fe6b81d60651f8fbec3d49fecc0).

### DIFF
--- a/src/Tribe/Views/V2/Template/Excerpt.php
+++ b/src/Tribe/Views/V2/Template/Excerpt.php
@@ -8,9 +8,9 @@
  */
 namespace Tribe\Events\Views\V2\Template;
 
-use Tribe__Template as Base_Template;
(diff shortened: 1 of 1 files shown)


bordoni:
Damn that is a hell of a solution! 🌮 🌮

PR #2911, opened 2019-12-10
Freemius uninstall

zbtirrell:
uninstall is firing and the reason is sent to Freemius.

see: 
https://central.tri.be/issues/138342 (100% of installs)
https://central.tri.be/issues/134066 (fix uninstall hook)

Note: the uninstall event is causing an email to be sent to the customer if the plugin is deleted soon after being installed. This is non-configurable and does not meet our quality standards, were reaching out to Freemius about this, but does not block progress on this work


bordoni:
What might be required is the correct path being set inside of the `get_caller_main_file_and_type` from Freemius or inside of the option set in `store_id_slug_type_path_map`.

src/Tribe/Main.php
```diff
@@ -3019,6 +3019,15 @@ public static function deactivate( $network_deactivating ) {
 			add_action( 'shutdown', array( $deactivation, 'deactivate' ) );
 		}
 
+		/**
+		 * plugin uninstall callback
+		 * @see register_uninstall_hook()
+		 */
+		public static function uninstall() {
+			Tribe__Events__Integrations__Manager::instance()->load_freemius();
+			tribe( 'events.integrations.freemius' )->uninstall();
```
bordoni: This method doesn't exist in the Freemius class or ours.

zbtirrell: yeah, I wanted it to fatal, but it never gets to this point in the code :(

zbtirrell:
updated PR with an adjusted approach. Edited description with current state of work

zbtirrell:
This is now fully functional!

src/Tribe/Integrations/Freemius.php
```diff
@@ -64,7 +64,10 @@ public function __construct() {
 			'tribe-help',
 		];
 
-		if ( ! in_array( $this->page, $valid_page ) && 'plugins.php' !== $pagenow ) {
+		if ( ! in_array( $this->page, $valid_page )
+			&& 'plugins.php' !== $pagenow
+			&& 'admin-ajax.php' !== $pagenow ) {
```
borkweb:
```php
if (
	! in_array( $this->page, $valid_page )
	&& 'plugins.php' !== $pagenow
	&& 'admin-ajax.php' !== $pagenow
) {
```

src/Tribe/Integrations/Freemius.php
```diff
@@ -64,7 +64,10 @@ public function __construct() {
...
+			echo 'page: ' . $pagenow;
```
borkweb: Is this debug output?

bordoni:
Waiting for tests to pass, and will merge.

bordoni:
Both fails are unrelated. One is a false positive, the other snapshots need to be regenerated on Views V2. Will look into this a little further on a separate PR.

PR #2928, opened 2019-12-16
 set the hidden input to readonly so iOS keyboard doesn't appear

zbtirrell:
I believe this is a totally safe way to hide the mobile keyboard in this scenario. JS can still change the value, but the keyboard won't pop up.

If this feels like a solid solution, I'll update the tests, but avoiding doing that until it has been reviewed.

(target for PR should be updated for wherever is most appropriate)

This was originally fixed here: https://central.tri.be/issues/134971 by removing the focus, but the focus was put back later for accessibility reasons.


paulmskim:
Does this work for keyboard navigation of the datepicker? If so, then I think that'll work.

zbtirrell:
just double checked - this still allows keyboard navigation in the picker

borkweb:
I updated snapshots 🕺

PR #2931, opened 2019-12-17
Filter get_past_url() and get_upcoming_url() calls

borkweb:
Filterbar hooks into `tribe_events_views_v2_url_query_args` filter to attach its query arguments to URLs. Our prev_url() and next_url() fire that filter, however, Map and Photo view's `get_past_url()` was not doing so. This changeset adds the filter so filterbar can work its magic on any of the URLs.

:movie_camera: http://p.tri.be/s6Gs7h

Related:
* https://github.com/moderntribe/events-pro/pull/1414
* https://github.com/moderntribe/events-filterbar/pull/217

see/138621
see/138632

### DIFF
--- a/src/Tribe/Views/V2/Views/List_View.php
+++ b/src/Tribe/Views/V2/Views/List_View.php
@@ -98,7 +98,6 @@ class List_View extends View {
 		$date           = $this->context->get( 'event_date', $default_date );
 		$event_date_var = $default_date === $date ? '' : $date;
 
-
 		$past = tribe_events()->by_args( $this->setup_repository_args( $this->context->alter( [
 			'event_display_mode' => 'past',
 			'paged'              => $page,
--- a/src/Tribe/Views/V2/Views/List_View.php
+++ b/src/Tribe/Views/V2/Views/List_View.php
@@ -188,16 +188,7 @@ class List_View extends View {
 				'tribe-bar-search' => $this->context->get( 'keyword' ),
 			];
 
-			/**
-			 * Filters the query arguments that will be used to build a View URL.
-			 *
-			 * @since TBD
-			 *
-			 * @param array          $query_args An array of query args that will be used to build the URL for the View.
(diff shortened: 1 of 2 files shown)


src/Tribe/Views/V2/Views/List_View.php
```diff
@@ -98,19 +98,34 @@ protected function get_past_url( $canonical = false, $page = 1 ) {
 		$date           = $this->context->get( 'event_date', $default_date );
 		$event_date_var = $default_date === $date ? '' : $date;
 
+
```
barryhughes: Non-essential thing: extra line break here that we probably don't need. Also not a functional concern :-)

src/Tribe/Views/V2/Views/List_View.php
```diff
@@ -98,19 +98,34 @@ protected function get_past_url( $canonical = false, $page = 1 ) {
...
+			/**
+			 * Filters the query arguments that will be used to build a View URL.
+			 *
+			 * @since TBD
+			 *
+			 * @param array          $query_args An array of query args that will be used to build the URL for the View.
+			 * @param View_Interface $this       This View instance.
+			 * @param bool           $canonical  Whether the URL should be the canonical one or not.
+			 */
+			$query_args = apply_filters( 'tribe_events_views_v2_url_query_args', $query_args, $this, $canonical );
```
barryhughes: Optional thing, just wanted to ask the question: in name and signature this is identical to what we see [down on line 200 (same file)](https://github.com/moderntribe/the-events-calendar/pull/2931/files#diff-a30e077e32723b54e7acf6fe03e4b685R200). Could it be worth moving it to its own method? That would keep things DRYer and, thinking of the potential for error during future maintenance, would reduce the potential for one to fall out of sync with the other.

borkweb: I've created a method in the View class to handle query_arg filtering and plugged it in to all the places! Good call :+1:

PR #2935, opened 2019-12-17
fix(Views v2) localized slugs

lucatume:
Ticket: https://central.tri.be/issues/138574

[Screencap](https://drive.google.com/open?id=1-pK81yqDR6fqGDblKEcH5tPZhfpL7sOU&authuser=luca@tri.be&usp=drive_fs)

This PR updates Views V2 code to handle 2 issues that were preventing localized slugs from correctly working:

1. during the creation of rewrite bases the `Tribe__Rewrite` class would only generate the not URL-encoded version of the slugs; this would, in turn, make it so that `/events/foto/página/2` would be picked up as valid, but the same slug, `urlencode`d during the AJAX request (`/events/foto/p%C3%A1gina/2`) would not.
2. prevent WordPress from redirecting to canonical URL on paginated Views; this would cause, inside `redirect_canonical` function, the `/events/foto/p%C3%A1gina/2` URL to be redirected to `/events/foto/p%C3%A1gina/2/page/2`.


src/Tribe/Views/V2/Rewrite.php
```diff
@@ -0,0 +1,65 @@
...
+	 * @param string $method The method that's being used to generate the bases; defaults to `regex`.
+	 *
+	 * @return array<string,array> The filtered rewrite rules, updated or modified if required.
+	 */
+	public function filter_raw_i18n_slugs( array $bases, $method ) {
+		if ( $method !== 'regex' ) {
+			return $bases;
+		}
+
+		$bases = $this->add_url_encoded_slugs($bases);
```
zbtirrell: spacing

borkweb:
Looks like there's one failing test.

lucatume:
Tests finally passed and good for another look.

PR #2945, opened 2019-12-19
Fixing snapshot that uses the datetime fix in common

borkweb:
With the updates in the tribe-common PR (with a branch name that matches this one), a datetime issue is resolved. This PR fixes the snapshot to match the desired reality.

Related: https://github.com/moderntribe/tribe-common/pull/1246

see/138795


lucatume:
If this is affected by https://github.com/moderntribe/tribe-common/pull/1246/, then changes should be reviewed.

PR #2946, opened 2019-12-19
feat(View V2) add filterable `container_data` tvar

lucatume:
Ticket: https://central.tri.be/issues/138619 

This PR adds a new, filterable, template var, `container_data`, and updates the templates to use it.
The template var is hydrated from an `array<string,string>` like:

```php  
$container_data = [
	'foo' => 23,
	'bar' => 89,
];
```

This will produce, on the View top-level container, the additional `data-view-` attributes:

```
data-view-foo="23"
data-view-bar="89"
```

The PR also modifies the `manager.js` to parse and read some of that informatin to sport shortcodes.
While shortcodes are a feature we add in PRO, the support for shortcodes is in TEC; that's why it's there.


src/resources/js/views/manager.js
```diff
@@ -278,11 +279,15 @@ tribe.events.views.manager = {};
...
 			prev_url: encodeURI( currentUrl ),
-			url: encodeURI( url ) ,
+			url: encodeURI( url ),
 			should_manage_url: shouldManageUrl,
-			_wpnonce: nonce
+			_wpnonce: nonce,
 		};
 
+		if ( shortcodeId ) {
+			data[ 'shortcode' ]	 = shortcodeId;
```
paulmskim:
```suggestion
			data[ 'shortcode' ] = shortcodeId;
```

PR #2948, opened 2019-12-19
VR > Include customization stylesheet.

juanfra:
🎥 https://www.loom.com/share/a90160eeb90d4a49bed091f6ca7bdabb

### DIFF
--- a/src/Tribe/Views/V2/Assets.php
+++ b/src/Tribe/Views/V2/Assets.php
@@ -318,21 +318,25 @@ class Assets extends \tad_DI52_ServiceProvider {
 			]
 		);
 
-		tribe_asset(
-			$plugin,
-			'tribe-events-views-v2-override-style',
-			Tribe__Events__Templates::locate_stylesheet( 'tribe-events/tribe-events.css' ),
-			[
-				'tribe-common-full-style',
-				'tribe-events-views-v2-skeleton',
-			],
-			'wp_enqueue_scripts',
-			[
-				'priority'     => 10,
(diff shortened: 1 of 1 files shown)


src/Tribe/Views/V2/Assets.php
```diff
@@ -316,6 +317,22 @@ public function register() {
 				'in_footer'    => false,
 			]
 		);
+
+		tribe_asset(
```
borkweb: Excuse my ignorance around tribe_asset(), but what is the behavior when the file is not located? Also, I have heard that tribe_asset is a bit more expensive than we’d like-I wonder if it’d be more performant to try to locate_stylesheet before invoking tribe_asset?

juanfra: If it's not found it doesn't load anything. Have to check the asset implementation but I feel that check must be part of the logic inside the function.

borkweb: I just dug into `tribe_asset()`. It most definitely does what you are saying, but after executing a bunch of things. Let's wrap this tribe_asset() call in an if where we check for the stylesheet first.

juanfra: Seems like we have that and other checks covered: https://github.com/moderntribe/tribe-common/blob/master/src/Tribe/Assets.php#L367

And, we wouldn't be adding more "cost", as it's how we currently do it: https://github.com/moderntribe/the-events-calendar/blob/master/src/Tribe/Assets.php#L316

PR #2949, opened 2019-12-19
Cease preventing redirects when we are TRYING to redirect

borkweb:
When redirecting from one view to another, we plop a `tribe_redirected` query argument onto the new URL. This changeset detects that and ensures that we can properly redirect.

:movie_camera: http://p.tri.be/ysdOT8

see/138864


lucatume:
I've merged the code from https://github.com/moderntribe/the-events-calendar/pull/2951 to try and see if this fixes the issues on this build.

PR #2954, opened 2019-12-23
TEC-2989 TEC-3033 - VR - Fix pwd protected post fields

lucatume:
Tickets: [TEC-2989], [TEC-3033]

This PR:

* fixes some password-protected fields issues I found on V1
* ports over V1 behavior, where possible, on V2

Screencaps:

1. [How it works on v1](https://drive.google.com/open?id=10ViBRIYfMnWzkAQ8YdX6rb9dumcaNwKY&authuser=luca@tri.be&usp=drive_fs); to provide a frame of reference about how it currently works
2. [How it works on v2](https://drive.google.com/open?id=10XwurNb5WbiUinoXLTUx1h4fwoF37o36&authuser=luca@tri.be&usp=drive_fs); the same in v2; there's an exception where the event day still shows (in pretty much all views) due to how v1 and v2 are different in terms of layout and UI building.
3. [How it works on V2 w/ object-caching enabled](https://drive.google.com/open?id=10ZAXQpMJQxQ5jsvBZqV-qzY33I6uXsGE&authuser=luca@tri.be&usp=drive_fs); same as the previous video, but w/ object-caching
4. [Differences between  Month on v1 and v2](https://drive.google.com/open?id=10jAL0rNOE5FUnvSCQx7PYrFdWva0zwae&authuser=luca@tri.be&usp=drive_fs) Month view is a good example of what I meant in point 2: it's not always clear when showing/hiding something is due to how the information is presented or due to a clear behavioral choice.

I think this PR should not be the last to tackle how password-protected events show in V2, yet I think it's a solid first one.
The next one should come after a "tour" of our existing Views and a clear definition, where V1 existing implementation is not a clear guide, of what should show/hide where.

[TEC-2989]: https://moderntribe.atlassian.net/browse/TEC-2989

[TEC-3033]: https://moderntribe.atlassian.net/browse/TEC-3033

### DIFF
--- a/src/Tribe/Views/V2/Hooks.php
+++ b/src/Tribe/Views/V2/Hooks.php
@@ -54,14 +54,11 @@ class Hooks extends \tad_DI52_ServiceProvider {
 		add_action( 'rest_api_init', [ $this, 'register_rest_endpoints' ] );
 		add_action( 'tribe_common_loaded', [ $this, 'on_tribe_common_loaded' ], 1 );
 		add_action( 'wp_head', [ $this, 'on_wp_head' ], 1000 );
-
-		// Hide sensitive event info if post is password protected.
-		add_action( 'the_post', array( $this, 'action_manage_sensitive_info' ) );
-
 		add_action( 'tribe_events_pre_rewrite', [ $this, 'on_tribe_events_pre_rewrite' ] );
 		add_action( 'wp_enqueue_scripts', [ $this, 'action_disable_assets_v1' ], 0 );
 		add_action( 'tribe_events_pro_shortcode_tribe_events_after_assets', [ $this, 'action_disable_shortcode_assets_v1' ] );
 		add_action( 'updated_option', [ $this, 'action_save_wplang' ], 10, 3 );
+		add_action( 'the_post', [ $this, 'manage_sensitive_info' ] );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/Tribe/Views/V2/Hooks.php
```diff
@@ -55,14 +54,11 @@ protected function add_actions() {
...
 		add_action( 'wp_head', [ $this, 'on_wp_head' ], 1000 );
-
-		// Hide sensitive event info if post is password protected.
-		add_action( 'the_post', array( $this, 'action_manage_sensitive_info' ) );
-
 		add_action( 'tribe_events_pre_rewrite', [ $this, 'on_tribe_events_pre_rewrite' ] );
 		add_action( 'wp_enqueue_scripts', [ $this, 'action_disable_assets_v1' ], 0 );
 		add_action( 'tribe_events_pro_shortcode_tribe_events_after_assets', [ $this, 'action_disable_shortcode_assets_v1' ] );
 		add_action( 'updated_option', [ $this, 'action_save_wplang' ], 10, 3 );
+		add_action( 'the_post', [ $this, 'manage_sensitive_info' ] );
```
borkweb: This action was actually causing a problem when a password protected post appears in the loop. With your value filtering that you've added elsewhere, I suspected that this action is not needed for v2, so I grabbed this branch and removed this hook to see what happens. What results is precisely what we want: sanitized content _without_ impacting the presence of the **Get Tickets**/**RSVP Now** links. See here:

**List View**

Screen Shot 2019-12-23 at 10 56 33 AM

**Single Post**

Screen Shot 2019-12-23 at 10 56 16 AM

borkweb: Follow-up: I tested against the **Default Events Template** and the action is still necessary for `is_single()` pages.

lucatume: Putting this back into Progress to address this.

lucatume:
After @borkweb comments on the PR I've updated with a more granular application of the filter on `the_post`.

[Screencap](https://drive.google.com/open?id=10lHYS4kmoZwFvl81F66emt4_njzi1_Yz&authuser=luca@tri.be&usp=drive_fs)

Please note that the ET branch to use is the `bucket/views-redesign-rc1
` one as the current ET `master` is not correctly checking for tickets existence (fixed in the `bucket/views-redesign-rc1` branch).

PR #2956, opened 2019-12-24
Several updates to the codebase

mitogh:
- Updates to the array sintax
- Update document references
- Remove non required `else` blocks
- Remove non required variables on return results

Task: https://central.tri.be/issues/116977
Ticket: [EA-71]
Ticket: [TEC-3093]

[EA-71]: https://moderntribe.atlassian.net/browse/EA-71

[TEC-3093]: https://moderntribe.atlassian.net/browse/TEC-3093


bordoni:
@mitogh lets make sure tickets are in the correct place for this PR. We should think about shipping this with B20.02.

mitogh:
Sounds good updating the tickets accordingly. Thanks @bordoni

src/Tribe/Aggregator/Service.php
```diff
@@ -543,13 +541,11 @@ public function get_image( $image_id, $record ) {
 		 * @since 4.6.18
 		 *
 		 * @param  array  $data      Which Arguments
-		 * @param  strng  $image_id  Image ID
+		 * @param  string  $image_id  Image ID
```
barryhughes: Don't know if it's in scope for this piece of work, but, since we're fixing this docblock: shouldn't we have three `@param` tags? One for the array being filtered, one for `$record` and one for `$image_id`?

mitogh: Let me do the update as we are already on this one :)

lucatume: Since we're doing this:
```suggestion
		 * @param string	      $image_id  The image post ID.
		 * @param array<string,mixed> $data      The image data.
```

PR #2964, opened 2020-01-02
Only check if the imported excerpt is not blank to return imported ex…

codingmusician:
…cerpt. Otherwise return post's excerpt

### DIFF
--- a/src/Tribe/Importer/File_Importer_Events.php
+++ b/src/Tribe/Importer/File_Importer_Events.php
@@ -456,7 +456,7 @@ class Tribe__Events__Importer__File_Importer_Events extends Tribe__Events__Impor
 		if ( $event_id ) {
 			$post_excerpt = get_post( $event_id )->post_excerpt;
 
-			return empty( $post_excerpt ) && ! empty( $import_excerpt ) ? $import_excerpt : $post_excerpt;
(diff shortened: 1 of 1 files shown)


codingmusician:
Ref: https://moderntribe.atlassian.net/browse/EA-48

We want uploaded excerpts (via CSV) to override the post's excerpt whether or not an excerpt for the post exists.

src/Tribe/Importer/File_Importer_Events.php
```diff
@@ -456,7 +456,7 @@ private function get_post_excerpt( $event_id, $import_excerpt ) {
 		if ( $event_id ) {
 			$post_excerpt = get_post( $event_id )->post_excerpt;
 
-			return empty( $post_excerpt ) && ! empty( $import_excerpt ) ? $import_excerpt : $post_excerpt;
+			return !empty( $import_excerpt ) ? $import_excerpt : $post_excerpt;
```
mitogh: Small suggestion to remove the need to use a negation operator here.

```suggestion
			return empty( $import_excerpt ) ? $post_excerpt : $import_excerpt;
```

mitogh: Another question about this one, here we are assuming `$post_excerpt` is not empty, is there any other check before this one to ensure the returned value is not empty in case `$post_excerpt` is not empty

PR #2958, opened 2019-12-26
Doc/133648 js docs

paulmskim:
https://central.tri.be/issues/133648

PR #2970, opened 2020-01-04
Fix/tec 3091 events bar input

paulmskim:
https://moderntribe.atlassian.net/browse/TEC-3091

events-bar-input

See https://github.com/moderntribe/the-events-calendar/pull/2958#discussion_r362939770

### DIFF
--- a/src/resources/js/views/events-bar-inputs.js
+++ b/src/resources/js/views/events-bar-inputs.js
@@ -80,7 +80,7 @@ tribe.events.views.eventsBarInputs = {};
 					return;
 				}
 
-				$input.off( 'change', obj.handleInputChange );
+				$input.off();
 			} );
 	};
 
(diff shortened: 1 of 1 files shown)


src/resources/js/views/events-bar-inputs.js
```diff
@@ -97,7 +97,7 @@ tribe.events.views.eventsBarInputs = {};
 		$container
 			.find( obj.selectors.inputWrapper )
 			.each( function( index, wrapper ) {
-				var inputWrapperClass = wrapper.className.match( /tribe-common-c-search__input-control--[a-z]+/ );
+				var inputWrapperClass = wrapper.className.match( /tribe-events-c-search__input-control--[a-z]+/ );
```
bordoni: I really think we should completely remove the requirement for this particular wrapper class. Possibly only depend on a particular "simpler" data attribute.

paulmskim: I agree, but I think we should tackle this in a separate ticket. In this, I'm just fixing something that should have been working but isn't.

bordoni: Sounds good to me. Would you mind logging that for the Backlog? and get it merged.

PR #2973, opened 2020-01-06
Change default list widget to unordered list

nielsranosch:
In most default themes, lists in widgets are expected to contain an unordered list rather than an ordered one. This change will make the widget more compatible with most free themes.

### DIFF
--- a/src/views/widgets/list-widget.php
+++ b/src/views/widgets/list-widget.php
@@ -33,7 +33,7 @@ $posts = tribe_get_list_widget_events();
 // Check if any event posts are found.
 if ( $posts ) : ?>
 
-	<ol class="tribe-list-widget">
+	<ul class="tribe-list-widget">
 		<?php
 		// Setup the post data for each event.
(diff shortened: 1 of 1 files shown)


nielsranosch:
I don't know if anyone will actually read this, but I don't know why the checks have failed. Looking at the error messages, I don't think it has anything to do with my commit.

bordoni:
Hi @nielsranosch,

We do read, but past week was quite action packed with the Version 5.0 release coming soon.
You are correct, the tests have failed in the `master` branch due to how the year was being embed in the snapshots, and last year it was not.

Not to worry, our team will review this and merge or not depending on the strategy decision around this change. It will most likely be something happening after 5.0 is released.

bordoni:
Hi @nielsranosch,

I did some further investigation, and we want to avoid making any changes to HTML elements for now on our Widgets, as they will get some changes in the near future, so we avoiding any changes that might make customers customizations with CSS break is ideal.

I am closing the ticket for the timebeing.

best regards,

PR #2975, opened 2020-01-07
VR > iCal link revamp

juanfra:
🎫 https://moderntribe.atlassian.net/browse/TEC-3016
🎥 https://www.loom.com/share/29f4bb6c9db24857843d989517bcd8cf


src/resources/postcss/components/full/_ical-link.pcss
```diff
@@ -3,3 +3,45 @@
...
+.tribe-events {
+
+	.tribe-events-c-ical__link {
+		align-items: center;
+		color: var(--color-accent-primary);
+		border: 1px solid var(--color-accent-primary);
+		border-radius: var(--border-radius-default);
+		display: inline-flex;
+		font-weight: var(--font-weight-bold);
+		padding: var(--spacer-2);
```
paulmskim: this should be in skeleton.

src/resources/postcss/components/full/_ical-link.pcss
```diff
@@ -3,3 +3,45 @@
...
+		justify-content: center;
+		width: 100%;
```
paulmskim: this should be in skeleton

src/resources/postcss/components/full/_ical-link.pcss
juanfra: Thanks for the review!

I left the padding on the full stylesheet because it really did to the styles of the button, I thought it was more appropriate there.

1. I see all with the same blue.
2. Nice, `twentyseventeen` was having something with the hover.
3. Not sure if we're gonna use this in any other place for now. If we move it, would be cool to make it part of the design system in invision, and find a name, so it doesn't get lost.

src/resources/postcss/components/full/_ical-link.pcss
```diff
@@ -3,3 +3,58 @@
...
+			color: var(--color-background);
+			background-color: var(--color-accent-primary);
+
+			&:before {
+				background-image: svg-inline(plus-focus);
+			}
+		}
+
+		.tribe-common--breakpoint-medium& {
+			width: auto;
```
paulmskim: this width should also move to skeleton.

src/resources/postcss/components/full/_ical-link.pcss
```diff
@@ -3,3 +3,58 @@
...
+		align-items: center;
+		color: var(--color-accent-primary);
+		border: 1px solid var(--color-accent-primary);
+		border-radius: var(--border-radius-default);
+		display: inline-flex;
+		font-weight: var(--font-weight-bold);
+		justify-content: center;
+
+		&:before {
+			background-image: svg-inline(plus);
```
paulmskim: this should be `plus-focus`.

src/resources/postcss/components/full/_ical-link.pcss
```diff
@@ -3,3 +3,58 @@
...
+		}
+
+		&:hover,
+		&:focus,
+		&:active {
+			color: var(--color-background);
+			background-color: var(--color-accent-primary);
+
+			&:before {
+				background-image: svg-inline(plus-focus);
```
paulmskim: this should be `plus-white`

PR #3021, opened 2020-01-24
Revert "Adjust the structure of next/prev URLs so they can be properly tested"

bordoni:
Reverts moderntribe/the-events-calendar#3007

These changes were creating: [TEC-3173]

[TEC-3173]: https://moderntribe.atlassian.net/browse/TEC-3173

PR #3022, opened 2020-01-24
fix(Views V2) URL building issue

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-3174

[Self-QA](https://drive.google.com/open?id=11gmKpvNDDEBH1pEAPTRojzH4Qnr60hwi&authuser=luca@tri.be&usp=drive_fs)
[Self QA for FBAR-64](https://drive.google.com/open?id=11piJSjPyr6dPYbPJfNX5ok2SyvVpWP-N&authuser=luca@tri.be&usp=drive_fs)

This PR fixes an issue introduced in 5.0.1 where list-like Views would
incorrectly build their URLs.
The issue was introduced in the context of the work to fix [FBAR-64](https://moderntribe.atlassian.net/browse/FBAR-64), so I've re-tested the code with this fix.


bordoni:
@borkweb this seems to be the solution from the problem that was reported by victor yesterday.

Luca we should PR this into B20.01, after removing the revert I had done yesterday night.
#3021

lucatume:
> Luca we should PR this into B20.01, after removing the revert I had done yesterday night.
> #3021

Ok, I will make sure this fixes it in 5.0.1 and then proceed to back-port on 5.0.

lucatume:
Back-port PR: #3023 3023

PR #3026, opened 2020-01-25
Fix/tec 3172 v2 a11y improvements

paulmskim:
[TEC-3172]

[TEC-3172]: https://moderntribe.atlassian.net/browse/TEC-3172


src/Tribe/Views/V2/View.php
```diff
@@ -863,6 +863,72 @@ public function prev_url( $canonical = false, array $passthru_vars = [] ) {
...
 	}
 
+	/**
+	 * Returns the top bar navigation label associated with this View logical, next view.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function top_bar_next_label() {
```
lucatume: In place of filtering into this method directly, I'd rather go with the same pattern we used for `View::setup_template_vars` or `View::setup_repository_arguments`.

Break this method down into two methods, I'm using the `View::top_bar_next_level` method as an example:

1. `View::top_bar_next_label` is the method that produces the default label ("Next")
2. `View::filter_top_bar_next_label` is the method that filters the label

In a case like the Day View, only the first method would be overridden, not the second. This de-clutters the code, slims it down and keeps the filters in one place only.

lucatume:
> Even if it means using a Trait that is added to all classes.

A lot of code here really belongs in the View. If we introduce a trait, we have to ensure that Trait will work with anything that is not a view.
Using `method_exists` to duck-type the object using a Trait can be a code smell that the code should be _composed_ (the code should live in a class that is a dependency of View, in this case) not _composed_.

If we create Traits that will work only with the View, then we're just scattering the code around for sake of readability we're losing.

lucatume:
> Even if it means using a Trait that is added to all classes.

This is a good idea, maybe not for immediate implementation, but still a good refactoring option.

This contradicts my previous comment about this not being a good idea. Brainfreeze.

paulmskim:
@bordoni @lucatume I've removed all the context stuff and copied templates for each view, after discussing with Gustavo. This seems like a cleaner approach rather than embedding constants in the context and adding more filters to maintain.

PR #3033, opened 2020-01-29
fix(Models/Post_Types/Event.php) tz perf and all-day

lucatume:
Ticket: https://moderntribe.atlassian.net/browse/TEC-3209

This fix adds logic, in the Event model, to handle the instance where:

* the timezone setting is "site wide"
* an event has a diff. timezone from the site one
* the event is an all-day one

When the above scenario is fullfilled, then all-day events have a chance
of becoming multi-day events too.

E.g. with a cut-off of 00:00, a New York (UTC-5), all-day event on 1/29
will start at UTC 1/29 5am and end on UTC 1/30 4:59AM, making it a
2-day, multiday event in a site that uses UTC timezone.

The fix includes logic to handle end-of-day cutoffs too; in the previous
example, given a 6am end-of-day cutoff, the event would NOT be
multi-day.


lucatume:
Closing this PR as deprecated by the work done in `release/B20.01.1`.

PR #3054, opened 2020-02-06
Correct SQL to use correct %s instead of %d + Safety checks | TEC-3233

bordoni:
🎟 [TEC-3233]
---
Modified the code to prevent checks on empty strings, `0` or `NULL` for `_EventEndDate` as well as the original culprit of the problem which was the prepare variable holder being `%d` instead of `%s` for post type.

This bug was originally found on:
https://wordpress.org/support/topic/website-pages-moved-to-trash-daily/

With some debugging we figured how it was happening.

[TEC-3233]: https://moderntribe.atlassian.net/browse/TEC-3233


tests/wpunit/Tribe/Events/Event_Cleaner_SchedulerTest.php
```diff
@@ -0,0 +1,134 @@
...
+	 * @return Cleaner_Scheduler
+	 */
+	protected function make_instance() {
+		return new Cleaner_Scheduler();
+	}
+
+	/**
+	 * @test
+	 */
+	public function it_should_delete_only_old_posts() {
```
borkweb:
```suggestion
	public function it_should_delete_only_old_events() {
```

PR #3056, opened 2020-02-07
Revert "Task/tec 3125 remove full styles"

Camwyn:
Reverts moderntribe/the-events-calendar#3053


Camwyn:
Reminder that we need to revert this revert the next sprint.

PR #3058, opened 2020-02-10
Feature/tec 3072 move live refresh controls

Camwyn:
Move live refresh to "Filter Activation".
Move "Filter Activation" to the Filter Bar settings if new views are active.

📸 https://d.pr/kiOBML

PR against B20.02 for now as that is what I branched from (minimal changeset)


src/Tribe/Editor/Compatibility.php
```diff
@@ -140,7 +140,7 @@ public function insert_toggle_blocks_editor_field( $fields = array() ) {
 			),
 		);
 
-		return Tribe__Main::array_insert_after_key( $insert_after, $fields, $insert_data );
+		return Tribe__Main::array_insert_before_key( $insert_before, $fields, $insert_data );
```
lucatume: Not required now, but sharing the mental note here: we should move this to the `Tribe__Utils__Array` class, redirect to that class and put a deprecated notice to note it.

PR #3062, opened 2020-02-11
Make sure 404s are respected from WordPress base.

bordoni:
🎟 [TEC-3242]
---
In branch QA happening here.
We need to make sure 404's are respected for which template is loaded.

[TEC-3242]: https://moderntribe.atlassian.net/browse/TEC-3242

### DIFF
--- a/src/Tribe/Views/V2/Template_Bootstrap.php
+++ b/src/Tribe/Views/V2/Template_Bootstrap.php
@@ -295,7 +295,7 @@ class Template_Bootstrap {
 			return false;
 		}
 
-		if ( is_404() ) {
+		if ( $query->is_404() ) {
 			static::$cache['should_load'][ $query->query_vars_hash ] = false;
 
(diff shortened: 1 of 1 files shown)


bordoni:
@vicskf just gave me the Ok on the QA for this.

src/Tribe/Views/V2/Template_Bootstrap.php
```diff
@@ -381,6 +389,10 @@ public function filter_add_body_classes( $classes ) {
 	 * @return string
 	 */
 	public function filter_template_file( $file, $name, $template ) {
+		if ( is_404() ) {
```
Camwyn: Don't need this one to be `$query->is_404()` as well?

bordoni: Not on this case because we have the global WP Query there already.

PR #3064, opened 2020-02-13
Fix/tec 3244 remove customizer text section

jesseeproductions:
_Ref:_ [TEC-3244](https://moderntribe.atlassian.net/browse/TEC-3244)

No artifact as it is deprecating class and moving functionality to general tab

Relates:
https://github.com/moderntribe/events-pro/pull/1500

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -615,8 +615,6 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 
 			if ( ! tribe_events_views_v2_is_enabled() ) {
 				tribe_singleton( 'tec.customizer.widget', new Tribe__Events__Customizer__Widget() );
-			} else {
-				tribe_singleton( 'tec.customizer.text', new Tribe__Events__Customizer__Text() );
 			}
 
 			/**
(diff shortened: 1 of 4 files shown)


src/Tribe/Main.php
```diff
@@ -615,8 +615,6 @@ public function bind_implementations(  ) {
 
 			if ( ! tribe_events_views_v2_is_enabled() ) {
 				tribe_singleton( 'tec.customizer.widget', new Tribe__Events__Customizer__Widget() );
-			} else {
-				tribe_singleton( 'tec.customizer.text', new Tribe__Events__Customizer__Text() );
 			}
```
jesseeproductions: Is there a way to deprecate tribe_singleton( 'tec.customizer.text', new Tribe__Events__Customizer__Text() );

Pro calls it and if they do not update Core and Pro at the same time it could trigger a fatal. ( if we do not require the latest version )

lucatume: I think your addition of a `_deprecated_file` later is fine.

src/deprecated/Tribe__Events__Customizer__Text.php
```diff
@@ -1,4 +1,6 @@
 <?php
+_deprecated_file( __FILE__, 'TBD' );
```
lucatume: We should suggest what to use in place of this file if we can.

jesseeproductions: Added the alternative to the class where the feature moved too.

PR #3070, opened 2020-02-17
Re-enable PHPCS GitHub Workflow with proper phpcs.xml

borkweb:
This removes some superfluous rule declarations (they are included in other rules), moves TribalScents to the last included rule, and properly excludes the short array syntax sniff.

This does not impact released code so will require no QA other than by the GitHub Workflow doing its thing.


phpcs.xml
```diff
@@ -1,20 +1,17 @@
...
 		<exclude name="Generic.Arrays.DisallowShortArraySyntax"/>
 	</rule>
+	<rule ref="TribalScents"></rule>
 
 	<!--exclude the rule for violation of direct DB calls as some have no alternative-->
 	<rule ref="WordPress.DB.DirectDatabaseQuery">
 		<exclude-pattern>src/Test.php</exclude-pattern>
 	</rule>
+
+	<exclude-pattern>*/tests/_support/_generated/*</exclude-pattern>
```
lucatume: Should we not exclude the whole tests folder?

borkweb: Perhaps. Though, it'd be nice if our tests also conformed to coding standards.

lucatume: Then, can we skip, for tests, the commenting ones?
Tests will have weird doc formats and tags.

borkweb: Added here: https://github.com/moderntribe/the-events-calendar/pull/3070/files#diff-6b216d983cf0a62e9c648934c5687456R5

lucatume: I'm sure more will pop up as we use this, but fine for the time being.

PR #3074, opened 2020-02-18
Freemius - Properly catch users email activation

bordoni:
🎟 [TEC-3218]
---
So there isn't a video here because I wasn't able to reproduce the problem locally in any capacity, my user gets redirected to the correct page, but I worked backward from what Joel sent me as the page he was getting redirected to.

The goal is to make sure the user lands into `tribe-common` admin page instead of `tribe-common-account` which doesn't exist.

I will ask Joel to do a round of in-branch QA on this since we dont have videos.

[TEC-3218]: https://moderntribe.atlassian.net/browse/TEC-3218

### DIFF
--- a/src/Tribe/Integrations/Freemius.php
+++ b/src/Tribe/Integrations/Freemius.php
@@ -88,6 +88,8 @@ class Tribe__Events__Integrations__Freemius {
 		if ( ! is_admin() ) {
 			return;
 		}
+		// Setup possible redirect.
+		add_action( 'wp_loaded', [ $this, 'action_redirect_incorrect_page' ] );
 
 		global $pagenow;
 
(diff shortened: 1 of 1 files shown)


src/Tribe/Integrations/Freemius.php
```diff
@@ -172,6 +174,41 @@ public function setup() {
...
+		}
+
+		$page = tribe_get_request_var( 'page', false );
+
+		if ( 'tribe-common-account' !== $page ) {
+			return;
+		}
+
+		$url = admin_url( 'admin.php' );
+		$url = add_query_arg( [
```
tr1b0t: :no_entry_sign: Opening parenthesis of a multi-line function call must be the last content on the line

<sub>hash: 6cd500bdd79aebed48371eed2ac30f58</sub>

src/Tribe/Integrations/Freemius.php
```diff
@@ -172,6 +174,41 @@ public function setup() {
...
+			'fs_action' => $action,
+			'page'      => $this->page,
+			'_wpnonce'  => tribe_get_request_var( '_wpnonce' ),
+		], $url );
```
tr1b0t: * :no_entry_sign: Only one argument is allowed per line in a multi-line function call
* :no_entry_sign: Closing parenthesis of a multi-line function call must be on a line by itself

<sub>hash: 58f617e4db713f40e599b0691b52b7cf</sub>

src/Tribe/Integrations/Freemius.php
```diff
@@ -172,6 +174,41 @@ public function setup() {
...
+
+		wp_safe_redirect( $url );
```
tr1b0t: :no_entry_sign: `wp_safe_redirect()` should almost always be followed by a call to `exit;`.

<sub>hash: da684f39406efe80900736295c73850d</sub>

lucatume:
@bordoni would [could this code](https://github.com/moderntribe/tribe-common/blob/23b901fc98542d2d5c82a69dda1b94cd0e3fc3a4/vendor/freemius/includes/class-freemius.php#L19897-L19900) be responsible?

I'm not really that expert when it comes to Freemius, but it seems like, in code like that and other places in the code, the `_get_admin_page_url` method is used to generate some return links, like the one in question.

I'd take a look at that code and see what return links are produced in different cases.

bordoni:
Most likely that is the responsible but I don’t think we can change that code. Or am I missing something?

PR #3082, opened 2020-03-04
add more to the documented function definition

zbtirrell:
Working with a customer, we needed these extra couple fields, but they weren't otherwise documented.


src/functions/advanced-functions/event.php
```diff
@@ -33,8 +33,10 @@
 	 * - EventShowMap bool - Set to true to embed the map in the Event view.
 	 * - EventCost string - Default cost of the Event.
 	 * - EventURL string - Link to the Event Website or Third-Party page
+ 	 * - FeaturedImage string - URL or ID of a featured image
```
tr1b0t: :no_entry_sign: Tabs must be used to indent lines; spaces are not allowed

<sub>hash: f3b103628f2247b71f57377a21e819ec</sub>

barryhughes: This is just an observation about the docblock in general—and not a comment on your change specifically—but, for cases like this, we should _probably_ be following this convention:

```php
/**
 * Create an event. ...
 *
 * @param array $args {
 *     @type string $EventStartDate Start date of event (required).
 *     @type string $EventEndDate End date of event (required).
 *     ...
 * }
 *
 * @otherTags
 */
```

I am pretty sure this jives nicely with WP Parser (thinking docs site). See: [make.wordpress.org "Inline Documentation Standards (parameters that are arrays)."](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-1-parameters-that-are-arrays)

lucatume: I confirm, proper doc block structure, while painful to type, has that structure.

Also: all doc-block lines should end in a full-stop, a question mark or an exclamation point.

src/functions/advanced-functions/event.php
```diff
@@ -62,16 +103,19 @@ function tribe_create_event( $args ) {
...
+	 * @category Events
 	 *
-	 * @return int|bool ID of the event that was created. False if update failed.
 	 * @link     http://codex.wordpress.org/Function_Reference/wp_update_post
+	 *
 	 * @see      wp_update_post()
 	 * @see      tribe_create_event()
-	 * @category Events
+	 *
+	 * @param int|bool   $postId  ID of the event to be modified.
```
tr1b0t, line 115: :no_entry_sign: Expected 1 spaces after parameter type; 3 found

<sub>hash: c26d7fe54130f3ffbe43aa3078a731d3</sub>

src/functions/advanced-functions/event.php
```diff
@@ -62,16 +103,19 @@ function tribe_create_event( $args ) {
...
+	 * @param array      $args    Args for updating the post. See {@link tribe_create_event()} for more info.
```
tr1b0t, line 116: :no_entry_sign: Expected 4 spaces after parameter type; 6 found

<sub>hash: 2b825b83791679d48f562959a7815e19</sub>

PR #3091, opened 2020-03-09
When removing a venue properly reset select fields as well.

bordoni:
🎟 [BTRIA-251]
---

@msawicki reported earlier last week a problem we have on our Event classic editor, where the possibility of removing venues that had a State from the US selected would trigger the creation of a new Venue with Empty Title.

See it in action here: http://p.tri.be/faDgnQ

[BTRIA-251]: https://moderntribe.atlassian.net/browse/BTRIA-251


lucatume:
@bordoni please create a ticket for this, add all the details in there and notify Strategy of it.
Then come back and add the ticket number in the changelog entry.

bordoni:
For sure will add the ticket. Talking ot Leah and Neill about it.

PR #3105, opened 2020-03-18
Task/price block split state

paulmskim:
**Ticket:** [TEC-3343]

Do not merge until QA passes.

This PR works on the block editor data management for TEC ([TEC-3021]). The specific block affected include the price block.

[TEC-3343]: https://moderntribe.atlassian.net/browse/TEC-3343


lucatume:
@paulmskim this could really use some notes for the reviewers about the over-arching changes you're making here.
No hand-holding, but a rope to follow.

src/modules/blocks/event-price/template.js
```diff
@@ -92,39 +88,43 @@ const renderLabel = ( props ) => {
...
+	attributes,
+	setAttributes,
+} ) => {
+	const setDescription = event => setAttributes( { costDescription: event.target.value } );
+
+	return (
+		<Dashboard isOpen={ isOpen }>
+			<Fragment>
+				<section className="tribe-editor__event-price__dashboard">
+					<input
```
lucatume: Not knowing the whole context this suggestion might be wrong; if that is the case, ignore it.

Might it be worth refactoring the inputs into a dedicated component that would allow us, down the line, to have one place to update the input code; e.g. to add a class or some attributes that we might need to identify them or to target them.

paulmskim: how do you suggest? i'm not sure i understand what the drawbacks of keeping it this way would be and the benefits of refactoring into a dedicated component.

lucatume: The drawback is that, if we have very different usage of input components, we might run into scenarios where we use the base, bloated component and have to remove stuff we do not need.
On the upside, if our inputs are pretty much the same across the board, if we unify them in a, say, `<Input>` component, we might benefit from enhancements to that base component across all plugins using it.
I'm thinking validation, masking, notices, type hints and so on.

This is, but a consideration that might be dismissed if we use inputs in a few places and updating the inputs would not take much time.

I've posted this as a question as I really do not have the same knowledge of the current components, and their usages, as you do.

paulmskim: Makes sense.

I only count 20 uses of `<input>` in the TEC block editor code. Not a small amount, but not overwhelming. I can see the case for things like type-hinting, etc., but I think things like validation are best left for the BE (FE is an insecure place to do it).

There are also places where `<input>` is used that don't relate to this PR, so I'd like to leave it until we want to put in some changes for the existing inputs.

PR #3108, opened 2020-03-18
Task/venue block split state

paulmskim:
**Ticket:** [TEC-3342]

Do not merge until QA passes.

This PR works on the block editor data management for TEC ([TEC-3021]). The specific block affected include the venue block.

[TEC-3342]: https://moderntribe.atlassian.net/browse/TEC-3342


src/modules/blocks/event-venue/template.js
```diff
@@ -86,14 +86,6 @@ class EventVenue extends Component {
 		}
 	}
 
-	componentWillUnmount() {
-		// TODO: this does not work as intended. If one deletes a block, then adds
```
bordoni: Was this addressed?

paulmskim: No, this comment is not actually correct, it was actually meant to occur like this:

* If a user removes the block with a venue attached to it, the event still has the venue, but does not show on the single. However, in any of the archive views it still shows up
* If a user removes the venue from the block before removing the block, it completely removes the venue.

PR #3110, opened 2020-03-19
fix(Views/V2/Hooks.php) more robust check before redirection

lucatume:
Issue: https://moderntribe.atlassian.net/browse/BTRIA-271

This commit fixes an issue that would cause Views v2 redirection logic
to apply in the wrong context if a plugin uses the "view" query argument
to build the request URL.

The fix addes two checks:

1. If the request is for an admin URL, then bail.
2. If the request is not for one of the enabled Views, then bail.

The fix includes tests.

Screencast: https://drive.google.com/open?id=14QEocZoBD0HVqbjCSZA60F-_7SCtf2lM&authuser=luca@tri.be&usp=drive_fs


lucatume:
@bordoni since we worked on this together I'm flagging you as a reviewer.
Is there anything in the PR that feels out of place?

src/Tribe/Views/V2/Hooks.php
```diff
@@ -432,6 +437,13 @@ public function filter_redirect_canonical( $redirect_url = null, $original_url =
 			return $redirect_url;
 		}
 
+		// If the request is not for one of the enabled views, then bail.
+		$enabled_views = tribe_get_option( 'tribeEnableViews', array() );
```
bordoni: Lets use the manager method to get this: `Manager::get_publicly_visible_views`

lucatume: Ah, yes, good catch.

lucatume:
This will require an update to make sure tests will pass.

lucatume:
Issue has been otherwise resolved, closing.

PR #3115, opened 2020-03-26
Preventing redirects from updated Views V2 to be too broad

bordoni:
A simple fix reported by the GravityView team when trying to do Gravity Forms Imports it would fail due to our redirect catching their request.

They have already tested that fix on their side, it works, I already tested this piece with the Embeds and Mobile redirects that this piece of code interacts with.

Not a lot to Screencast here, I this should be tested on our normal full pass.


lucatume:
Sorry, I've closed this by mis-click.

PR #3122, opened 2020-04-01
Feature/past events view fe

paulmskim:
[TEC-3386]
[TEC-3387]

This PR covers the FE for past events view. Still waiting on BE to hook up context.

[TEC-3386]: https://moderntribe.atlassian.net/browse/TEC-3386
[TEC-3387]: https://moderntribe.atlassian.net/browse/TEC-3387


lucatume:
> I think this looks good.
> I think we are going to end up removing the components and top bar as recent past view will append to an existing view such as the list view.
> @lucatume can you please review and I think we can merge into the branch and then do a strategy review with Leah to make the final changes.

The current look is definitely weird: 
recent-past-events

But, as long as we're iterating over it, it's fine.

jesseeproductions:
@paulmskim I am going to merge and I will handle the changes from here. Thanks!

PR #3129, opened 2020-04-08
Task/datetime block split state

paulmskim:
**Ticket:** [TEC-3341]

Do not merge until QA passes.

This is based off of https://github.com/moderntribe/the-events-calendar/pull/3127 currently because it depends on https://github.com/moderntribe/the-events-calendar/pull/3127. This PR will be rebased to the bucket branch once https://github.com/moderntribe/the-events-calendar/pull/3127 is merged. The diff will be largely the same.

EDIT: This is now based on the bucket branch.

This PR works on the block editor data management for TEC ([TEC-3021]). The specific block affected include the datetime block.

[TEC-3341]: https://moderntribe.atlassian.net/browse/TEC-3341


src/modules/blocks/event-datetime/controls/container.js
```diff
@@ -11,33 +11,36 @@ import {
...
 	separatorDate: dateTimeSelectors.getDateSeparator( state ),
 	separatorTime: dateTimeSelectors.getTimeSeparator( state ),
-	showTimeZone: dateTimeSelectors.getTimeZoneVisibility( state ),
 	timeZone: dateTimeSelectors.getTimeZone( state ),
 } );
 
-const mapDispatchToProps = ( dispatch ) => ( {
-	...bindActionCreators( dateTimeActions, dispatch ),
-	onTimeZoneVisibilityChange: onTimeZoneVisibilityChange( dispatch ),
+const mapDispatchToProps = ( dispatch, ownProps ) => ( {
```
alliuca, line 28: I'm a fan of destructuring whenever possible. Just coding style here, doesn't really matter especially if inconsistent with the rest of the project.

```suggestion
const mapDispatchToProps = ( dispatch, { setDateTimeAttributes } ) => ( {
```

and then down below `setDateTimeAttributes( { separatorDate: value } )`

src/modules/blocks/event-datetime/controls/template.js
```diff
@@ -29,65 +29,63 @@ import {
...
-			separatorDate,
-			separatorTime,
-			setSeparatorDate,
-			setSeparatorTime,
-			setTimeZone,
-			showTimeZone,
-			timeZone,
-			isEditable,
-		} = this.props;
+	return isEditable && (
```
alliuca, line 45: I think we'd have to be super sure that `isEditable` is required here. If not, I believe it would error out saying something like nothing is return from the render. I usually err on the side of caution by using a conditional ternary operator and return null as fallback.

```suggestion
	return isEditable ? ( ... ) : null
```

src/modules/blocks/event-datetime/subblocks/content/template.js
```diff
@@ -42,219 +43,219 @@ FORMATS.date = settings() && settings().dateWithYearFormat
...
+const renderSeparator = ( props, type, className ) => {
+	const { separatorDate, separatorTime } = props;
 
-		// Bail when not classic
-		if ( ! editor() || ! editor().isClassic ) {
+	switch ( type ) {
+		case 'date-time':
+			return (
+				<span className={ classNames( 'tribe-editor__separator', className ) }>
+					{ ' '.concat( separatorDate, ' ' ) }
```
alliuca, line 60: If we're adding a space before and after, I think we could just do
```suggestion
					{` ${separatorDate} ` }
```

src/modules/blocks/event-datetime/subblocks/content/template.js
```diff
@@ -42,219 +43,219 @@ FORMATS.date = settings() && settings().dateWithYearFormat
...
-	/**
-	 * Renders a separator based on the type called
-	 *
-	 * @param {string} type - The type of separator
-	 *
-	 * @returns {ReactDOM} A React Dom Element null if none.
-	 */
-	renderSeparator = ( type, className ) => {
-		const { separatorDate, separatorTime } = this.props;
+	return showTimeZone && (
```
alliuca, line 162: Same as above regarding ternary `{var} ? (...) : null`

src/modules/data/blocks/datetime/__tests__/sagas.test.js
```diff
@@ -181,7 +181,12 @@ describe( 'Event Date time Block sagas', () => {
 
 		describe( 'onHumanReadableChange', () => {
 			test( 'On change handler when dates are null', () => {
-				const gen = sagas.onHumanReadableChange();
+				const action = {
```
alliuca, line 184: Wonder if this and the next could be cleaned up with a `beforeEach` at the top, with the action object definition and possibly the saga declaration

src/modules/blocks/event-datetime/controls/template.js
paulmskim: `isEditable` will always be passed here, it's initialized before the blocks are loaded. If it is `false`, then it renders `false`, which is nothing. If `true`, then it'll render the template.

PR #3131, opened 2020-04-08
Spotfix/missing periods

andrasguseo:
Found these inconsistencies while looking at translations. Should be an easy approve.
Sorry, no Jira ticket for this.
If needed, then I guess it could be added to: BTRIA-307


andrasguseo:
Note: I'm out this week so might not see the approval, thus someone else might need to merge.

PR #3160, opened 2020-04-28
Code that fixes the incorrect data being passed.

Camwyn:
And code that checks for site data if event data is missing.

The attr() function was being passed an extra param, so it was returning incorrect default values.
The timezone info is saved to the block as attributes - so it's not saved to meta and it's not saved if the block isn't changed. This means that if we get a default value back, we have to check for it in Event->Settings->General.

[TEC-2964]

[TEC-2964]: https://moderntribe.atlassian.net/browse/TEC-2964

### DIFF
--- a/tests/wpunit/Tribe/Events/Editor/TemplateTest.php
+++ b/tests/wpunit/Tribe/Events/Editor/TemplateTest.php
@@ -1,0 +1,99 @@
+<?php
+namespace Tribe\Events\Editor;
+
+use Tribe__Events__Editor__Template;
+
+
+class TemplateTest extends \Codeception\TestCase\WPTestCase {
+	public function setUp() {
+		// before
+		parent::setUp();
+
+		// your set up methods here
+	}
+
+	public function tearDown() {
+		// your tear down methods here
+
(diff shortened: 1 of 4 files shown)


tests/wpunit/Tribe/Events/Editor/TemplateTest.php
```diff
@@ -0,0 +1,114 @@
+<?php
+namespace Tribe\Events\Editor;
+
+use Tribe__Events__Editor__Template;
+
+
+class TemplateTest extends \Codeception\TestCase\WPTestCase {
+	public function setUp() {
```
lucatume, line 8: If this is empty, then remove it

tests/wpunit/Tribe/Events/Editor/TemplateTest.php
```diff
@@ -0,0 +1,114 @@
...
+		// before
+		parent::setUp();
+
+		// your set up methods here
+	}
+
+	public function tearDown() {
```
lucatume, line 15: Remove empty method

PR #3166, opened 2020-05-01
Add EA History filter and fix counts for other content types

sc0ttkclark:
This PR is for [EXT-182] so that we can customize the HTML shown in the source column (to include our custom filters and other custom text). It also fixes the counting logic so that it uses the content type for the record, not hardcoded for events only.

image

[EXT-182]: https://moderntribe.atlassian.net/browse/EXT-182


sc0ttkclark:
FYI I'm very open to adding the Content Type content directly into TEC here. The extension has to show it because we have multiple content types, but it really feels like it could be approached in TEC itself.

sc0ttkclark:
Since this is a nice-to-have for the extension, I'll let Blue team take it from here with a ticket, deciding on whether to add Content Type, and testing instructions.

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1294,13 +1294,19 @@ public function get_event_count( $type = null ) {
 			return 0;
 		}
 
+		$activity_type = 'event';
+
+		if ( ! empty( $this->meta['content_type'] ) ) {
+			$activity_type = $this->meta['content_type'];
```
bordoni, line 1300: Should we have a allowed content type array?

src/Tribe/Aggregator/Record/List_Table.php
```diff
@@ -520,6 +520,16 @@ public function column_source( $post ) {
 			$html[] = '</dl></div>';
 		}
 
+		/**
+		 * Customize the Events > Import > History > Source column HTML.
+		 *
+		 * @since TBD
+		 *
+		 * @param array                                       $html   List of HTML details.
+		 * @param Tribe__Events__Aggregator__Record__Abstract $record The record object.
```
bordoni, line 529: Pass `use Tribe__Events__Aggregator__Record__Abstract as Record_Abstract;` to make docblock smaller.

src/Tribe/Aggregator/Record/Abstract.php
sc0ttkclark: I'm not sure we need one at this point in the code.

src/Tribe/Aggregator/Record/List_Table.php
sc0ttkclark: I usually do that but if I don't use the class in PHP then PhpStorm will remove it on cleanup/auto-format of the file -- docblocks don't count in their checks there.

I can do that here though.

PR #3170, opened 2020-05-06
Remove incompatible usage of func_get_args for PHP 7.0 

bordoni:
🎟 [TEC-3104]
---
No changes in behavior just add compatibility for version PHP 7.0 and above. 
Originally added here: https://github.com/moderntribe/the-events-calendar/pull/2750

[TEC-3104]: https://moderntribe.atlassian.net/browse/TEC-3104

### DIFF
--- a/src/Tribe/Templates.php
+++ b/src/Tribe/Templates.php
@@ -543,8 +543,8 @@ if ( ! class_exists( 'Tribe__Events__Templates' ) ) {
 		 **/
 		public static function getTemplateHierarchy( $template, $args = array() ) {
 			if ( ! is_array( $args ) ) {
+				$passed        = [ $template, $args ];
 				$args          = array();
-				$passed        = func_get_args();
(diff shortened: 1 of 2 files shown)


src/Tribe/Templates.php
```diff
@@ -543,8 +543,8 @@ public static function showInLoops( $query ) {
 		 **/
 		public static function getTemplateHierarchy( $template, $args = array() ) {
 			if ( ! is_array( $args ) ) {
+				$passed        = [ $template, $args ];
```
sc0ttkclark, line 546: This does not seem to be backwards compatible with the $backwards_map usage below where it still supports other arguments coming from func_get_args()

sc0ttkclark:
```suggestion
				$passed        = func_get_args();
```

bordoni: True, I didn't consider that piece. Because we don't have any internal usage for that it seems. Do you feel like just moving this up would solve the issue?

PR #3177, opened 2020-05-19
Spotfix - prevent undefined notices when force cache refresh in tribe…

jesseeproductions:
…_get_events

Resolves an issue when force a cache refresh.

Found when Gustavo and I were creating a test and updating an event.


bordoni:
No changes to the rest API at all, just Travis CI being annoying.

PR #3189, opened 2020-06-03
Filter Event Label in V2 Views & Prevent PHP Notice

jesseeproductions:
_Ref:_ [VE-9](https://moderntribe.atlassian.net/browse/VE-9)
_Ref:_ [VE-111](https://moderntribe.atlassian.net/browse/VE-111)

For VE-9 this filter enables Events to be filtered to Virtual Events on the Views.
FOR VE-111 the url contains a percentage sign causing sprintf to throw a too few argument. This moves the url to an argument to prevent it. 

Related PR:
https://github.com/moderntribe/events-pro/pull/1559


jesseeproductions:
@lucatume We had some coverage of the area in here including add featured to the wp title:
views_integration/Tribe/Events/Views/V2/Template/TitleTest.php

I added a couple more tests to help.

PR #3192, opened 2020-06-05
Add existing-post to linked post option data and use that to determine if 'new fields' should be shown

sc0ttkclark:
Typically when you are not logged in as the owner of a venue or organizer, you may get an empty edit link for a linked post type. This is more pronounced when you are logged out using the Community Events submission form when anonymous submissions are enabled.

This PR addresses this by introducing a new data attribute to know which options are an existing linked post and which are brand new. Only existing linked posts will get the data attribute as when typing to create a new one will have that attribute not set ensuring it's treated as intended (as a new linked post to create with the fields).

[TEC-3481]

# Screenshots from CE usage

All screenshots below are from the event submission form while logged out (with anonymous submissions enabled).

## The venue dropdown initial state

Screen Shot 2020-06-04 at 8 05 04 PM

## The venue dropdown with a new venue being typed (not yet marked for creation)

Screen Shot 2020-06-04 at 8 05 23 PM

## The venue dropdown with a new venue created showing 'new' fields as expected

Screen Shot 2020-06-04 at 8 05 29 PM

## The venue dropdown with an existing venue selected showing no fields as expected

Screen Shot 2020-06-04 at 8 05 44 PM

## BEFORE: The venue dropdown after selecting an existing venue, then going to create a new one showing edit link

Screen Shot 2020-06-04 at 8 34 48 PM

## AFTER: The venue dropdown after selecting an existing venue, then going to create a new one showing no edit link

Screen Shot 2020-06-04 at 8 36 45 PM

[TEC-3481]: https://moderntribe.atlassian.net/browse/TEC-3481

### DIFF
--- a/src/resources/js/events-admin.js
+++ b/src/resources/js/events-admin.js
@@ -387,8 +387,6 @@ jQuery( document ).ready( function( $ ) {
 			existingPost = !! $selected.data( 'existingPost' );
 		}
 
-		$edit.hide();
-
 		if (
 			! existingPost &&
 			'-1' !== value &&
@@ -407,6 +405,8 @@ jQuery( document ).ready( function( $ ) {
 			$group.parents( '.tribe-section' ).addClass( 'tribe-is-creating-linked-post' );
 
 		} else {
+
+			$edit.hide();
 			// Hide all fields and remove their values
 			$group.find( '.linked-post' ).hide().find( 'input, select' ).val( '' );
 
@@ -812,5 +812,4 @@ jQuery( document ).ready( function( $ ) {
 			$el.val( tribeDateFormat( $el.datepicker( 'getDate' ), 'tribeQuery' ) );
 		} );
 	} );
-
-});
+} );
--- a/src/resources/js/events-admin.js
+++ b/src/resources/js/events-admin.js
@@ -387,6 +387,8 @@ jQuery( document ).ready( function( $ ) {
 			existingPost = !! $selected.data( 'existingPost' );
 		}
 
+		$edit.hide();
+
 		if (
 			! existingPost &&
 			'-1' !== value &&
@@ -405,8 +407,6 @@ jQuery( document ).ready( function( $ ) {
 			$group.parents( '.tribe-section' ).addClass( 'tribe-is-creating-linked-post' );
 
 		} else {
-			$edit.hide();
-
 			// Hide all fields and remove their values
 			$group.find( '.linked-post' ).hide().find( 'input, select' ).val( '' );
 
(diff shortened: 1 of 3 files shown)


src/resources/js/events-admin.js
```diff
@@ -373,21 +373,26 @@ jQuery( document ).ready( function( $ ) {
...
+		const selectedVal = $selected.val();
+		let editLink = '';
+		let existingPost = false;
+
+		if ( selectedVal === value ) {
+			editLink = $selected.data( 'editLink' );
+			existingPost = !! $selected.data( 'existingPost' );
 		}
 
+		$edit.hide();
```
sc0ttkclark, line 390: Without this here, the edit link would keep showing after you've selected a venue and go to create a new venue.

src/resources/js/events-admin.js
```diff
@@ -373,21 +373,26 @@ jQuery( document ).ready( function( $ ) {
...
+		const $edit = $group.find( '.edit-linked-post-link a' );
+		const value = $select.val();
+		const $selected = $select.find( ':selected' );
+		const selectedVal = $selected.val();
+		let editLink = '';
+		let existingPost = false;
+
+		if ( selectedVal === value ) {
+			editLink = $selected.data( 'editLink' );
+			existingPost = !! $selected.data( 'existingPost' );
```
sc0ttkclark, line 387: I'm good with adjusting this to remove the `!!` but I wanted to make sure it enforced a boolean even if it was undefined or an integer.

src/resources/js/events-admin.js
bordoni: Lets add a comment about that?

src/resources/js/events-admin.js
bordoni: I like, @paulmskim do you have anything on this?

src/resources/js/events-admin.js
sc0ttkclark, line 390:
```suggestion
		// Always hide the edit link unless we have an edit link to show (handled below).
		$edit.hide();
```

src/resources/js/events-admin.js
paulmskim: What are the possible values for `data-existing-post`? Is it always available? I would keep the `!!`, probably the best way other than using `Boolean()`.

sc0ttkclark: Wow I did a smart thing!

Also the existing post value is output via PHP in this PR as "1" so it's only ever going to be "1" or undefined.

paulmskim: Then this is fine. If you also had `'0'` as an option and you meant that to be false then `!!` would not work. But in this case this is fine.

sc0ttkclark:
Not sure if the Travis fail here is related to this PR

PR #3210, opened 2020-06-15
fix vertical align for month view styles

paulmskim:
Fixes a small style issue pointed out in VE DQA.


bordoni:
Closing this as it seems like it's stale. If we need it @paulmskim please re-open towards B20.10.

PR #3224, opened 2020-06-22
Handle case that rest auth is null (i.e. for SAML)

ahaurat:
In cases where alternate authentication (such as SAML) is being used on a WordPress site, `rest_authentication_errors` may return `null` per https://developer.wordpress.org/reference/hooks/rest_authentication_errors/

Not handling the `null` case causes bugs on sites with alternate authentication because it is being treated the same as `false` (unauthorized) or `WP_Error`. This manifests as 403 errors on the frontend whenever a call is made to the REST API such as when switching from List to Month view on the events page (see screenshot)

Screen Shot 2020-06-22 at 12 01 53 PM


bordoni:
@ahaurat Thanks so Much for your PR.

Sorry for the delay in our response, we had a big release past 3 or 4 days for our Virtual Events.

I added a couple of labels to the PR and our team will handle the Changelog and getting it tested to merged on our next release, this seems like it solves the problem correctly.

bordoni:
🎟 [BTRIA-417]

PR #3232, opened 2020-07-08
Small style tweaks via MattCR

Camwyn:
Make the "band" white.
Reduce the spacing above the "band".

📷 http://p.tri.be/6CNpWr

### DIFF
--- a/src/resources/postcss/freemius.pcss
+++ b/src/resources/postcss/freemius.pcss
@@ -95,6 +95,10 @@
 				&:last-child {
 					margin-bottom: 0;
 				}
+
+				&:last-of-type {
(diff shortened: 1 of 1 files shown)


src/resources/postcss/freemius.pcss
```diff
@@ -95,6 +95,10 @@
 				&:last-child {
 					margin-bottom: 0;
 				}
+
+				&:last-of-type {
```
Camwyn, line 99: @paulmskim the above rule (last-child) doesn't apply because the Freemius logo in the upper corener is actually the last child - so we have extra margin here: http://p.tri.be/t8dow2

PR #3243, opened 2020-07-18
Fix JS console issues, improve shortcode theme  builder support

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-3553

This PR fixes a number of issues I had introduced in my previous
work to support shortcodes in theme builders context, reverts some
of those changes and improves on them to make the code more robust.

[Screencast](https://drive.google.com/open?id=1A5EUWOFiAweq4nN8jjvPU3EeL7WgW1Qm&authuser=luca%40tri.be&usp=drive_fs)

Key points of the PR:

1. Theme builders, when the shortcode is rendered as part of a text field, will make the View HTML output go through the `wptexturize` function.
2. Together w/ the View HTML code we print two `<script>` tags: the `tribe-events-view-data` one, that contains a View template vars, and the one for the View events JSON-LD data.
3. The `wptexturize` function will **not**, by default, escape `<script>` tags, so we should be fine, right? 
4. Nope. First of all, in the `tribe-events-view-data` data we would re-print the JSON-LD data. This would screw the `wptexturize` regular-expression based handling completely by having a `<script>`-ish thing (the JSON-LD data) printed into the data `<script>`.
5. We print some objects in the data (e.g. the View object and the dates) that, once cast to string, would create more noise that would further mess w/ the `wptexturize` regular expressions.
6. I've solved 4 and 5 by removing the JSON-LD data (it's printed by the `components/json-ld-data.php` template anyway) and any object that is not a simple value object from the data.
7. So this works now? Nope.
8. In our `components/data.php` template we had `<script ... \n>`, the script closing bracket on the next line. Guess what? That too messes up the `wptexturize` regular expressions.
9. I've fixed 8 by putting the closing `>` of the opening `<script>` tag on the same line.

Wow.

In the PR, the snapshot updates are following the `components/data` updates.


src/Tribe/Views/V2/Utils/View.php
```diff
@@ -54,4 +54,35 @@ public static function get_past_event_display_key() {
...
+			return $view_data;
+		}
+
+		/*
+		 * Remove the JSON-LD data, it's already printed by the `components/json-ld-data.php` template. Printing a
+		 * `<script>`, the JSON-LD data, inside a `<script>`, the data, will cause issues.
+		 */
+		$view_data = array_diff_key( $view_data, array_flip( [ 'json_ld_data' ] ) );
+
+		// Remove objects that should not be printed on the page, keep data objects.
```
tr1b0t, line 81: * :no_entry_sign: Opening parenthesis of a multi-line function call must be the last content on the line
* :no_entry_sign: Only one argument is allowed per line in a multi-line function call

<sub>hash: 1fd5dafc1b16dfb153b417b665d4a3f5</sub>

src/Tribe/Views/V2/Utils/View.php
```diff
@@ -54,4 +54,35 @@ public static function get_past_event_display_key() {
...
+		$view_data = array_filter( $view_data, static function ( $value ) {
+			return ! is_object( $value ) || $value instanceof \stdClass;
```
tr1b0t, line 83: :no_entry_sign: Closing parenthesis of a multi-line function call must be on a line by itself

<sub>hash: 9b3c6cc2e0752d173f1e3afff828cc10</sub>

PR #3253, opened 2020-08-07
Remove redundant query control

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-3530

This PR removes the Event Query Controller code (including its otherwise unused ancestor and related tests) to solve the issue.

While radical as an approach, it solves the issue and does not seem to interfere with Views v2.
Originally, we had introduced the query controller to make sure the `tribe_suppress_query_filters` property would be set to `true` when running repository queries from the Views.
As the code evolved we've:
1. set that `tribe_suppress_query_filters` explicitly in the Repositories
2. check for that `tribe_suppress_query_filters` property in the `Tribe__Events__Query` methods that would modify the quries.

The two steps could have made the Events Query Controller redundant and, as reported in the issue, overreaching.


src/Tribe/Views/V2/V1_Compat.php
```diff
@@ -56,10 +55,41 @@ public function register() {
...
 	public function remove_v1_filters() {
-		$backcompat        = V1_Backcompat::instance();
-		$tec_bar           = tribe( 'tec.bar' );
+		$filters_to_remove = $this->get_filters_to_remove();
+
+		foreach ( $filters_to_remove as $tag => $filters ) {
+			foreach ( $filters as $filter_data ) {
+				$callback = $filter_data['callback'];
+				$priority = isset( $filter_data['priority'] ) ? $filter_data['priority'] : 10;
+				/*
```
tr1b0t, line 68: :no_entry_sign: Empty line required before block comment

<sub>hash: 601a2faa9d8948406ea188d5fb2ca316</sub>

lucatume:
Closing this PR: the approach taken here is not the correct one.

PR #3260, opened 2020-08-13
VE-150 - move iCal output method to its own method for support of VE …

jesseeproductions:
…exports.

_Ref:_ [VE-150](https://moderntribe.atlassian.net/browse/VE-150)

This moved the iCal out to its own public method.

This enables the output to be tested.
And I added array keys for each field to be able to find fields by the key instead of searching for partial values. 

These changes are for:
https://github.com/moderntribe/events-virtual/pull/177

Quick overview:
https://cloudup.com/cfcNwPmVhiu

PR #3262, opened 2020-08-17
Debug restv1 suite tests.

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-3550

As it happened before, at times the `restv1` test suite will fail seemingly at random.
The purpose of this PR is to debug and explore this failure.

Related PR: #3260

The first fix is one to make sure our code is a bit more robust and resistant to critical runs in the context of nonce generation.
The second issue proved to be a change, that I have located in effect but not in source, in either WordPress with version `5.5`, or Travis CI: uploaded images have inconsistent paths that will yield a number of issues.

To solve the second one, I've opted for a temporary skip of the assertions (already redundant), that I will restore with more time and research.


lucatume:
> Looks fine - any reason we don't just use uopz for this though?

Two reasons:
1. We're not using `uopz` in tests on TEC as we're not using `tric` yet.
2. REST API tests are acceptance type of tests where the scope of the tests and the scope of the code handling the tests requests is not the same: they are two distinct and separate PHP processes.

lucatume:
Closing this PR in favor of #3265

PR #3266, opened 2020-08-20
Fixes non-numeric value encountered warnings

dgwatkins:
The signature for `get_i18n_strings_for_domains` doesn't match what we have here: 
-the 4th parameter is supposed to be $flags, not $locale.

It looks like we wanted to use `get_i18n_strings` instead.

Fixes errors like this show when activating the plugin:
PHP Warning:  A non-numeric value encountered in wp-content/plugins/the-events-calendar/src/Tribe/I18n.php on line 272


lucatume:
Rebasing on `release/B20.09`.

lucatume (approved):

PR #3267, opened 2020-08-20
Handle the translation of settings from wp_options table.

dgwatkins:
We have a “documented bug” here:
https://theeventscalendar.com/knowledgebase/k/creating-translations-for-events-using-wpml/#Translations_in_the_WordPress_admin

and here:
https://wpml.org/errata/the-events-calendar-admin-texts-do-not-translate/

This PR adds that code to WPML integrations class.


lucatume:
Re-basing on `release/B20.09`.

PR #3277, opened 2020-09-03
Feature/fbar 135 show hide filters button

paulmskim:
**Ticket:** [FBAR-135]

This PR removes all mention of filters and tabs from TEC.

**Screencast:** Everything working as it did before

https://drive.google.com/file/d/1cAJZcEs7716jEHWiKSOblbDv647zgEUN/view?usp=sharing

[FBAR-135]: https://moderntribe.atlassian.net/browse/FBAR-135


src/views/v2/components/events-bar.php
```diff
@@ -41,15 +41,11 @@
 		<?php $this->template( 'components/events-bar/search-button' ); ?>
 
 		<div
-			class="tribe-events-c-events-bar__search-filters-container"
-			id="tribe-events-search-filters-container"
-			data-js="tribe-events-search-filters-container"
+			class="tribe-events-c-events-bar__search-container"
```
lucatume, line 44: This change will break back-compat with existing CSS modifications, was this intentional?

src/views/v2/components/events-bar/filters.php
```diff
@@ -1,15 +0,0 @@
-<?php
```
lucatume, line 1: Same as above, removing this file will create issues with back-compatibility, intentional?

lucatume (commented):
I have questions about back-compatibility and how updating or removing CSS classes or templates might affect existing overrides and customizations.

src/views/v2/components/events-bar.php
bordoni: To answer your question I think B20.09 will be a feature release for both TEC and Filterbar

lucatume (approved):

PR #3278, opened 2020-09-07
Allow the accordion script to run out of FE context

lucatume:
This PR modifies the auto-binding function of the accordion script to avoid issues when not using it in the Views Front-end context.

### DIFF
--- a/src/resources/js/views/accordion.js
+++ b/src/resources/js/views/accordion.js
@@ -298,6 +298,11 @@ tribe.events.views.accordion = {};
 	 * @return {void}
 	 */
 	obj.ready = function() {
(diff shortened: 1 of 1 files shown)


src/resources/js/views/accordion.js
```diff
@@ -298,6 +298,11 @@ tribe.events.views.accordion = {};
 	 * @return {void}
 	 */
 	obj.ready = function() {
+		if ( ! tribe.events.views.manager ) {
+			// The script might  be used outside of the Views v2 context, if that is the case, skip the auto-binding.
```
paulmskim, line 302:
```suggestion
			// The script might be used outside of the Views v2 context, if that is the case, skip the auto-binding.
```

PR #3280, opened 2020-09-14
Fix some WPML integration issues and add default value for translation flags.

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-3578

This PR fixes some smaller issues I found while inspecting the issue that would make the translation
of the site slugs incorrect when using the WPML integration.

[Screencast (no audio)](https://drive.google.com/open?id=1D3Z08GM1ucS69CtZAsL70kd6P-Q4RTtP&authuser=luca%40tri.be&usp=drive_fs)

### DIFF
--- a/src/Tribe/Rewrite.php
+++ b/src/Tribe/Rewrite.php
@@ -317,6 +317,9 @@ class Tribe__Events__Rewrite extends Tribe__Rewrite {
 		 */
 		$bases = apply_filters( 'tribe_events_rewrite_i18n_slugs_raw', $bases, $method, $domains );
 
+		// Again, make sure the bases are unique.
+		$bases = array_map( 'array_unique', $bases );
+
 		if ( 'regex' === $method ) {
 			foreach ( $bases as $type => $base ) {
 				// Escape all the Bases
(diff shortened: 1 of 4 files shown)


src/Tribe/Rewrite.php
```diff
@@ -317,6 +317,9 @@ public function get_bases( $method = 'regex' ) {
 		 */
 		$bases = apply_filters( 'tribe_events_rewrite_i18n_slugs_raw', $bases, $method, $domains );
 
+		// Again, make sure the bases are unique.
+		$bases = array_map( 'array_unique', $bases );
```
Camwyn, line 321: This is good here, but I think you can remove it from above (line 286) - it's all filters except for `merge_localized_bases()` - which _also_ runs array_unique on `$bases`?

Not needed for this PR 😉 just an observation.

Camwyn (approved):
Looks legit.

src/Tribe/Rewrite.php
lucatume: It's actually required as filters can alter (and they do) the bases to add redundancy.

PR #3289, opened 2020-09-24
Ensure List View displays in ASC order

lucatume:
Issue: https://moderntribe.atlassian.net/browse/ECP-467 

[Screencast](https://drive.google.com/open?id=1Fi6u79JEMep5HskfRIVZmQI7uoAAWNzD&authuser=luca%40tri.be&usp=drive_fs)
[Screeenshot of PRO tests passing](https://drive.google.com/open?id=1FWKM3vg9ETuWLyXevCuZzycf13syLmPV&authuser=luca%40tri.be&usp=drive_fs)

This PR fixes an issue where List View would not display events in ASCending order when the PRO setting "Recurring Event Instances" > "Show only the first instance of each recurring event" is activated.
The issue originated from the fact that List View would assume the default repository order direction would be set to ASC (correct in most cases), but would not enforce it.
The order would, but, be DESC when that setting is activated (due to how that setting works) and List View would neither check nor enforce that.

Since the fix for a PRO issue is in TEC, I've provided a screenshot of the passing tests on PRO to show that nothing is breaking there.

The fix simply enforces that order.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -224,7 +224,8 @@ Remember to always make a backup of your database and files before updating!
 
 = [5.2.0] 2020-09-28 =
 
-* Fix - Fix a PHP error that would arise when during the translation, by the WPML plugin, of some strings. [TEC-3454]
+* Fix - Fix the order of the List View when the PRO is set to show only the first instance of a Recurring Event. [ECP-467]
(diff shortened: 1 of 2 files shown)


jesseeproductions:
I am good with this change. Can this be set to Ready for Review?

PR #3308, opened 2020-10-21
Fix regular expression for parsing column names

mitogh:
Expresions such as:

`keyword(s)` where replaced into `keywords)`

This due the regular expresion was not global and only the first
ocurrence was removed from the text, in this case a global expresion is
required in order to make:

`keyword(s)`  into `keywords`

Artifact: https://d.pr/2HRySa


lucatume (commented):
Could you provide a before/after screenshot or screencast? It's not really clear what this is affecting.

lucatume (changes requested):
The fix is fine, thanks. Please add a changelog entry.

lucatume (dismissed):

lucatume:
Merging this one, it looks like the jslint issue is not related to the changes and is having hiccups.

PR #3309, opened 2020-10-21
Removal of non required variable

mitogh:
The variable it was never used other than for directly accesing inside
of the filter call, in this case having a variable sitting around for no
use does not provide any additional value.


lucatume (commented):
Looks fine, but hold it a bit while the next release goes out.

lucatume (approved):

PR #3314, opened 2020-10-23
Ensure Cron schedules is an array.

mmcachran:
This error occurred on a vanilla install of WP. An error was thrown for the Event Aggregator because $schedules was not being passed into the hook as an array.


lucatume:
@mmcachran thanks for taking the time to debug, fix and open a PR. Sorry for the delay in the reply: this went under my radar.
I'm re-basing the PR on the next release branch and let the tests run, I will handle it from here and will include it in a "package" release to make sure all boxes are checked.

Again, thanks.

PR #3317, opened 2020-10-26
Use $this->slug when generating prev/next URLs for List View | TEC-3648

borkweb:
With this small change (which is present on a couple of other views), we make the List View extendable when creating new views—which should make new views more DRY.

:movie_camera: http://p.tri.be/LsTnFg
:ticket: [TEC-3648]

[TEC-3648]: https://moderntribe.atlassian.net/browse/TEC-3648


Camwyn (approved):
Nice!

lucatume (approved):

PR #3323, opened 2020-10-29
Wrap and document PR #3314

lucatume:
This PR just wraps, adding documentation, [the PR from @mmcachran](https://github.com/moderntribe/the-events-calendar/pull/3314)
to add changelog entry and tracking for it.


src/Tribe/Aggregator/Cron.php
```diff
@@ -192,18 +192,21 @@ public function action_register_cron() {
...
 	 * to make sure we can insert without having to expire cache.
 	 *
-	 * @param  array $schedules
+	 * @param array $schedules Current list of schedules.
 	 *
-	 * @return array
+	 * @return array            Modified list of schedules.
 	 */
-	public function filter_add_cron_schedules( array $schedules ) {
+	public function filter_add_cron_schedules( $schedules = [] ) : array {
```
Camwyn, line 199:
```suggestion
	public function filter_add_cron_schedules( $schedules = [] ) {
```
That's a php 7+ syntax - unless we're now requiring php 7+ we shouldn't do this?

src/Tribe/Aggregator/Cron.php
```diff
@@ -192,18 +192,21 @@ public function action_register_cron() {
...
+		// Ensure schedules is an array.
+		$schedules = is_array( $schedules ) ? $schedules : [];
```
Camwyn, line 201: php 7

Camwyn (changes requested):

src/Tribe/Aggregator/Cron.php
lucatume: Correct, thanks @Camwyn

Camwyn (approved):
Looks good 👍

PR #3336, opened 2020-11-11
Sync Bucket Branch with the release branch.

mitogh:


bordoni (dismissed):

lucatume:
@mitogh correct me if I'm wrong, please; this is just an automatic sync branch, right? Let me know if/when you would actually rqeuire a code review on this?

mitogh:
Actually this one does not require a CR as is basically all the changes already approved and merged into `release/B20.11` into the bucket brach for batch pushing as any `bucket/*` branch is protected we neeed to create a PR in order to merge into that branch so we should be good with moving forward with this one. 

Let me knwo in case you have questions abour this PR, thanks. @lucatume

PR #3343, opened 2020-11-16
Fix/TEC-3635 default limit by number

codingmusician:
JIRA Ticket: [TEC-3635](https://moderntribe.atlassian.net/browse/TEC-3635)

* Having the default import limit setting be by number instead of date range is a better UX for our customers


mitogh (changes requested):
This looks good but looks like you are missing this behavior.

> New Default = by number of events (1000)

^ This is from the AC of the ticket. https://moderntribe.atlassian.net/browse/EA-339

The default is `200` at the moment.

codingmusician:
Good catch. Thanks.

mitogh (dismissed):

codingmusician:
@mitogh Looks like something happened with the base branch earlier that is requiring your approval again. Thanks!

mitogh (dismissed):

mitogh:
That usually happens every time new things are merged into the base branch :) 

Let me know in case you face the same issue again :+1:

src/Tribe/Aggregator/Settings.php
```diff
@@ -657,7 +657,7 @@ public function get_import_limit_count_default() {
 		 *
 		 * @param int
 		 */
-		return apply_filters( 'tribe_aggregator_import_count_default', 200 );
+		return apply_filters( 'tribe_aggregator_import_count_default', 1000 );
```
bordoni, line 660: Why are we changing to 1000? It might break peoples imports, because of timeouts on the server level.

bordoni (changes requested):

src/Tribe/Aggregator/Settings.php
mitogh: This is a change for new installs, so users with TEC installed will remain with the current value of 200, so is backwards compatible. Reasons are described on the ticket as User Stories. 

But you might want to jump into q call with me an Geoff to discuss your concerns with the new default :) so we can find a smaller number. 

Let me know if that makes sense, thanks for your question 🙏

bordoni (approved):

PR #3352, opened 2020-11-19
Fix/tec 3649 theme overrides 2

paulmskim:
**Ticket:** [TEC-3649]

This PR fixes theme overrides that were incorrectly applied. The WooCommerce plugin would add the body class `theme-twentyseventeen`, for example, to the body. However, this was an incorrect selector to use, and has been removed.

**Artifacts:**

Astra:
image

Enfold:
image

Twenty Seventeen:
image

[TEC-3649]: https://moderntribe.atlassian.net/browse/TEC-3649


paulmskim:
merging this to keep work moving, if there are any changes to make pls let me know.

PR #3360, opened 2020-11-24
Update logic so v2 widget constant is not filtered

paulmskim:
**Ticket:** N/A

Fixes logic so that widget v2 constant is not filtered.

**Artifact:** (using a screencast of another ticket, but uses this logic) https://drive.google.com/file/d/1FTsHsdhy6zT8SAosKXBbjhVValN-Bchu/view?usp=sharing


src/functions/views/provider.php
```diff
@@ -131,21 +131,23 @@ function tribe_events_widgets_v2_is_enabled() {
...
 	/**
-	 * Allows toggling of the v2 widget views via a filter.
+	 * Allows toggling of the v2 widget views via a filter. Defaults to true.
 	 *
 	 * @since TBD
 	 *
 	 * @return boolean Do we enable the widget views?
 	 */
-	return apply_filters( 'tribe_events_widgets_v2_is_enabled', $enabled );
+	return apply_filters( 'tribe_events_widgets_v2_is_enabled', true );
```
Camwyn, line 152: So the constant doesn't get filtered this way?

paulmskim: Wasn't that the result of the discussion in our products dev call?

Camwyn: I had to re-read it to be sure. This is fine.

lucatume (approved):

Camwyn (approved):

PR #3306, opened 2020-10-20
TEC-3633, EA-343: Batch pushing implementation on TEC

mitogh:
Updates on the TEC code to support the new batch pushing system.

Tickets:
-  [TEC-3633]
-  [EA-343]

[TEC-3633]: https://moderntribe.atlassian.net/browse/TEC-3633

[EA-343]: https://moderntribe.atlassian.net/browse/EA-343

PR #3368, opened 2020-11-25
Batch Pushing

mitogh:
Updates to support batch pushing as `Cron` mechanism for EA Client.


src/Tribe/Aggregator/Cron.php
```diff
@@ -19,7 +19,7 @@ class Tribe__Events__Aggregator__Cron {
 	 * Limit of Requests to our servers
 	 * @var int
 	 */
-	private $limit = 5;
+	private $limit = 25;
```
bordoni, line 22: This could be problematic, depending on the server. I wonder why specifically we are changing this.

src/Tribe/Aggregator/Cron.php
```diff
@@ -388,6 +390,81 @@ public function verify_child_record_creation() {
...
+	 * @since TBD
+	 */
+	public function start_batch_pushing_records() {
+		if ( ! tribe( 'events-aggregator.main' )->is_service_active() ) {
+			return;
+		}
+
+		$records = Tribe__Events__Aggregator__Records::instance();
+
+		$query = $records->query(
```
bordoni, line 406: Move this query into it's own method.

src/Tribe/Aggregator/Cron.php
```diff
@@ -388,6 +390,81 @@ public function verify_child_record_creation() {
...
+			[
+				'post_status'    => Tribe__Events__Aggregator__Records::$status->pending,
+				'posts_per_page' => 250,
```
bordoni, line 409: This could be extremely painful depending on what we get back. There are 3 meta values here.

src/Tribe/Aggregator/Processes/Batch_Imports.php
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+/**
+ * Class Tribe__Events__Aggregator__Processes__Batch_Imports
+ *
+ * Add custom hooks in order to support batch pushing.
+ *
+ * @since TBD
+ */
+class Tribe__Events__Aggregator__Processes__Batch_Imports {
```
bordoni, line 18: If we are introducing classes we need to use namespacing.

src/Tribe/Aggregator/Processes/Batch_Imports.php
```diff
@@ -0,0 +1,102 @@
...
+		// Make sure $parent_id is always an integer.
+		if ( $parent_id instanceof WP_Post ) {
+			$parent_id = $parent_id->ID;
+		}
+
+		$allow_batch_pushing = get_post_meta(
+			$parent_id,
+			Tribe__Events__Aggregator__Record__Abstract::$meta_key_prefix . 'allow_batch_push',
+			true
+		);
```
bordoni, line 74: Lets have a method to check this for a particular post.

src/Tribe/Aggregator/Processes/Batch_Imports.php
```diff
@@ -0,0 +1,102 @@
...
+			return $service_supports_batch_push;
+		}
+
+		return false;
+	}
+
+	/**
+	 * Update the args used for imports.
+	 *
+	 * TODO: Update EventBrite to use batch pushing to deliver events instead.
```
bordoni, line 86: Make the todo a `@todo`, possibly set a ticket in Jira with the label `dev-todo`

src/Tribe/Aggregator/Processes/Service_Provider.php
```diff
@@ -16,6 +16,8 @@ public function register() {
 		tribe_register( 'events-aggregator.record-items', 'Tribe__Events__Aggregator__Record__Items' );
 		tribe_register( 'events-aggregator.processes.import-events', 'Tribe__Events__Aggregator__Processes__Import_Events' );
 		tribe_singleton( 'events-aggregator.queue-control', 'Tribe__Events__Aggregator__Processes__Queue_Control' );
+		$batch_imports = new Tribe__Events__Aggregator__Processes__Batch_Imports();
+		tribe_singleton( Tribe__Events__Aggregator__Processes__Batch_Imports::class, $batch_imports );
```
bordoni, line 20: Not a good idea to trigger the initialization of the class here, it defeats the purpose of using a service provider.
```suggestion
		tribe_singleton( Tribe__Events__Aggregator__Processes__Batch_Imports::class, Tribe__Events__Aggregator__Processes__Batch_Imports::class );
```

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1519,7 +1551,12 @@ public function insert_posts( $items = [] ) {
 			if ( ! empty( $event['Venue']['Venue'] ) ) {
 				$event['Venue']['Venue'] = trim( $event['Venue']['Venue'] );
 
-				if ( ! empty( $item->venue->global_id ) || in_array( $this->origin, array( 'ics', 'csv', 'gcal', 'ical' ) ) ) {
+				if ( ! empty( $item->venue->global_id ) || in_array( $this->origin, [
+						'ics',
+						'csv',
+						'gcal',
+						'ical',
+					] ) ) {
```
bordoni, line 1559: Save this into a variable, this style of line breaking makes it hard to read.

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -2015,14 +2052,17 @@ public function insert_posts( $items = [] ) {
...
+			 * @param self  $record The record we are dealing with
 			 */
 			do_action( 'tribe_aggregator_after_insert_post', $event, $item, $this );
 		}
 
-		remove_filter( 'tribe-post-origin', array( Tribe__Events__Aggregator__Records::instance(), 'filter_post_origin' ), 10 );
+		remove_filter( 'tribe-post-origin', [
+			Tribe__Events__Aggregator__Records::instance(),
+			'filter_post_origin',
+		], 10 );
```
bordoni, line 2065: Either all arguments are in a new line or none.

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+use Tribe__Events__Aggregator__Record__Queue_Cleaner;
+use Tribe__Events__Aggregator__Record__Queue_Interface;
+use Tribe__Events__Aggregator__Records;
+use Tribe__Events__Aggregator__Service;
+use Tribe__Events__Main;
+use WP_Error;
+use WP_Post;
+
+// Prevent to access the file directly.
+defined( 'WPINC' ) || die;
```
bordoni, line 26: No need, that is an old requirement of WP.

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+	 * @since TBD
+	 *
+	 * @return $this
+	 */
+	protected function complete() {
+		try {
+			// Updates the Modified time for the Record Log.
+			$args = [
+				'ID'            => $this->record->post->ID,
+				'post_modified' => ( new DateTime( 'now', new DateTimeZone( 'UTC' ) ) )->format( Tribe__Date_Utils::DBDATETIMEFORMAT ),
```
bordoni, line 188: Use `Tribe__Date_Utils::build_date_object` method so we can avoid the try catch.

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+	 * @throws Exception
+	 *
+	 * @param null $batch_size The batch size is ignored on batch import as is controlled via the initial filtered value.
+	 *
+	 * @return self|Tribe__Events__Aggregator__Record__Activity
+	 */
+	public function process( $batch_size = null ) {
+		// This batch has not started yet, make sure to initiate this import.
+		if ( empty( $this->record->meta['batch_started'] ) ) {
+			$now = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
```
bordoni, line 217: Use `Tribe__Date_Utils::build_date_object`

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+
+			return;
+		}
+
+		$version                 = $service->api['version'];
+		$service->api['version'] = 'v2.0.0';
+
+		$body = [
+			'batch_size'       => apply_filters( 'event_aggregator_event_batch_size', 10 ),
+			'batch_interval'   => apply_filters( 'event_aggregator_event_batch_interval', 10 ),
```
bordoni, line 271: Needs doc blocks, and likely needs to be on it's own method.

src/Tribe/Aggregator/Service.php
```diff
@@ -161,8 +161,18 @@ public function build_url( $endpoint, $data = [] ) {
...
+		 *
+		 * @since TBD
+		 *
+		 * @param string   $url      The completed URL generated.
+		 * @param string   $endpoint The path of the endpoint inside of the base url.
+		 * @param stdClass $api      An object representing the properties of the API.
+		 *
+		 * @return string The generated URL.
+		 */
+		$url = apply_filters( 'tribe_events_aggregator_build_url', "{$api->domain}{$api->path}{$api->version}/{$endpoint}/", $endpoint, $api );
```
bordoni, line 175: I wonder if this should be a method.

bordoni (changes requested):

src/Tribe/Aggregator/Cron.php
mitogh: With batch pushing we are increasing the number of requests back to the server with a regular set of imports we only have a single call and the delivery is processed on the client, in this case if a user has several schedule imports with batch pushing the client will increase the number of calls back to EA server if we set this as it is is likely that those imports would fail with an error due the number of requests, right now we remove this limit on loxi due this scenario with batch pushing that was creating a problem when pulling data from EA server (batch pushing)

src/Tribe/Aggregator/Cron.php
mitogh: By painful do you mean slow?

Actually this one is a bit more conservative that the regular cron for EA client.

https://github.com/moderntribe/the-events-calendar/blob/674025ca52e6cab552b78026186f73bd027dbfde/src/Tribe/Aggregator/Cron.php#L483-L484

src/Tribe/Aggregator/Processes/Batch_Imports.php
mitogh: Sure, will move to a ticket but we want to address it if it happens to be a failure during QA :)

src/Tribe/Aggregator/Processes/Service_Provider.php
mitogh: Great question, however this is actually better as the instance is provided to the service provider instead of a string to be resolved, more details can be found here: https://github.com/moderntribe/the-events-calendar/pull/3306#discussion_r523962403

src/Tribe/Aggregator/Service.php
mitogh: I'm not sure if I get correctly, which part specifically the URL construction or the URL value?

src/Tribe/Aggregator/Record/Batch_Queue.php
mitogh: Great point, about the method tho I'm not sure the difference or benefit in this instance, would you mind clarify a bit more, thanks

src/Tribe/Aggregator/Record/Abstract.php
mitogh: Great point :+1:

src/Tribe/Aggregator/Record/Batch_Queue.php
mitogh: Fantastic doing the clean up :)

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+	 * Tribe__Events__Aggregator__Record__Queue constructor.
+	 *
+	 * @since TBD
+	 *
+	 * @param int|Tribe__Events__Aggregator__Record__Abstract       $record
+	 * @param Tribe__Events__Aggregator__Record__Queue_Cleaner|null $cleaner
+	 */
+	public function __construct( $record, Tribe__Events__Aggregator__Record__Queue_Cleaner $cleaner = null ) {
+		if ( is_numeric( $record ) ) {
+			$record = Tribe__Events__Aggregator__Records::instance()->get_by_post_id( $record );
```
lucatume, line 77: Here and elsewhere, use aliases and shorten the code.

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+
+			return;
+		}
+
+		$this->record = $record;
+		$this->activity();
+	}
+
+	/**
+	 * GEt the activity if a call to a dynamic attribute is taking place in this case `$this->>activity`
```
lucatume, line 100:
```suggestion
	 * Get the activity if a call to a dynamic attribute is taking place in this case `$this->>activity`
```

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+	 * @since TBD
+	 *
+	 * @return boolean
+	 */
+	public function is_fetching() {
+		return $this->is_in_progress();
+	}
+
+	/**
+	 * Shortcut to check how many items are going to be processed next
```
lucatume, line 150:
```suggestion
	 * Shortcut to check how many items are going to be processed next.
```

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+			$args = [
+				'ID'            => $this->record->post->ID,
+				'post_modified' => ( new DateTime( 'now', new DateTimeZone( 'UTC' ) ) )->format( Tribe__Date_Utils::DBDATETIMEFORMAT ),
+				'post_status'   => Tribe__Events__Aggregator__Records::$status->success,
+			];
+			wp_update_post( $args );
+		} catch ( Exception $e ) {
+			do_action( 'tribe_log', 'debug', __METHOD__, [
+				'message'   => $e->getMessage(),
+				'record'    => $this->record,
```
lucatume, line 195: Here and elsewhere: do not log objects that have a lot of nested dependencies or the log will just fill up and be unusable. Log only meaningful scalars, ideally.

src/Tribe/Aggregator/Record/Batch_Queue.php
lucatume: Usually we alias it to `Dates`

src/Tribe/Aggregator/Record/Batch_Queue.php
lucatume: Usually we dedicate a method to the filtering, e.g. `$batch_size = $this->get_batch_size()` and move the filtering there.

src/Tribe/Aggregator/Record/Batch_Queue.php
```diff
@@ -0,0 +1,422 @@
...
+
+	/**
+	 * Returns the primary post type the queue is processing
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_queue_type() {
+		$item_type = Tribe__Events__Main::POSTTYPE;
```
lucatume, line 370: Usually aliased to `TEC::POSTTYPE`.

src/Tribe/Aggregator/Record/Queue_Cleaner.php
```diff
@@ -90,7 +90,7 @@ public function remove_duplicate_pending_records_for( Tribe__Events__Aggregator_
 	 */
 	public function maybe_fail_stalled_record( Tribe__Events__Aggregator__Record__Abstract $record ) {
 		$pending = Tribe__Events__Aggregator__Records::$status->pending;
-		$failed = Tribe__Events__Aggregator__Records::$status->failed;
+		$failed  = Tribe__Events__Aggregator__Records::$status->failed;
```
lucatume, line 93: Since you're in there: alias.

src/Tribe/Aggregator/Record/Queue_Cleaner.php
```diff
@@ -143,5 +169,7 @@ public function get_stall_limit() {
 	 */
 	public function set_stall_limit( $stall_limit ) {
 		$this->stall_limit = $stall_limit;
+
```
lucatume, line 172: This could be a good spot to add a filter.

src/Tribe/Aggregator/Record/Queue_Realtime.php
```diff
@@ -135,9 +135,10 @@ public function ajax() {
 		 * @var \Tribe__Events__Aggregator__Record__Queue_Interface $current_queue
 		 */
 		$current_queue = $this->queue_processor->current_queue;
-		$done = $current_queue->is_empty() && empty( $current_queue->is_fetching );
-		$percentage    = $current_queue->progress_percentage();
- 		$this->ajax_operations->exit_data( $this->get_progress_message_data( $current_queue, $percentage, $done ) );
+		$done          = $current_queue->is_empty() && empty( $current_queue->is_fetching() );
+
```
lucatume, line 139:
```suggestion
```

src/Tribe/Aggregator/Records.php
```diff
@@ -561,14 +561,14 @@ public function action_do_import() {
 		$json = json_encode( $_POST );
```
lucatume, line 561: Use `tribe_get_request_var` here.

lucatume (changes requested):

src/Tribe/Aggregator/Cron.php
mitogh: That's correct. 

However this can be filterable otherwise users can change it via the same filter, and batch imports is not specif to an option of imports as it can be available only for a set of imports, we can enable it dynamically before executing a batch import or processing the batch import, however I think the type of logic might be a bit overkill if eventually all imports are going to rely on batch pushing imports.

lucatume: Makes sense, thanksn.

lucatume (dismissed):

src/Tribe/Aggregator/Record/Queue_Cleaner.php
```diff
@@ -127,6 +128,32 @@ public function maybe_fail_stalled_record( Tribe__Events__Aggregator__Record__Ab
...
 	}
 
+	/**
+	 * Allow external caller to define the amount of the time to live in seconds.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $time_to_live Live time in seconds default to 12 hours.
+	 *
+	 * @return $this
```
lucatume, line 138: Missing doc comment.

src/Tribe/Aggregator/Record/Queue_Cleaner.php
```diff
@@ -127,6 +128,32 @@ public function maybe_fail_stalled_record( Tribe__Events__Aggregator__Record__Ab
...
+
+		return $this;
+	}
+
+	/**
+	 * Get the current value of time to live setting an integer in seconds, default to 12 hours.
+	 *
+	 * @since TBD
+	 *
+	 * @return int
```
lucatume, line 151: Missing doc comment.

src/Tribe/Aggregator/Record/Queue_Cleaner.php
lucatume: No filter?

lucatume (changes requested):
Smaller adjustments then all good.

lucatume (approved):

PR #3375, opened 2020-11-27
Customizer style overrides

alliuca:
**Description:** Implements the style overrides for the new Customizer V2 settings.

**Tickets:**
https://moderntribe.atlassian.net/browse/TEC-3651
https://moderntribe.atlassian.net/browse/TEC-3673
https://moderntribe.atlassian.net/browse/TEC-3334
https://moderntribe.atlassian.net/browse/TEC-3682
https://moderntribe.atlassian.net/browse/TEC-3675

**Screenshots:**

tec tribe_events_list_

Screen Shot 2020-11-27 at 3 03 11 PM

Screen Shot 2020-11-27 at 3 03 17 PM

tec tribe_events_list_ (1)

Screen Shot 2020-11-27 at 3 04 23 PM

tec tribe_events_list_(iPhone X)

tec tribe_events_month_(iPhone X)

tec tribe_event_made-with-block-editor_

tec tribe_event_multi-day-featured-2_

Screen Shot 2020-11-30 at 9 56 29 AM

Screen Shot 2020-11-30 at 9 56 38 AM

PR #3393, opened 2020-12-07
Feature/tec 3276 v2 map pin 2

jesseeproductions:
_Ref:_ [TEC-3276](https://theeventscalendar.atlassian.net/browse/TEC-3276)

Customizer setting: https://cloudup.com/iufgrSmQ-3T

Changes the Map V2 View and Single View. 

Venue V2 does not change and looks to require a change from using an iFrame to display the map. 

Map View: https://cloudup.com/iWv97Z-x_87
Single View: https://cloudup.com/iPkGYxZJdOr
Venue ( does not show ): https://cloudup.com/iTS8p4_Jrv1
https://developers.google.com/maps/documentation/embed/get-started#optional_parameters

Related PR: https://github.com/moderntribe/events-pro/pull/1641


paulmskim (commented):
The PR should be based off of https://github.com/moderntribe/the-events-calendar/pull/3375, as there were many changes made to the customizer. However, this setting should only be set up in PRO, not in TEC, so that makes this PR moot. Let's move this work to PRO instead.

jesseeproductions:
Closing per the last comment

lucatume (changes requested):
From the code point of view, all fine.
Please, update the code with Paul's comment.

jesseeproductions:
Hi @paulmskim,

I confirmed in the ticket that both TEC and Pro use this feature. 

https://theeventscalendar.atlassian.net/browse/TEC-3276?focusedCommentId=94345

For TEC it is the single events. 

So I believe this PR is still relevant as TEC should have the option as you can change single events without Pro. 

For Pro the change is for the Map View and the Pro PR handles that. 

Does that make sense?

paulmskim:
@jesseeproductions gotcha, I hadn't thought about the event single. If we're allowing Google Maps API to be set in TEC, then this looks good.

jesseeproductions:
@lucatume Please take another look. I believe this is set.

PR #3395, opened 2020-12-08
Feature/EA-310

codingmusician:
JIRA Ticket: [EA-310](https://moderntribe.atlassian.net/browse/EA-310)

- Add link to filter event list by aggregator record in Scheduled imports tab

Note: This is a resubmission after QA returned fail.

### DIFF
--- a/src/Tribe/Admin_List.php
+++ b/src/Tribe/Admin_List.php
@@ -9,7 +9,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 use Tribe__Events__Main as TEC;
-use Tribe__Events__Aggregator__Records as TEC_AR;
 
 if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 	class Tribe__Events__Admin_List {
--- a/src/Tribe/Admin_List.php
+++ b/src/Tribe/Admin_List.php
@@ -9,6 +9,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 use Tribe__Events__Main as TEC;
+use Tribe__Events__Aggregator__Records as TEC_AR;
 
 if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 	class Tribe__Events__Admin_List {
--- a/src/Tribe/Admin_List.php
+++ b/src/Tribe/Admin_List.php
@@ -190,7 +190,7 @@ if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 
 			// Add the record meta filter if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias , '/' ) . '\\s/', $clauses['where'] ) ) {
-				$sub_query = $wpdb->prepare("SELECT {$wpdb->posts}.`ID` FROM {$wpdb->posts} WHERE `post_parent` = {$parent_record_id}");
+				$sub_query = $wpdb->prepare( "SELECT {$wpdb->posts}.`ID` FROM {$wpdb->posts} WHERE `post_parent` = {$parent_record_id}" );
 				$clauses['where'] .= " AND {$table_alias}.meta_value IN ( {$sub_query} ) ";
 			}
 
--- a/src/Tribe/Admin_List.php
+++ b/src/Tribe/Admin_List.php
@@ -190,7 +190,8 @@ if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 
 			// Add the record meta filter if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias , '/' ) . '\\s/', $clauses['where'] ) ) {
-				$clauses['where'] .= " AND {$table_alias}.meta_value IN ( SELECT {$wpdb->posts}.ID FROM {$wpdb->posts} WHERE {$wpdb->posts}.post_parent = {$parent_record_id} ) ";
+				$sub_query = $wpdb->prepare("SELECT {$wpdb->posts}.`ID` FROM {$wpdb->posts} WHERE `post_parent` = {$parent_record_id}");
+				$clauses['where'] .= " AND {$table_alias}.meta_value IN ( {$sub_query} ) ";
 			}
 
 			return $clauses;
--- a/src/Tribe/Admin_List.php
+++ b/src/Tribe/Admin_List.php
@@ -175,18 +175,18 @@ if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 			}
 
 			// Check if filtering by aggregator record.
-			$record_num = (int) tribe_get_request_var( 'aggregator_record', 0 );
-			if ( 0 === $record_num ) {
+			$parent_record_id = (int) tribe_get_request_var( 'aggregator_record', 0 );
+			if ( 0 === $parent_record_id ) {
 				return $clauses;
 			}
 
 			global $wpdb;
 
-            $child_args = array(
-                'post_type'   => TEC_AR::$post_type,
-                'post_parent' => $record_num,
-            );
-            $child_ids = implode( ",", array_keys( get_children( $child_args, ARRAY_A ) ) );
+            		$child_args = array(
+                		'post_type'   => TEC_AR::$post_type,
+                		'post_parent' => $parent_record_id,
+            		);
+            		$child_ids = implode( ",", array_keys( get_children( $child_args, ARRAY_A ) ) );
 
 			$table_alias = 'ea_record_' . substr( uniqid( 'ea_record', true ), 0, 10 );
 			// Add the record meta query if it is missing.
--- a/src/Tribe/Admin_List.php
+++ b/src/Tribe/Admin_List.php
@@ -181,16 +181,22 @@ if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 			}
 
 			global $wpdb;
-			
+
+            $child_args = array(
+                'post_type'   => TEC_AR::$post_type,
+                'post_parent' => $record_num,
+            );
+            $child_ids = implode( ",", array_keys( get_children( $child_args, ARRAY_A ) ) );
+
 			$table_alias = 'ea_record_' . substr( uniqid( 'ea_record', true ), 0, 10 );
 			// Add the record meta query if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias, '/' ) . '\\s/', $clauses['join'] ) ) {
-				$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} AS {$table_alias} ON {$wpdb->posts}.ID = {$table_alias}.post_id AND {$table_alias}.meta_key = '_tribe_aggregator_parent_record' ";
+				$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} AS {$table_alias} ON {$wpdb->posts}.ID = {$table_alias}.post_id AND {$table_alias}.meta_key = '_tribe_aggregator_record' ";
 			}
 
 			// Add the record meta filter if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias , '/' ) . '\\s/', $clauses['where'] ) ) {
-				$clauses['where'] .= " AND {$table_alias}.meta_value = '{$record_num}' ";
+				$clauses['where'] .= " AND {$table_alias}.meta_value IN ({$child_ids}) ";
 			}
 
 			return $clauses;
--- a/src/Tribe/Admin_List.php
+++ b/src/Tribe/Admin_List.php
@@ -183,12 +183,6 @@ if ( ! class_exists( 'Tribe__Events__Admin_List' ) ) {
 
 			global $wpdb;
 
-            		$child_args = array(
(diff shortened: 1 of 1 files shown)


codingmusician:
This had been approved before, but I realized that I should have been searching by the child aggregator record, not the parent. Not all posts have the parent assigned to them.

src/Tribe/Admin_List.php
```diff
@@ -181,16 +181,22 @@ public static function filter_by_aggregator_record( array $clauses, WP_Query $wp
 			}
 
 			global $wpdb;
-			
+
```
mitogh, line 184: Looks like indentation is a bit off over here.

src/Tribe/Admin_List.php
```diff
@@ -181,16 +181,22 @@ public static function filter_by_aggregator_record( array $clauses, WP_Query $wp
...
+            $child_args = array(
+                'post_type'   => TEC_AR::$post_type,
+                'post_parent' => $record_num,
+            );
+            $child_ids = implode( ",", array_keys( get_children( $child_args, ARRAY_A ) ) );
```
mitogh, line 189: I think you can avoid this one by updating the where clause instead.

mitogh: _See example on `where` section_

src/Tribe/Admin_List.php
```diff
@@ -181,16 +181,22 @@ public static function filter_by_aggregator_record( array $clauses, WP_Query $wp
...
 			// Add the record meta query if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias, '/' ) . '\\s/', $clauses['join'] ) ) {
-				$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} AS {$table_alias} ON {$wpdb->posts}.ID = {$table_alias}.post_id AND {$table_alias}.meta_key = '_tribe_aggregator_parent_record' ";
+				$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} AS {$table_alias} ON {$wpdb->posts}.ID = {$table_alias}.post_id AND {$table_alias}.meta_key = '_tribe_aggregator_record' ";
 			}
 
 			// Add the record meta filter if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias , '/' ) . '\\s/', $clauses['where'] ) ) {
-				$clauses['where'] .= " AND {$table_alias}.meta_value = '{$record_num}' ";
+				$clauses['where'] .= " AND {$table_alias}.meta_value IN ({$child_ids}) ";
```
mitogh, line 199:
```suggestion
				$clauses['where'] .= " AND {$wpdb->posts}.post_parent = ({$record_num}) ";
```

mitogh: Additionally I would change `$record_num` with `$parent_record_id`

mitogh (changes requested):

src/Tribe/Admin_List.php
codingmusician: We're not looking at the parent for the event, we're looking at the aggregator record ID in the meta of the event. The aggregator record's parent is the integer that gets passed into the page. We have to pull the children IDs separately and search for any events that have those IDs in their meta. The event's parent doesn't play a part in this.

src/Tribe/Admin_List.php
codingmusician: Fixed

src/Tribe/Admin_List.php
codingmusician: I renamed the variable.

src/Tribe/Admin_List.php
mitogh: Oh right, makes sense then.

What's the value or where is located `TEC_AR` I'm not able to see it on the file.

src/Tribe/Admin_List.php
```diff
@@ -175,22 +175,28 @@ public static function filter_by_aggregator_record( array $clauses, WP_Query $wp
...
 				return $clauses;
 			}
 
 			global $wpdb;
-			
+
+            		$child_args = array(
+                		'post_type'   => TEC_AR::$post_type,
+                		'post_parent' => $parent_record_id,
+            		);
```
mitogh, line 188: Additionally I think you need to set a limit to the number of results here based on the number of events per page. I think `get_option( 'posts_per_page' );` should work.  As I think this one returns all the available results and in some instances can be harmful. 

But need to work with the previous and future pages if any as well.

src/Tribe/Admin_List.php
codingmusician: Now, THAT would be a mistake on my part. Good catch.

mitogh: Great usually that one is called `Records` I think would be ideal to keep consistent with other files like this one: 

- https://github.com/moderntribe/the-events-calendar/blob/d1902db0aef3f14a2503410f3fdb5e86bd089add/src/Tribe/Aggregator/Cron.php#L5

src/Tribe/Admin_List.php
codingmusician: Hmm. That's true, but using 'posts_per_page' wouldn't be accurate. We would want events from all child records of the parent id. Maybe it would be better to alter the SQL statement to use a temp table. Let me work something up... I'll ping you when I have something.

codingmusician: I took the get_children() call out of there and handled it via the SQL statement. Not sure why I didn't think of it before.

src/Tribe/Admin_List.php
```diff
@@ -9,6 +9,7 @@
 }
 
 use Tribe__Events__Main as TEC;
+use Tribe__Events__Aggregator__Records as TEC_AR;
```
mitogh, line 12:
```suggestion
use Tribe__Events__Aggregator__Records as Records;
```

https://github.com/moderntribe/the-events-calendar/pull/3395#discussion_r538862446

mitogh: Looks like this one is no longer required.

src/Tribe/Admin_List.php
```diff
@@ -175,22 +176,22 @@ public static function filter_by_aggregator_record( array $clauses, WP_Query $wp
...
 			// Add the record meta query if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias, '/' ) . '\\s/', $clauses['join'] ) ) {
-				$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} AS {$table_alias} ON {$wpdb->posts}.ID = {$table_alias}.post_id AND {$table_alias}.meta_key = '_tribe_aggregator_parent_record' ";
+				$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} AS {$table_alias} ON {$wpdb->posts}.ID = {$table_alias}.post_id AND {$table_alias}.meta_key = '_tribe_aggregator_record' ";
 			}
 
 			// Add the record meta filter if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias , '/' ) . '\\s/', $clauses['where'] ) ) {
-				$clauses['where'] .= " AND {$table_alias}.meta_value = '{$record_num}' ";
+				$clauses['where'] .= " AND {$table_alias}.meta_value IN ( SELECT {$wpdb->posts}.ID FROM {$wpdb->posts} WHERE {$wpdb->posts}.post_parent = {$parent_record_id} ) ";
```
mitogh, line 194: I think this one also faces the same "issue" where a parent record can have let's say 100 records but in this particular case you are interested on a chunk let's say the first 10 or the first 20 depending on the page you are looking for, is that correct?

src/Tribe/Admin_List.php
codingmusician: Correct. I removed it.

src/Tribe/Admin_List.php
codingmusician: I do understand the concern, but there's not a way to know how many child records to pull. Most aggregator records only pull events that are in the future, so if we were to only pull (for example) 1 record, then ALL events in the past wouldn't be pulled. There might be only one event that would show if only one event were imported for that record, but there could be thousands of past events from previous imports that need to be detected to display the correct pagination numbers. By using the temporary table, we're filtering out events that don't include a child record ID as the import record. We're not pulling the child records into memory like e were before.  The query itself will still be limited to a certain number of events and only display the number of events based on the pagination settings.

mitogh: Right, I agree this is a great approach, however we are pulling all the records where the parent is `_parent_record_id`  which I think is great, however we don't need to pull all of them we just need to pull the first 20 (if this is the first page and the number of results per page are 20) meaning we are missing a `LIMIT` and  `OFFSET` depending of the current page and results per page into the sub query so it only returns that subset of records instead of all the ones associated with the parent ID.

Let me know if that makes sense. 

Additionally I think it might be good if we use `wpdb->prepare` to make sure expected values are passed into those variables. 

Great job, thanks and let me know if you have additional questions.

codingmusician: Actually, I'm not altering the built in LIMIT and OFFSET at all. I'm only altering the 'where' and 'join' clauses. Not the 'limits' clause. It's the one that includes the LIMIT and OFFSET.

When outputting the 'limits' clause, it reads:
'limits' => 'LIMIT 0, 10'

My page limit is set to 10.

I hope that clears up the confusion.

mitogh: Correct, however the subquery (inside of the where) does not have this limit in consideration.

```sql
SELECT {$wpdb->posts}.ID FROM {$wpdb->posts} WHERE {$wpdb->posts}.post_parent = {$parent_record_id} )
```

So if this subquery will return all the records when not required so in this case you only need the first 10 and so on so this one is the one that might be missing the `LIMIT`

codingmusician: Correct. That is because we want to only include records that are a part of ANY of the child records. Not just the first 20 or so. We are not "pulling" rows here. Only querying against the IDs that have a certain post_parent value.

mitogh: Great makes sense then, wanted to make sure that was the case :)

mitogh (approved):

src/Tribe/Admin_List.php
```diff
@@ -175,22 +175,22 @@ public static function filter_by_aggregator_record( array $clauses, WP_Query $wp
 			}
 
 			// Check if filtering by aggregator record.
-			$record_num = (int) tribe_get_request_var( 'aggregator_record', 0 );
-			if ( 0 === $record_num ) {
+			$parent_record_id = (int) tribe_get_request_var( 'aggregator_record', 0 );
+			if ( 0 === $parent_record_id ) {
```
mitogh, line 179:
```suggestion
			if ( 0 <= $parent_record_id ) {
```

mitogh, line 179: Or:

```suggestion
			$parent_record_id = absint( tribe_get_request_var( 'aggregator_record', 0 ) );
			if ( 0 === $parent_record_id ) {
```

mitogh (changes requested):

src/Tribe/Admin_List.php
codingmusician: Looking at it from a different vantage point: If we were to use LIMIT and OFFSET in the sub-query, (assuming posts per page is set to 20) when they go to page 2 of events, our subquery would then look for records 21 - 40. If there are only 5 aggregator records, it would return no children IDs and therefore all the events would disappear.

codingmusician: I did add $wpdb->prepare() for the subquery.

src/Tribe/Admin_List.php
```diff
@@ -175,22 +175,23 @@ public static function filter_by_aggregator_record( array $clauses, WP_Query $wp
...
 			// Add the record meta query if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias, '/' ) . '\\s/', $clauses['join'] ) ) {
-				$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} AS {$table_alias} ON {$wpdb->posts}.ID = {$table_alias}.post_id AND {$table_alias}.meta_key = '_tribe_aggregator_parent_record' ";
+				$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} AS {$table_alias} ON {$wpdb->posts}.ID = {$table_alias}.post_id AND {$table_alias}.meta_key = '_tribe_aggregator_record' ";
 			}
 
 			// Add the record meta filter if it is missing.
 			if ( ! preg_match( '/\\s' . preg_quote( $table_alias , '/' ) . '\\s/', $clauses['where'] ) ) {
-				$clauses['where'] .= " AND {$table_alias}.meta_value = '{$record_num}' ";
+				$sub_query = $wpdb->prepare( "SELECT {$wpdb->posts}.`ID` FROM {$wpdb->posts} WHERE `post_parent` = {$parent_record_id}" );
```
mitogh, line 193:
```suggestion
				$sub_query = $wpdb->prepare( "SELECT %s.`ID` FROM %s WHERE `post_parent` = %d", $wpdb->posts, $wpdb->posts, $parent_record_id );
```

mitogh (changes requested):

mitogh (approved):

PR #3397, opened 2020-12-09
TEC-3684 This adds the single_event control I missed.

Camwyn:
This adds the single_event control I missed.

[TEC-3684]

old: http://p.tri.be/sNOqaC
new: http://p.tri.be/FiE0pP

Will require some logic tweaks where we use the single-event color, the same as the ones for the global elements.

[TEC-3684]: https://moderntribe.atlassian.net/browse/TEC-3684

### DIFF
--- a/src/Tribe/Customizer/Single_Event.php
+++ b/src/Tribe/Customizer/Single_Event.php
@@ -43,7 +43,10 @@ final class Tribe__Events__Customizer__Single_Event extends Tribe__Customizer__S
 			';
 		}
 
-		if ( $customizer->has_option( $this->ID, 'post_title_color' ) ) {
+		if (
+			! $customizer->has_option( $this->ID, 'post_title_color_choice' )
+			&& $customizer->has_option( $this->ID, 'post_title_color' )
+		) {
 			$template .= '
 				.single-tribe_events .tribe-events-single-event-title {
 					color: <%= single_event.post_title_color %>;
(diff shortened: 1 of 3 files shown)


src/Tribe/Views/V2/Customizer.php
```diff
@@ -45,7 +45,7 @@ public function include_global_elements_settings( $section, $manager, $customize
 				$manager,
 				$customizer->get_setting_name( 'event_title_color', $section ),
 				[
-					'label'    => esc_html__( 'Event Title', 'the-events-calendar' ),
+					'label'    => esc_html__( 'Event Title G', 'the-events-calendar' ),
```
lucatume, line 48: This looks like a typo.

lucatume (changes requested):
Smaller fix.

src/Tribe/Customizer/Single_Event.php
```diff
@@ -43,7 +43,10 @@ public function get_css_template( $template ) {
 			';
 		}
 
-		if ( $customizer->has_option( $this->ID, 'post_title_color' ) ) {
+		if (
+			! $customizer->has_option( $this->ID, 'post_title_color_choice' )
```
paulmskim, line 47: @Camwyn this will break the V1 customizer. If a user switches to V2, sets the `post_title_color_choice`, then switches back to V1, this will always have this option. Instead, return early above if V2.

PR #3402, opened 2020-12-10
Make sure the attributes are rendered for TEC

mitogh:
Make sure the tag is actually executed so the attributes are render for TEC.


lucatume:
Rebasing this on `release/B20.12` to see if it could fit in.

PR #3404, opened 2020-12-14
fix(Repositories/Event) make sure event UTC dates are correctly set

lucatume:
This PR fixes an issue where the Event ORM would set the UTC dates incorrectly depending on the site timezone settings.

Artifacts: see tests.


bordoni:
@lucatume This was automatically closed due to being in a deleted branch. We should make sure we move it to the correct branch.

PR #3410, opened 2020-12-21
Fix timezone issue in setup_repository_args

therealgilles:
tribe_beginning_of_day/tribe_end_of_day require a Dates::build_date_object to take the timezone into account.

This bug will only show up if PHP runs on a different timezone than the timezone set in the Wordpress settings. This is not unusual on shared hosting platforms where the server/PHP timezones are set to UTC.

This bug is not easy to reproduce as it requires to be in the time range where the WP timezone and the PHP timezone put the event on a different day, with the end of day cutoff taken into account.

All the other places where tribe_beginning_of_day/tribe_end_of_day are called, they are passed a Dates::build_date_object, so hopefully whoever wrote the code can see it's missing here. Or feel free to reject if you think the proposed change is wrong.

### DIFF
--- a/src/Tribe/Views/V2/Views/Day_View.php
+++ b/src/Tribe/Views/V2/Views/Day_View.php
@@ -111,7 +111,8 @@ class Day_View extends View {
 		$date = Arr::get( $context_arr, 'event_date', 'now' );
 		$event_display = Arr::get( $context_arr, 'event_display_mode', Arr::get( $context_arr, 'event_display' ), 'current' );
 
-		$args['date_overlaps'] = [ tribe_beginning_of_day( $date ), tribe_end_of_day( $date ) ];
+		$current_date = Dates::build_date_object( $date );
+		$args['date_overlaps'] = [ tribe_beginning_of_day( $current_date->format('Y-m-d') ), tribe_end_of_day( $current_date->format('Y-m-d') ) ];
 
 		/**
 		 * @todo  @bordoni We need to consider fetching events on a given day from a cache


lucatume:
@therealgilles thanks for your PR.
As you _might_ have guessed, this one slipped through the cracks and was not correctly handled; for this, I apologize.

I'm refreshing the code and checking it out in the next couple of days; I will let you know.

lucatume:
I'm wrapping this PR in [this one](https://github.com/the-events-calendar/the-events-calendar/pull/3524) to add tests and changelog entries.

lucatume:
@therealgilles I'm following up on this PR in [this one](https://github.com/the-events-calendar/the-events-calendar/pull/3524). I've not changed your code, just broke down the long line, and added tests for it.

Thanks again for your PR: you will see this being closed once I merge mine (where I credit you, of course).

lucatume:
> PS: Unrelated but I feel like the TEC codebase has become unnecessarily convoluted. Every time I update the plugin, I now have to fix my site either because of css changes now overriding mine, some race condition in javascript (reported, still not fixed), some new javascript incompatibility with optimization tools. It has been a frustrating experience lately and I don't remember things being this bad earlier on. Sorry for the rant :/

@therealgilles could you point out specific examples of this? I'm asking not to defend a position, but to get some unbiased feedback. If there's an opportunity to improve, it's not lack of will or disdain that is keeping us from doing it: sometimes we simply never get word of it.
If, and when, you have time, I would appreciate you letting me know.

therealgilles:
@lucatume: Here are some thoughts and facts mixed together:

• I've been using the plugins for years and my experience has deteriorated.
• The way I see it, at the end of the day, it's a list of events with attributes to be queried and displayed. It's very useful and should not be very complicated.
• The code base is scary huge.
• There should only be a single way to query events used everywhere, even internally. It ought to be well documented.
• There is a (fancy) ORM, with bells and whistles, even advertising, and yet there is pretty much zero user/developer documentation for it.
• There are different options to include CSS styling. But I cannot tell what I am really getting with either options.
• CSS styling in version N+1 seems to be different from CSS styling in version N, although the styling is referred the same way with the same name.
• The javascript code used to be compatible with JS optimization tools but it's not anymore.
• View / Context / Repository. I acknowledge I'm not sure what they are for. What I know if that they make debugging so much harder than it should be. I surely hope they make things easier on your side.
• My present experience with the plugins is that at every new update, I wonder what is going to be broken this time. And there always seems to be something, which is frustrating and tiring.
• I just tried to access the support ticket system and could not. https://support.theeventscalendar.com redirected me to https://theeventscalendar.com/support/, even though I have a valid license.
• I've filed more tickets than I wanted with the support team. They always have trouble reproducing the (real) issues I find, and they are quick to blame the theme and other plugins. Case in point, I filed this bug months ago with the support team. Which tells me that the bugs are very complex to debug, pointing to an overcomplicated code base.

Maybe this won't be convincing from your perspective. I also understand that it's an evolving code base with a long history. I cannot even imagine its technical debt. I hope someone is able to spend some time thinking: "how can we simplify this?"

Best :)

Also this:

**PENDING BUGS**  (all filed with support with no action taken)
BUG #1 [messes up other plugins]
the-events-calendar/src/Tribe/Views/V2/View.php
Code:
    // Set the $_SERVER['REQUEST_URI'] as many WordPress functions rely on it to correctly work.
   $_SERVER['REQUEST_URI'] = $this->get_request_uri();
Corrected code:
   if ( !isset( $_SERVER['REQUEST_URI'] ) ) { $_SERVER['REQUEST_URI'] = $this->get_request_uri(); }
Note: Never a good idea to modify $_SERVER['REQUEST_URI'].

BUG #2 [javascript race condition]
the-events-calendar/src/resources/js/views/manager.js
Code:
     $container.trigger( 'afterSetup.tribeEvents', [ index, $container, data ] );
Workaround:
     setTimeout(() => $container.trigger( 'afterSetup.tribeEvents', [ index, $container, data ] ), 0);
Note: There is a race condition on the afterSetup.tribeEvents event, as it's used in multiple locations.

BUG #3 (javascript warning)
the-events-calendar/src/resources/js/views/manager.js
Code:
```JS
     if ( ! data.prev_url ) {
         data.prev_url = containerData.prev_url;
     }
```
Corrected code:
```JS
     if ( ! data.prev_url && containerData ) {
          data.prev_url = containerData.prev_url;
     }
```
Note: Somehow containerData is sometimes undefined.

PR #3412, opened 2020-12-22
Correctly backup locale in Tribe\I18n

lucatume:
Issue: https://theeventscalendar.atlassian.net/browse/TEC-3711

Artifact: see tests.

Before version 4.7 WordPress would use nested arrays to keep track of filters.
This PR updates our code, now requiring a minimum WP version of 4.9, to assume hook objects should be instances of the `WP_Hook` class and setting the default (empty and backup) to a new instance of that class.

Furthermore, the PR removes some duplicated code in the `Tribe\Events\I18n::with_locale` method that would, incorrectly, prevent the method to be publicly usable; after the cleanup, I've updated the method visibility to `public`.

Also, this PR modifies the build script to make sure that, in the not-so-frequent instances where the `chrome` container used by `tric` to run the tests does not start immediately (due to limited resources, I guess), the container is started before-hand to make sure it will actually be up when required.


bordoni (approved):
I am sure you checked but the unloading that was removed here, it's not needed right ?

lucatume:
> I am sure you checked but the unloading that was removed here, it's not needed right?

It was a duplicated code. From an inspection of the git history, I've moved that logic to the `compile_translations` method and then forgot to remove it from there.
Debugging the code I've seen the code will run, but will not change the output in any way, just run again.

PR #3416, opened 2021-01-04
Fix and restore views_ui suite

lucatume:
This PR fixes and issue that would prevent the `views_ui` suite from correctly running in the release branch and restores all tests along with it.

- build(.github) restore all tests
- build(test/_support) regen
- test(views_ui) fix use of Chrome URL
- test(views_ui) do not skip views_ui test


bordoni (approved):

borkweb (approved):
:+1:

PR #3419, opened 2021-01-07
Task / update bucket customizer dqa adjustments

alliuca:
**Description:** Updates the bucket/customizer-dqa-adjustments branch to reference the latest common hash


paulmskim:
I know it's only a common update, but let's get a `tric run` screenshot here since GHA aren't running them fully, something like https://github.com/the-events-calendar/events-virtual/pull/226#issuecomment-756117528

PR #3426, opened 2021-01-21
properly enqueue manager script as last script to be enqueued

paulmskim:
**Ticket:** [ECP-690]

This fixes an issue with the countdown widget JS not loading properly, but loading after the manager. This enqueues the manager as the last JS asset to allow all the other JS files to work properly.

**Artifact:**

image

### DIFF
--- a/src/Tribe/Views/V2/Assets.php
+++ b/src/Tribe/Views/V2/Assets.php
@@ -191,7 +191,7 @@ class Assets extends \tad_DI52_ServiceProvider {
 			[
 				'priority'     => 20,
 				'conditionals' => [ $this, 'should_enqueue_frontend' ],
-				'groups'       => [ static::$group_key, static::$widget_group_key ],
+				'groups'       => [ static::$group_key ],
 			]
 		);
 
--- a/src/Tribe/Views/V2/Widgets/Widget_Abstract.php
+++ b/src/Tribe/Views/V2/Widgets/Widget_Abstract.php
@@ -51,6 +51,7 @@ abstract class Widget_Abstract extends \Tribe\Widget\Widget_Abstract {
 
 		add_filter( 'tribe_events_views_v2_view_template_vars', [ $this, 'filter_widget_template_vars' ], 20, 2 );
 		add_filter( "tribe_events_views_v2_view{$this->view_slug}_template_vars", [ $this, 'filter_widget_template_vars' ], 20, 2 );
+		// Dequeue and enqueue manager JS to ensure it is the last JS file to be added.
 		add_action( 'tribe_events_views_v2_widget_after_enqueue_assets', [ $this, 'action_enqueue_manager' ], 10, 3 );
 	}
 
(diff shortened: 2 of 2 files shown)


src/Tribe/Views/V2/Widgets/Widget_Abstract.php
```diff
@@ -51,6 +51,7 @@ public function setup() {
 
 		add_filter( 'tribe_events_views_v2_view_template_vars', [ $this, 'filter_widget_template_vars' ], 20, 2 );
 		add_filter( "tribe_events_views_v2_view{$this->view_slug}_template_vars", [ $this, 'filter_widget_template_vars' ], 20, 2 );
+		add_action( 'tribe_events_views_v2_widget_after_enqueue_assets', [ $this, 'action_enqueue_manager' ], 10, 3 );
```
lucatume, line 54: Either here or in the method long description, could you add an explanation of why this is done? 
It's not apparent from the code.

src/Tribe/Views/V2/Assets.php
```diff
@@ -191,7 +191,7 @@ public function register() {
 			[
 				'priority'     => 20,
 				'conditionals' => [ $this, 'should_enqueue_frontend' ],
-				'groups'       => [ static::$group_key, static::$widget_group_key ],
+				'groups'       => [ static::$group_key ],
```
lucatume, line 194: Why remove this?

src/Tribe/Views/V2/Widgets/Widget_Abstract.php
paulmskim: For our widgets, when we enqueue JS files, we enqueue the viewport, breakpoints, and manager. After that, we add the assets for the specific widget. However, if we have a JS file (as in the countdown widget), we need the countdown widget JS to be enqueued before the manager (priority does not seem to take effect when force enqueuing). This also becomes a problem when we have multiple widgets, all with their own JS. So this was my solution to the problem.

lucatume: I understand the solution, I just would like the quick description noted in the code to avoid future-dev come in, see a clunky code, and remove it.

paulmskim: OK, will add.

Camwyn: If the manager is always enqueued, why not use it as a dependency of the widget js - then manager has to be loaded first?

paulmskim:
@lucatume addressed your note, looks like wpunit is failing but not related to this PR. Merging this so we can move on.

src/Tribe/Views/V2/Widgets/Widget_Abstract.php
paulmskim: The manager needs to be loaded last, as it triggers events that other JS files use.

PR #3429, opened 2021-02-01
Feature / TEC-3753 event header

alliuca:
**Description:** Implements style updates for the event title and the metadata below it

**Tickets:**
- https://theeventscalendar.atlassian.net/browse/TEC-3751
- https://theeventscalendar.atlassian.net/browse/TEC-3752
- https://theeventscalendar.atlassian.net/browse/TEC-3753
- https://theeventscalendar.atlassian.net/browse/TEC-3755

**Screenshots:**

Screen Shot 2021-02-01 at 12 03 06 PM

Screen Shot 2021-02-01 at 2 36 13 PM


src/views/single-event.php
```diff
@@ -22,7 +22,7 @@
 
 ?>
 
-<div id="tribe-events-content" class="tribe-events-single">
+<div id="tribe-events-content" class="tribe-events-single tribe-events-view-single"> <!-- TODO: tribe-events-view-single should be added under condition -->
```
paulmskim, line 25: you can use `tribe_events_views_v2_is_enabled()` to check if v2 is enabled

paulmskim (commented):
This is good stuff so far. One thing I'd check is the breakpoint in v1 styles. IIRC, it's a max-width of 769px? I don't remember exactly, but at a viewport of 768px and 769px something funky might happen.

src/views/single-event.php
alliuca: will have to ask Gustavo, I think we're going to use a new constant for this

alliuca: @paulmskim I removed that class from the div and added a body class `single-tribe_events-v2` via my theme's functions.php for now. I'm using that to target stuff in the style files. Gustavo should be able to properly add that class soon

src/Tribe/Views/V2/Assets.php
```diff
@@ -423,11 +453,6 @@ public function register() {
 	 * @return void
 	 */
 	public function disable_v1() {
-		// Dont disable V1 on Single Event page
-		if ( tribe( Template_Bootstrap::class )->is_single_event() ) {
-			return;
-		}
```
bordoni, line 429: Just to confirm our intention is to remove V1 styles completely correct?

src/Tribe/Views/V2/Assets.php
```diff
@@ -507,4 +532,25 @@ public function should_enqueue_full_styles() {
 		 */
 		return apply_filters( 'tribe_events_views_v2_assets_should_enqueue_full_styles', $should_enqueue );
 	}
+	
+	/**
+	 * Verifies if we are on V2 and on Event Single in order to enqueue the override styles for Single Event
```
bordoni, line 537:
```suggestion
	 * Verifies if we are on V2 and on Event Single in order to enqueue the override styles for Single Event.
```

bordoni (changes requested):
One small question and a docblock style change. 💙

src/Tribe/Views/V2/Assets.php
alliuca: @bordoni yes

bordoni (dismissed):

bordoni (approved):

PR #3430, opened 2021-02-02
Feature/ec 624 fe assets

Camwyn:
No ticket, see https://github.com/the-events-calendar/events-pro/pull/1685#discussion_r568013424

Added some model tests per https://github.com/the-events-calendar/the-events-calendar/pull/3430#pullrequestreview-580956930


paulmskim (commented):
Code looks good, though are there any tests we need to update? If there are none, we should probably add some for this.

Camwyn:
> Code looks good, though are there any tests we need to update? If there are none, we should probably add some for this.

There were **no** model tests - there are now!

bordoni (approved):

PR #3433, opened 2021-02-03
Ensure the "Latest Past" View is not public

borkweb:
This view is not intended to be a publicly selectable view. Flipping the property to `false`. While the Event Settings Display page does not render this view, calls to the view's `is_publicly_visible()` method should reflect our intention.

The artifact for this is a comparison of these two outputs of the method in question:

*Before*:
Selection_104

*After*:
Selection_105


bordoni:
lets make sure we have a changelog here.

PR #3442, opened 2021-02-11
Feature / event single related events updates

alliuca:
**Description:** Adds styles to update the way Related Events block displays

**Note:** This also modifies the whole single event layout to re-order blocks according to the designs. It also separates part of the `.recurringinfo` styles to display things a bit differently between the top header and the related events metadata.

**Ticket:** https://theeventscalendar.atlassian.net/browse/ECP-673

**Screenshots:**

Screen Shot 2021-02-11 at 12 08 11 PM

Screen Shot 2021-02-11 at 12 08 29 PM

### DIFF
--- a/src/resources/postcss/views/skeleton/_single-event.pcss
+++ b/src/resources/postcss/views/skeleton/_single-event.pcss
@@ -300,6 +300,7 @@
 .tribe-events-single ul.tribe-related-events {
 	display: flex;
 	flex-direction: column;
+	margin-bottom: 0;
 	
 	@media (--viewport-medium) {
 		flex-direction: row;
--- a/src/resources/postcss/views/skeleton/_single-event.pcss
+++ b/src/resources/postcss/views/skeleton/_single-event.pcss
@@ -297,7 +297,7 @@
 	width: 100%;
 }
 
-ul.tribe-related-events {
+.tribe-events-single ul.tribe-related-events {
 	display: flex;
 	flex-direction: column;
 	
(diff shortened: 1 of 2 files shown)


src/resources/postcss/views/skeleton/_single-event.pcss
```diff
@@ -118,6 +117,31 @@
...
+	
+	&:before {
+		@media (--viewport-medium) {
+			left: 16px;
+		}
+	}
+	
+	.event-is-recurring {
+		@media (--viewport-medium) {
+			letter-spacing: normal;
```
paulmskim, line 139: this is probably a full style

src/resources/postcss/views/skeleton/_single-event.pcss
```diff
@@ -260,3 +284,83 @@
...
+				a {
+					display: block;
+					overflow: hidden;
+					position: relative;
+					padding-top: 75%; /* 4:3 aspect ratio */
+				}
+				
+				img {
+					height: 100%;
+					object-fit: cover;
```
paulmskim, line 337: this is not supported in IE11, we should probably find another way to do this.

paulmskim (changes requested):
a couple small things, but looks good otherwise.

src/resources/postcss/views/skeleton/_single-event.pcss
alliuca: will remove the logic for now and revisit in a different PR, Rachel said it's more of a nice to have at this point

PR #3445, opened 2021-02-12
Feature / event single additional updates

alliuca:
**Description:** Polishes the new Event Single view to match the current designs more closely.

**Ticket:** none in particular, different tickets under https://theeventscalendar.atlassian.net/browse/TEC-3693

**Screenshots:**

Screen Shot 2021-02-12 at 11 46 48 AM


src/resources/postcss/views/skeleton/_single-event.pcss
```diff
@@ -276,6 +307,8 @@
 	
 	/* Make sure the gmap div has this specific height on desktop */
```
Camwyn, line 308: Should maybe change this comment - since we're enforcing a max-height on mobile too?
```suggestion
	/* Make sure the gmap div has these specific heights */
```

src/resources/postcss/views/skeleton/_single-event.pcss
```diff
@@ -357,19 +390,8 @@
...
-			display: inline-block;
-			height: 20px;
-			vertical-align: middle;
-			width: 11px;
-		}
-		
-		span {
-			display: none;
-		}
+	a span {
```
Camwyn, line 393: What is this hiding? Generic tags with `display: none` make me uneasy...

Camwyn (commented):
One suggestion, one question.

src/resources/postcss/views/skeleton/_single-event.pcss
alliuca: @Camwyn that's hiding the right angle quote, unfortunately we can't touch the templates to remove it in this project 
Screen Shot 2021-02-12 at 4 59 22 PM

Camwyn (approved):

PR #3459, opened 2021-02-22
Translate calendar view in language switcher

dgwatkins:
The current code is overwriting the translation of the calendar slug in the URL.

With these changes, the WPML language switcher displays correct URLs.


bordoni:
Hi @dgwatkins,

I am working on a couple of internal tickets to track and QA this work, we will make sure it's merged into our next release sprint. We have a release coming this week which I wasn't able to get this fix merged and tested into.

PR #3461, opened 2021-02-22
Update Event Single Re-Skin bucket branch

alliuca:
**Description:** Updates the bucket branch with the latest B21.03


src/Tribe/Views/V2/Assets.php
```diff
@@ -551,21 +551,11 @@ public function should_enqueue_full_styles() {
...
 		
-		// Bail if not Single Event.
-		if ( ! tribe( Template_Bootstrap::class )->is_single_event() ) {
-			return false;
-		}
-		
-		// Bail if Block Editor.
-		if ( has_blocks( get_queried_object_id() ) ) {
-			return false;
-		}
```
paulmskim, line 567: should we not be keeping this? we don't want to enqueue single event styles if it's not a single event page, nor do we want to enqueue if it's a block editor single right?

src/functions/views/provider.php
```diff
@@ -170,6 +171,16 @@ function tribe_events_single_view_v2_is_enabled() {
...
+	
+	// Bail if not Single Event.
+	if ( ! tribe( Template_Bootstrap::class )->is_single_event() ) {
+		return false;
+	}
+	
+	// Bail if Block Editor.
+	if ( has_blocks( get_queried_object_id() ) ) {
+		return false;
+	}
```
paulmskim, line 183: Oh I see, you moved the logic here. I don't think this should live here. If we needed to know if single view v2 is enabled but didn't care if it was a single event or if it had blocks or not, then this doesn't allow us to separate that logic. We should move it back to the assets file.

paulmskim (changes requested):

bordoni (approved):

PR #3475, opened 2021-03-12
Fix / classic details block create organizer

alliuca:
**Description:** Fixes the classic-details-block's organizer creation and, in the process:
- Adds a couple of tests
- Refactors / cleans up the original component a bit
- Fixes the form layout by using WordPress' `TextControls` component instead of a standard input

**Ticket:** https://theeventscalendar.atlassian.net/browse/TEC-3698

**Screenshots:**

Screen Shot 2021-03-12 at 11 30 22 AM


src/modules/elements/organizer-form/element.js
```diff
@@ -93,29 +97,20 @@ class OrganizerForm extends Component {
 				console.warning( 'Invalid creation of organizer:', newPost );
 			}
 
-			this.props.addOrganizer( newPost );
+			this.props.addOrganizer( newPost.id, newPost );
```
paulmskim, line 100: Did we change the function signature somewhere? I'm not seeing it in this PR.

paulmskim (approved):
One question, otherwise looks good to me.

src/modules/elements/organizer-form/element.js
alliuca: @paulmskim I don't know why, but it wasn't working at all before. I found out the function requires those arguments

alliuca: but I'll double check that this same component isn't used by the standard organizer block, maybe that's why

alliuca: yeah it works everywhere

paulmskim: hah... well probably why it wasn't working at all

PR #3494, opened 2021-04-06
Allow status=any in filter_post_status_list (-> next_rest_url requests)

maxm123:
When you perform an authenticated request (user has capability to edit events) to an archive list REST endpoint, the 'status' parameter defaults to 'any'. This value is included in the generated next_rest_url/previous_rest_url fields in the response. But when those urls are used, the API reports _rest_invalid_param_ for status.

There are multiple ways to cope with that error. One could simply remove status=any from the response url (in Archive_Base->get_current_rest_url) but that would be rather inconsistent.

This PR just allows 'any' as value for 'status' in the corresponding sanitizing method (Archive_Base->filter_post_status_list) which is called by e.g. Archive_Event->get.


maxm123:
Uhm.. Your tr1b0t seems not to be working correctly..

bordoni (approved):

PR #3498, opened 2021-04-10
TEC-3882: Filter the accepted file types for EA client

mitogh:
Only allow accepted file types for EA client as the other allowed type
does not provide any real value for EA Customers when having a large set
of files on the directory of your computer from where to select the
file.

Artifact: https://d.pr/v/24wMZN
Ticket: TEC-3882


src/Tribe/Aggregator/Page.php
```diff
@@ -199,8 +199,23 @@ public function filter_plupload_default_settings( $settings ) {
...
 		}
 
-		if ( ! empty( $settings['filters']['mime_types'] ) ) {
-			unset( $settings['filters']['mime_types'] );
+
+		if ( isset( $settings['filters']['mime_types'] ) ) {
+			$extensions = [
+				'csv', // comma-separated values
+				'xls', // Excel 97- Excel 2003 Workbook
+				'xlsx' // Excel Workbook
```
borkweb, line 207:
```suggestion
				'xlsx', // Excel Workbook
```

borkweb (changes requested):
A small tweak and a changelog entry is all that is needed here. This is a great change!

PR #3499, opened 2021-04-12
Fix Organizer saving problem on Blocks editor.

bordoni:
🎫 [TEC-3844]
---
Resolved the problem by no longer filtering only to Events Rest APIs.

[TEC-3844]: https://theeventscalendar.atlassian.net/browse/TEC-3844


borkweb (dismissed):
@bordoni says that the test fixes live in alfajor. Otherwise, this code looks good.

PR #3501, opened 2021-04-16
 Move all common classes to the beginning of the list

Camwyn:
They should be "overridden" by the element class(es). 
Order doesn't really matter for this - _but_ it's easier to tell a user "grab the last class on the element and use that (then they don't accidentally override a common class) when you know the "last class" will be the most specific.

Element-specific classes should be listed in order of specificity:
Old: `'tribe-events-header__messages tribe-events-c-messages tribe-common-b2'`
New: `'tribe-common-b2 tribe-events-c-messages tribe-events-header__messages'`

This is also prep for using something like Tailwind or Atomic CSS - where the "utility classes" come first and the final class listed is a specific target class.

Also cleans up our implementations of `tribe_classes()`. To order the arrays in the same way.


Camwyn:
Closing as this has become part of B21.danish!

PR #3504, opened 2021-04-20
Correct hover colors on export button.

Camwyn:
[TEC-3781]

Normal: https://d.pr/i/AkrBwA
Hover: https://d.pr/i/c0Lz65

[TEC-3781]: https://theeventscalendar.atlassian.net/browse/TEC-3781

### DIFF
--- a/src/Tribe/Views/V2/Customizer.php
+++ b/src/Tribe/Views/V2/Customizer.php
@@ -555,7 +555,7 @@ class Customizer {
 
 			// overrides for tec components/full/_ical-link.pcss.
 			$css_template .= "
-				$tribe_events .tribe-events-c-ical__link {
+				$tribe_events .tribe-events-c-ical__link:not(:hover):not(:focus):not(:active) {
 					border-color: <%= global_elements.accent_color %>;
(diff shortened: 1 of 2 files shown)


src/Tribe/Views/V2/Customizer.php
```diff
@@ -555,7 +555,7 @@ public function filter_global_elements_css_template( $css_template, $section, $c
 
 			// overrides for tec components/full/_ical-link.pcss.
 			$css_template .= "
-				$tribe_events .tribe-events-c-ical__link {
+				$tribe_events .tribe-events-c-ical__link:not(:hover):not(:focus):not(:active) {
```
paulmskim, line 558: I would add a `color` style to the `hover`, `focus`, and `active` case below to override this style (I assume the CSS files do this too) rather than tagging every single case that this shouldn't target (increases specificity and if there are any cases we need to add it here too).

Camwyn: I did - the issue is that the color has to be hard coded as the custom props aren't available to the Customizer - yet.

bordoni (approved):

PR #3505, opened 2021-04-21
Correct implementation of tribe_get_event_website_link()

Camwyn:
The issue: 
The event website link label is not being filtered correctly.

Reason:
The filters rely on a valid post ID - and we don't always pass one.

Correction:

1. Find our uses of tribe_get_event_website_link() and include the post ID - it's a param and since it's available in every case we should pass it.
2. Add functionality to _get_ the post ID if it's not passed (for backwards compatibility and customizations)

https://d.pr/i/YJlL0I

[ECP-761]

Related: https://github.com/the-events-calendar/events-pro/pull/1734

[ECP-761]: https://theeventscalendar.atlassian.net/browse/ECP-761

### DIFF
--- a/src/Tribe/Views/V2/Hooks.php
+++ b/src/Tribe/Views/V2/Hooks.php
@@ -969,12 +969,6 @@ class Hooks extends \tad_DI52_ServiceProvider {
 	 * @return string
 	 */
 	public function filter_single_event_details_event_website_label( $label, $post_id = null ) {
-		bdump('filter_single_event_details_event_website_label');
-		bdump([
-			is_null( $post_id ),
-				! tribe_events_single_view_v2_is_enabled(),
-				has_blocks( $post_id ),
-		]);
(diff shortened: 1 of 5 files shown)


src/Tribe/Views/V2/Hooks.php
```diff
@@ -969,6 +969,12 @@ public function enqueue_customizer_controls_styles() {
 	 * @return string
 	 */
 	public function filter_single_event_details_event_website_label( $label, $post_id = null ) {
+		bdump('filter_single_event_details_event_website_label');
+		bdump([
+			is_null( $post_id ),
+				! tribe_events_single_view_v2_is_enabled(),
+				has_blocks( $post_id ),
+		]);
```
bordoni, line 977:
```suggestion
```

bordoni (changes requested):

bordoni (approved):

PR #3514, opened 2021-05-04
Month mobile view changes

lucatume:
- feat(v2/month) mobile messages below grid
- feat(Views/Month_View) display no messages on day

Issues:
* https://theeventscalendar.atlassian.net/browse/TEC-3793 
* https://theeventscalendar.atlassian.net/browse/TEC-3812 

This PR updates the Month mobile view to show the "No events found" for the month message below the grid in mobile view.
Furthermore, the View will, when the user selects a day that has no events, hide or show a message informing the user of that.
Since using `flex-order` around the wrapper turned out to be a terrible pain, I've resolved to print the messages area 2 times and to hide the second print to screen readers by default.

[Screencap](https://www.dropbox.com/s/a8wz1hswkq6r28r/Screen%20Recording%202021-05-06%20at%2015.49.35.mov?dl=0)

Edit: updated screencap link.


lucatume:
To reviewers: I'm not sure if the `eslint` step is broken or I broke it.
Please, let me know if you know more than I do!

Camwyn:
@lucatume don't worry about the eslint - but fix the views_integration tests?

src/resources/js/views/month-mobile-events.js
```diff
@@ -56,6 +59,8 @@ tribe.events.views.monthMobileEvents = {};
 	 * @return {void}
 	 */
 	obj.closeAllEvents = function( $container ) {
+		var hasDays = false;
```
paulmskim, line 62: seems like you don't use `hasDays`, maybe remove?

src/resources/js/views/month-mobile-events.js
```diff
@@ -69,12 +74,55 @@ tribe.events.views.monthMobileEvents = {};
 			if ( contentId ) {
 				$content = $container.find( '#' + contentId );
 				tribe.events.views.accordion.closeAccordion( $header, $content );
+				hasDays = true;
```
paulmskim, line 77: seems like you don't use `hasDays`, maybe remove?

src/resources/js/views/month-mobile-events.js
```diff
@@ -140,6 +188,8 @@ tribe.events.views.monthMobileEvents = {};
 			$content = $container.find( '#' + contentId );
 		}
 
+		obj.handleMobileDayClick( contentId );
```
paulmskim, line 191: `handleMobileDayClick()` expects a `boolean` as the first argument, but `contentId` is a string, is this what you intended?

src/resources/js/views/month-mobile-events.js
```diff
@@ -225,11 +275,19 @@ tribe.events.views.monthMobileEvents = {};
...
+		if ( ! isMobile ) {
+			$( obj.selectors.mobileEventsDefaultNotices ).hide();
+			$( obj.selectors.mobileEventsDynamicNotices ).hide();
+
+			if ( ! state.desktopInitialized ) {
+				obj.closeAllEvents( $container );
+				state.desktopInitialized = true;
+			}
+		} else {
+			obj.handleMobileDayClick();
```
paulmskim, line 287: `handleMobileDayClick()` expects a `boolean` as the first argument, is this what you intended?

src/views/v2/month.php
```diff
@@ -75,6 +76,30 @@ class="tribe-events-calendar-month"
...
+		<?php
+		$this->template(
+			'components/messages',
+			[
+				'attributes' => [
+					'aria-hidden' => 'true',
+					'data-js-mobile-notice' => 'default',
+				],
+			]
+		); ?>
```
mitogh, line 89:
```suggestion
		); 
		?>
```

src/resources/js/views/month-mobile-events.js
```diff
@@ -69,12 +74,55 @@ tribe.events.views.monthMobileEvents = {};
...
+	 *
+	 * @since TBD
+	 *
+	 * @param {bool} viewHasDays Whether the day selected by the user has days in it or not.
+	 *
+	 * @return {void} The method does not return a value and will hide and show message elements.
+	 */
+	obj.handleMobileDayClick = function( hasDays ) {
+		var $defaultNotices = $( obj.selectors.mobileEventsDefaultNotices );
+		var $dynamicNotices = $( obj.selectors.mobileEventsDynamicNotices );
```
paulmskim, line 95: Everything we do in the JS should be scoped to the `$container`, so instead, we call something like `$container.find( obj.selectors.mobileEventsDefaultNotices )`. If this is being called from another function, we should pass down `$container` so the function has access to it.

src/resources/js/views/month-mobile-events.js
```diff
@@ -69,12 +74,55 @@ tribe.events.views.monthMobileEvents = {};
...
+
+		if ( hasDays ) {
+			$defaultNotices.hide();
+			$dynamicNotices.hide();
```
paulmskim, line 99: Instead of using `hide()` and `show()` methods, let's add and remove the class `tribe-common-a11y-hidden`. We can add it to the selectors above and do something like `$defaultNotices.addClass( obj.selectors.hidden.className() )`

src/resources/js/views/month-mobile-events.js
```diff
@@ -225,11 +275,19 @@ tribe.events.views.monthMobileEvents = {};
...
+			if ( ! state.desktopInitialized ) {
+				obj.closeAllEvents( $container );
+				state.desktopInitialized = true;
+			}
+		} else {
+			obj.handleMobileDayClick();
+			if ( state.desktopInitialized ) {
+				state.desktopInitialized = false;
+			}
 		}
```
mitogh, line 291: Maybe a personal preference but you can get ride of the `! isMobile `

```suggestion
		if ( isMobile ) {
			obj.handleMobileDayClick();
			if ( state.desktopInitialized ) {
				state.desktopInitialized = false;
			}
		} else {
			$( obj.selectors.mobileEventsDefaultNotices ).hide();
			$( obj.selectors.mobileEventsDynamicNotices ).hide();

			if ( ! state.desktopInitialized ) {
				obj.closeAllEvents( $container );
				state.desktopInitialized = true;
			}
		}
```

src/resources/js/views/month-mobile-events.js
```diff
@@ -69,12 +74,55 @@ tribe.events.views.monthMobileEvents = {};
...
+		var visibleMessages = 0;
+
+		if ( dynamicNoticesList.length > 0 ) {
+			dynamicNoticesList.each( function( index, item ) {
+				if ( item.getAttribute( 'data-key' ) !== 'no-events-in-day' ) {
+					item.style.display = 'none';
+				} else {
+					item.style.display = '';
+					visibleMessages++;
+				}
```
mitogh, line 113: Same as the other areas inverting the conditional can remove the `!==` operator

```suggestion
				if ( item.getAttribute( 'data-key' ) === 'no-events-in-day' ) {
					item.style.display = '';
					visibleMessages++;				
				} else {
					item.style.display = 'none';
				}
```

src/Tribe/Views/V2/Views/Month_View.php
```diff
@@ -537,4 +540,38 @@ protected function setup_ical_repository_args( $per_page ) {
...
+		 *
+		 * @since TBD
+		 *
+		 * @param array<string,array<string|int,string>> A map from message types to messages for
+		 *                                                each type.
+		 * @param Month_View $view                        A reference to the View instance that is filtering its mobile messages.
+		 */
+		$mobile_messages = apply_filters( 'tribe_events_views_v2_month_mobile_messages', $mobile_messages, $this );
+
+		return $mobile_messages;
```
mitogh, line 575:
```suggestion
		return apply_filters( 'tribe_events_views_v2_month_mobile_messages', $mobile_messages, $this );
```

mitogh (changes requested):

src/resources/js/views/month-mobile-events.js
paulmskim, line 113: same here, let's add/remove the `tribe-common-a11y-hidden` class. Also, if for some reason there was another notice with a key that is not `no-events-in-day`, that would never show. Is that on purpose?

src/views/v2/month.php
```diff
@@ -75,6 +76,30 @@ class="tribe-events-calendar-month"
...
+		$this->template(
+			'components/messages',
+			[
+				'attributes' => [
+					'aria-hidden' => 'true',
+					'data-js-mobile-notice' => 'dynamic',
+				],
+				'messages'   => $mobile_messages,
+			] );
+		?>
```
paulmskim, line 101:
```suggestion
		<?php
		$this->template(
			'components/messages',
			[
				'attributes' => [
					'aria-hidden' => 'true',
					'data-js-mobile-notice' => 'default',
				],
			]
		);
		?>

		<?php
		$this->template(
			'components/messages',
			[
				'attributes' => [
					'aria-hidden' => 'true',
					'data-js-mobile-notice' => 'dynamic',
				],
				'messages'   => $mobile_messages,
			]
		);
		?>
```

paulmskim: Also curious why we have the default messages, it seems like we never use it (unless someone removes the dynamic messages)?

paulmskim: Also, I wonder if it makes sense, instead of having this one message, which seems like it's meant for each day, output only once and handle it via JS, we add a message component to each mobile day. If there are no events, then we output a message component with the mobile messages. This way, we won't need to do any extra JS, CSS, and just output a message component. It allows for a message for each day (if a user wanted), and all the showing/hiding is handled for you already.

paulmskim (changes requested):
Some changes suggested, but also mainly a question around possible approach for the FE side here, let me know your thoughts.

src/resources/postcss/views/skeleton/_header.pcss
```diff
@@ -50,9 +50,20 @@
...
 		.tribe-common--breakpoint-medium& {
 			margin-bottom: var(--spacer-7);
 			order: 1;
+
+			&:not([aria-hidden]) {
+				display: flex;
+			}
+			&[aria-hidden] {
+				display: none;
+			}
```
paulmskim, line 66: we can probably remove these, right?

src/views/v2/month.php
```diff
@@ -75,6 +75,13 @@ class="tribe-events-calendar-month"
 
 		</div>
 
+		<!-- This is the second occurrence of this, let's never show it to screen readers. -->
+		<?php
+		$this->template(
+			'components/messages', [
+			'attributes' => [ 'aria-hidden' => 'true', 'data-js-type' => 'default' ]
```
paulmskim, line 82: we can probably remove `aria-hidden` here and rely on the `tribe-common-a11y-hidden` class, since we don't actually change this in the JS, it hurts us more than not.

mitogh (dismissed):

src/views/v2/month/mobile-events/mobile-day.php
```diff
@@ -52,16 +53,24 @@
...
+			<?php $this->template( 'month/mobile-events/mobile-day/mobile-event', [ 'event' => $event ] ); ?>
 
-	<?php $this->template( 'month/mobile-events/mobile-day/more-events', [ 'more_events' => $day['more_events'], 'more_url' => $day['day_url'] ] ); ?>
+		<?php endforeach; ?>
+
+		<?php $this->template( 'month/mobile-events/mobile-day/more-events', [ 'more_events' => $day['more_events'], 'more_url' => $day['day_url'] ] ); ?>
+
+	<?php else : ?>
+
+		<?php $this->template( 'components/messages', [ 'attributes' => [ 'aria-hidden' => 'true' ], 'messages' => $mobile_messages ] ); ?>
```
paulmskim, line 73: we can probably remove `aria-hidden` here, since we don't actually change this in the JS, it hurts us more than not.

paulmskim (changes requested):
a few more smaller things, but looks good otherwise

src/views/v2/month/mobile-events/mobile-day.php
```diff
@@ -52,16 +53,27 @@
...
+		<?php endforeach; ?>
+
+		<?php $this->template( 'month/mobile-events/mobile-day/more-events', [ 'more_events' => $day['more_events'], 'more_url' => $day['day_url'] ] ); ?>
+
+	<?php else : ?>
+
+		<?php $this->template( 'components/messages', [
+			'classes' => [ 'tribe-events-header__messages--mobile','tribe-events-header__messages--day' ],
+			'messages' => $mobile_messages ]
+		); ?>
```
paulmskim, line 76:
```suggestion
		<?php
		$this->template(
			'components/messages',
			[
				'classes' => [ 'tribe-events-header__messages--mobile', 'tribe-events-header__messages--day' ],
				'messages' => $mobile_messages,
			]
		);
		?>
```

paulmskim (approved):

PR #3516, opened 2021-05-05
Correctly restore loop vars in View

lucatume:
- feat(v2/month) mobile messages below grid
- fix(V2/View.php) correctly restore the loop vars

Issue: https://theeventscalendar.atlassian.net/browse/ECP-466

This PR fixes the issue reported in the ticket and, potentially, another set of others.

The specific issues was caused by the fact that the form would set its `action` attribute to the current `$_SERVER['REQUEST_URI']` value.  
If the form shortcode came, in the HTML rendering order, **after** our shortcode (really **any** shortcode using a View), then that variable would have been altered in the context of the `View::setup_the_loop` method and, theoretically, restored after it.  

Since the `$_SERVER['REQUEST_URI']` variable was **not** restored correctly, the form, and in a similar way any other code rendering after a View and relying on the server request URI, would find the incorrect one.

[Screencap](https://www.dropbox.com/s/wv790tonpbg5b8e/Screen%20Recording%202021-05-05%20at%2015.05.07.mov?dl=0)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -223,7 +223,7 @@ Remember to always make a backup of your database and files before updating!
 
 = [5.6.1] TBD =
 
-
+* Fix - Resolve conflicts with Gravity Forms plugin that would prevent correct submission of forms data. [ECP-466]
 
 = [5.6.0] 2021-04-29 =
 
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -1004,8 +1004,8 @@ class View implements View_Interface {
 		global $wp_query;
 
 		$this->global_backup = [
-			'wp_query'   => $wp_query,
-			'$_SERVER'   => isset( $_SERVER ) ? $_SERVER : []
+			'globals::wp_query'   => $wp_query,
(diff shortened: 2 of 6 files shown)


tests/views_integration/Tribe/Events/Views/V2/Views/Latest_Past_ViewTest.php
```diff
@@ -10,6 +10,13 @@
 class Latest_Past_ViewTest extends ViewTestCase {
 	use With_Post_Remapping;
 
+	/**
```
jesseeproductions, line 13: @lucatume The files above related to the change look good. Did you intend to include the changes to the tests for latest past, HTML_cachetest, and iCalendar Request test?

bordoni (approved):

tests/views_integration/Tribe/Events/Views/V2/Views/Latest_Past_ViewTest.php
lucatume: Yes, the change is intentional. While working on this, I found out some tests would not behave the same depending on their running alone, in the context of their test case or in the context of the full test suite. The changes address that.

PR #3517, opened 2021-05-05
TEC-3797: Minification of TEC javascript

mitogh:
Make sure the break points are executed as soon as the DOM is ready
to query.

Ticket: TEC-3797
Artifact: https://d.pr/v/wy0dqW


readme.txt
```diff
@@ -226,6 +226,7 @@ Remember to always make a backup of your database and files before updating!
 * Tweak - Move messages below the calendar grid in the mobile version of Month View. [TEC-3793]
 * Tweak - Display a message to let visitors know the selected Month View day has no events in mobile. [TEC-3812]
 * Tweak - Add compatibility for the Full Site Editor navigation block. [TEC-3850]
+* Fix - Apply responsive breakpoints once the page has fully load with all JavaScript and Elements on the page [TEC-3797]
```
Camwyn, line 229:
```suggestion
* Fix - Apply responsive breakpoints once the page has fully loaded with all JavaScript and Elements on the page [TEC-3797]
```

src/views/v2/components/breakpoints.php
```diff
@@ -20,7 +20,7 @@
 }
 ?>
 <script class="tribe-events-breakpoints">
-	(function(){
+	window.addEventListener('DOMContentLoaded', function(){
```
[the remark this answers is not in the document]
mitogh: In the context of fitlerbar looks lik we can use `window.tribe.filterBar.filterBarState.setup ` what is the equivalent on `TEC` I think just firing the same function as an `iIFE` and wrap it around  `DOMContentLoaded` ensures the code is executed as expected.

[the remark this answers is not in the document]
paulmskim: `window.tribe.events.views.breakpoints.setup`. It's exactly the same as below, except instead of returning early, we add the event listener.

src/views/v2/components/breakpoints.php
```diff
@@ -20,32 +20,49 @@
 }
 ?>
 <script class="tribe-events-breakpoints">
-	(function(){
-		if ( 'undefined' === typeof window.tribe ) {
-			return;
-		}
+	(function () {
```
paulmskim, line 23: we put spaces after and before parentheses for JS:
```suggestion
	( function () {
```

src/views/v2/components/breakpoints.php
```diff
@@ -20,32 +20,49 @@
...
+		var completed = false;
 
-		if ( 'undefined' === typeof window.tribe.events ) {
-			return;
-		}
+		function initBreakpoints() {
+			if (completed === true) {
```
paulmskim, line 27:
```suggestion
			if ( completed ) {
```

src/views/v2/components/breakpoints.php
```diff
@@ -20,32 +20,49 @@
...
+				// This was fired already and completed no need to attach to the event listener.
+				document.removeEventListener('DOMContentLoaded', initBreakpoints);
```
paulmskim, line 29:
```suggestion
				document.removeEventListener( 'DOMContentLoaded', initBreakpoints );
```
and so on

paulmskim:
looks like you'll need to update snapshots @mitogh

PR #3523, opened 2021-05-10
Do not remove archive body class in post tag and cat archive

lucatume:
Issue: https://theeventscalendar.atlassian.net/browse/TEC-3846

This commit fixes an issue where our code would remove the `archive`
body class if the request is not for a taxonomy page.

To detect whether a request is for a taxonomy page or not, we would use
the `is_tax` template tag that is, though, not the correct way to do it.

The `is_tax` template tag will return `true` if:
1. The current request is for a **custom** taxonomy term page.

Since the post `tag` and `category` terms are **not** part of a custom
taxonomy, the call to `is_tax` would return `false` and we would
erroneously remove the `archive` body class in that case.

[Screencap](https://www.dropbox.com/s/n232jlb7dsvj49u/Screen%20Recording%202021-05-10%20at%2011.04.21.mov?dl=0)


tests/views_wpunit/Tribe/Events/Views/V2/Template_BootstrapTest.php
```diff
@@ -457,4 +457,88 @@ protected function setup_event_query() {
...
+	}
+
+	/**
+	 * It should correctly filter body classes for tax archives when using page template
+	 *
+	 * @test
+	 * @dataProvider page_template_tax_archive_body_classes_provider
+	 */
+	public function should_correctly_filter_body_classes_for_tax_archives_when_using_page_template
+	(
```
borkweb, line 532:
```suggestion
	public function should_correctly_filter_body_classes_for_tax_archives_when_using_page_template(
```

borkweb (approved):
The change looks good. Just a couple minor spacing things and this can be merged :dancers:

PR #3530, opened 2021-05-12
TEC-3890: Prevent to trigger `xhr` requests for hash changes.

mitogh:
Make sure hash changes on the URL are just to move the viewport to
the correct section of the page but on the same URL, for that
scenario we don't need to request a new request as the URL
didn't change.

Artifact: https://d.pr/v/5GeZUb
Ticket: TEC-3890

### DIFF
--- a/src/resources/js/views/manager.js
+++ b/src/resources/js/views/manager.js
@@ -30,7 +30,7 @@ tribe.events.views.manager = {};
  * @return {void}
  */
 ( function( $, _, obj ) {
-	const $window = $( window );
+	var $window = $( window );
 
 	/**
 	 * Selectors used for configuration and setup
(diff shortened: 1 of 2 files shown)


src/resources/js/views/manager.js
```diff
@@ -377,6 +377,11 @@ tribe.events.views.manager = {};
 		var url = target.location.href;
 		var $container = obj.getLastContainer();
 
+		// A pop state from a hash inside of the same URL has been trigger like #content  so no request is required.
+		if ( url && url.split( '#' ).length >= 2 ) {
```
paulmskim, line 381: 2 things here:

1. We can check whether a hash is empty or not by checking `target.location.hash`
2. This takes care of the cases when a url did not have a hash but now has a hash or when a url has a hash and the hash changed to something else. However, it does not address the situation where a url has a hash but now does not., in which case we should not trigger an ajax request either.

paulmskim (changes requested):

src/resources/js/views/manager.js
mitogh: 1. I wasn't aware of that property, so let's use that one instead.
2. I think that's a valid point let me take a look to see if we can put something for that use case.

borkweb (approved):
Nice changes :t-rex: :dash:

PR #3532, opened 2021-05-12
Fix Day View time separator time

lucatume:
Issue: https://theeventscalendar.atlassian.net/browse/TEC-3877

This fixes an issue where the Day View time separator HTML would use the
event start time, without consideration of the Event Timezone and of the
site timezone settings (Events > Settings > "Timezone Settings" >
"Timezone Mode".

The fix consists in using the already available
`$event->dates->display_date` object that will be set, depending on the
site timezone mode settings, to the correct date and timezone to use
when the event details require, as the name implies, displaying.

[Screencap](https://www.dropbox.com/s/h99d1bd3ajyc9cq/TEC-3867-self-qa-2.mov?dl=0)


borkweb (approved):
:dog2: :dash:

PR #3533, opened 2021-05-12
Check to ensure linked_post_types is an array

baden03:
Allows devs to completely remove the standard venu and organization cpts. Prevents Warning: array_keys() expects parameter 1 to be array, null given in Linked_Posts.php on line 67
Warning: Invalid argument supplied for foreach() in linked_Posts.php on line 72


lucatume:
Hi @baden03 ,

thanks for taking the time to open a PR.

I'm changing the branch to the current release one; the fix might not make it into the current release code, yet this will ensure it will get some attention.

lucatume:
PR merged, thanks again.

PR #3534, opened 2021-05-12
Fix single event iCalendar export

lucatume:
Issue: https://theeventscalendar.atlassian.net/browse/TEC-3886

Fix the single event export by taking into account the request for the
single event.

Before we would build the View directly on the `view` context location
return value, when looking at single events, that return value would be
`single-event` that, not resolving to any View, would default to the
current default View (usually List).

This fix specially handles the iCalendar export request coming from
single event to return that single event ID only.

[Screencap](https://www.dropbox.com/s/dktmrh3con0bl1z/TEC-3886-self-qa.mov?dl=0)


borkweb (approved):
:hotdog:

PR #3539, opened 2021-05-14
Fix the customizer CSS template string for v2 views

apmatthews:
Some elements of v2 views are not properly styled according to customizer settings. This is due to a missing curly brace in the CSS template string, which causes a large portion of customizer CSS to fail. The most obvious example is the "Find Events" search button, which should respect the `global_elements.accent_color` setting from the customizer.


lucatume:
Hi @apmatthews,

thanks for taking the time to open a PR and provide the fix.

I'm changing the base of this PR to target the current release branch: it _might_ not be merged into this release, but it's a good first step.

Camwyn (dismissed):
Great catch!

lucatume:
Thanks again @apmatthews!

Camwyn:
Hey @apmatthews!

Thanks for catching this! I'm creating a new PR for this to ensure you get attribution in the changelog - I want you to get credit for the fix!
https://github.com/the-events-calendar/the-events-calendar/pull/3547
 
We've been working on improving the amount of control and ease of use of the Customizer for users, so expect even more changes soon!

apmatthews:
Awesome, thank you both for the quick turn around!

I'm loving the new views by the way. Looking forward to seeing what comes next!

@Camwyn @lucatume

welbinator:
Hello!

We're reaching out because we've recently released an update to our plugins that we believe should fix one or more bugs reported in this thread.

Please make sure all your Events Calendar related plugins are updated to the latest version and let us know if your issue still persists after the update.

Thank you for your patience as we work through these issues!

Current plugin versions:

The Events Calendar 5.7.0
Events Calendar Pro 5.7.0
Filter Bar 5.1.2
Virtual Events 1.4.1

PR #3552, opened 2021-05-26
Make sure not found messages are render on mobile

mitogh:
Make sure the list view on mobile displays the not found message when no
events are present on mobile layout.

Ticket: TEC-3812

This is done on top of https://github.com/the-events-calendar/the-events-calendar/pull/3514 to make sure this is not a regressive issue.

### DIFF
--- a/src/views/v2/list.php
+++ b/src/views/v2/list.php
@@ -53,6 +53,7 @@ if ( empty( $disable_event_search ) ) {
 
 		<header <?php tribe_classes( $header_classes ); ?>>
 			<?php $this->template( 'components/messages' ); ?>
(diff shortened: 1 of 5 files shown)


src/views/v2/list.php
```diff
@@ -53,6 +53,7 @@
 
 		<header <?php tribe_classes( $header_classes ); ?>>
 			<?php $this->template( 'components/messages' ); ?>
+			<?php $this->template( 'components/messages', [ 'classes' => [ 'tribe-events-header__messages--mobile' ] ] ); ?>
```
borkweb, line 56: G and I are looking at this in a call. We think you may have a file you haven't committed where `$classes` gets merged into the array of classes being output. `src/views/v2/components/messages.php`. We're guessing that the change just wasn't pushed up to this PR

borkweb (approved):
:ballet_shoes:

PR #3555, opened 2021-05-28
[TEC-3937] Fix for single view and Event Tickets compatibility when using `twentynineteen`.

juanfra:
🎫  [TEC-3937]

The styles that we're applying for the 2019 overrides are taking over the `entry` of any other thing that's used in that page. In our case (Event Tickets) it's adding a margin top on every item of the tickets block. I assume for WooCommerce or any other thing that uses the post classes it'll happen the same.

On hold until is ticketed and assigned for a given sprint.

https://www.loom.com/share/e9dc59b71ef94f2382edb3af72cbdbb1

[TEC-3937]: https://theeventscalendar.atlassian.net/browse/TEC-3937

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -223,7 +223,7 @@ Remember to always make a backup of your database and files before updating!
 
 = [5.7.1] TBD =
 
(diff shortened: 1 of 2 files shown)


bordoni:
Once this finds a release, we are good to merge.

borkweb (approved):

PR #3557, opened 2021-06-06
Trigger update to _EventCost when updating events via the Block Editor | TEC-3141

borkweb:
Similar to the Event API call that is done on event additions/updates in the classic editor, we need to give the opportunity for ET/ET+ and other sources to influence the cost fields.

🎥 https://d.pr/v/tMH4z7
🎫 [TEC-3141]

[TEC-3141]: https://theeventscalendar.atlassian.net/browse/TEC-3141

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -223,6 +223,8 @@ Remember to always make a backup of your database and files before updating!
 
 = [5.7.1] TBD =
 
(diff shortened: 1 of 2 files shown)


borkweb:
The tests are currently failing because it is Sunday. I'll re-run them tomorrow.

juanfra (approved):

PR #3559, opened 2021-06-09
Remove aria attributes from Featured and Recurring icons.

Camwyn:
Removes the incorrect aria-label attribute from Recurring and Feared icons.

Requires https://github.com/the-events-calendar/tribe-common/pull/1608


src/views/v2/month/calendar-body/day.php
```diff
@@ -110,9 +115,9 @@ class="tribe-events-calendar-month__mobile-events-icon tribe-events-calendar-mon
 			?>
 			<em
 				class="tribe-events-calendar-month__mobile-events-icon tribe-events-calendar-month__mobile-events-icon--event"
-				aria-label="<?php echo esc_attr( $has_events_label ); ?>"
 				title="<?php echo esc_attr( $has_events_label ); ?>"
 			>
+				<title><?php echo esc_html( $has_events_label ); ?></title>
```
paulmskim, line 120: This probably won't be valid. Best to add a visually hidden span with the `$has_events_label` as text.

paulmskim (changes requested):

src/views/v2/month/calendar-body/day.php
Camwyn: It's valid: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title

paulmskim: So that only provides an accessible text for SVG container or graphic element, neither of which `<em>` is a part of.

Camwyn: Oh, dang, I thought I'd put it _in_ the svg 🤦🏼

bordoni (approved):

PR #3562, opened 2021-06-15
Cast tribe_get_request_var to array.

Camwyn:
Since the URL param is not namespaced, we cannot assume we will get back _our_ info.

This caused a plugin conflict and fatal here as Analytify also uses the `view_data` param - and it's a simple string.

Possibly a larger question here as to how/why the function is being called in the first place on an admin page that isn't ours?

[TEC-3946]

[TEC-3946]: https://theeventscalendar.atlassian.net/browse/TEC-3946

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -225,7 +225,7 @@ Remember to always make a backup of your database and files before updating!
 
 * Fix - Properly recalculate event cost when creating events via the Block Editor. [TEC-3141]
 * Fix - Resolve a compatibility issue with the new single view and the tickets block when using the `twentynineteen` theme. [TEC-3937]
-* Fix - Resolve a compatibility issue with the Analytify dashboard. [TEC-3946]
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -225,6 +225,7 @@ Remember to always make a backup of your database and files before updating!
 
 * Fix - Properly recalculate event cost when creating events via the Block Editor. [TEC-3141]
 * Fix - Resolve a compatibility issue with the new single view and the tickets block when using the `twentynineteen` theme. [TEC-3937]
+* Fix - Resolve a compatibility issue with the Analytify dashboard. [TEC-3946]
```
borkweb, line 228: Perhaps a broadly applicable changelog entry:

```suggestion
* Fix - Ensure that `view_data` is an array when fetching values from the request. [TEC-3946]
```

borkweb (changes requested):
One tiny wording change

borkweb (approved):
:+1:

PR #3567, opened 2021-06-24
Kadence hooks

Camwyn:
This adds a collection of hooks to the single event templates per the ticket.

[TEC-3948]

[TEC-3948]: https://theeventscalendar.atlassian.net/browse/TEC-3948

### DIFF
--- a/src/views/single-event-blocks.php
+++ b/src/views/single-event-blocks.php
@@ -31,9 +31,7 @@ if ( ! empty( $event_id ) && function_exists( 'tribe_is_recurring_event' ) ) {
 	<?php if ( $is_recurring ) { ?>
 		<?php $this->template( 'single-event/recurring-description' ); ?>
 	<?php } ?>
-	<?php do_action( 'tribe_events_single_event_block_before_the_content' ) ?>
 	<?php $this->template( 'single-event/content' ); ?>
-	<?php do_action( 'tribe_events_single_event_block_after_the_content' ) ?>
 	<?php $this->template( 'single-event/comments' ); ?>
 	<?php $this->template( 'single-event/footer' ); ?>
 </div>
--- a/src/views/single-event-blocks.php
+++ b/src/views/single-event-blocks.php
@@ -31,7 +31,9 @@ if ( ! empty( $event_id ) && function_exists( 'tribe_is_recurring_event' ) ) {
 	<?php if ( $is_recurring ) { ?>
 		<?php $this->template( 'single-event/recurring-description' ); ?>
 	<?php } ?>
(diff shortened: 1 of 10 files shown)


borkweb (dismissed):
Just needs a changelog!

src/views/single-event-blocks.php
```diff
@@ -31,7 +31,9 @@
 	<?php if ( $is_recurring ) { ?>
 		<?php $this->template( 'single-event/recurring-description' ); ?>
 	<?php } ?>
+	<?php do_action( 'tribe_events_single_event_block_before_the_content' ) ?>
```
bordoni, line 34: In this case here they can use the FIlters from the template system.

src/views/single-event-blocks.php
```diff
@@ -31,7 +31,9 @@
...
 	<?php $this->template( 'single-event/content' ); ?>
+	<?php do_action( 'tribe_events_single_event_block_after_the_content' ) ?>
```
bordoni, line 36: In this case here they can use the FIlters from the template system.

src/views/v2/default-template.php
```diff
@@ -15,5 +15,25 @@
...
 get_header();
+
+/**
+ * Provides an action that allows for the injection of HTML at the top of the template after the header.
+ *
+ * @since TBD
+ *
+ * @param string $eventDisplay The string representation (slug) of the displayed view - "month".
+ */
+do_action( 'tribe_default_events_views_v2_template_after_header', $eventDisplay );
```
bordoni, line 26: Event Display is not a variable in V2. And on this case we can use a entry point here.

src/views/v2/default-template.php
```diff
@@ -15,5 +15,25 @@
...
 echo tribe( Template_Bootstrap::class )->get_view_html();
+
+/**
+ * Provides an action that allows for the injections of HTML at the bottom of the template before the footer.
+ *
+ * @since TBD
+ *
+ * @param string $eventDisplay The string representation (slug) of the displayed view - "month".
+ */
+do_action( 'tribe_default_events_views_v2_template_before_footer', $eventDisplay );
```
bordoni, line 37: Event Display is not a variable in V2. And on this case we can use a entry point here.

bordoni (changes requested):

src/views/v2/default-template.php
Camwyn: Changed the param, but entry_point isn't available here.

src/views/v2/default-template.php
Camwyn: Changed the param, but entry_point isn't available here.

bordoni (approved):

PR #3569, opened 2021-07-01
Remove trailing comma that fatals in php5.6

Camwyn:


jesseeproductions (approved):
looks good merge away

PR #3570, opened 2021-07-01
Tec 3952/bugfix random border appearing on avada

EugeneKyale:


EugeneKyale:
Artifact: This PR fixes the issue of a random border appearing for Avada users on the single events page. I've done some dev testing to confirm that CSS snippet wasn't being used anywhere else. 
image

the-events-calendar.php
```diff
@@ -73,4 +73,4 @@ function tribe_events_not_php_version_plugin_name( $names ) {
 Tribe__Events__Main::instance();
 
 register_activation_hook( TRIBE_EVENTS_FILE, array( 'Tribe__Events__Main', 'activate' ) );
-register_deactivation_hook( TRIBE_EVENTS_FILE, array( 'Tribe__Events__Main', 'deactivate' ) );
+register_deactivation_hook( TRIBE_EVENTS_FILE, array( 'Tribe__Events__Main', 'deactivate' ) );
```
Camwyn, line 76: Not a problem, but I'd avoid this kind of thing in commits. There's not real change here - just thrash due to your IDE removing the end line?

Camwyn (approved):

src/resources/postcss/views/full/_single-event.pcss
```diff
@@ -234,7 +234,6 @@
 
 	/* TODO: move .event-tickets declaration to ET in the future if still needed,
 	/* depending on template updates */
-	&:before,
```
paulmskim, line 237: Removing this will remove it for all themes, not just Avada. If this is an Avada specific issue, create a theme override at the bottom of this file for Avada overrides.

paulmskim (changes requested):

the-events-calendar.php
EugeneKyale: Hi @Camwyn I had a situation where the tests weren't running because I only made changes to the CSS file. I removed the extra line intentionally to trigger the tests. What's the recommended way to trigger tests in this scenario?

borkweb (approved):
:+1:

PR #3575, opened 2021-07-07
Add check for incorrect context

Camwyn:
Add an additional check in get_view_html() to ensure that we still show the single event when the context is wrong.

Need to explore further _why_ the context is wrong :/

[TEC-3975]

[TEC-3975]: https://theeventscalendar.atlassian.net/browse/TEC-3975

### DIFF
--- a/src/Tribe/Views/V2/Template_Bootstrap.php
+++ b/src/Tribe/Views/V2/Template_Bootstrap.php
@@ -210,8 +210,10 @@ class Template_Bootstrap {
 			return $pre_html;
 		}
 
+		$is_single_event = 'single-event' === $view_slug || ( $query->is_singular && $query->tribe_is_event_query );
+
 		$should_display_single = (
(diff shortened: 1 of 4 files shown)


src/Tribe/Views/V2/Template_Bootstrap.php
```diff
@@ -210,8 +210,10 @@ public function get_view_html() {
 			return $pre_html;
 		}
 
+		$is_single_event = 'single-event' === $view_slug || ( $query->is_singular && $query->tribe_is_event_query );
```
Camwyn, line 213: When we get here, `$view_slug` has defaulted to "month" for some reason yet `$query->is_singular `& `$query->tribe_is_event_query` are both true.

bordoni (changes requested):
Changelog.

bordoni (approved):

PR #3576, opened 2021-07-08
TEC-3987 Ensure Block Editor saves cost data when submitted

EugeneKyale:
[TEC-3987]

[TEC-3987]: https://theeventscalendar.atlassian.net/browse/TEC-3987


EugeneKyale:
Artifact: https://www.loom.com/share/5d8e601dc58e4f89ba3e969b7344dc0c

src/Tribe/Editor/Meta.php
```diff
@@ -194,7 +194,19 @@ public function meta_backwards_compatibility( $data, $post, $request ) {
 	public function update_cost( $post_data, $request ) {
 		$post_id = $request->get_param( 'id' );
 
-		\Tribe__Events__API::update_event_cost( $post_id );
+		// Fetch cost data from the request (if set).
+		$json = $request->get_json_params();
+
+		// Set an appropriate default for the cost.
+		$default = (array) tribe_get_cost( $post_id );
```
Camwyn, line 201: Suggestion: I'd use `$cost` or something similar, rather than `$default`, since we may change it (and then it's no longer the default?)

Probably above as well 🤔

src/Tribe/API.php
```diff
@@ -480,11 +480,12 @@ protected static function prepare_event_date_meta( $event_id, $data ) {
...
 		 *
 		 * @param $event_id
+		 * @param array $default The default cost for the Event.
 		 */
-		public static function update_event_cost( $event_id ) {
+		public static function update_event_cost( $event_id, $default = [] ) {
 			// Loads current event costs, on construct
 			// Tribe__Events__Tickets__Tickets->get_ticket_prices() adds them to this filter
-			$event_cost = (array) apply_filters( 'tribe_events_event_costs', [], $event_id );
+			$event_cost = (array) apply_filters( 'tribe_events_event_costs', $default, $event_id );
```
Camwyn, line 488: Needs a docblock

Camwyn (changes requested):
The docblock is necessary, the other is just a suggestion.

paulmskim:
@EugeneKyale a few questions:
1. Towards the end of your screencast, you had the ticket, and you added the price range. When you refreshed in the FE, the range showed correctly as 50 - 100, but when you refreshed in the block editor, it only showed up as 50. Is this intended/expected?
2. The price block you used the price 50 - 100, where 100 was the price of one of the tickets. What happens when the price is something like 30-50? Or just 50?
3. Leah left a comment here in the ticket: https://theeventscalendar.atlassian.net/browse/TEC-3987?focusedCommentId=98409 and it seems that there are some scenarios with tickets that may need checking, were you able to check these?

src/Tribe/Editor/Meta.php
```diff
@@ -194,7 +194,19 @@ public function meta_backwards_compatibility( $data, $post, $request ) {
...
+
+		// If the cost is set in the submitted data, set THAT as the default.
+		if ( isset( $json['meta']['_EventCost'] ) ) {
+			$default = [ $json['meta']['_EventCost'] ];
+		}
```
paulmskim, line 206: We have a utility class called `Tribe__Utils__Array` that we use to get values from an array when we don't know whether the values are set, and return a default if not. It might look something like this:

```
$json = $request->get_json_params();
$meta = Tribe__Utils__Array::get( $json, 'meta', [] );

if ( isset( $meta['_EventCost'] ) ) {
    ...
}
```

If the meta is set, it'll get the value, otherwise return an empty array.

paulmskim, line 206: we can write this as a ternary, rather than setting, then overwriting.

paulmskim (changes requested):

src/Tribe/Editor/Meta.php
EugeneKyale: @paulmskim here's a video which covers all the scenarios described in the ticket. https://www.loom.com/share/74e05e0df370412988f1722ee27c7046

In regard to this question **_Towards the end of your screencast, you had the ticket, and you added the price range. When you refreshed in the FE, the range showed correctly as 50 - 100, but when you refreshed in the block editor, it only showed up as 50. Is this intended/expected?_** I spoke to @borkweb about this and he mentioned that this issue has been there for a while so I think it needs a separate bug ticket.

paulmskim (dismissed):
just need to resolve stephen's note

src/Tribe/Editor/Meta.php
EugeneKyale: @Camwyn I've updated this to $cost

bordoni (dismissed):

bordoni:
Lets make sure tests are passing, but the code is looking good.

src/Tribe/API.php
Camwyn: Yeah, sorry - I operate under the "if you change it you own it" 😉 rule - it was not that you did something wrong @EugeneKyale

Camwyn (dismissed):

Camwyn (approved):

PR #3578, opened 2021-07-09
Inno/view labels filter

andrasguseo:
The added filter(s) should allow the user to change the view Labels to anything they like easily, with the following snippet:

```add_filter( 'tribe_events_views_v2_view_label', 'my_custom_view_labels' );

function my_custom_view_labels( $view_label ) {
	$new_labels = [
		'Month'   => 'Grid',
		'List'    => 'Listed',
		'Day'     => 'One day',
		'Week'    => '7 days',
		'Photo'   => 'Masonry',
		'Map'     => 'Location',
		'Summary' => 'Condensed',
	];

	if ( array_key_exists( $view_label, $new_labels ) ) {
		$view_label = $new_labels[ $view_label ];
	}

	return $view_label;
}
```

Works in any language, even on multilingual sites.
Though for non-English languages the user can just change the translation itself, they can also do it with the snippet.
Changing labels in multiple languages only requires the addition of all variants. For example for month view in English and Spanish the snippet would need to be adjusted like:

```
	$new_labels = [
		'Month'   => 'Grid',
		'Mes'      => 'Red',
		...
	];
```

### DIFF
--- a/src/views/v2/components/events-bar/views.php
+++ b/src/views/v2/components/events-bar/views.php
@@ -23,6 +23,15 @@ $view_selector_classes = [
 	'tribe-events-c-view-selector--labels' => empty( $disable_event_search ),
 	'tribe-events-c-view-selector--tabs'   => $is_tabs_style,
 ];
+
+/**
+ * Filters the labels for the views.
+ *
+ * @since TBD
+ *
+ * @param string $view_label  Current label of the view.
+ */
+$view_label = apply_filters( 'tribe_events_views_v2_view_label', $view_label );
 ?>
 <div class="tribe-events-c-events-bar__views">
 	<h3 class="tribe-common-a11y-visual-hide">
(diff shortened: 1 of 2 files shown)


bordoni:
I wonder if we should make use of `tribe_events_views_v2_view_public_views` or `tribe_events_views_v2_publicly_visible_views`, which both would allow filtering of the Label.

andrasguseo:
My 2 cents:
I tried `tribe_events_views_v2_publicly_visible_views`. It does work.

```add_filter( 'tribe_events_views_v2_publicly_visible_views', 'mycv' );
function mycv ( $views ){
	$views['month']->view_label = "Grid";
	$views['week']->view_label = "7-days";
	return $views;
}```

The above did the trick. 
However, if  I don't have week view enabled, then
- The month view label will not get changed
- week view will be added to the dropdown
- the added week view will not work

That probably could be avoided by adding different checks. But that makes it more complex.
Based on this I find this solution more prone to errors and more complex.

My recommendation allows a simpler execution.

All in all I'm okay if this gets rejected, it was a nice exercise for me. :)

borkweb:
This is stale and archived elsewhere. Closing this PR and deleting the branch.

PR #3580, opened 2021-07-12
fix(Customizer/General_Theme) get value if set

lucatume:
Issue: n/a

This PR fixes an issue where a Customizers setting value would be fetched without checking if it's set.

Related to: https://github.com/the-events-calendar/tribe-common/pull/1619 and https://app.travis-ci.com/github/the-events-calendar/tribe-common/builds/232502340#L2084


src/Tribe/Customizer/General_Theme.php
```diff
@@ -33,7 +33,9 @@ public static function instance() {
 	 * @return string
 	 */
 	protected function get_button_bg_color( $settings = [] ) {
-		$scheme = $this->sanitize_featured_color_choice( $settings['featured_color_scheme'] );
+		$scheme = isset( $settings['featured_color_scheme'] )
```
Camwyn, line 36: This works, but we should also just not call this if it's not set?

Camwyn: I pushed a tweak.

Camwyn (approved):
Approving but I added code.

borkweb (dismissed):
Tests appear to be failing

PR #3581, opened 2021-07-12
Add check for shortcodes that fly under the radar.

Camwyn:
[ECP-888]

This code is better in TEC since ECP relies on the function here for asset loading.

[ECP-888]: https://theeventscalendar.atlassian.net/browse/ECP-888


src/Tribe/Views/V2/Template_Bootstrap.php
```diff
@@ -323,15 +324,38 @@ public function should_load( $query = null ) {
...
+				if ( ! empty( $queried_object->post_content ) ) {
+					// Classic Editor
+					$is_shortcode = has_shortcode( 'tribe_events', $queried_object->post_content );
+					// Bollocks
+					$is_shortcode = $is_shortcode || false !== stripos( $queried_object->post_content, '[tribe_events');
+				}
+			}
+
+			$should_load = $should_load || $is_shortcode;
+		}
```
bordoni, line 350: I see what this does, but I am not entirely sure why this is needed for shortcodes in specific.

I would love to dig a bit more into why we are doing it, instead of the actual fix which is valid, but not super "clean".

bordoni (changes requested):

Camwyn:
Moved: https://github.com/the-events-calendar/events-pro/pull/1798

PR #3583, opened 2021-07-14
[TEC-3952] Target Avada Fusion Builder class to remove border

EugeneKyale:
@Camwyn This is an updated PR where I'm targeting avada fusion builder class to hide random border appearing here: 
image


Camwyn (dismissed):
@EugeneKyale  - just to be sure this change still affects all the views?

Camwyn (approved):

PR #3600, opened 2021-08-04
Fix the issue of the Event Date Time block > Timezone dropdown taking up a lot of space. [TEC-4029]

EugeneKyale:
Here's a description of what this PR fixes - https://www.loom.com/share/c2e5c1e4a7a148d4b0835025e453d10a

Jira Ticket - https://theeventscalendar.atlassian.net/browse/TEC-4029

### DIFF
--- a/src/modules/blocks/event-datetime/style.pcss
+++ b/src/modules/blocks/event-datetime/style.pcss
@@ -145,7 +145,7 @@
 	align-items: center;
 }
 
-// We need to enforce this override the value set by the SelectControl component. [TEC-4029]
+// We need to enforce this rule to override the value set by the SelectControl component. [TEC-4029]
 .tribe-editor__date-time__time-zone-setting {
     height: auto !important;
 }
--- a/src/modules/blocks/event-datetime/style.pcss
+++ b/src/modules/blocks/event-datetime/style.pcss
@@ -145,7 +145,8 @@
 	align-items: center;
 }
 
-// We need to enforce this rule to override the value set by the SelectControl component. [TEC-4029]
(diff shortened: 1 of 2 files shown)


src/modules/blocks/event-datetime/style.pcss
```diff
@@ -144,3 +144,9 @@
 	display: flex;
 	align-items: center;
 }
+
+/* We need to enforce this rule to override the value set by the SelectControl component. [TEC-4029] */
+
+.tribe-editor__date-time__time-zone-setting {
+    height: auto !important;
```
Camwyn, line 151: I hate using important...
Couple things - 

- I can reproduce locally (which gives me an edge you didn't have).
- While I suspected it may come from our webpack code, it appears the issue is the class from either the gutenberg plugin - or the core block editor: https://d.pr/i/8cqH4E
- You can get rid of the `!important` by using a more precise selector: https://d.pr/v/849efI
```css
.components-select-control[class*="tribe-editor"] {
    height: auto;
}
```
This has the added benefit of targeting _all_ our selects, in case this is happening somewhere else and we haven't caught it?

CC: @paulmskim

EugeneKyale: Thanks @Camwyn, I've tested the issue using your selector and it works well. I've made the update.

Camwyn (dismissed):

Camwyn (approved):

EugeneKyale:
Hey @paulmskim this PR is approved but I'm unable to merge it myself because the tests are not running since its just a CSS update. Do you have the necessary permissions to merge it without tests?

PR #3605, opened 2021-08-06
Ensure the year is displayed in the block editor event single page for events that are occurring in a different year from the current one. [TEC-3740]

EugeneKyale:
This PR fixes the issue described in this ticket : https://theeventscalendar.atlassian.net/secure/RapidBoard.jspa?rapidView=163&projectKey=BTPLAN&modal=detail&selectedIssue=TEC-3740&assignee=5feb71d1332c3a0107acdab7

Here's a video describing what the PR entails: https://www.loom.com/share/3dbed484df6240b68558af6f452b2196


src/views/blocks/event-datetime.php
```diff
@@ -35,10 +35,14 @@
 );
 
 $time_format    = tribe_get_time_format();
-$date_format    = tribe_get_date_format( $use_yearless_format );
+$date_format    = tribe_get_date_format( true );
```
paulmskim, line 38: It looks like a logic fault, couldn't we just do the following?

```suggestion
$date_format    = tribe_get_date_format( ! $use_yearless_format );
```

paulmskim (changes requested):

paulmskim:
It would also be good to show the case where the year shouldn't display, and make sure that's working properly

EugeneKyale:
> It would also be good to show the case where the year shouldn't display, and make sure that's working properly

@paulmskim just checked this and found out that the events currently don't respect any date format other than F j, Y. This is the case for the classic editor as well. This probably requires a new ticket because its happening in all editors. What do you think?

src/views/blocks/event-datetime.php
EugeneKyale: Yeah that neater, I've updated this.

paulmskim (dismissed):

paulmskim (approved):

PR #3625, opened 2021-08-30
[TEC-3332] Ensure that venue state or province are displayed next to the city in list view.

EugeneKyale:
Ticket : https://theeventscalendar.atlassian.net/browse/TEC-3332
Artifact : https://www.loom.com/share/4bb7d07f03954fa8bfd9b2458ef54351

### DIFF
--- a/src/views/v2/list/event/venue.php
+++ b/src/views/v2/list/event/venue.php
@@ -32,12 +32,12 @@ $address              = $venue->address . ( $venue->address && $append_after_add
 	<span class="tribe-events-calendar-list__event-venue-address">
 		<?php echo esc_html( $address ); ?>
 
-		<?php if ( $venue->city ) : ?>
-			<?php echo esc_html( $venue->city . $separator ) ?>
+		<?php if ( ! empty( $venue->city ) ) : ?>
+			<?php echo esc_html( $venue->city . $separator ); ?>
 		<?php endif; ?>
 		
 		<?php if ( $append_after_address ) : ?>
-			<?php echo esc_html( reset( $append_after_address ) ) ?>
+			<?php echo esc_html( reset( $append_after_address ) ); ?>
 		<?php endif; ?>
 	</span>
 </address>
--- a/src/views/v2/list/event/venue.php
+++ b/src/views/v2/list/event/venue.php
@@ -30,14 +30,16 @@ $address              = $venue->address . ( $venue->address && $append_after_add
 		<?php echo wp_kses_post( $venue->post_title ); ?>
 	</span>
 	<span class="tribe-events-calendar-list__event-venue-address">
-		<?php echo esc_html( $address ); ?>
+		<?php 
+			echo esc_html( $address ); 
 
-		<?php if ( ! empty( $venue->city ) ) : ?>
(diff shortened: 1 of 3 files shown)


src/views/v2/list/event/venue.php
```diff
@@ -31,8 +31,13 @@
 	</span>
 	<span class="tribe-events-calendar-list__event-venue-address">
 		<?php echo esc_html( $address ); ?>
+
+		<?php if ( $venue->city ) : ?>
+			<?php echo esc_html( $venue->city . $separator ) ?>
+		<?php endif; ?>
+		
 		<?php if ( $append_after_address ) : ?>
```
paulmskim, line 39: a lot of open and close php tags, let's make this under one open and close tag

src/views/v2/list/event/venue.php
```diff
@@ -31,8 +31,13 @@
 	</span>
 	<span class="tribe-events-calendar-list__event-venue-address">
 		<?php echo esc_html( $address ); ?>
+
+		<?php if ( $venue->city ) : ?>
```
paulmskim, line 35: check if empty rather than just checking venue city

src/views/v2/list/event/venue.php
```diff
@@ -31,8 +31,13 @@
...
+			<?php echo esc_html( $venue->city . $separator ) ?>
+		<?php endif; ?>
+		
 		<?php if ( $append_after_address ) : ?>
-			<?php echo esc_html( reset( $append_after_address ) ); ?>
+			<?php echo esc_html( reset( $append_after_address ) ) ?>
```
paulmskim, line 40: keep semicolon

paulmskim (changes requested):

src/views/v2/list/event/venue.php
EugeneKyale: I've refactored this to be under one open and close tag

src/views/v2/list/event/venue.php
EugeneKyale: I've updated this to use if ! empty

src/views/v2/list/event/venue.php
EugeneKyale: should be updated

src/views/v2/list/event/venue.php
```diff
@@ -22,17 +22,24 @@
...
+			echo esc_html( $address ); 
+
+			if ( ! empty( $venue->city ) ) : 
+				echo esc_html( $venue->city . $separator ); 
+		 	endif;
+		
+		 	if ( $append_after_address ) : 
+				echo esc_html( reset( $append_after_address ) ); 
+		 	endif; 
+		?>
```
paulmskim, line 43:
```suggestion
		<?php 
		echo esc_html( $address ); 

		if ( ! empty( $venue->city ) ) : 
			echo esc_html( $venue->city . $separator ); 
		endif;

		if ( $append_after_address ) : 
			echo esc_html( reset( $append_after_address ) ); 
		endif; 
		?>
```

paulmskim (approved):

PR #3628, opened 2021-09-01
Unescape the title more thoroughly

Camwyn:
Use `html_entity_decode` instead of `htmlspecialchars_decode` so that other special characters get unescaped as well.
Also fixed apostrophes in the page title.

This prevents anything else hooking into the_title from double-escaping characters.

All this gets escaped later, as seen below:
:camera: https://d.pr/v/vYJ7nJ

[TEC-4049]
[TEC-4058]

[TEC-4049]: https://theeventscalendar.atlassian.net/browse/TEC-4049
[TEC-4058]: https://theeventscalendar.atlassian.net/browse/TEC-4058


src/Tribe/Views/V2/View.php
```diff
@@ -1778,7 +1778,7 @@ public function get_title( array $events = [] ) {
 		 */
 		$title = apply_filters( "tribe_events_views_v2_view_{$slug}_title", $title, $this );
 
-		return htmlspecialchars_decode($title);
+		return html_entity_decode($title, ENT_QUOTES);
```
skyshab, line 1781: Space inside parentheses 

```suggestion
		return html_entity_decode( $title, ENT_QUOTES );
```

skyshab (changes requested):

EugeneKyale:
Looks good, we only need a change log.

EugeneKyale (approved):

PR #3640, opened 2021-09-10
Fix the error from missing `$local_time_zone` var.

Camwyn:
It appears that this var got used in some code then copied to somewhere it's not defined.

[TEC-3791]

This copies the definition from `Tribe__Events__Admin__Timezone_Updater` so we don't throw `Undefined variable` warnings.

🎥 https://d.pr/v/nemEKf 

[TEC-3791]: https://theeventscalendar.atlassian.net/browse/TEC-3791


EugeneKyale (approved):

EugeneKyale:
Change log added

Camwyn:
This was a patch - it did not need a changelog.

EugeneKyale (approved):

PR #3657, opened 2021-09-30
[TEC-3042] - Ensure a map preview is displayed on the venue block in the admin area when using TEC's default Google API Key.

EugeneKyale:
This PR ensures that a map preview is displayed on the venue block in the admin area when using TEC's default Google API Key. According to the recommendations in the ticket, we will display an iframe whenever a user is using a our default key. 

Ticket : https://theeventscalendar.atlassian.net/browse/TEC-3042
Artifact : https://www.loom.com/share/c588459bdc5e4ee6bd70170efda9a3c6


src/modules/elements/google-map/element.js
```diff
@@ -367,6 +368,16 @@ export default class GoogleMap extends Component {
 		);
 	}
 
+	renderIframe() {
+		return (
+			<Iframe
+				url="https://www.google.com/maps/embed?pb=!1m10!1m8!1m3!1d65369183.36050215!2d0!3d0!3m2!1i1024!2i768!4f13.1!5e0!3m2!1sen!2sus!4v1633005420084!5m2!1sen!2sus" // eslint-disable-line max-len
+				width="450px"
```
Camwyn, line 375: Do we have a style to cover smaller mobile widths?
I know this is the admin, but still... a `max-width: 100%` could go a long way toward avoiding frustration and a support request for a too-wide iframe in a narrow page? (sounds like a side-scroll nightmare?)

Camwyn (approved):
Looks good, added a question that's more strategy/design so this PR doesn't depend on it.

src/modules/elements/google-map/element.js
EugeneKyale: @Camwyn We currently don't have a mobile design for this. I believe this will be addressed in the next version of the re-skin. Ideally we should have the iframe in a single column same as on the frontend.

PR #3672, opened 2021-10-22
Comment error handler warning for undefined key on array

stratease:
This one I was not able to 100% recreate, I did not know which pages or specific steps were causing the error. I was able to recreate the warning by unsetting the `type__in` key in the array, and then applied a fix for that scenario.

This probably was not showing previously because it is an escalated error level in [php 8](https://www.php.net/manual/en/migration80.incompatible.php).

[TEC-4093]

https://www.loom.com/share/300247ee42e14fccaf2321adb2cc5a8b

[TEC-4093]: https://theeventscalendar.atlassian.net/browse/TEC-4093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stratease:
> Looks good, but has some merge conflicts to resolve.

Merged!

jesseeproductions (approved):
All Set!

PR #3677, opened 2021-10-28
Added shortcode parser on hook that overwrites the_content

stratease:
The content hook for the Venue, Organizer (and other pages potentially?) were never parsing the content for shortcodes, bypassing that step via logic [here](https://github.com/the-events-calendar/the-events-calendar/blob/master/src/Tribe/Views/V2/Template/Page.php#L325) and [here](https://github.com/the-events-calendar/tribe-common/blob/master/src/functions/template-tags/post.php#L36) that caused it to never happen.

[ECP-956]

https://www.loom.com/share/69054e9a6916485f8f5d5da37c722b38

[ECP-956]: https://theeventscalendar.atlassian.net/browse/ECP-956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Camwyn (approved):
🩳

PR #3695, opened 2021-11-11
Add body to selector so customizations carry through view refreshes.

Camwyn:
The js targets get re-created on view change. This ensures we have at least one that will exist after a view change.

This might down the line require some fine tuning. We may be able to remove the other selectors, but for now I'd prefer to keep them.


jesseeproductions (approved):

paulmskim:
Let's be sure to add a ticket and an artifact next time.

PR #3697, opened 2021-11-12
Target a body class only added when in the Customizer preview.

Camwyn:
Stop relying on targets that are only present at the whim of themes and filters 😉 

[TEC-4055]

Instead of relying on classes added by themes, or core (and removable via a filter) let's add our own.

[TEC-4055]: https://theeventscalendar.atlassian.net/browse/TEC-4055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Views/V2/Customizer/Configuration.php
+++ b/src/Tribe/Views/V2/Customizer/Configuration.php
@@ -41,7 +41,7 @@ class Configuration {
 	 * @return string The selector string.
 	 */
 	public static function get_selector() {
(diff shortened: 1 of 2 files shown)


stratease (dismissed):

stratease:
Looks good!

stratease (approved):

PR #3701, opened 2021-11-16
Related Events Block Customizer Integration

EugeneKyale:
This PR ensures that the `related events title` and `event titles` within the single event page for the block editor make use the customizer font settings.

Ticket : https://theeventscalendar.atlassian.net/browse/TEC-4125
Artifact : https://www.loom.com/share/e290b16bac174e068689d6a124266ed5


paulmskim:
Does this fix need to apply for both the block editor reskin and non-reskinned? Or just the reskin?

EugeneKyale:
> Does this fix need to apply for both the block editor reskin and non-reskinned? Or just the reskin?

Only for the reskin. The issue is as a result of the related events block not being consistent with the rest of the blocks that went through the reskin earlier.

paulmskim (approved):

PR #3705, opened 2021-11-17
Prevent setting application on event manager page [ECP-948]

EugeneKyale:
This PR adds a check to ensure that the `Disable the Event Search Bar` setting doesn't get applied on the Event Manager page.

Ticket : https://theeventscalendar.atlassian.net/browse/ECP-948
Artifact : https://www.loom.com/share/84431c66ce51409882ecb876d7235a53

Related to : https://github.com/the-events-calendar/events-pro/pull/1857


src/Tribe/Views/V2/View.php
```diff
@@ -1498,7 +1499,7 @@ static function () {
 			'before_events'        => tribe( Advanced_Display::class )->get_before_events_html( $this ),
 			'after_events'         => tribe( Advanced_Display::class )->get_after_events_html( $this ),
 			'display_events_bar'   => $this->filter_display_events_bar( $this->display_events_bar ),
-			'disable_event_search' => tribe_is_truthy( tribe_get_option( 'tribeDisableTribeBar', false ) ),
+			'disable_event_search' => Admin_Helpers::instance()->is_screen( 'tribe-admin-manager' ) ? false : tribe_is_truthy( tribe_get_option( 'tribeDisableTribeBar', false ) ),
```
bordoni, line 1502: Same exact problem as in the other PR, but in this case we already have the filters so that you can modify just for the manager instead of adding the overhead here.

bordoni (changes requested):

src/Tribe/Views/V2/View.php
EugeneKyale: @bordoni Question regarding this approach: Since the Events Manager is a ECP specific feature ( and that's where the filter lives ) what's the best way to include it in TEC?

As things stand, it works as expected when both TEC and ECP are active but immediately I deactivate ECP things stop working.

src/Tribe/Views/V2/View.php
```diff
@@ -1498,7 +1498,7 @@ static function () {
 			'before_events'        => tribe( Advanced_Display::class )->get_before_events_html( $this ),
 			'after_events'         => tribe( Advanced_Display::class )->get_after_events_html( $this ),
 			'display_events_bar'   => $this->filter_display_events_bar( $this->display_events_bar ),
-			'disable_event_search' => tribe_is_truthy( tribe_get_option( 'tribeDisableTribeBar', false ) ),
+			'disable_event_search' => tribe_is_truthy( tribe_get_option( 'tribeDisableTribeBar', false ) && apply_filters( 'tribe_apply_settings_on_event_manager_page', false ) ),
```
bordoni, line 1501: We should use the same pattern I suggested here:
https://github.com/the-events-calendar/events-pro/pull/1857/files#r759356356

bordoni (changes requested):

bordoni (approved):

PR #3709, opened 2021-11-24
Wrong rel value causing HTML 5 and WCAG 2.1 compliance error [TEC-4129]

EugeneKyale:
This PR adds a valid rel value to the link tag for TEC REST API support in order to improve HTML 5 and WCAG 2.1 compliance.

Ticket : https://theeventscalendar.atlassian.net/browse/TEC-4129
Artifacts : 
Before Fix;
image
After Fix; 
image

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -223,6 +223,7 @@ Remember to always make a backup of your database and files before updating!
 
 = TBD =
 
+* Fix - Add a valid rel value to the link tag for TEC REST API support in order to improve HTML 5 and WCAG 2.1 compliance. [TEC-4129]
 * Tweak - Ensure the `related events title` and `event titles` within the single event page for the block editor make use the customizer font settings. [TEC-4125]
 
 = [5.11.0] 2021-11-17 =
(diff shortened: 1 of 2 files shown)


EugeneKyale:
> This PR adds a valid rel value to the link tag for TEC REST API support in order to improve HTML 5 and WCAG 2.1 compliance.
> 
> Ticket : https://theeventscalendar.atlassian.net/browse/TEC-4129 Artifacts : Before Fix; image After Fix; image

@lucatume as a precaution please take a look at this modification to determine whether there may be any unwanted trickle down effects. You seem to have added that URL intentionally so I don't want to miss anything.

jesseeproductions (approved):
The change looks good and tests are passing. I am approving, but good to hear if Luca has any comments before merging.

EugeneKyale:
> The change looks good and tests are passing. I am approving, but good to hear if Luca has any comments before merging.

Great, thanks @jesseeproductions I'll wait to hear from Luca before merging.

lucatume (approved):
Reading a bit the code and the [`rel` attribute definition](https://www.w3schools.com/tags/att_link_rel.asp), I'd say the change is correct. My suspect is the original value was just a forgotten placeholder.

PR #3734, opened 2021-12-16
Ensure the modifications made through the `tribe_events_views_v2_category_title` are applied. [TEC-4110]

EugeneKyale:
This PR ensures that users are able to use the `tribe_events_views_v2_category_title` filter to modify the page title on category archive pages.

Ticket : https://theeventscalendar.atlassian.net/browse/TEC-4110
Artifact : https://www.loom.com/share/b62845c18ef243a7bad65411342a207f


src/Tribe/Views/V2/Template/Title.php
```diff
@@ -140,7 +140,7 @@ public function build_title( $current_title = '', $depth = true ) {
 		}
 
 		$term = $context->get( TEC::TAXONOMY, false );
-		if ( false !== $term && $depth ) {
+		if ( false !== $term && $depth = true ) {
```
Camwyn, line 143:
```suggestion
		if ( false !== $term && $depth === true ) {
```

Camwyn: You are setting it here rather than checking it!

src/Tribe/Views/V2/Template/Title.php
```diff
@@ -140,7 +140,7 @@ public function build_title( $current_title = '', $depth = true ) {
 		}
 
 		$term = $context->get( TEC::TAXONOMY, false );
-		if ( false !== $term && $depth ) {
+		if ( false !== $term && true !== $depth ) {
```
Camwyn, line 143: I don't think this is the right solution either.

EugeneKyale: @Camwyn Any suggestions?

src/Tribe/Views/V2/Template/Title.php
```diff
@@ -366,7 +366,7 @@ protected function build_day_title( $event_date ) {
 	 * @return string The built category archive title.
 	 */
 	protected function build_category_title( $title, $cat ) {
-		$new_title = '<a href="' . esc_url( tribe_get_events_link() ) . '">' . $title . '</a>';
```
bordoni, line 369: I think the category title still needs the link, we might need to figure out how to exclude it in the specific places we need it to be removed.

bordoni (changes requested):

src/Tribe/Views/V2/Template/Title.php
EugeneKyale: @bordoni I don't think there's a use case for having a linked category on the page title. I can't find anywhere else where this title is being used. 
Screenshot 2022-01-04 at 22 14 49

bordoni: Ahh i see that is good to know. You are correct here, we need revise the depth problem that stephen brought up.

bordoni (approved):

bordoni (approved):

bordoni (approved):

PR #3739, opened 2021-12-23
Load the correct mobile default view when site home is set to the calendar.

Camwyn:
This corrects an issue where the mobile default view is not respected on the home page.
Due to some older code setting the default view in the query, "default" gets overwritten with a non-mobile view.

This also moves all mobile-view handling out to ECP as it is not supported by TEC alone.

Required by: https://github.com/the-events-calendar/events-pro/pull/1870

🎥 https://d.pr/v/UZoo2Z
🎥 (shortcode): https://d.pr/v/4DpFbZ

[TEC-3862]

[TEC-3862]: https://theeventscalendar.atlassian.net/browse/TEC-3862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Main.php
```diff
@@ -2604,13 +2628,6 @@ public function default_view() {
...
 			}
 
-			// If the stored option is no longer available, pick the first available one instead
-			$first_view = array_shift( $available_views );
-			$view       = $first_view['displaying'];
-
-			// Update the saved option
-			Tribe__Settings_Manager::instance()->set_option( 'viewOption', $view );
-
 			return $view;
```
Camwyn, line 2631: @Camwyn there is a bug in the logic here... 🤦🏼

bordoni (approved):

PR #3751, opened 2022-01-24
Removing Legacy v1 view files from Beta 2.1

borkweb:
This removes the v1 files from 6.0beta2.1.


lucatume:
Dev tested and merged into `feature/rbe-beta-2.1`.

PR #3752, opened 2022-01-24
Add a Legacy Views provider and force v2 to be active

borkweb:
The core bits of this PR are:
* Forcing views, widgets, and single to be v2 versions.
* This adds some cached notification of v1 files existing within themes.

That's pretty much it, other than removing a superfluous check on v2 being enabled in the Custom_Tables provider.


lucatume (approved):

lucatume:
Merged into `feature/rbe-beta-2.1`

PR #3753, opened 2022-01-25
Filter Bar Upsell Tab [TEC-4238]

EugeneKyale:
Adds a Filter Bar upsell banner to the Filters tab that displays when the Filter Bar plugin isn't active. 

Ticket : https://theeventscalendar.atlassian.net/browse/TEC-4238
Artifact : https://www.loom.com/share/72494dfd1d62419ba3e6ff08b7fff62d


src/Tribe/Assets.php
```diff
@@ -393,6 +393,20 @@ public function register() {
 				],
 			]
 		);
+
```
jesseeproductions, line 396: Should we load the asset like we do for the upgrade tab only if the upsell is active?

https://github.com/the-events-calendar/the-events-calendar/blob/b889b8762d5fd2f38885326653ceebf0c897a5eb/src/Tribe/Main.php#L1311

EugeneKyale: Ooh yeah, it's better to load them only when the upsell is active. I've made the changes.

src/Tribe/Main.php
```diff
@@ -6133,5 +6134,58 @@ public function register_plugin_autoload_paths( ) {
...
+			tribe_assets(
+				self::instance(),
+				[
+					[
+						'tec-admin-filterbar-upsell',
+						'tec-admin-filterbar-upsell.css',
+					],
+				],
+				'admin_enqueue_scripts',
+			);
```
bordoni, line 6158: This is just one asset use `tribe_asset`

src/Tribe/Main.php
```diff
@@ -6133,5 +6134,58 @@ public function register_plugin_autoload_paths( ) {
...
+
+			tribe_assets(
+				self::instance(),
+				[
+					[
+						'tec-admin-filterbar-upsell',
+						'tec-admin-filterbar-upsell.css',
+					],
+				],
+				'admin_enqueue_scripts',
```
bordoni, line 6157: We need to make sure this only enqueues on the correct page, so use `conditional` argument.

src/Tribe/Main.php
```diff
@@ -6133,5 +6134,58 @@ public function register_plugin_autoload_paths( ) {
...
+			];
+			
+			/**
+			* Allows the fields displayed in the Filter Bar upsell tab to be modified.
+			*
+			* @since TBD
+			*
+			* @param array $fbar_upsell_tab Array of fields used to setup the Filter Bar upsell Tab.
+			*/
+			$fbar_upsell_fields = apply_filters( 'tribe_upgrade_fields', $fbar_upsell_tab );
```
bordoni, line 6178: We need to namespace this correctly, `tec_events_` and have some more naming related to filterbar.

src/Tribe/Main.php
```diff
@@ -6133,5 +6134,58 @@ public function register_plugin_autoload_paths( ) {
...
+						'tec-admin-filterbar-upsell',
+						'tec-admin-filterbar-upsell.css',
+					],
+				],
+				'admin_enqueue_scripts',
+			);
+
+			ob_start();
+			include_once $this->plugin_path . 'src/admin-views/filterbar/banners/filterbar-upsell.php';
+			$fbar_upsell_tab_html = ob_get_clean();
```
bordoni, line 6162: This needs o use a template engine.

bordoni (changes requested):
We cannot add these to the Main file at all.
Create a new Provider in the Admin folder to add this from.

src/Tribe/Main.php
EugeneKyale: updated and moved to a new provider.

src/Tribe/Main.php
EugeneKyale: conditional added

src/Tribe/Main.php
EugeneKyale: namespace updated

src/Tribe/Main.php
EugeneKyale: this has been updated

EugeneKyale:
I've created a ticket to resolve the failing tests on the release branch. 
REF: https://theeventscalendar.atlassian.net/browse/BTPLAN-172

src/Tribe/Main.php
Camwyn: @EugeneKyale it wasn't the $variable that needed namespacing - it was the hook `tribe_upgrade_fields` -
which is also already in use [elsewhere](https://github.com/the-events-calendar/the-events-calendar/blob/master/src/Tribe/Main.php#L1361), so someone using it might break something unrelated.

Maybe `tec_events_filterbar_upgrade_content` would be more appropriate. Also, the use of the word "fields" here is misleading - there are no fields, just HTML content.

Camwyn (changes requested):
namespace the hook correctly. For namespacing rules, see https://the-events-calendar.github.io/products-engineering/docs/plugins/namespacing/

Also please change "fields" to something more accurate (perhaps "content")

src/Tribe/Main.php
EugeneKyale: Thanks @Camwyn I've namespaced the hook correctly.

Camwyn (approved):

EugeneKyale:
> We cannot add these to the Main file at all. Create a new Provider in the Admin folder to add this from.

This has been addressed.

src/Tribe/Admin/Filter_Bar/Provider.php
```diff
@@ -0,0 +1,143 @@
+<?php
+namespace Tribe\Events\Admin\Filter_Bar;
+
+use Tribe__Events__Main;
+use Tribe__Settings_Tab;
+use Tribe__Admin__Helpers;
+
+// Bail if Filter Bar is already installed.
+if ( ! class_exists( 'Tribe__Events__Filterbar__View' ) ) {
```
bordoni, line 9: This class exists will cause fatale, we should have this conditional on the register method, so that if this class doesn’t exist it doesn’t continue to do the rest

bordoni (changes requested):

src/Tribe/Admin/Filter_Bar/Provider.php
EugeneKyale: @bordoni I've moved the conditional to the register method.

src/Tribe/Admin/Filter_Bar/Provider.php
```diff
@@ -0,0 +1,144 @@
...
+class Provider extends \tad_DI52_ServiceProvider {
+
+	/**
+	 * Binds and sets up implementations.
+	 *
+	 * @since TBD
+	 */
+	public function register() {
+		// Bail if Filter Bar is already installed.
+		if ( ! class_exists( 'Tribe__Events__Filterbar__View' ) ) {
```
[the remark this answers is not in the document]
EugeneKyale: Done.

bordoni (approved):

PR #3762, opened 2022-02-06
Improve accessibility based on @mrwweb suggestions

bordoni:
On [this Gist](https://gist.github.com/mrwweb/6f5ba48873d9bd24dfc23eed7ad934df) @mrwweb  makes some suggestions on Accessibility that we can add to this release.

Not all suggestions are included in this PR as some of them are decisions we made in the past because they are better suited for the event structure. Example, Date/Time appearing before the Title of the Event on List view.

### DIFF
--- a/src/views/v2/latest-past/event/date-tag.php
+++ b/src/views/v2/latest-past/event/date-tag.php
@@ -24,7 +24,7 @@ $event_year      = $event->dates->start_display->format_i18n( 'Y' );
 $event_date_attr = $event->dates->start_display->format( Dates::DBDATEFORMAT );
 ?>
 <div class="tribe-events-calendar-latest-past__event-date-tag tribe-common-g-col">
-	<time class="tribe-events-calendar-latest-past__event-date-tag-datetime" datetime="<?php echo esc_attr( $event_date_attr ); ?>">
+	<time class="tribe-events-calendar-latest-past__event-date-tag-datetime" datetime="<?php echo esc_attr( $event_date_attr ); ?>" aria-hidden="true">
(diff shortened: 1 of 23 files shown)


borkweb (approved):
:+1:

EugeneKyale (approved):

Camwyn (approved):

mrwweb:
Hey @bordoni. These changes are super exciting! I'm so glad they're headed into the main plugin and I won't have to worry about making these in the future!

---

Looking at this PR, I think `aria-hidden` was missing and should be added in [src/views/v2/list/event/featured-image.php](https://github.com/the-events-calendar/the-events-calendar/pull/3762/files#diff-d54d76faa47c3fc2d18b8e57bdc62f36b8d9e8c737fb53ec87691e4e57e8f71b). I'm a little short on time and in a rush right now, but take a look at that and make sure it matches the other changes.

To make sure you understand the change, this [article from Sara Soueidan is excellent](https://www.sarasoueidan.com/blog/keyboard-friendlier-article-listings/) and where I got the pattern from.

---

> decisions we made in the past because they are better suited for the event structure. Example, Date/Time appearing before the Title of the Event on List view.

I just want to call out that the visual display is fine this way, but the source order is inaccessible. There are ways to fix the source order without impacting the visual appearance, so I hope you'll look into that in the future.

As it stands, the heading structure on the Event list implies that the date/time of the event belongs to the previous event.

---

And a final note: I mostly use the List view so that's my focus. I suspect that there are similar changes that could be made to nearly all of your views. I hope that my explanations of _why_ these changes are important were clear and can be applied more broadly across the plugin.

PR #3760, opened 2022-02-03
Removal of Legacy Views on 6.0.0 beta 2

bordoni:
## [Documentation on what will be removed](https://github.com/the-events-calendar/the-events-calendar/blob/feature/remove-legacy-views-beta-2.2/legacy-views.md)

PR #3788, opened 2022-02-23
Further tweaks to the massive legacy views removal PR

borkweb:
These are additional tweaks / adjustments to the deprecation PR #3760


src/functions/template-tags/general.php
```diff
@@ -1307,6 +1300,26 @@ function tribe_events_tab_index() {
...
+	 *
+	 * @return bool
+	 **/
+	function tribe_events_is_view_enabled( $view ) {
+		$manager = tribe( \Tribe\Events\Views\V2\Manager::class );
+		$public_views = $manager->get_publicly_visible_views();
+
+		$enabled = isset( $public_views[ $view ] );
+
+		return apply_filters( 'tribe_events_is_view_enabled', $enabled, $view, array_keys( $public_views ) );
```
bordoni, line 1319: Docblock

src/functions/template-tags/day.php
```diff
@@ -1,51 +1,38 @@
...
- * Get the previous day's date
- *
- * @category Events
- *
- * @return string
- */
-function tribe_get_previous_day_date( ) {
-	return null;
-}
+		return apply_filters( 'tribe_get_day_link', $tribe_ecp->getLink( 'day', $date ), $date );
```
bordoni, line 36: This is the V1  way of getting a link `tribe_events_get_url`

bordoni (changes requested):
Two small changes. I mentioned that particular doc block because it felt like one that deserves explanation, but feel free to ignore.

src/functions/template-tags/day.php
borkweb: Update it to the v2 way and push those changes.

PR #3789, opened 2022-02-25
Fix broken styling in Elementor preview mode. [ECP-1065]

EugeneKyale:
Load the assets required by the `Event View` Elementor widget in preview mode.

Ticket : https://theeventscalendar.atlassian.net/browse/ECP-1065

https://www.loom.com/share/9ce90a65a5604d63aa28dcf209a3ed9a

### DIFF
--- a/src/Tribe/Views/V2/Assets.php
+++ b/src/Tribe/Views/V2/Assets.php
@@ -462,7 +462,7 @@ class Assets extends \tad_DI52_ServiceProvider {
 				'tec-variables-full',
 				'tec-variables-skeleton',
 			],
-			'enqueue_block_editor_assets',
+			'enqueue_block_assets',
 			[
 				'priority'     => 15,
 				'groups'       => [ static::$single_group_key ],
--- a/src/Tribe/Views/V2/Assets.php
+++ b/src/Tribe/Views/V2/Assets.php
@@ -528,6 +528,20 @@ class Assets extends \tad_DI52_ServiceProvider {
 			return $this->should_enqueue_frontend;
 		}
 
+		// Detect plugin. For frontend only
+		include_once ABSPATH . 'wp-admin/includes/plugin.php';
+
+		/**
+		 * Checks whether the page is being viewed in Elementor preview mode.
+		 * 
+		 * @since TBD
+		 * 
+		 * @return bool $should_enqueue Should the frontend assets be enqueued.
+		 */
+		if ( is_plugin_active( 'elementor/elementor.php' ) && \Elementor\Plugin::$instance->preview->is_preview_mode() ) {
+			return $this->should_enqueue = true;
+		}
+
 		$should_enqueue = tribe( Template_Bootstrap::class )->should_load();
 
 		/**
--- a/src/Tribe/Views/V2/Assets.php
+++ b/src/Tribe/Views/V2/Assets.php
@@ -528,9 +528,6 @@ class Assets extends \tad_DI52_ServiceProvider {
 			return $this->should_enqueue_frontend;
 		}
 
-		// Detect plugin. For frontend only
-		include_once ABSPATH . 'wp-admin/includes/plugin.php';
-
 		/**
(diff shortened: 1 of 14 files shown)


src/Tribe/Views/V2/Assets.php
```diff
@@ -528,6 +528,20 @@ public function should_enqueue_frontend() {
...
+		/**
+		 * Checks whether the page is being viewed in Elementor preview mode.
+		 * 
+		 * @since TBD
+		 * 
+		 * @return bool $should_enqueue Should the frontend assets be enqueued.
+		 */
+		if ( is_plugin_active( 'elementor/elementor.php' ) && \Elementor\Plugin::$instance->preview->is_preview_mode() ) {
+			return $this->should_enqueue = true;
+		}
```
bordoni, line 543: We should use the same conditional as the code from the integrations. To determine if elementor is active.
```
		if ( ! defined( 'ELEMENTOR_PATH' ) || empty( ELEMENTOR_PATH ) ) {
			return;
		}
```

bordoni (changes requested):

src/Tribe/Views/V2/Assets.php
EugeneKyale: @bordoni I've updated the conditional

src/Tribe/Views/V2/Assets.php
```diff
@@ -528,6 +528,17 @@ public function should_enqueue_frontend() {
...
 		}
 
+		/**
+		 * Checks whether the page is being viewed in Elementor preview mode.
+		 * 
+		 * @since TBD
+		 * 
+		 * @return bool $should_enqueue Should the frontend assets be enqueued.
+		 */
+		if ( ! ( ! defined( 'ELEMENTOR_PATH' ) || empty( ELEMENTOR_PATH ) ) && \Elementor\Plugin::$instance->preview->is_preview_mode() ) {
```
Camwyn, line 538: Are we sure that we should _always_ load this in The Elementor preview? This bypasses **all** the checks in Template_Bootstrap

Camwyn: If so - the logic could be simplified a lot.
```suggestion
		if (
			defined( 'ELEMENTOR_PATH' )
			&& ! empty( ELEMENTOR_PATH )
			&& \Elementor\Plugin::$instance->preview->is_preview_mode()
        ) {
```

Camwyn: ...and the call to `is_preview_mode` can probably be boiled down to (for our purposes)

`if ( isset( $_GET[ 'elementor-preview' ] ) && $_GET[ 'elementor-preview' ] === get_the_ID() )`

removing the requirement for calling a third-party class?

Sorry, to be clearer, try:
```
if (
	defined( 'ELEMENTOR_PATH' )
	&& ! empty( ELEMENTOR_PATH )
	&& isset( $_GET[ 'elementor-preview' ] )
	&& $_GET[ 'elementor-preview' ] === get_the_ID()
) {
```

Camwyn (changes requested):

src/Tribe/Views/V2/Assets.php
EugeneKyale: @Camwyn we need to load it on the elementor preview when the event view widget is added to the page.

EugeneKyale: I've updated the conditional but haven't added the last one i.e. `$_GET[ 'elementor-preview' ] === get_the_ID()`. `isset( $_GET[ 'elementor-preview' ] )` works as expected.

Camwyn (approved):
🥇

src/Tribe/Views/V2/Assets.php
Camwyn: Yeah, the `$_GET[ 'elementor-preview' ] === get_the_ID()` part was pulled directly from `is_preview_mode()` this is really nice and simple now 😄

PR #3795, opened 2022-03-07
Updating the tests to match the new ical subscribe logic for month view

borkweb:
This is a spotfix to adjust tests to ensure that the date of the text excution is taken into account when generating the expected values to test against when looking at the current month.

Related: https://github.com/the-events-calendar/the-events-calendar/pull/3794


EugeneKyale (approved):

Camwyn (approved):
Looks great :)

PR #3796, opened 2022-03-08
Correct useage of currency code

Camwyn:
[TEC-4212]

This corrects our usage of currency code i.e. `priceCurrency` in our JSON-LD by adding a field for a default value to the admin and an event-specific one to both the classic and block editors.

Previously we were incorrectly using the currency _symbol_ but the ISO-4217 three-letter code is what's expected (as $ can be USD, AUD, CAD, Mexican Peso, etc...)

**Blocks:**
https://d.pr/i/Dzgjxk
https://d.pr/i/MqzhXU

**Classic**
https://d.pr/i/10IDFO
https://d.pr/i/T3Bw6l
(output in validator)
https://d.pr/i/8Eg8H3


EugeneKyale (commented):
The Blocks portion looks good.

EugeneKyale (approved):

lucatume (dismissed):
All fine and dandy, thanks!

PR #3799, opened 2022-03-14
Correct the calculation on the month ics dates so that it doesn't rely on "now"

Camwyn:
… but uses actual dates.

This was causing today "2022-03-11" to be compared to "now".

[TEC-4223]

[TEC-4223]: https://theeventscalendar.atlassian.net/browse/TEC-4223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


borkweb (approved):
The logic looks reasonable to me!

PR #3807, opened 2022-04-03
Fix a bug in url's in non-english based website

shisho585:
In non-english based sites, when slugs are translated to a language which is encoded with '%' inside, the rewrite rule adds, in addition to the translated string, another translated string without the '%' char, which is being removed by 'get_i18n_strings_for_domains' function that uses wordpress's 'sanitize_key' function.

For example, on an Hebrew based site, month slug will be saved in rewrite rules as '(?:month|d797d795d793d7a9|%d7%97%d7%95%d7%93%d7%a9)' instead of '?:month|%d7%97%d7%95%d7%93%d7%a9)'.

This causes the url for hebrew version of the site to be ugly ('d797d795d793d7a9' which is meaningless, instead of 'חודש' which is the actual translation), and also causes problems with links on the english version of the site, because it can't solve the rewrite rule correctly.

This PR solves the first problem for sure, and somehow also fix the second problem on my site.

It can be written nicer but this way it does'nt break any compatibily.


shisho585:
Hi @Camwyn,
Can you or someone else please review my PR?
Each update of the plugin cause the same errors in my website again until i fix it by myself.
Thank's

Camwyn:
@shisho585 - yes I'll take a look and try to get this updated as soon as I reasonably can - thanks for the submission! (and the mention - things get missed, sorry)

Camwyn:
@shisho585 I have updated this PR to an upcoming release branch - ETA early October. I'm going to add some of our required stuff (like readme entries) as well.

Note that using this branch in production is _not recommended_ until it releases as some of the release branch code is still in flux. (also it includes significant changes to our event data structure!)

Camwyn:
@shisho585: Actually, it's likely simpler I create a new PR but you can watch it here: https://github.com/the-events-calendar/the-events-calendar/pull/3967

PR #3812, opened 2022-04-11
Ensure that we get the event description (or post content) if it exists.

Camwyn:
Two things here:

1. The logic was inverted (so it was trying to get it if it was empty)
2. It's not always stored on the event as event->description, so if that's empty we try to use event->post_content

[TEC-4315]

[TEC-4315]: https://theeventscalendar.atlassian.net/browse/TEC-4315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


readme.txt
```diff
@@ -223,6 +223,10 @@ Remember to always make a backup of your database and files before updating!
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Ensure Google Calendar links include the event description if it exists. [TEC-4315]
```
jesseeproductions, line 228: Looks good, once the readme conflict is fixed it should be all set.

jesseeproductions (approved):

src/Tribe/Views/V2/iCalendar/Links/Google_Calendar.php
```diff
@@ -115,7 +115,8 @@ public function generate_single_url( $post = null ) {
 			return '';
 		}
 
-		$event_details = empty( $event->description ) ? urlencode( $event->description ) : '';
+		$event_details = ! empty( $event->description ) ? $event->description : $event->post_content;
```
lucatume, line 118: Question: do we care that the event might have an empty `post_content`?

lucatume (approved):
Just one question.

src/Tribe/Views/V2/iCalendar/Links/Google_Calendar.php
Camwyn: Good point, we should cover that possibility thoroughly. I'll revise.

jesseeproductions (approved):

PR #3738, opened 2021-12-22
Events/Tickets Menus > Events menu

juanfra:
### 🎫 Ticket

[ET-1335] 

### 🗒️ Description

- Implementation of top of https://github.com/the-events-calendar/tribe-common/pull/1691

### 🎥 Artifacts 

- Coming soon.

### ✔️ Checklist
- [ ] I've included a changelog entry in the readme.txt file. 
- [ ] My code is tested. 
- [ ] My code has [proper inline documentation](https://the-events-calendar.github.io/products-engineering/docs/code-standards/).

[ET-1335]: https://theeventscalendar.atlassian.net/browse/ET-1335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3813, opened 2022-04-13
Further adjustments to the admin pages structure for TEC

borkweb:
This is an iteration on top of the work from #3738

:movie_camera: https://d.pr/v/WZkMyM

### DIFF
--- a/src/Tribe/Admin/Event_Settings.php
+++ b/src/Tribe/Admin/Event_Settings.php
@@ -1,13 +1,16 @@
 <?php
+namespace Tribe\Events\Admin;
 
 /**
  * Manages the admin settings UI in relation to events configuration.
  */
+use Tribe__App_Shop;
+use Tribe__Settings;
+use Tribe__Settings_Tab;
 use Tribe__Events__Main as Plugin;
 use Tribe\Admin\Troubleshooting as Troubleshooting;
 
-
-class Tribe__Events__Admin__Event_Settings {
+class Event_Settings {
 
 	/**
 	 * The Events Calendar settings page slug.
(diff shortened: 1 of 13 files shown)


src/Tribe/Admin/Event_Settings.php
```diff
@@ -1,13 +1,16 @@
...
  */
+use Tribe__App_Shop;
+use Tribe__Settings;
+use Tribe__Settings_Tab;
 use Tribe__Events__Main as Plugin;
 use Tribe\Admin\Troubleshooting as Troubleshooting;
 
-
-class Tribe__Events__Admin__Event_Settings {
+class Event_Settings {
```
juanfra, line 13: I think we're trying to use the plural for wherever we use Events or Tickets, right? Maybe setting this one as `Events_Settings` will make more sense.

juanfra (commented):
This is awesome stuff, we should probably do the same for tickets. I have one only comment/question about the naming of the class.

src/Tribe/Admin/Event_Settings.php
```diff
@@ -369,11 +359,13 @@ public function settings_ui( $admin_page ) {
 			return;
 		}
 
-		include_once Tribe__Main::instance()->plugin_path . 'src/admin-views/tribe-options-general.php';
-		include_once Tribe__Main::instance()->plugin_path . 'src/admin-views/tribe-options-display.php';
+		include_once \Tribe__Main::instance()->plugin_path . 'src/admin-views/tribe-options-general.php';
+		include_once \Tribe__Main::instance()->plugin_path . 'src/admin-views/tribe-options-display.php';
+		include_once tribe( 'tec.main' )->plugin_path . 'src/admin-views/tribe-options-general.php';
+		include_once tribe( 'tec.main' )->plugin_path . 'src/admin-views/tribe-options-display.php';
```
juanfra, line 365: Are we including these twice?

src/Tribe/Admin/Provider.php
```diff
@@ -0,0 +1,30 @@
+<?php
+namespace Tribe\Events\Admin;
+
+class Provider extends \tad_DI52_ServiceProvider {
+	/**
+	 * Register implementations.
+	 */
```
juanfra, line 7:
```suggestion
	/**
	 * Register implementations.
	 *
	 * @since TBD
	 */
```

src/Tribe/Admin/Provider.php
```diff
@@ -0,0 +1,30 @@
...
+	public function register() {
+		tribe_singleton( Event_Settings::class, Event_Settings::class );
+
+		$this->add_hooks();
+	}
+
+	/**
+	 * Add hooks.
+	 */
```
juanfra, line 16:
```suggestion
	/**
	 * Add hooks.
	 *
	 * @since TBD
	 */
```

src/Tribe/Admin/Event_Settings.php
borkweb: No, one is coming from common, the other from TEC

src/Tribe/Admin/Event_Settings.php
borkweb: _Event_, in this case, acts as an adjective with _Settings_ being the noun that should be pluralized.

juanfra: Ok, gotcha. Although, with that naming, as a dev that comes at this for the first time I'd interpret that it's related to settings for a particular event rather than the "events" settings. Just a thought.

borkweb: Interesting perspective. I can't bring myself to go with `Events_Settings` because it isn't an English convention and my brain breaks. That being said, I get your point, so I propose `Tribe/Admin/Events/Settings.php` since the directories are intended to be nouns as well.

juanfra (approved):
👏

PR #3814, opened 2022-04-13
Support Currency Code Field in CSV Imports

jesseeproductions:
Adds support the the currency code field in csv imports.

_Ref:_ [TEC-4341](https://theeventscalendar.atlassian.net/browse/TEC-4341)

Artifact
https://cloudup.com/cZtaHEbJeXv


lucatume (commented):
All fine and dandy, but I have one question: what if the currency code is invalid?
Say `FOO`?

jesseeproductions:
> All fine and dandy, but I have one question: what if the currency code is invalid? Say `FOO`?

@lucatume There appears to be no validation on the field in the admin either. If I type in "egewghewhe" then that is what will display as my currency code.

Camwyn (approved):

PR #3816, opened 2022-04-13
Ensure that max returns an int since we're doing math with it.

Camwyn:
[TEC-4322]

If fed a string like "hello" and it equates to a larger number than the others, `max()` will return the string, not the integer representation of it. `"hello" - 1` will fail (and fatal)

[TEC-4322]: https://theeventscalendar.atlassian.net/browse/TEC-4322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -1224,6 +1224,7 @@ class View implements View_Interface {
 	 * This allows us to save a query when determining pagination for list-like views.
 	 *
 	 * @since 5.0.0
+	 * @since TBD ZEnsure our max() gets all ints for math reasons
 	 *
 	 * @param null|int $offset_override Offset override value.
 	 * @param \WP_Query $query WP Query object.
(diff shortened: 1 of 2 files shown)


src/Tribe/Views/V2/View.php
```diff
@@ -1224,6 +1224,7 @@ protected function setup_repository_args( Context $context = null ) {
 	 * This allows us to save a query when determining pagination for list-like views.
 	 *
 	 * @since 5.0.0
+	 * @since TBD ZEnsure our max() gets all ints for math reasons
```
jesseeproductions, line 1227: Typo in Ensure.

Camwyn: doh!

lucatume (approved):

PR #3820, opened 2022-04-18
CSV Import Fix on PHP 8.0

jesseeproductions:
n PHP 8.0.15 they added this patch: [Fixed bug #75917 ](https://bugs.php.net/bug.php?id=75917)(SplFileObject::seek broken with CSV flags).

That appears to be the source of the issue. 
Before that patch this coding would return the last line of a csv file:
$this->file->seek( $this->file->getSize() );
$this->lines = $this->file->key();

It is similar to an example of getting the last line for [PHP: Hypertext Preprocessor](http://php.net/) 
https://www.php.net/manual/en/splfileobject.seek.php#121666

In PHP 8.0.18 this patch was added: [Fixed bug #8121](https://github.com/php/php-src/pull/8138) (SplFileObject - seek and key with csv file inconsistent).

_Ref:_ [TEC-4312](https://theeventscalendar.atlassian.net/browse/TEC-4312)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -231,6 +231,7 @@ Remember to always make a backup of your database and files before updating!
 * Fix - Correctly calculate Event duration when the Event crosses the daylight saving date and time. [TEC-4336]
 * Fix - don't try to validate "raw" as a date. [TEC-4254]
 * Fix - Ensure the Views don't try to do math with strings. [TEC-4322]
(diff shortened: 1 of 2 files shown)


Camwyn (dismissed):
Looks legit although I'd love to see a link to the issue documented somewhere (I'd _personally_ love some context)

Camwyn (dismissed):

Camwyn (approved):

PR #3826, opened 2022-04-27
Fix the venue link/url in the event venue form

Camwyn:
`tribe_get_venue_website_link` returns a formatted HTML anchor
we need a URL - we get that from `tribe_get_venue_link` with the second param set to false.

I also indexed the params for `tribe_get_organizer_website_link` and `tribe_get_venue_website_link` while testing and since we're supposed to always used indexed params, I left that in place.


jesseeproductions (approved):
Approved.

PR #3828, opened 2022-04-28
Fix/tec 4314 v2 tag archives

jesseeproductions:


src/Tribe/Taxonomy/Event_Tag.php
```diff
@@ -0,0 +1,48 @@
...
+			return $actions;
+		}
+
+		$link = tribe_events_get_url( [ 'tag' => $tag->slug, 'post_type' => 'tribe_events', 'eventDisplay' => 'default' ] );
+		if ( is_wp_error( $link ) ) {
+			return $actions;
+		}
+
+		//todo translate
+		$actions['event-view'] = '<a href="' . esc_url( $link ) . '" rel="tag">Event View</a>';
```
bordoni, line 44: Should we do the translation here before mergin?

bordoni (changes requested):

src/Tribe/Taxonomy/Event_Tag.php
jesseeproductions: Yeah... let me follow my todo and add that now. Thanks

PR #3831, opened 2022-05-02
Put back the event website label for Classic editor.

Camwyn:
This was intentionally removed as part  of the single event redesign. Yet its removal was not in any of the design files I could find.

📸 https://d.pr/i/wVL5Ma

[TEC_4334]


stratease (approved):
Looks good!

lucatume (changes requested):
The code changes are fine, but the snapshot tests will need to be re-generated.
Remember to `composer install --no-dev` in Common before doing that.

lucatume (approved):
My brain got confused, all fine here.

PR #3833, opened 2022-05-03
This fixes the application of the TOD filter in FBAR.

Camwyn:
**No longer in strudel! This is getting punted to the next MR to implement Luca's suggestions. Don't merge until the base branch has been changed!**

The problem: Events created in the block editor are filtered out by any TOD (time of day) filter.
The cause: Events created in the blocks editor do not have the `_EventDuration` meta set.
The solution: hook into event save and the model `build_properties()` calculate the duration and save it.

As part of this fix, I also discovered that somewhere `Event->build_properties()` is being called on a venue.
So I put in some sanity checks to prevent that.

Original fix. At the end you'll see a debugged duration of 0 - which lead me to the sanity checks.
🎥  https://d.pr/v/L2LmxS

[FBAR-255]

[FBAR-255]: https://theeventscalendar.atlassian.net/browse/FBAR-255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bordoni:
Is this ready for Code Review?

src/Tribe/Editor/Hooks.php
```diff
@@ -27,6 +30,22 @@ public function register() {
 	 */
 	protected function add_actions() {
 		add_action( 'current_screen', [ $this, 'add_widget_resources' ] );
+		add_action( 'save_post_' . Tribe__Events__Main::POSTTYPE, [ $this, 'calculate_duration' ], 20 );
```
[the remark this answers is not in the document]
lucatume: A proposal to deal with that:
```php
if( empty(tribe_get_option( 'fix_duration')){
    global $wpdb;
    $query = "insert into wp_postmeta (post_id, meta_key, meta_value) select p.ID, '_EventDuration', TIME_TO_SEC(TIMEDIFF(end.meta_value, start.meta_value)) from wp_posts p join wp_postmeta start on start.post_id = p.ID and start.meta_key = '_EventStartDateUTC' join wp_postmeta end on end.post_id = p.ID and end.meta_key = '_EventEndDateUTC' left join wp_postmeta as duration on p.ID = duration.post_id and  duration.meta_key = '_EventDuration' where p.post_type = 'tribe_events' and duration.meta_value is null";
    $fixed = $wpdb->query($query);
    if($fixed) tribe_update_option('fix_duration', 1);
}
```

The query will only pick up events that have the correct pre-conditions. To reduce the scope of the query events with an `_EventDuration` meta are not touched.
Format and add all the usuals bells and whistles.
I've typed directly in the comment, formatting and code should be checked.

lucatume (changes requested):
Add tests to include the across-daylight-saving case. Add system to fix all the events, not just the new and updated ones.

src/Tribe/Editor/Hooks.php
```diff
@@ -27,6 +30,76 @@ public function register() {
...
+		if ( empty( $start_date_utc ) || empty( $end_date_utc ) ) {
+			return;
+		}
+
+		$utc_timezone          = new \DateTimezone( 'UTC' );
+		$start_date_utc_object = Dates::immutable( $start_date_utc, $utc_timezone );
+		$end_date_utc_object   = Dates::immutable( $end_date_utc, $utc_timezone );
+		$calculated_duration   = $end_date_utc_object->getTimestamp() - $start_date_utc_object->getTimestamp();
+
+		if ( empty( $saved_duration ) || (int) $saved_duration !== (int) $calculated_duration ) {
```
lucatume, line 65: This check should be done before, right after you fetch the `$saved_duration` value and avoid all the wasted computation that got you here.

lucatume: I mean the `empty( $saved_duration )` check.

src/Tribe/Editor/Hooks.php
```diff
@@ -27,6 +30,76 @@ public function register() {
...
+						and start.meta_key = %s join $wpdb->postmeta end
+						on end.post_id = p.ID and end.meta_key = %s
+						left join $wpdb->postmeta as duration
+						on p.ID = duration.post_id and  duration.meta_key = %s
+						where p.post_type = %s and duration.meta_value is null",
+						'_EventDuration',
+						'_EventStartDateUTC',
+						'_EventEndDateUTC',
+						'_EventDuration',
+						TEC::POSTTYPE
```
lucatume, line 95:
```suggestion
			$query = $wpdb->prepare( "INSERT INTO $wpdb->postmeta (post_id, meta_key, meta_value)
						SELECT p.ID, %s, TIME_TO_SEC(TIMEDIFF(end.meta_value, start.meta_value))
						from $wpdb->posts p
						JOIN $wpdb->postmeta start ON start.post_id = p.ID
						AND start.meta_key = %s JOIN $wpdb->postmeta end
						ON end.post_id = p.ID AND end.meta_key = %s
						LEFT JOIN $wpdb->postmeta duration
						ON p.ID = duration.post_id AND duration.meta_key = %s
						WHERE p.post_type = %s AND duration.meta_value IS NULL",
						'_EventDuration',
						'_EventStartDateUTC',
						'_EventEndDateUTC',
						'_EventDuration',
						TEC::POSTTYPE
```

src/Tribe/Editor/Hooks.php
```diff
@@ -27,6 +30,76 @@ public function register() {
...
+					);
+
+			$fixed = $wpdb->query( $query );
+
+			if( $fixed ) {
```
lucatume, line 100: The `wpdb::query` method will return the number of affected rows for a query of this type. No affected rows is a good result that would keep the fix looping; specialize the check:
```suggestion
			if( false !== $fixed ) {
```

lucatume (changes requested):
Update the check. Add an integration test for the fix function.

src/Tribe/Editor/Hooks.php
Camwyn: No.
I create an event. The duration is calculated.
I then go in and change the dates - the meta exists but the duration needs to be recalculated.

bordoni (dismissed):

src/Tribe/Models/Post_Types/Event.php
```diff
@@ -46,15 +52,13 @@ protected function build_properties( $filter ) {
...
-			$end_date_utc          = isset( $post_meta['_EventEndDateUTC'][0] ) ? $post_meta['_EventEndDateUTC'][0] : null;
-			$duration              = (int) isset( $post_meta['_EventDuration'][0] ) ? $post_meta['_EventDuration'][0] : null;
+			$start_date            = get_post_meta( $post_id, '_EventStartDate', true );
+			$start_date_utc        = get_post_meta( $post_id, '_EventStartDateUTC', true );
+			$end_date              = get_post_meta( $post_id, '_EventEndDate', true );
+			$end_date_utc          = get_post_meta( $post_id, '_EventEndDateUTC', true );
+			$duration              = get_post_meta( $post_id, '_EventDuration', true );
 			$timezone_string       = Timezones::get_event_timezone_string( $post_id );
-			$all_day               = tribe_is_truthy( isset( $post_meta['_EventAllDay'][0] ) ? $post_meta['_EventAllDay'][0] : null );
+			$all_day               = tribe_is_truthy( get_post_meta( $post_id, '_EventAllDay', true ) );
```
bordoni, line 61: Why did we change this?

bordoni (changes requested):

src/Tribe/Models/Post_Types/Event.php
Camwyn: https://github.com/the-events-calendar/the-events-calendar/pull/3833/files/470c7f19ca6ddc8caf94940ca7b807c306224f66#diff-8d783ab4e9c9e65f92a01f9ba840456d5b38559ca6c3650f42969c36d7be60e3

bordoni: @lucatume I would like us to revert to the old way here, I don't think it's worth the refactor here. Since the reason why we changed in the past was due to some problems with performance.

lucatume: After some Slack conversation, we agreed on going back to the `get_post_meta( $post_id )` version. A bit less readable, but the change was made for a performance gain.

Camwyn: Prior PR for ref: https://github.com/the-events-calendar/the-events-calendar/pull/2987

src/Tribe/Editor/Hooks.php
```diff
@@ -27,6 +30,22 @@ public function register() {
 	 */
 	protected function add_actions() {
 		add_action( 'current_screen', [ $this, 'add_widget_resources' ] );
+		add_action( 'save_post_' . Tribe__Events__Main::POSTTYPE, [ $this, 'calculate_duration' ], 20 );
+	}
+
+	public function calculate_duration( $post_id ) {
+		$post_meta             = get_post_meta( $post_id );
```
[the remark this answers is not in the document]
Camwyn: For the get_post_meta, see https://github.com/the-events-calendar/the-events-calendar/pull/3833#discussion_r874949189

tests/event_status/Tribe/Events/Event_Status/JSON_LDTest.php
```diff
@@ -73,6 +74,7 @@ public function it_should_have_event_status_for_canceled() {
 
 	/**
 	 * @test
+	 * @skip The meta tested is not being hydrated correctly by get_mock_event
```
lucatume, line 77: This should not be skipped anymore, right?

tests/event_status/Tribe/Events/Event_Status/JSON_LDTest.php
```diff
@@ -52,6 +52,7 @@ public function it_sets_offline_attendance_mode_default() {
...
```
lucatume, line 55: This should not be skipped anymore, right?

lucatume (changes requested):

tests/event_status/Tribe/Events/Event_Status/Views/__snapshots__/Single_ViewTest__it_should_render_single_canceled_name_no_reason_and_html__1.php
```diff
@@ -1,10 +1 @@
-<?php return '<div class="tribe-common-b2 tribe-events-status-single-notice">
```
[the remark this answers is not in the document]
Camwyn: Dangit, no.

This has been punted to next the release, so I have some more time to dig.

tests/event_status/Tribe/Events/Event_Status/JSON_LDTest.php
Camwyn: Nope - still fails because the eventStatus doesn't get added to the mock object as expected

Camwyn: `1) JSON_LDTest: It_should_have_event_status_for_canceled
 Test  tests/event_status/Tribe/Events/Event_Status/JSON_LDTest.php:it_should_have_event_status_for_canceled
Function modify_event did not add eventStatus.
Failed asserting that object of class "stdClass" has attribute "eventStatus".
#1  /var/www/html/wp-content/plugins/the-events-calendar/tests/event_status/Tribe/Events/Event_Status/JSON_LDTest.php:64
#2  Tribe\Events\Event_Status\JSON_LDTest->it_should_have_event_status_for_canceled
#3  {closure}`

tests/event_status/Tribe/Events/Event_Status/Views/__snapshots__/Single_ViewTest__it_should_render_single_canceled_name_no_reason_and_html__1.php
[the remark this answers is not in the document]
Camwyn: Ah, those should also have been skipped - same reason.

tests/event_status/Tribe/Events/Event_Status/JSON_LDTest.php
Camwyn: now fixed

src/Tribe/Editor/Hooks.php
```diff
@@ -27,6 +30,40 @@ public function register() {
...
+
+	/**
+	 * Calculate the event duration based on the saved start and end dates.
+	 * This occurs during post save, and checks the saved value against the calculated value on updates.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $post_id The event we are setting a duration for.
+	 */
+	public function calculate_duration( $post_id ) {
```
lucatume, line 44: The `Hooks` class is a service provider, it should not define or run business logic if possible. This method's business logic should be moved to a dedicated class.

src/Tribe/Editor/Hooks.php
```diff
@@ -27,6 +30,40 @@ public function register() {
...
+		// Sanity check.
+		if ( TEC::POSTTYPE !== get_post_type( $post_id ) ) {
+			return;
+		}
+
+		$saved_duration = get_post_meta( $post_id, '_EventDuration', true  );
+		$start_date_utc = get_post_meta( $post_id, '_EventStartDateUTC', true );
+		$end_date_utc   = get_post_meta( $post_id, '_EventEndDateUTC', true  );
```
lucatume, line 52: Working on the CT1 migration and related fixes, I found out that many sites will have varying combinations of localized start dates and UTC start dates; some due to being old, some due to plugins that will insert events directly in the database and will leave out meta we require. To deal with that, in CT1 code, [we use thist logic](https://github.com/the-events-calendar/the-events-calendar/blob/7179da28512f3bd23948f0a16f15510a04749588/src/Events/Custom_Tables/V1/Migration/Process_Worker.php#L847-L888) -- I think the best course of action here would be to either unify this logic and write it only once, or to make this logic more robust and be able to deal with mixed scenarios (e.g. `_EventStartDate` and `_EventEndDateUTC`).

src/Tribe/Models/Post_Types/Event.php
```diff
@@ -39,6 +40,11 @@ class Event extends Base {
 	 * {@inheritDoc}
 	 */
 	protected function build_properties( $filter ) {
+		// Sanity check.
```
lucatume, line 43: This sanity check is a welcome one, but why would this be called while setting up a Venue object? Did you find the origin of that bug?

lucatume (changes requested):

bordoni (approved):

src/Tribe/Models/Post_Types/Event.php
Camwyn: I did not yet. I'm going to ticket it separately.

Camwyn:
This is getting punted to the next MR to implement Luca's suggestions.

PR #3840, opened 2022-05-16
Add Outlook ICS Export

jesseeproductions:
Adds an Outlook ICS Export option that removes the X-WR-CALNAME from the ics file.

Fix is similar to what Andy Fragen's Plugin does: https://wordpress.org/plugins/the-events-calendar-outlook-import-fix/
However, his plugin removes it from the regular ics file creation, while this creates an outlook specific one to prevent issues with the regular iCal download.

Artifact:
https://cloudup.com/cSjYunRSWAT

_Ref:_ [TEC-4188](https://theeventscalendar.atlassian.net/browse/TEC-4188)


jesseeproductions:
This requires tests, but will be QA'ed in branch first. Tests are coming, but the coding can be reviewed in the mean time.

readme.txt
```diff
@@ -226,8 +226,9 @@ Remember to always make a backup of your database and files before updating!
 = [TBD] TBD =
 
 * Feature - Add subscribe links for Outlook 365 and Outlook Live to the event views. [TEC-4168]
-* Feature - Add export links for Outlook 365 and Outlook Live to single events. [TEC-4168]
+* Feature - Add import links for Outlook 365 and Outlook Live to single events. [TEC-4168]
```
Camwyn, line 229: IMHO export was correct but 🤷🏼

src/Tribe/Views/V2/iCalendar/Links/Outlook_Export.php
```diff
@@ -0,0 +1,79 @@
...
+	public static $slug = 'outlook-ics';
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function register() {
+		$this->label = __( 'Export Outlook .ics file', 'the-events-calendar' );
+		$this->single_label = $this->label;
+
+		add_filter( 'tec_views_v2_subscribe_link_outlook_ics_visibility', [ $this, 'filter_tec_views_v2_subscribe_link_outlook_ics_visibility'], 10, 2 );
```
Camwyn, line 34: This isn't going to work - the slug is `outlook-ics` so this would have to be `tec_views_v2_subscribe_link_outlook-ics_visibility` (default to "doesn't show on single events")

src/Tribe/Views/V2/iCalendar/iCalendar_Handler.php
```diff
@@ -35,9 +36,10 @@ class iCalendar_Handler extends \tad_DI52_ServiceProvider {
 	protected $default_feeds = [
 		Google_Calendar::class,
 		iCal::class,
-		iCalendar_Export::class,
 		Outlook_365_Export::class,
 		Outlook_Live_Export::class,
```
Camwyn, line 40: It's not a big deal - and it may not have been clear, but the "_Export" class(es) were originally just for the download, specifically to help differentiate between iCal/iCalendar (export vs. subscribe)

Camwyn (changes requested):
Fix the non-working filter.

src/Tribe/Views/V2/iCalendar/Links/Outlook_Export.php
jesseeproductions: Ok good, because for now it will be hidden on single events.

src/Tribe/Views/V2/iCalendar/iCalendar_Handler.php
jesseeproductions: Good to know. I renamed them.

src/Tribe/iCal.php
```diff
@@ -197,48 +206,62 @@ public function maybe_add_link() {
...
+			$this->type = 'outlook';
+		}
+
+		/**
+		 * Action fired before the creation of the feed is started, helpful to set up methods and other filters used
+		 * on this class.
+		 *
+		 * @since 4.6.11
+		 */
+		do_action( 'tribe_events_ical_before' );
```
bordoni, line 232: What variables can we pass to the hook here?

src/Tribe/iCal.php
```diff
@@ -197,48 +206,62 @@ public function maybe_add_link() {
...
+		 * Allows filtering the event IDs after the `Tribe__Events__ICal` class
+		 * tried to fetch them from the current request.
+		 *
+		 * @since 4.6.0
+		 *
+		 * @param array<int>|false Either a list of requested event post IDs or `false`
+		 *                         if the current request does not specify the event post
+		 *                         IDs to fetch.
+		 */
+		$event_ids = apply_filters( 'tribe_ical_template_event_ids', $event_ids );
```
bordoni, line 250: Namespace is wrong here, I think.

bordoni (changes requested):

src/Tribe/iCal.php
jesseeproductions: This is an existing hook, I could add another, but I prefer not to change this one.

src/Tribe/iCal.php
jesseeproductions: This is an existing hook, I could add another, but I prefer not to change this one.

jesseeproductions:
merging to the bucket branch to reduce open outlook branches

PR #3863, opened 2022-06-07
Load stylesheet in Elementor preview mode. [ECP-1105]

EugeneKyale:
I'm adding a check to the `should_enqueue_single_event_block_editor_styles` method to enable the `tribe-events-v2-single-blocks` stylesheet to load in Elementor preview mode.

No change log entry required since one was already included in the original PR [here](https://github.com/the-events-calendar/the-events-calendar/pull/3860).

https://theeventscalendar.atlassian.net/browse/ECP-1105

Styles not loading in Elementor preview mode.
image

Fixed version
image


src/Tribe/Views/V2/Assets.php
```diff
@@ -650,6 +650,23 @@ public function should_enqueue_single_event_styles() {
...
 	 */
 	public function should_enqueue_single_event_block_editor_styles() {
+		/**
+		 * Checks whether the page is being viewed in Elementor preview mode.
+		 *
+		 * @since TBD
+		 *
+		 * @return bool Should the assets be enqueued.
+		 */
+		if (
```
Camwyn, line 660: This is fine - but if we're going to be doing this more we should consider creating a function for it. We do something similar in https://github.com/the-events-calendar/events-pro/blob/master/src/Tribe/Integrations/Manager.php#L125

Camwyn (approved):

PR #3871, opened 2022-06-13
Alter our FSE template addition 

Camwyn:
1. to give the correct path, and
2. to not affect single events.

### DIFF
--- a/src/Events/Editor/Full_Site/Hooks.php
+++ b/src/Events/Editor/Full_Site/Hooks.php
@@ -76,6 +76,10 @@ class Hooks extends \tad_DI52_ServiceProvider {
 			return $query_result;
 		}
 
+		if ( is_singular( 'tribe_events' ) ) {
+			return $query_result;
+		}
+
 		return $this->container->make( Templates::class )->add_events_archive( $query_result, $query, $template_type );
 	}
 
--- a/src/Events/Editor/Full_Site/Templates.php
+++ b/src/Events/Editor/Full_Site/Templates.php
@@ -73,7 +73,7 @@ class Templates {
 	 */
 	public function get_template_events_archive() {
 		$template_content = file_get_contents(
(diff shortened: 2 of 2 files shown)


src/Events/Editor/Full_Site/Hooks.php
```diff
@@ -76,6 +76,10 @@ public function filter_include_templates( $query_result, $query, $template_type
 			return $query_result;
 		}
 
+		if ( is_singular( 'tribe_events' ) ) {
```
bordoni, line 79: Use constant

src/Events/Editor/Full_Site/Templates.php
```diff
@@ -73,7 +73,7 @@ public function add_events_archive( array $query_result, $query, string $templat
 	 */
 	public function get_template_events_archive() {
 		$template_content = file_get_contents(
-			dirname( \Tribe__Main::instance()->plugin_path ) . '/src/Events/Editor/Full_Site/Templates/archive-events.html'
+			dirname( TRIBE_EVENTS_FILE ) . '/src/Events/Editor/Full_Site/Templates/archive-events.html'
```
bordoni, line 76: Why not use the plugin path? I would love to avoid using the main file constant outside of the base loading

bordoni (dismissed):

src/Events/Editor/Full_Site/Templates.php
Camwyn: Because the plugin path is giving an incorrect path:
`dev/plugins/src/Events/Editor/Full_Site/Templates/archive-events.html`

Camwyn: Ah, it's the `dirname` causing the issue 🤦🏼 
Tribe__Main::instance()->plugin_path . /src/Events/Editor/Full_Site/Templates/archive-events.html'
works correctly

PR #3876, opened 2022-06-16
Stopgap fix for aggregator csv imports

Camwyn:
[TEC-4070]

This is a stopgap to get the import running on all admin pages. We should shift to using action scheduler, but that has to wait for RBE/V1-removal.

1. Ensure that import continues when we're in the admin: https://d.pr/i/TxrEif
   - The intent is to hide the notice on non-aggregator pages, but w/o it the import stalls.
2. Use the built-in classes for notices, since we're copying their styling (colors, etc) 

[TEC-4070]: https://theeventscalendar.atlassian.net/browse/TEC-4070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Aggregator/Cron.php
+++ b/src/Tribe/Aggregator/Cron.php
@@ -166,18 +166,12 @@ class Tribe__Events__Aggregator__Cron {
 		// Fetch the initial Date and Hour
 		$date = date( 'Y-m-d H' );
 
-		// Based on the Minutes construct a Cron
+		// Based on the Minutes construct a wp_cron
 		$minutes = (int) date( 'i' );
-		if ( $minutes < 15 ) {
-			$date .= ':00';
-		} elseif ( $minutes >= 15 && $minutes < 30 ) {
-			$date .= ':15';
-		}elseif ( $minutes >= 30 && $minutes < 45 ) {
-			$date .= ':30';
-		} else {
-			$date .= ':45';
-		}
-		$date .= ':00';
+		// Get minutes / 15 with no remainder.
+		$minutes = intdiv( $minutes, 15 ) * 15;
+		// Format it & insert into date string. Add 0 seconds.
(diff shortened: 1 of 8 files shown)


src/Tribe/Aggregator/Cron.php
```diff
@@ -166,18 +166,12 @@ public function action_register_cron() {
...
-		}elseif ( $minutes >= 30 && $minutes < 45 ) {
-			$date .= ':30';
-		} else {
-			$date .= ':45';
-		}
-		$date .= ':00';
+		// Get minutes / 15 with no remainder.
+		$minutes = intdiv( $minutes, 15 ) * 15;
+		// Format it & insert into date string. Add 0 seconds.
+		$date .= sprintf(':%02d', $minutes) . ':00';
```
bordoni, line 174: Spacing.

src/Tribe/Aggregator/Record/Queue_Realtime.php
```diff
@@ -64,10 +65,6 @@ public function update_loop_vars() {
 	}
 
 	public function render_update_message() {
-		if ( ! Tribe__Events__Aggregator__Page::instance()->is_screen() ) {
-			return false;
-		}
```
bordoni, line 69: Let's make sure it doesn't show on all pages.

bordoni (changes requested):

src/Tribe/Aggregator/Cron.php
Camwyn, line 174:
```suggestion
		$date .= sprintf( ':%02d', $minutes ) . ':00';
```

bordoni (approved):

PR #3883, opened 2022-06-28
Date formatting critical error fix.

juliangumenita:
On some of the WordPress sites that use this plugin, it throws a critical error due to not checking the nullability of the date object.

With this fix, we are checking if the date is set and, if not, returning an empty string.


Camwyn:
@juliangumenita I've created a tweaked version of your fix and added a readme entry for it. If possible it will go out in the next release, if not it will go in the following one.

Thank you!

PR #3892, opened 2022-07-07
Don't load aggregator notice (and scripts) on new/edit post pages to prevent datepicker conflicts.

Camwyn:
Also, don't try to set a variable that shouldn't exist. Unless it does.

Notes:

1. The js changes just prevent a fatal if `tribe_ev` isn't defined - the datepicker conflict will still occur if the script is loaded on a page w/another datepicker. Thus the new PHP function...
2. `tec_aggregator_should_load_scripts()` prevents loading on edit/new post pages. It also has a hook in it so folks can prevent it in other places as needed.

[BTRIA-1285](https://theeventscalendar.atlassian.net/browse/BTRIA-1285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

### DIFF
--- a/src/Tribe/Aggregator/Page.php
+++ b/src/Tribe/Aggregator/Page.php
@@ -176,7 +176,7 @@ class Tribe__Events__Aggregator__Page {
 			'admin_enqueue_scripts',
 			[
 				'conditionals' => [
-					[ $this, 'tec_aggregator_should_load_scripts' ],
+					[ $this, 'aggregator_should_load_scripts' ],
 				],
 				'localize'     => (object) $localize_data,
 			]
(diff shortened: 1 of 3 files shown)


src/Tribe/Aggregator/Page.php
```diff
@@ -236,6 +239,41 @@ public function action_request( $screen ) {
...
 
+	/**
+	 * Basically an edited version of is_screen(), below,
+	 * that allows for loading on non-post edit pages.
+	 *
+	 * @since 5.16.2.1
+	 *
+	 * @return boolean
+	 */
+	public function tec_aggregator_should_load_scripts() {
```
bordoni, line 250: I don't think we need to namespace the method

PR #3895, opened 2022-07-08
V2 Styling > Month view: add CSS selector tribe-events-othermonth. [TEC-4034]

EugeneKyale:
Add a CSS class i.e. `tribe-events-othermonth` to past and future month dates in the month view to allow easy targeting similar to what we had in v1. 

https://theeventscalendar.atlassian.net/browse/TEC-4034

image

### DIFF
--- a/src/views/v2/month/calendar-body/day.php
+++ b/src/views/v2/month/calendar-body/day.php
@@ -53,7 +53,7 @@ if ( $today_date > $day_date ) {
  * @since 5.16.2
  */
 if ( $day[ 'month_number' ] < date( 'm', strtotime( $today_date ) ) ) {
-	$day_classes[] = 'tribe-events-calendar-month__day--past-month tribe-events-othermonth';
+	$day_classes[] = 'tribe-events-calendar-month__day--past-month';
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/views/v2/month/calendar-body/day.php
```diff
@@ -53,7 +53,7 @@
  * @since 5.16.2
  */
 if ( $day[ 'month_number' ] < date( 'm', strtotime( $today_date ) ) ) {
-	$day_classes[] = 'tribe-events-calendar-month__day--past-month';
+	$day_classes[] = 'tribe-events-calendar-month__day--past-month tribe-events-othermonth';
```
Camwyn, line 56:
```php
/**
 * Add a unique CSS class to any date that is not in the current month.
 *
 * @since TBD
 */
if ( $day[ 'month_number' ] !== date( 'm', strtotime( $today_date ) ) ) {
	$day_classes[] = 'tribe-events-calendar-month__day--other-month';
}
```

and you only have to do it once.

Camwyn (changes requested):
Simplify, and change the class so it matches our existing classes.

src/views/v2/month/calendar-body/day.php
Camwyn: This is meant as a separate, standalone conditional instead of mixing ti with the future/past ones.

EugeneKyale: Ooh yeah this works as well. I thought my initial approach or reusing the existing conditionals would be more efficient 🙂

EugeneKyale:
> Simplify, and change the class so it matches our existing classes.

The reason for using the specific class `tribe-events-calendar-month__day--other-month` is to enable v1 users who are migrating to v2 to be able to target the same class they had before.

Camwyn (dismissed):

Camwyn (approved):

PR #3899, opened 2022-07-12
Theme based CSS class [TEC-4391]

EugeneKyale:
Add a theme based CSS class to the HTML body tag when the `Default Page Template` setting is enabled under Events > Settings > Display. e.g. the CSS class would be `page-template-twenty-twenty-one` if the active theme is Twenty Twenty One.

https://theeventscalendar.atlassian.net/browse/TEC-4391

image

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -225,6 +225,7 @@ Remember to always make a backup of your database and files before updating!
 
 == [TBD] TBD ==
 
+* Fix - Add a theme based CSS class to the HTML body tag when the `Default Page Template` setting is enabled under Events > Settings > Display. [TEC-4391]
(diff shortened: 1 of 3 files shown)


Camwyn:
It's an _**old**_ theme, but we have at least one targeted style for 2017 based on that class: https://github.com/the-events-calendar/the-events-calendar/blob/master/src/resources/postcss/tribe-events-full.pcss#L2383

EugeneKyale:
> It's an _**old**_ theme, but we have at least one targeted style for 2017 based on that class: https://github.com/the-events-calendar/the-events-calendar/blob/master/src/resources/postcss/tribe-events-full.pcss#L2383
> 
> We should alter that so it uses the new class target 😉

For sure, nice catch there.

PR #3900, opened 2022-07-12
Split off the notice js from the fields js

Camwyn:
Also take care of the pertinent localized data.

### DIFF
--- a/src/resources/js/aggregator-fields.js
+++ b/src/resources/js/aggregator-fields.js
@@ -60,8 +60,6 @@ tribe_aggregator.fields = {
 		20000
 	],
 
-	progress: {},
-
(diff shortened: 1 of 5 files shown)


src/resources/js/aggregator-fields.js
```diff
@@ -286,7 +284,7 @@ tribe_aggregator.fields = {
 		}
 
 		if ( 'object' === typeof tribe_aggregator_save ) {
-			obj.progress.init();
+			$(document).trigger( 'tribe_aggregator_init_notice' )
```
borkweb, line 287:
```suggestion
			$(document).trigger( 'tribe_aggregator_init_notice' );
```

borkweb (approved):

PR #3926, opened 2022-07-29
Transient check smarter about Custom Fields and Table Schema version changes

bordoni:


src/Events/Custom_Tables/V1/Schema_Builder/Schema_Builder.php
```diff
@@ -48,6 +48,37 @@ public function get_field_schemas_that_need_updates() {
...
+
+			// Skip if the version constant is not defined.
+			if ( ! defined( $constant_name ) ) {
+				continue;
+			}
+
+			$versions[ $class_name ] = constant( $constant_name );
+		}
+
+		return md5( json_encode( $versions ) );
```
stratease, line 79: Not likely to be a problem, but we may want to sort the array to ensure the order doesn't vary and cause false hash changes.

stratease (changes requested):

src/Events/Custom_Tables/V1/Schema_Builder/Schema_Builder.php
bordoni: Resolved.

bordoni:
I am going to add some tests to this later today, to make sure it behaves correctly.

stratease (approved):

PR #3940, opened 2022-08-10
[Tribe__Events__Linked_Posts] Always initialize `$linked_post_types`

zosterops-lateralis:
Hello,

Thanks for the work on this plugin!

I've just stumbled upon a minor issue: in case all linked post types are disabled by the `tribe_events_register_default_linked_post_types` filter, the `$linked_post_types` member variable remains uninitialized, which then triggers errors or warnings in other class methods which expect it to be an array.

To reproduce:
1. Have a `tribe_events_register_default_linked_post_types` filter which returns an empty array in order to disable all linked post types.
2. Open the event editor page (by creating a new event or editing an existing one).
3. The `foreach()` on `$this->linked_post_types` in the `render_meta_box_sections()` method will then trigger an “Invalid argument supplied for `foreach()`” warning, as `$this->linked_post_types` is indeed not an array, but `null`: https://github.com/the-events-calendar/the-events-calendar/blob/ef733088decc926a386a8b98b15b5d3c03277967/src/Tribe/Linked_Posts.php#L1210-L1215

This tiny PR ensures that the `$linked_post_types` is always initialized to an array, even if the `register_linked_post_type()` method is never called.

Thanks!

Zosterops


bordoni:
Thanks so much for the PR, I will make sure it finds its way into a Release soon.

PR #3942, opened 2022-08-10
[Tribe__Events__Main] Remove `comments` from the default post type supports

zosterops-lateralis:
Hello,

It seems to me that the “Show comments” (`showComments`) configuration option is not correctly taken into account: even when it is unchecked, I can still see a “Comments” column in the event admin list, and a “Allow Comments” checkbox in the quick-edit box for events.

Looking at the code, it seems that comments are always part of the default supports for the event post type, regardless of the value of the `showComments` option. In fact, if the option is set, the `comments` support will be added to the event post type: https://github.com/the-events-calendar/the-events-calendar/blob/ef733088decc926a386a8b98b15b5d3c03277967/src/Tribe/Main.php#L2123-L2125
… but the converse is not true.

This PR suggests to simply remove `comments` from the default supports in the event post type declaration:
- if `showComments` is set, they will be reinstated later by the above code;
- and if it is unset, all comment-related stuff will be hidden in the admin pages, as one might expect when disabling this setting.

Thanks!

Zosterops


bordoni:
Thanks so much for the PR, I will make sure it finds its way into a Release soon.

PR #3943, opened 2022-08-11
fix(CT1) update earliest/latest markers correctly

lucatume:
When an Event is updated or created, then TEC will update the earliest
and latest dates and date markers.
The markers are the post IDs of the Events that should be used as
"signposts", the CT1 one code was not updating those creating an issue
where Views, like the Week one, would find those signposts empty and
would not set up navigation correctly.


stratease (approved):
Groovy.

PR #3946, opened 2022-08-11
Add an `is_now` property to events.

Camwyn:
This allows checking the event against the current time easily via `$event->is_now`
to allow for things like this filter:
https://d.pr/i/C0QXFH
https://d.pr/i/QnmxPP

This is from a conversation in Slack: https://lw.slack.com/archives/C01SYM19N7K/p1660147997503949

### DIFF
--- a/src/Tribe/Models/Post_Types/Event.php
+++ b/src/Tribe/Models/Post_Types/Event.php
@@ -173,6 +173,14 @@ class Event extends Base {
 			$end_site           = $end_date_object->setTimezone( $site_timezone );
 			$use_event_timezone = Timezones::is_mode( Timezones::EVENT_TIMEZONE );
 
+			// Determine if current datetime is between event start and end dates.
+			$is_now             = Dates::range_coincides(
+				$now->format( 'U' ),
+				$now->add( $one_second )->format( 'U' ),
+				$start_date_utc_object->format( 'U' ),
+				$end_date_utc_object->format( 'U' )
+			);
+
 			$properties = [
 				'start_date'             => $start_date,
 				'start_date_utc'         => $start_date_utc,
@@ -192,6 +200,7 @@ class Event extends Base {
 				'duration'               => $duration,
 				'multiday'               => $multiday,
 				'is_past'                => $start_date_object < $now,
+				'is_now'                 => $is_now,
 				'all_day'                => $all_day,
 				'starts_this_week'       => $starts_this_week,
 				'ends_this_week'         => $ends_this_week,
--- a/src/functions/template-tags/event.php
+++ b/src/functions/template-tags/event.php
@@ -175,7 +175,7 @@ if ( ! function_exists( 'tribe_get_event' ) ) {
 			 */
 			$post = apply_filters( 'tribe_get_event', $post, $output, $filter );
 
-			// Dont try to reset cache when forcing.
+			// Don't try to reset cache when forcing.
 			if ( ! $force ) {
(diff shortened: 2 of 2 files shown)


src/Tribe/Models/Post_Types/Event.php
```diff
@@ -173,6 +173,14 @@ protected function build_properties( $filter ) {
 			$end_site           = $end_date_object->setTimezone( $site_timezone );
 			$use_event_timezone = Timezones::is_mode( Timezones::EVENT_TIMEZONE );
 
+			// Determine if current datetime is between event start and end dates.
+			$is_now             = Dates::range_coincides(
+				$now->format( 'U' ),
+				$now->add( $one_second )->format( 'U' ),
+				$start_date_utc_object->format( 'U' ),
+				$end_date_utc_object->format( 'U' )
+			);
```
bordoni, line 182: make this into a callback method no? To avoid caching problems.

bordoni (changes requested):

src/Tribe/Models/Post_Types/Event.php
Camwyn: Smart

lucatume (changes requested):
Add a test to check on the `is_now` property of the event, not much to make sure it works now, but to make sure it will keep working and it will not be silently compromised in a regression.

jesseeproductions (approved):

tests/wpunit/Tribe/Events/Models/Event_ModelTest.php
```diff
@@ -46,6 +48,28 @@ public function it_should_return_past() {
...
+		$event = Event::from_post( $mock )->to_post();
+
+		$this->assertFalse( $event->is_now );
+	}
+
+	/**
+	 * @test
+	 */
+	public function it_should_return_is_now_true() {
+		// Here, we're just overriding the function in common to assure it returns true.
```
lucatume, line 65: I would not do this: while there is some value in testing that, given the `is_now` from common will return a value the `$event->is_now` property will reflect that, there is more value in writing a real integration test where an Event is set up to happen now, using `tribe_events()->set_args( ... )->create()` and test what comes out of it.

tests/wpunit/Tribe/Events/Models/Event_ModelTest.php
```diff
@@ -46,6 +48,28 @@ public function it_should_return_past() {
 		$this->assertTrue( $event->is_past );
 	}
 
+	/**
+	 * @test
+	 */
+	public function it_should_return_is_now() {
+		$mock  = $this->get_mock_event( 'events/single/1.json' );
```
lucatume, line 55: No need to use a mock event for anything that is not a snapshot: use a real Event created using `tribe_events()->set_args( ... )->create()`.

tests/wpunit/Tribe/Events/Models/Event_ModelTest.php
```diff
@@ -46,6 +48,28 @@ public function it_should_return_past() {
 		$this->assertTrue( $event->is_past );
 	}
 
+	/**
+	 * @test
+	 */
+	public function it_should_return_is_now() {
```
lucatume, line 54: This test, and the following one, could be conflated in a `test_is_now_property` test that would be fed data from a data provider to create an Event happening at a certain offset, with respect to the current moment, and an expected `is_now` value.

E.g. `test_is_now_property( string $when, int $duration, bool $expected ) : void`.

lucatume (changes requested):

lucatume (approved):

PR #3949, opened 2022-08-12
Break out day class logic into its own function.

Camwyn:
From a Slack conversation: https://lw.slack.com/archives/C01T0TSD4SC/p1659734059492149

This takes the logic that was cluttering up the `src/views/v2/month/calendar-body/day.php` template and moves it to a function alongside the similar one we use for multiday classes.

All logic is now contained in the function.
this logic is now changed so that the comparison date is the request date, rather than _today_ - so when viewing a past/future month you get the same visual effects you would on the current month.

Two new filters:
`tec_month_day_classes_comparison_date` - so folks can filter the date used for class determination comparisons.
`tec_month_day_classes` - so folks can filter the actual class list before it gets passed to the template.

[TEC-4457]

[TEC-4457]: https://theeventscalendar.atlassian.net/browse/TEC-4457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Views/V2/functions/classes.php
```diff
@@ -94,3 +94,76 @@ function month_multiday_classes( $event, $day_date, $is_start_of_week, $today_da
...
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string> $day_classes     The classes to add to the day "cell".
+	 * @param string        $comparison_date The date that was used for comparisons.
+	 * @param array<mixed>  $day             The current day data.
+	 *
+	 * @return array<string> $day_classes The final list of classes to add to the day "cell".
+	 */
+	return apply_filters( 'tec_month_day_classes', $day_classes, $comparison_date, $day );
```
bordoni, line 168: `tec_events_`

src/Tribe/Views/V2/functions/classes.php
```diff
@@ -94,3 +94,76 @@ function month_multiday_classes( $event, $day_date, $is_start_of_week, $today_da
...
+		$day_classes[] = 'tribe-events-calendar-month__day--past';
+	}
+
+	/**
+	 * Add unique CSS class(es) to any date that is not in the current month.
+	 *
+	 * @since 5.16.3
+	 * @since TBD moved to month_day_classes function.
+	 */
+	$month = date( 'm', strtotime( $comparison_date ) );
```
bordoni, line 143: Avoid Date usage.

src/Tribe/Views/V2/functions/classes.php
```diff
@@ -94,3 +94,76 @@ function month_multiday_classes( $event, $day_date, $is_start_of_week, $today_da
...
+ * @since TBD
+ *
+ * @param array<mixed> $day          The current day data.
+ * @param string       $day_date     The current day date, in the `Y-m-d` format.
+ * @param DateTime     $request_date The request date for the view.
+ * @param string       $today_date   Today's date in the `Y-m-d` format.
+ *
+ * @return array<string> $day_classes The classes to add to the day "cell".
+ */
+function month_day_classes( $day, $day_date, $request_date, $today_date ) {
```
bordoni, line 111: All of the conditionals here could be simplified if we passed an array of the conditionals.

bordoni, line 111: Type hint every param and the return.

bordoni (changes requested):

src/Tribe/Views/V2/functions/classes.php
Camwyn: Do you mean parameters?

Camwyn: oh!

lucatume:
The [most recent failure in test](https://github.com/the-events-calendar/the-events-calendar/runs/8222748334?check_suite_focus=true) was fixed in `release/B22.scone` release. Fixing it should be just a question of merging `release/B22.scone` into `release/B22.strudel`. I will reach out to point it out.

lucatume (approved):
Code and fix look fine, wait for the tests to pass.

src/Tribe/Views/V2/functions/classes.php
```diff
@@ -94,3 +96,64 @@ function month_multiday_classes( $event, $day_date, $is_start_of_week, $today_da
...
+	 *
+	 * @param string       $comparison_date The date used for comparisons.
+	 * @param DateTime     $request_date    The request date for the view.
+	 * @param string       $day_date        The current day date, in the `Y-m-d` format.
+	 * @param array<mixed> $day             The current day data.
+	 */
+	$comparison_date =  apply_filters( 'tec_events_month_day_classes_comparison_date', $comparison_date, $request_date, $day_date, $day  );
+
+	// Convert it to a date object.
+	$comparison_date = Dates::build_date_object( $comparison_date );
```
lucatume, line 130: Use the immutable version of the object:
```suggestion
	$comparison_date = Dates::immutable( $comparison_date );
```
When the reference to a date object is passed around, as is the case in the filter following this code, then there's a chance the filtering code will modify it by reference for all the following filters.
You should always use immutable date-time objects, and use the non-mutable version only when in complete control of the flow. Too many bugs to number emerged during CT1 development due to this.

lucatume (changes requested):

lucatume (dismissed):

PR #3950, opened 2022-08-12
Make datepicker pay attention to RTL and language.

Camwyn:
This checks the document for `lang` and `rtl` attributes and passes them on to the datepicker (because, why would it do that check itself? :P )

It also tweaks the styling to ensure that the datepicker is entirely on the screen on mobile.

List View: 
📱  https://d.pr/i/Wn1Tuy
💻 https://d.pr/i/B31rRV

Month View:
📱 https://d.pr/i/0WIVg2
💻 https://d.pr/i/vSbSXE

[TEC-4452]

[TEC-4452]: https://theeventscalendar.atlassian.net/browse/TEC-4452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resources/js/views/datepicker.js
```diff
@@ -642,6 +642,16 @@ tribe.events.views.datepicker = {};
 		obj.options.beforeShowMonth = obj.filterMonthCells;
 		obj.options.beforeShowYear = obj.filterYearCells;
 
+		// A tweak for RTL languages.
+		if ( document.dir === 'rtl' ) {
```
bordoni, line 646: Check if this is defined, just in case.

src/resources/js/views/datepicker.js
```diff
@@ -642,6 +642,16 @@ tribe.events.views.datepicker = {};
...
+			obj.options.rtl = true;
+		}
+
+		// ...and for document language.
+		if ( document.lang ) {
```
bordoni, line 651: Check if this is defined, just in case.

bordoni (changes requested):

src/resources/js/views/datepicker.js
Camwyn: That's all this is doing 😉 - if it exists, we set it.

lucatume (approved):

bordoni (approved):

PR #3955, opened 2022-08-17
This prevents network deactivation from overwriting subsite settings …

Camwyn:
…with main site (cached) settings.

Of note:
This creates a `get_reset_version()` in Updater so we can not rely on hardcoded values in tests.
Also fixes a couple asserts that had the expected & actual params swapped.

[TEC-3738]

[TEC-3738]: https://theeventscalendar.atlassian.net/browse/TEC-3738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Updater.php
```diff
@@ -69,6 +69,18 @@ protected function get_updates() {
...
+	 * Getter for the private reset version.
+	 * Mainly for tests.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The reset version number.
+	 */
+	public function get_reset_version() {
+		return $this->reset_version;
+	}
```
bordoni, line 82: Type hint missing.

bordoni (changes requested):

bordoni (approved):

PR #3956, opened 2022-08-19
Fix for errant group by.

stratease:
Fix for RBE issue 450.

On the shortcode edit view, the group by was being added when the table was not being joined. Adding some introspection to avoid this.


bordoni (approved):

lucatume:
I've ran both suites locally (the `restv1` one and the `views_integration` one, and both are passing. Merging.

lucatume (approved):

PR #3965, opened 2022-08-31
Fix for occurrence duration timezone validator.

stratease:
[Fix for ECP-1270](https://theeventscalendar.atlassian.net/browse/ECP-1270).

No timezone exists in an occurrence, so we need to pull the timezone from the event.
Added new validator to handle this, that extends the other duration validator that assumes the timezone is a property of the model.

This showed up as two different errors. One complained about the incorrect duration being computed, the other buries that error inside the `Unexpected number of Occurrences found: expected 1, found 0.`

Screenshot 2022-08-31 155433


stratease:
Looks like there are several tests failing because of the relative comparison, and not using the right timezone. Starting to fix some, but if we re-run the tests tomorrow they "should" pass.

src/Events/Custom_Tables/V1/Models/Validators/Occurrence_Duration.php
```diff
@@ -0,0 +1,77 @@
...
+	 * @since TBD
+	 *
+	 * @param Occurrence $model The occurrence to fetch it's timezone for.
+	 *
+	 * @return string|null
+	 */
+	protected function get_occurrence_timezone( Occurrence $model ): ?string {
+		$event = Event::find( $model->event_id );
+		if ( $event instanceof Event ) {
+
```
lucatume, line 71:
```suggestion
```

lucatume (changes requested):
The PR code seems fine; add a test to cover the migration process in the `ct1_migration` suite, please.

src/Events/Custom_Tables/V1/Models/Validators/Occurrence_Duration.php
stratease: Added new test. We don't seem to have much doing specific validator testing. New test will cover this as well as duration validation, in an integration test.

lucatume (approved):

PR #3967, opened 2022-09-01
Fix translated url escaping

Camwyn:
This is a copy of https://github.com/the-events-calendar/the-events-calendar/pull/3807 so I could add a readme entry and docblock. All props for the code itself go to @shisho585

> In non-English based sites, when slugs are translated to a language which is encoded with '%' inside, the rewrite rule adds, in addition to the translated string, another translated string without the '%' char, which is being removed by 'get_i18n_strings_for_domains' function that uses WordPress' 'sanitize_key' function.
> 
> For example, on an Hebrew based site, month slug will be saved in rewrite rules as '(?:month|d797d795d793d7a9|%d7%97%d7%95%d7%93%d7%a9)' instead of '?:month|%d7%97%d7%95%d7%93%d7%a9)'.
> 
> This causes the url for hebrew version of the site to be ugly ('d797d795d793d7a9' which is meaningless, instead of 'חודש' which is the actual translation), and also causes problems with links on the English version of the site, because it can't solve the rewrite rule correctly.
> 
> This PR solves the first problem for sure, and somehow also fix the second problem on my site.
> 
> It can be written nicer but this way it doesn't break any compatibility.


shisho585:
Thanks @Camwyn.

src/Tribe/I18n.php
```diff
@@ -156,6 +156,60 @@ public function get_i18n_strings_for_domains( $strings, $languages, $domains = [
...
+	 *
+	 * WARNING: This function is slow because it deals with files, so don't overuse it!
+	 * Differently from the `get_i18n_strings` method this will not use any domain that's not specified.
+	 *
+	 * This function is same as above one, but instead of sanitizing with 'sanitize_key()' which removes '%',
+	 * it uses 'sanitize_title()'.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $strings    An array of strings (required).
```
lucatume, line 170:
```suggestion
	 * @param array<string> $strings    An array of strings (required).
```

src/Tribe/I18n.php
```diff
@@ -156,6 +156,60 @@ public function get_i18n_strings_for_domains( $strings, $languages, $domains = [
...
+	 * @param array $languages Which l10n to fetch the string (required).
```
lucatume, line 171:
```suggestion
	 * @param array<string> $languages Which l10n to fetch the string (required).
```

src/Tribe/I18n.php
```diff
@@ -156,6 +156,60 @@ public function get_i18n_strings_for_domains( $strings, $languages, $domains = [
...
+	 * @param array $domains   Possible domains to re-load.
```
lucatume, line 172:
```suggestion
	 * @param array<string> $domains   Possible domains to re-load.
```

lucatume (approved):

PR #3970, opened 2022-09-04
Updating some test methods to properly be declared as static

borkweb:
This resolves some PHP 7.3 compatibility issues with methods being called statically when they weren't being declared as such.

I'm PR-ing this against `release/B22.royal` on purpose.


lucatume (dismissed):

bordoni:
I've moved this PR from `release/B22.royal` to `master`, please review and point it to the correct branch. The next sprint will live in `release/B22.wagashi`.

PR #3975, opened 2022-09-07
Fix and tweak CT1 migration process

lucatume:
WiP!


stratease (approved):
Groovy. Marginally concerned about the query performance, but I think it should be fine. Might be worth doing an EXPLAIN on a worst case scenario...

PR #3976, opened 2022-09-07
Include migration state into Support System Information

bordoni:
🎫 [ECP-1316]
---

Leah has asked for the status of the Custom Tables migration to be included in the Support data.

📷  http://i.bordoni.me/XV3AlxTHCuFEwMMFbSR2

_Note: When reviewing, please remove whitespace changes to see only the lines that matter to this code_

[ECP-1316]: https://theeventscalendar.atlassian.net/browse/ECP-1316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


readme.txt
```diff
@@ -231,6 +231,9 @@ Remember to always make a backup of your database and files before updating!
 
 = [6.0.1] TBD =
 
+* Tweak - Ensure the Fast-forward link uses the correct, customizable color as other links. [TEC-3962]
```
Camwyn, line 234: did we have a reversion?

src/Events/Custom_Tables/V1/Activation.php
```diff
@@ -86,6 +86,35 @@ public static function init() {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param array $info
+	 *
+	 * @return array
+	 */
+	public static function filter_include_migration_to_system_info( array $info = [] ): array {
+		$phase = tribe( State::class )->get_phase();
+		$incomplete_label = esc_html__( 'Incomplete', 'the-events-calendar' );
```
Camwyn, line 100: Should we add something about what state it _is_ in here? Might be useful to know they are in the middle of a reversion, etc

Camwyn: Also - translations should always use _x() functions - add some context for the translators 🙏🏼

Camwyn (changes requested):
Some small tweaks and a question.

readme.txt
bordoni: Not sure. What do you mean? I conflict resolved and I think it ended up re-ordering.

src/Events/Custom_Tables/V1/Activation.php
bordoni: Not for the which state, but the `_x()` is a good idea

bordoni: I decided to go with untranslated strings since this is only for us.

src/Events/Custom_Tables/V1/Activation.php
```diff
@@ -86,6 +86,36 @@ public static function init() {
...
 	}
 
+	/**
+	 * Includes the Migration state into the System Information for support.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $info
+	 *
+	 * @return array
```
lucatume:
```suggestion
	 * @return array<string,mixed> The filtered support information map.
```

src/Events/Custom_Tables/V1/Activation.php
```diff
@@ -86,6 +86,36 @@ public static function init() {
 		}
 	}
 
+	/**
+	 * Includes the Migration state into the System Information for support.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $info
```
lucatume:
```suggestion
	 * @param array<string,mixed> $info The support information map.
```

lucatume (approved):

PR #3979, opened 2022-09-09
Notify before/after Views container replacement

lucatume:
Ticket: [ECP-1308](https://theeventscalendar.atlassian.net/browse/ECP-1308).

Required by [ECP PR](https://github.com/the-events-calendar/events-pro/pull/2037).

This PR updates the Views v2 manager code to make it dispatch native JS events when the container HTML is replaced using jQuery.

The jQuery `element.replaceWith` method will not only replace the element HTML, but **it will remove all bound event listeners**.

Dispatching a native event before and after the container HTML is replaced will allow listeners to gracefully unbind and re-bind events before and after the container is replaced.

Artifacts: [ECP PR Screencast](https://share.cleanshot.com/dJSrQL)

### DIFF
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -27,6 +27,7 @@ use Tribe__Events__Rewrite as TEC_Rewrite;
 use Tribe__Events__Venue as Venue;
 use Tribe__Repository__Interface as Repository;
 use Tribe__Utils__Array as Arr;
+use WP_Post;
 
 /**
  * Class View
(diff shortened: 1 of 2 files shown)


bordoni:
I am wondering if we should start making use of `wp-hook` here since they are they standard now for actions and filters on JS for WP.

lucatume:
> I am wondering if we should start making use of `wp-hook` here since they are they standard now for actions and filters on JS for WP.

I'm not a fan of using a library when the native API is perfectly capable of delivering the same value. I'm open to suggestions, though.

bordoni (approved):

PR #3987, opened 2022-09-19
Feature/admin pages menu

Camwyn:
Draft so folks can see and comment on where I am with the new admin page/menu architecture.

Requires https://github.com/the-events-calendar/tribe-common/pull/1810


src/Events/Menus/Event.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Class Events CPT admin/menu.
+ *
+ * @since TBD
+ *
+ * @package TEC\Events\Menus
+ */
+class Event extends Abstract_Menu {
+	use Submenu, CPT;
```
lucatume, line 26: Following along the lines of my proposal in the Common PR: I think it would make more sense to have a base, `abstract`, custom-post-type, sub-menu class in the Common portion that would be used here.

lucatume (changes requested):

src/Events/Menus/Event.php
borkweb: Haha! I asked those same questions in a call with Stephen yesterday :D He explained that he thinks the Traits make sense when combinations of the following are possible:

Menu + CPT + Admin Bar
Menu + CPT + Admin Bar + Settings
Menu + CPT
Submenu + CPT + Admin Bar
Submenu + Admin Bar
Submenu + Settings
etc.

A bunch of combinations are possible and we use a number of those combinations, so Traits feel like they make sense here.

Side note: I think the menu handling code could be spun out to its own StellarWP library!

Camwyn: > Side note: I think the menu handling code could be spun out to its own StellarWP library!

That's my hope 🤞🏼

PR #3988, opened 2022-09-20
Handle Main deprecated method and testing

lucatume:
Ticket: [ECP-1326](https://theeventscalendar.atlassian.net/browse/ECP-1326)

Artifacts: tests.

This PR fixes the issue where the `Tribe__Events__Main::get_closest_event` might not have been gracefully deprecated in `6.0.0.` and proposes an approach to deprecation and its testing to be used in this case and others that might require it.

The idea is to create, in the `src/deprecated/Traits` directory, traits dedicated to the support of deprecated methods in classes that require it.  
The naming convention would be `<class_name>_Deprecated.php` and the trait would be used in the class.

E.g. the `Tribe__Events__Main` would use a `Tribe__Events__Main_Deprecated.php` trait that would define all its deprecated methods. This would avoid keeping the deprecated methods definitions and handling in the class and would only require the class to `use` the trait dedicated to its deprecated methods.

As part of of this PR, and the proposal for this approach, I've laid out the `deprecated` suite: this only contains a test that will cover the deprecated methods defined in each `X_Deprecated.php` trait automatically, without requiring the writing of a test dedicated to it. See the test to know more.

- refactor(src) remove unused var
- test(integration) redirect and update test from deprecated class
- refactor(src,tests) move deprecated methods to trait
- build(.github) add the deprecated suite


borkweb (approved):
I love the idea of moving deprecated methods into a trait!

This looks good 🐘:dash:

Camwyn (dismissed):
Ooh, I _like_ this!

💡 can we test the deprecated_version against the current plugin version and fail if there is a major version diff?

bordoni (dismissed):
I really like the idea of the Trait here.

lucatume:
> 💡 can we test the deprecated_version against the current plugin version and fail if there is a major version diff?

Added in last commit, good one!

PR #3989, opened 2022-09-21
No type-hinting on public hooks

lucatume:
Ticket: [ECP-1343](https://theeventscalendar.atlassian.net/browse/ECP-1343)

Artifacts: tests are already covering the chanded code.

Refactor the CT1 portion of the code to make sure method and functions
hooked on public (non-TEC) actions and filters will not cause fatals
when the type-hinted expected type is not met during the application of
a filter or the dispatch of an action.

This might happen if other plugins code is hooking on the filter before
our code and does not respect the expected output types; the Filters API
does not have type-checking.


stratease (dismissed):
Groovy, thanks!

PR #3991, opened 2022-09-22
Revert access level to prevent uncatchable fatal on Transient Unserialization

bordoni:


borkweb (approved):
Approved. Once tests pass, and they will.

PR #3994, opened 2022-09-26
Avoid wasteful queries in multi-site context

lucatume:
Ticket: [BTRIA-1370](https://theeventscalendar.atlassian.net/browse/BTRIA-1370)

Artifacts: tests and [screencast](https://share.cleanshot.com/Wxp0Py).

When in multi-site installation context, we would run the
`Schema_Builder::up` method on each `switch_blog` action dispatch. That
action is called a number of times, in multi-site context, both when
navigating the site Admin UI and the site front-end.

This commit updates the logic to:
1. Set the activation transient that will prevent that check from
   happening more than 1/day.
2. Add a second layer of protection (if transients are deactivated or
   cleared more frequently) to cache the call to the `up` method on a
   per-request base and avoid calling it more than once.


stratease (dismissed):
Groovy!

PR #3990, opened 2022-09-22
feat(REST API) suport strict_dates on /events

lucatume:
By default, unless the site administrator has filtered the
`tribe_events_rest_use_inclusive_start_end_dates` filter to return a
falsy value, a REST API request for the `/events` endpoint will cast the
`start_date` parameter to the start of the day, and the `end_date`
parameter to the end of the day.

This adds a `strict_dates` supported REST API argument to allow REST API
clients to control, in the context of a request, the default value of
the value filtered in the
`tribe_events_rest_use_inclusive_start_end_dates` filter.

The filtered value will still override this value if explicitly set.

Furthermore, I've fixed the description of the `end_date` parameter: it
will limit the fetch by Events ending before a date and time; it's not
about the start date.

Artifacts: tests

[TEC-4493]

[TEC-4493]: https://theeventscalendar.atlassian.net/browse/TEC-4493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3995, opened 2022-09-26
Feature/tec 4493 rest params

moraleida:
(copy of #3990, rebased oustide of strudel)

By default, unless the site administrator has filtered the
tribe_events_rest_use_inclusive_start_end_dates filter to return a
falsy value, a REST API request for the /events endpoint will cast the
start_date parameter to the start of the day, and the end_date
parameter to the end of the day.

This adds a strict_dates supported REST API argument to allow REST API
clients to control, in the context of a request, the default value of
the value filtered in the
tribe_events_rest_use_inclusive_start_end_dates filter.

The filtered value will still override this value if explicitly set.

Furthermore, I've fixed the description of the end_date parameter: it
will limit the fetch by Events ending before a date and time; it's not
about the start date.

==== (Ricardo's notes below)

To get the ET+ ap working properly we also needed to allow the /events endpoint to work with `starts_after`, `ends_before` (...) params. These were enabled and tests were added to check results both for these types of simple relative queries as well as queries mixing set dates (`start_date`, `end_date`) and the new params.  

Artifacts: tests

[TEC-4493](https://theeventscalendar.atlassian.net/browse/TEC-4493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)


stratease (approved):
Groovy

src/Tribe/REST/V1/Endpoints/Archive_Event.php
```diff
@@ -24,6 +24,10 @@ class Tribe__Events__REST__V1__Endpoints__Archive_Event
 		'status'      => 'post_status',
 		'post_parent' => 'post_parent',
 		'include'     => 'post__in',
+		'starts_before' => 'starts_before',
```
lucatume, line 27: Add argument documentation for these parameters in he `READ_args` method of this class.

lucatume (changes requested):
Just missing documentation in the archive endpoing `READ_args` method.

rafsuntaskin (dismissed):

rafsuntaskin:
`restv1` tets are still failing.

moraleida:
oh, that's new... looking into it

moraleida:
finally! :tada:

lucatume (approved):

PR #3997, opened 2022-09-27
CT1 Migration - do not throw on others' notices and warnings

lucatume:
Ticket: [ECP-1318](https://theeventscalendar.atlassian.net/browse/ECP-1318)

Artifacts: tests and [screencast](https://share.cleanshot.com/Ebqn5Q)

Update the error handler logic we set during the migration process to
not throw due to notices (`E_NOTICE` and `E_USER_NOTICE`) and warnings
(`E_WARNING` and `E_USER_WARANING`) generated by non-TEC plugins.

The previous, stricter, code would stop the Event migration when notices
and warnings are raised during Event migration. Other plugins and themes
not as strict would routinely emit those and would, thus, prevent the
migration from happening.

The PR does not alter how notices and warnings coming from TEC plugins
are handled: they will raise an exception. This is a decision based on
the fact the TEC team cannot control others' code, but can control, thus
fix, its own code.


stratease (approved):
Groovy.

PR #3966, opened 2022-09-01
Fix CT1 issue 492

lucatume:
Fix CT1 migration issue 492 by making sure we rebuild an Event date information,
the `_Event(Start|End)Date(UTC)*` meta, where possible.

The PR will throw an exception when unable to do so that will end up in the
migration report to suggest the user a course of action to remedy the issue.

Artifacts: tests in ECP PR.

PR #3998, opened 2022-09-29
Prevent GCalendar events from creating duplicates

bordoni:
🎫 [TEC-4497]
---
Kept creating new instances until the new code.

Screen Shot 2022-09-29 at 8 36 05 AM

Kinda reverts code from #3966

[TEC-4497]: https://theeventscalendar.atlassian.net/browse/TEC-4497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lucatume (commented):
Since it was broken and no one noticed and it’s now fixed and no one noticed, I would add tests on the `Aggregator/Record/Abstract` object.
Check the `wpunit` suite failure.

From the PR description, it's not clear to me what the issue was and, thus, what the fix is.

lucatume (changes requested):
Add tests to cover the fix from future regression.

bordoni:
Added tests

tests/wpunit/Tribe/Events/Aggregator/Record/AbstractTest.php
```diff
@@ -282,6 +281,69 @@ public function should_correctly_insert_posts() {
 		$this->assertEquals( 0, $activity->count( Main::POSTTYPE,'skipped' ) );
 	}
 
+	/**
+	 * It should correctly insert posts
```
lucatume, line 285: Copy and pasta here?

tests/wpunit/Tribe/Events/Aggregator/Record/AbstractTest.php
```diff
@@ -282,6 +281,69 @@ public function should_correctly_insert_posts() {
...
+			]
+		] );
+
+		$manual_test_record = new \Tribe__Events__Aggregator__Record__Manual_Test( $record );
+		$manual_test_record->meta['origin'] = 'manual-test';
+
+		// Enforce this particular Static variable.
+		\Tribe__Events__Aggregator__Record__Manual_Test::$unique_id_fields = \Tribe__Events__Aggregator__Record__Abstract::$unique_id_fields;
+
+		$data = [
```
lucatume, line 335: Formatting a bit wonky here.

tests/wpunit/Tribe/Events/Aggregator/Record/AbstractTest.php
```diff
@@ -282,6 +281,69 @@ public function should_correctly_insert_posts() {
...
+			[
+				'uid' => $uid,
```
lucatume, line 337: Adding, either here or in a dedicated test, a `uid` pointing to an event that is not existing, would test the mixed, and more realistic nature, of this code.

PR #3999, opened 2022-09-29
Restore date-based filtering of Evt queries

lucatume:
Ticket: [TEC-4470](https://theeventscalendar.atlassian.net/browse/TEC-4470)

Artifacts: tests and [screencast](https://share.cleanshot.com/jpR4qz)

Related: [ECP PR](https://github.com/the-events-calendar/events-pro/pull/2087); this deals with the effect of this change in ECP context.

In 6.0, the previous query filtering was removed to rely on the one provided by Views v2,
now the default view in non-CT1 context, and on the one adopted by CT1 when in CT1 context.  

One of the functionality provided by such code was filtering **any** Event query, whether it
came from Views or not, to make sure that Event would be sorted, by default, in ASCending start
date order. 
With the removal of that general purpose code, queries implicitly requiring on that sorting logic
as the ones set up by visual theme builders or the RSS feed, would show Events in `post_date` order
resulting in out-of-order (in respect to the start date one) list of Events.

This PR restores that catch-all functionality to make sure it will apply, as it did before, to
all queries. Views v2 will keep overriding that logic, and CT1 will leverage it to redirect it
to the custom tables.

Since the logic would influence CT1 and non-CT1 queries, test are duplicated in both contexts.


bordoni (approved):
I really like the tests here, make the cde much easier to review.

src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
```diff
@@ -297,20 +301,49 @@ public function order_by_occurrence_id( $order_by, $query = null ) {
...
+			$meta_query_clauses = $this->meta_query->get_clauses();
+			foreach ( $normalized_order_by as $original_key => $direction ) {
+				if ( $original_key === 'meta_value' ) {
+					// Handle queries with on meta value.
+					$original_key = array_key_first( $meta_query_clauses );
+				}
+
+				if ( in_array( $original_key, [ 'ID', $wpdb->posts . '.ID' ], true ) ) {
+					// If the order is by post ID, order by post ID and occurrence ID.
+					$buffer[] = "ID $direction, $occurrences.occurrence_id $direction";
```
stratease, line 324: Shouldn't these be parameterized? `$wpdb->prepare($query, $table)` ?

src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
```diff
@@ -297,20 +301,49 @@ public function order_by_occurrence_id( $order_by, $query = null ) {
...
+					continue;
+				}
+
+				if ( ! ( is_string( $original_key ) && isset( $meta_query_clauses[ $original_key ] ) ) ) {
+					// Not a key we redirect or handle.
+					$buffer[] = $original_key . ' ' . $direction;
```
stratease, line 330: Here to

stratease (changes requested):
Couple small SQL comments

src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
lucatume: If we prepared the table name and the direction, then they would both be prepared as strings, it would result in a SQL string like `ID 'ASC', 'wp_tec_occurrences'.occurrence_id 'ASC'` that would be invalid SQL. SQL fragments like these are typically composed like this and in a controlled context. The table name comes from a function we control.

src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
lucatume: See my reply to the previous comment.

PR #4002, opened 2022-09-30
CT1 - set properties on original query

lucatume:
Ticket: [BTRIA-1385](https://theeventscalendar.atlassian.net/browse/BTRIA-1385)

Artifacts: tests and [result of running the `ct1_integration`](https://share.cleanshot.com/svIkj0) and [`ct1_migration` suite](https://share.cleanshot.com/pATgBj) on ECP. I've run the tests on ECP as this fix has an impact on all CT1 queries.

When taking over a query in the context of CT1, set the `found_posts`
and `max_num_pages` properties correctly on the original query.

In CT1 context, a `WP_Query` for Events request will not run; in its
place will run a `Custom_Tables_Query` built on top of it that would
not, before this fix, correctly set all the expected properties on the
original `WP_Query` object.


readme.txt
```diff
@@ -243,6 +243,7 @@ Remember to always make a backup of your database and files before updating!
 * Fix - Restore erroneously removed Events query filters that would result in out-of-order results. [TEC-4770]
 * Tweak - Update the organizer website field to type URL. [TEC-4395]
 * Tweak - Update the venue website field to type URL. [TEC-4349]
+* Fix - Correctly set `found_posts` ans `max_num_pages` when redirecting a query to the custom tables. [BTRIA-1385]
```
Camwyn, line 246:
```suggestion
* Fix - Correctly set `found_posts` and `max_num_pages` when redirecting a query to the custom tables. [BTRIA-1385]
```

Camwyn (changes requested):

readme.txt
Camwyn: The rest looks 👨🏼‍🍳

PR #4007, opened 2022-10-05
Divi Organizer ID failing due to Integration filter

bordoni:
On [TEC-3814] we introduced this bug while Divi is active since we filter out the values for the Event ID, but this particular method is used for a lot of other interactions in our codebase, so we need to take all of that into consideration.

See the original PR:
https://github.com/the-events-calendar/the-events-calendar/pull/3963

I completely missed this on my Original Code Review.

Currently, I am commenting this out to prove the concept that this solves the problem, but I need @Camwyn to take over and get this to the finish line as he understands the original ticket a little better. cc @EugeneKyale 

[TEC-3814]: https://theeventscalendar.atlassian.net/browse/TEC-3814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Integrations/Divi/Service_Provider.php
+++ b/src/Tribe/Integrations/Divi/Service_Provider.php
@@ -41,7 +41,7 @@ class Service_Provider extends \tad_DI52_ServiceProvider {
 	 * @since 6.0.1
 	 */
 	protected function hooks() {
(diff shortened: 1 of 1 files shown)


bordoni:
Ahh ok this was solved by the latest code from @lucatume

PR #4008, opened 2022-10-05
Functions used in `apply_filters` _must_ return a value.

Camwyn:
This is to fix a bug introduced in strudel


stratease (approved):
Groovy!

PR #4009, opened 2022-10-06
Fix handling of query filters from repositories

lucatume:
Ticket: [n/a]()

Artifacts: tests

Correctly apply the JOIN query filters that might have been applied by
other plugins.

The `Custom_Tables_Query::from_wp_query` method handles the case
where the `WP_Query` instance controlled by the Custom Tables Query
has been built by one of the repositories. If so, then the method will
"transfer" the query filters originally attached to the `WP_Query`
instance built by the repository over to the `Custom_Tables_Query`
instance that is being built. This is to make sure the `JOIN`, `WHERE`
and other clause filters that would have applied to the `WP_Query`
instance will, instead, apply to the `Custom_Tables_Query` instance.

Previously, though, the code would make an exception and would instruct
the query filters to apply all filters **but** the `JOIN` ones. This to
avoid duplication of `JOIN` queries that would lead to SQL errors.

This all-or-nothing approach would cut out, though, `JOIN` clause
filters applied by the repositories schema resulting in malfunctioning
queries.

This commit updates the code to remove that limit and let the query
filters, all of them including the `JOIN` one, apply to the
`Custom_Tables_Query` instance and adds a step to deduplicate the
`JOIN` clauses before applying them.


moraleida:
Moved to a new base branch at #4010

PR #4011, opened 2022-10-07
CT1 - migrate some events while fetching the Migration UI report

lucatume:
Ticket: n/a

The Migration UI poll the backend to fetch the current migration status
HTML.
This update hooks on that report request to migrate some non migrated
Events.
This processing on the back of the XHR request to the backend will start
by migrating the not yet migrated event and will, then start migrating
Events that have been claimed by Action Scheduler actions are are
pending (not yet started).

Since this processing works on the premise of the Migration UI polling
the backend, the code does not exclude Action Scheduler and will,
instead, rely on it should the user close the tab/window/browser.

Artifacts:
* tests
* [screencast](https://share.cleanshot.com/vE0GFM)


src/Events/Custom_Tables/V1/Migration/State.php
```diff
@@ -349,4 +349,17 @@ public function save() {
...
+	 * Returns whether the current phase is a migration dry-run or not.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool Whether the current phase is a migration dry-run or not.
+	 */
+	public function is_dry_run(): bool {
+		$phase = $this->get_phase();
+
+		return $phase === self::PHASE_PREVIEW_IN_PROGRESS;
```
stratease, line 363: We may want to reference a few more states here, off the top of my head I think these are technically where we go back to a dry run.

`PHASE_REVERT_COMPLETE, PHASE_PREVIEW_PROMPT, PHASE_CANCEL_COMPLETE` - not 100% positive, going off my memory.

stratease: Since this check is in the State object, the context of `is_migrating` shouldn't be assumed

stratease (changes requested):
Awesome idea.

stratease:
Created fork for this ticket off current branch, working to move this to strudel.
https://github.com/the-events-calendar/the-events-calendar/pull/4013

PR #4012, opened 2022-10-07
Fix for fatal, assuming WP_CLI exists.

stratease:
[TEC-4470](https://theeventscalendar.atlassian.net/browse/TEC-4470)

Fixes the fatal:
``` PHP Fatal error:  Uncaught Error: Class 'WP_CLI' not found in...```

Custom plugin in the ticket that uses the `empty_custom_tables` function will now activate without error.

image

### CODE REVIEW

File: src/Events/Custom_Tables/V1/Schema_Builder/Schema_Builder.php [hunk 1]
```diff
@@ -308,7 +308,9 @@ public function empty_custom_tables() {
 		$schemas = $this->get_registered_table_schemas();
 		foreach ( $schemas as $custom_table ) {
 			/** @var Table_Schema_Interface $custom_table */
-			WP_CLI::debug( 'Emptying table ' . $custom_table::table_name(), 'TEC' );
+			if ( class_exists( WP_CLI::class ) ) {
```
bordoni, line 311:
```suggestion
			if ( class_exists( 'WP_CLI' ) ) {
```

bordoni (changes requested):

Camwyn (approved):

PR #4017, opened 2022-10-13
Create integration to solve problem of Single Event caching

bordoni:


src/Events/Custom_Tables/V1/Integrations/DotCom/Clear_Event_Cache.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @since TBD
+	 *
+	 * @param WP_Query|null           $wp_query    A reference to the `WP_Query` instance that is currently running.
+	 * @param array<WP_Post|int>|null $posts       The filter input value, it could have already be filtered by other
+	 *                                             plugins at this stage.
+	 *
+	 * @return null|array<WP_Post|int> The filtered value of the posts, injected before the query actually runs.
+	 */
+	public function filter_posts_pre_query( $posts = null, $wp_query = null ) {
+		$cache_hash = md5( serialize( $wp_query->query ) );
```
Camwyn, line 44: check for null first so this doesn't potentially throw an error?

Camwyn (changes requested):

stratease (changes requested):
`DotCom` should be `Dot_Com` right? Also, wouldn't it be better to make it the full name? Or will everyone know it's for wordpress.com ?

stratease (approved):
Just needs artifacts! Thanks

Camwyn (approved):

PR #4019, opened 2022-10-17
Order only Events by date

lucatume:
Ticket: [TEC-4470](https://theeventscalendar.atlassian.net/browse/TEC-4470)

Artifacts: tests and [screencast](https://share.cleanshot.com/tQX7Rq)

This fixes an issue where the event-date based order, based on the
`_EventStartDate` or `_EventEndDate` meta, would be erroneously applied
to all post types, not just Events.

Notably, this would result in tickets related to Events not being
correctly displayed on the front-end.

The cause of the issue was the Query code would hook the filtering
and ordering function on any query for a TEC post type. The detection,
in the Context component, of the post type is based on the **request**
variables.
So, a request to render the Event single would be a main query for a TEC
post type, any accessory query for any TEC post type (e.g. Venues or
Tickets) would then be filtered.

The fix consists in checking, per-query, if the query is for only Events
or not.


readme.txt
```diff
@@ -262,6 +262,7 @@ Remember to always make a backup of your database and files before updating!
 * Tweak - Added filters: `tec_events_today_button_label` and `tec_events_view_{view_slug}_today_button_label` to filter the output of the `tec_events_get_today_button_label` function
                          `tec_events_today_button_title` and `tec_events_{view_slug}_view_today_button_title` to filter the link title and aria-label. [TEC-4458]
 * Tweak - Speed up the Custom Tables v1 migration process if a browser window is open on the migration UI. [TEC-4517]
+* Fix - Co
```
Camwyn, line 265: ?

Camwyn (changes requested):
Weird stub of a readme entry, otherwise 🏁

readme.txt
lucatume: No idea, I blame my cat.

Camwyn (approved):

bordoni (approved):

PR #4024, opened 2022-10-25
Update post type model code

lucatume:
Ticket: [TEC-4379](https://theeventscalendar.atlassian.net/browse/TEC-4379)

Based on [Common PR](https://github.com/the-events-calendar/tribe-common/pull/1821) 

Artifacts: tests and [screencast](https://share.cleanshot.com/2ktUfb) 

This commit updates the code used to build and return decorated post
models for Events, Venues and Organizers when dealing with object
caching.

The update follows the one done in common in the branch by the same name
to avoid serialization and unserialization issues.

### DIFF
--- a/src/Tribe/Models/Post_Types/Event.php
+++ b/src/Tribe/Models/Post_Types/Event.php
@@ -14,6 +14,8 @@ use DatePeriod;
 use DateTimeZone;
 use Tribe\Events\Collections\Lazy_Post_Collection;
 use Tribe\Models\Post_Types\Base;
+use Tribe\Utils\Date_I18n;
+use Tribe\Utils\Date_I18n_Immutable;
 use Tribe\Utils\Lazy_Boolean;
 use Tribe\Utils\Lazy_Collection;
 use Tribe\Utils\Lazy_String;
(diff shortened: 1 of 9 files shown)


src/functions/template-tags/event.php
```diff
@@ -137,28 +136,12 @@ function tribe_get_event( $event = null, $output = OBJECT, $filter = 'raw', $for
...
 		$cache_key = 'tribe_get_event_' . md5( wp_json_encode( $key_fields ) );
 
-		if ( ! $force ) {
-			$initial_unserialize_callback = ini_get( 'unserialize_callback_func' );
-			// Prevent warning from happening.
-			ini_set( 'unserialize_callback_func', '__return_false' );
-
-			$post = $cache->get( $cache_key, Tribe__Cache_Listener::TRIGGER_SAVE_POST );
+		// Try getting the memoized value.
+		$post = $cache[ $cache_key ];
```
stratease, line 140: Do we need a ?
```
$post = $cache[ $cache_key ] ?? false;
```

src/functions/template-tags/venue.php
```diff
@@ -57,20 +57,17 @@ function tribe_get_venue_object( $venue = null, $output = OBJECT, $filter = 'raw
...
-		/** @var Tribe__Cache $cache */
-		$cache = tribe( 'cache' );
-		$post  = $cache->get( $cache_key, Tribe__Cache_Listener::TRIGGER_SAVE_POST );
-	}
+	/** @var Tribe__Cache $cache */
+	$cache = tribe( 'cache' );
+	$cache_key = 'tribe_get_venue_object_' . md5( json_encode( [ $venue, $output, $filter ] ) );
+
+	// Try getting the memoized value.
+	$post = $cache[ $cache_key ];
```
stratease, line 65: Do we need a isset check?

```$post = $cache[ $cache_key ] ?? false;```

File: src/functions/template-tags/event.php (hunk 1 above)
Comment (lucatume) [reply src/functions/template-tags/event.php:140]: No, the `Tribe__Cache` object implements the `ArrayAccess` interface, and the `offsetGet` method will call the `Tribe__Cache::get` method that already embeds that check. If the key is not set, it will return `false`. 
The `ArrayAccess` API is the one used to quickly access and set the memoized values.

File: src/functions/template-tags/venue.php (hunk 2 above)
Comment (lucatume) [reply src/functions/template-tags/venue.php:65]: Same as above.

File: src/Tribe/Models/Post_Types/Event.php [hunk 3]
```diff
@@ -333,4 +335,59 @@ public static function prime_cache( array $events = [] ) {
...
+
+			if ( $date instanceof \DateTime ) {
+				return new Date_I18n( $date->format( 'Y-m-d H:i:s' ), $date->getTimezone() );
+			}
+
+			return $date;
+		}, $properties['dates'] ?? [] );
+
+		return $properties;
+	}
```
bordoni, line 392: Where are these two beign used?

bordoni (approved):

src/Tribe/Models/Post_Types/Event.php
lucatume: Left over from another iteration, no longer required, nice catch.

lucatume: They are used in the Common `Base` during serialization for the purpose of object caching, and during unserialization from object caching.

PR #4025, opened 2022-10-25
Move the filtering and translation of view labels to the View class.

Camwyn:
In the View Manager class, we have been passing the label and text domain as variables. `__( $text, $domain )`

This does not work.
See: http://ottopress.com/2012/internationalization-youre-probably-doing-it-wrong/

This changeset moves that handling to the View class itself - and deprecates the offending methods and filters. This only requires the (non-abstract) subclasses (specific views) have a construct method that defines the label in the standard `__( string, string )` manner.

**Exceptions/TODO:**
 - Latest Past  - as I believe that inherits it's label from the original view.

**Tests:** none yet. Will return to this shortly.
**Checks:** I have generated a test .pot file and confirmed the missing strings are available for translation. (not included in the PR)

**Required by:** https://github.com/the-events-calendar/events-pro/pull/2104

[TEC-4485]

[TEC-4485]: https://theeventscalendar.atlassian.net/browse/TEC-4485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lucatume (dismissed):
This is a good improvement. I _think_ snapshot tests will need updating; make sure the current snapshot tests already cover the code in charge of the View slugs.

Edit: I wanted to leave this as Comment, not an approval, fat fingers.

lucatume (dismissed):
I've run again the tests that were failing due to unrelated issues, all are passing and this should be good to go.
Some tests will show up as if they did not run, I manually ran the workflows in GH Actions and all are passing.

src/Tribe/Views/V2/View.php
```diff
@@ -663,11 +667,59 @@ public function get_html() {
...
+	 * Returns the view label value after filtering.
+	 *
+	 * This is the method you want to overwrite to replace the label for a view with translations.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public static function get_view_label(): string {
+		return static::filter_view_label( static::$label );
```
stratease, line 680:
```suggestion
		return (string)static::filter_view_label( static::$label );
```
Probably less a problem here, but in general I'd do this on these type cast functions that rely on output of filters. People can "accidentally" null it or something that is going to fatal here.

stratease: Actually this is probably more relevant in the `filter_view_label` function, but you get what I mean.

stratease (changes requested):
Some type casting safety nets.

src/Tribe/Views/V2/View.php
```diff
@@ -663,11 +667,59 @@ public function get_html() {
...
+		$label = apply_filters( 'tec_events_views_v2_view_label', $label, $slug );
+
+		/**
+		 * Filters the label that will be used on the UI for views listing.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $label        Label of the Current view.
+		 */
+		return apply_filters( "tec_events_views_v2_{$slug}_view_label", $label );
```
Camwyn, line 715:
```suggestion
		return (string) apply_filters( "tec_events_views_v2_{$slug}_view_label", $label );
```

src/Tribe/Views/V2/Views/Day_View.php
```diff
@@ -46,6 +47,24 @@ class Day_View extends View {
...
+	 */
+	protected static $label = 'Day';
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function get_view_label(): string {
+		static::$label = _x( 'Day', 'The text label for the Day View.', 'the-events-calendar' );
+
+		return static::filter_view_label( static::$label );
```
Camwyn, line 65:
```suggestion
		return (string) static::filter_view_label( static::$label );
```

src/Tribe/Views/V2/Views/List_View.php
```diff
@@ -44,6 +45,24 @@ class List_View extends View {
...
+	 */
+	protected static $label = 'List';
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function get_view_label(): string {
+		static::$label = _x( 'List', 'The text label for the List View.', 'the-events-calendar' );
+
+		return static::filter_view_label( static::$label );
```
Camwyn, line 63:
```suggestion
		return (string) static::filter_view_label( static::$label );
```

src/Tribe/Views/V2/View.php
Camwyn: See: https://github.com/the-events-calendar/the-events-calendar/pull/4025/files#diff-b5ac6d676b19cb6b889de2de40d22903ba5bd8d989292eac14d03900e64560e4R715

stratease (approved):

PR #4030, opened 2022-11-01
Fix CT1 Builder::first method type check

lucatume:
Ticket: [ECP-1402](https://theeventscalendar.atlassian.net/browse/ECP-1402)

Artifacts:
* the code is already covered by tests in TEC
* an additional test has been [added to ECP to cover this](https://github.com/the-events-calendar/events-pro/pull/2109)

This adds type-checking to the `Builder::first` method to make sure the returned type will be either the Model one or `null`.


src/Events/Custom_Tables/V1/Models/Builder.php
```diff
@@ -762,11 +762,22 @@ public function find_all( $value, $column = null ) {
...
 
-		return reset( $results );
+		$result = reset( $results );
+
+		switch ( $this->output_format ) {
+			case OBJECT:
+			default:
+				return $result instanceof $this->model ? $result : null;
+			case ARRAY_N:
+				return is_array( $result ) ? array_values( $result ) : null;
```
stratease, line 777: Should update return doc too, like:
```
* @return Model|null|array
```

stratease (changes requested):
Minor tweak.

bordoni (dismissed):

stratease (dismissed):

lucatume:
I've updated this PR to make sure the Custom Tables v1 queries will play nice with WordPress 6.1 caching of queries

stratease (dismissed):

PR #4033, opened 2022-11-02
Fix issue where early queries can interfere with template redirects. | TEC-4554

Camwyn:
A query run on 'init' that simply contains the suppress_filters param will break the view passed via context and show the wrong view on a refresh: https://d.pr/i/jKaJqR or a direct hit to any view other than the default view.

This changes the param we ask tribe_context for to ensure we're getting the value expected: https://d.pr/i/4a7Mqs

PR is set to tiramisu as I think this should go out as soon as possible. WP 6.1 makes this a priority

https://d.pr/v/67Za8k

[BTRIA-1474]

[BTRIA-1474]: https://theeventscalendar.atlassian.net/browse/BTRIA-1474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stratease (approved):

lucatume (approved):
The fix looks ok, but I would like to ask you to run the test suites that were interrupted and post a result from them.
Alternatively, to have a smooth run, re-generate the currently not passing snapshots using `slic` and WordPress 6.1 to bring them in line with CI.

borkweb (dismissed):

borkweb:
This fix is going to go into the hotfix. I cherry-picked the commits into a new PR: #4036

I'll close this one.

PR #4037, opened 2022-11-04
Prevent the caching of occurrence queries

borkweb:
We do our own caching of occurrence results, so we don't need WordPress core to do it.


borkweb:
I'm running the integration tests manually since we are being interrupted by the date issue with restv1.

borkweb:
Alright, the only failures are the 2 time-based integration tests that need to be rewritten so timezones don't impact them. Merging.

borkweb:
Gustavo reviewed this on a call.

PR #4048, opened 2022-11-16
Fix/ecp 1372 non series events bleed

stratease:
[ECP-1372](https://theeventscalendar.atlassian.net/browse/ECP-1372)

This is a [dependency for this PR](https://github.com/the-events-calendar/events-pro/pull/2121).

Adds a new filter to allow global filters.

Artifacts:

See tests.


bordoni:
I've moved this PR from `release/B22.tiramisu` to `master`, please review and point it to the correct branch. The next sprint will live in `release/B22.wagashi`.

src/Tribe/Views/V2/View.php
```diff
@@ -2562,6 +2576,31 @@ protected function get_repository_args() {
...
+			 * Will filter any repository args to be applied globally on the various repository queries on this view.
+			 *
+			 * @since TBD
+			 *
+			 * @param array<string,mixed> Events Repository args that will be applied globally to all event repository queries.
+			 * @param View $this The View object being rendered.
+			 *
+			 * @return array<string,mixed> The repository args to be applied.
+			 */
+			$this->global_repository_args = apply_filters( 'tec_events_views_v2_view_global_repository_args', [], $this );
```
lucatume, line 2598: The usual practice, in instances like this one, is to have 2 filters: a generic one (the one already there it's fine) and a specific one for the view slug.
You will see this pattern applied elsewhere in the Views V2 implementation.
I would add a `tec_events_views_v2_{$this->slug}_view_global_repository_args` filter.

src/Tribe/Views/V2/View.php
```diff
@@ -1267,6 +1278,9 @@ protected function setup_repository_args( Context $context = null ) {
 			'view_override_offset' => true,
 		];
 
+		// Merge our global repository args with the primary event args.
+		$args = array_merge($args, $this->get_global_repository_args());
```
lucatume, line 1282:
```suggestion
		$args = array_merge( $args, $this->get_global_repository_args() );
```

The override order, where global repository arguments would override the View ones, seems odd; I think it should be the opposite, where global args are overridden by the View-specific ones.

src/Events/Custom_Tables/V1/WP_Query/Modifiers/Events_Series_Relationship_Modifier.php
```diff
@@ -23,24 +22,7 @@
...
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function unhook() {
-		remove_filter( 'posts_join', [ $this, 'join_on_series_relationships_table' ] );
-		remove_filter( 'posts_where', [ $this, 'where_event_is_related_to_series' ] );
-	}
+class Events_Series_Relationship_Modifier extends Occurrences_Series_Relationship_Modifier {
```
lucatume, line 25: Should this class be here at all? Series relationships and its effects are not a concern that should be handled in TEC at all.
It was erroneously here before, but this looks like an excellent chance to move it to ECP for good.

lucatume (changes requested):

src/Events/Custom_Tables/V1/WP_Query/Monitors/Query_Monitor.php
```diff
@@ -77,6 +86,37 @@ public function __construct( tad_DI52_Container $container = null ) {
...
+	 * Will filter and retrieve the list of WP_Query_Modifier implementations. Any implementation filters
+	 * must be applied before the init hook is completed.
+	 *
+	 * @since TBD
+	 *
+	 * @return array<WP_Query_Modifier> List of WP_Query_Modifier implementations.
+	 */
+	public function get_implementations(): array {
+		// Keep running filter until init is finished.
+		if ( did_action( 'init' ) > 0 || empty( $this->implementations ) ) {
```
lucatume, line 99:
```suggestion
		if ( doing_action( 'init' ) || ! did_action( 'init' ) ) {
```
Do no check for empty values: filtering code might have emptied the implementations for some legit reasons, just keep filtering until the `init` action has not been done.
Check for `doing_action ( 'init' )` as `did_action( 'init' )` will return `1` if the current action being dispatched is the `init` one, the value that will be read  by `did_action` is set before callbacks are applied.
Finally keep filtering if the `init` action has not been done yet.

src/Events/Custom_Tables/V1/WP_Query/Monitors/Query_Monitor.php
```diff
@@ -77,6 +86,37 @@ public function __construct( tad_DI52_Container $container = null ) {
...
+			/**
+			 * Filters the Query Modifier implementations that will be used in the Query Monitor parsing.
+			 *
+			 * @since TBD
+			 *
+			 * @param array<WP_Query_Modifier>  The list of Query Modifier implementations that will be used in the
+			 *                                  query parsing.
+			 * @param Query_monitor This instance of the Query Monitor retrieving implementations.
+			 *
+			 * @return array<WP_Query_Modifier> The filtered list of Query Modifier implementations.
```
lucatume, line 109: Filter doc-block should not include the `@return` tag.

src/Events/Custom_Tables/V1/WP_Query/Monitors/WP_Query_Monitor.php
```diff
@@ -26,20 +25,6 @@
...
 	use Query_Monitor;
 
-	/**
-	 * A list of possible modifiers implementations.
-	 *
-	 * @since 6.0.0
-	 *
-	 * @var array<string>
-	 */
-	private $implementations = [
```
lucatume, line 36: I would not remove this property from the class, just remove the ECP entries. See my comment in the trait.

src/Events/Custom_Tables/V1/WP_Query/Monitors/Query_Monitor.php
```diff
@@ -25,6 +25,15 @@
...
 	use With_WP_Query_Introspection;
 
+	/**
+	 * A list of possible modifiers implementations.
+	 *
+	 * @since TBD
+	 *
+	 * @var array<string>
+	 */
+	private $implementations = [];
```
lucatume, line 35: I would not define this property in the trait. It forces complexity, in the form of classes having to filter to add their own implementations, that is not required in the using classes.

src/Events/Custom_Tables/V1/WP_Query/Provider.php
```diff
@@ -48,15 +51,41 @@ public function register() {
 			add_action( 'tribe_repository_events_init', [ $this, 'replace_repository_query_filters' ] );
 		}
 
+		//
+		if ( ! has_filter( 'tec_events_custom_tables_v1_query_modifier_implementations', [
```
lucatume, line 55: See my comment in the trait and classes using it: this adds complexity that is not required. The `WP_Query_Monitor` and `Custom_Tables_Query` monitor in ECP are the base, and it's fine for them to define the base implementation set.

src/Events/Custom_Tables/V1/WP_Query/WP_Query_Monitor_Filters.php
```diff
@@ -0,0 +1,45 @@
...
+use TEC\Events\Custom_Tables\V1\WP_Query\Monitors\WP_Query_Monitor;
+
+/**
+ * Class WP_Query_Monitor_Filters
+ *
+ * @since   TBD
+ *
+ * @package TEC\Events\Custom_Tables\V1\WP_Query
+ */
+class WP_Query_Monitor_Filters {
```
lucatume, line 26: This adds complexity, I would remove it, see my comments above.

lucatume (changes requested):

src/Events/Custom_Tables/V1/WP_Query/Monitors/Query_Monitor.php
stratease: Fair point. I would have put it in an abstract class, but the refactor is ballooning. I'll leave it undefined for now, as an optional definition. 

I think typically we should have a base/abstract class these systems should leverage. Helps in scenarios like this, where it's an optional definition.

lucatume: If you define the property in the trait, you cannot override it in the classes using it.
If you think an abstract class would be better, go for it.

src/Events/Custom_Tables/V1/WP_Query/Monitors/Query_Monitor.php
stratease: @lucatume I had similar to that originally but that logic failed because it was not run until after `init` - so it would never pass those conditions, hence the empty check e.g. "did it run yet?". 

I can check for null vs an array instead of empty? i.e. null means run it once at least, empty array skips.

lucatume (approved):

PR #4056, opened 2022-11-19
Avoid a Join for Hiding from Upcoming which was happening on every request.

bordoni:
`EventHideFromUpcoming` is a big problem for performance since it is a Join that we are doing for all of our requests on views, but it most of the time contains no items. Here I am introducing a caching of which posts are marked for hiding and then including a `post__not_in` which is much more performant.

| Before | After |
| - | - |
| image | image |


src/Tribe/Repositories/Event.php
```diff
@@ -1781,14 +1781,19 @@ public function order_by( $order_by, $order = 'ASC' ) {
 	 */
 	public function filter_by_hidden_on_upcoming( $hidden ) {
 		$hidden = tribe_is_truthy( $hidden );
+		$hidden_posts = tribe( \Tribe\Events\Views\V2\Query\Hide_From_Upcoming_Controller::class )->get_hidden_post_ids();
```
stratease, line 1784: I am a lil worried about hitting some edge case limits on this. 

Any issues/concerns about grabbing all? Will this hit a wall at like 1000 hidden posts etc?

stratease (commented):
Just a general question / push back on the unbound query.

src/Tribe/Repositories/Event.php
bordoni: Potentially, I don't think you are wrong that this could have an upper limit problem.

Wonder if we should have a filter that allows someone to disable this and move it back to the normal query.

stratease: Yeah, I've hit several issues with similar queries. Sometimes queries get so big they will fail because of the size of the query text itself.

I guess considering time just doing a flag to use the old is an option, but it will not be easy for anyone to know when to use it.

Any thoughts on creating a temporary table to join on? Never done something like that in a WP world.

Alternatively we move this to a custom table to help optimize it, but that's a bigger project. Maybe this is the long term goal, a set of more optimizations in the custom table move.

stratease: Other thing I wanted to point out, in the WP hooks somewhere someone may look at the `post__in` field on the query args array and iterate on some large list.

stratease (approved):
Going to approve this, I think the direction this should go is a bigger project and might deserve scoping out later. Perhaps bringing over several meta fields into Events table etc.... Some things worth an engineering discussion.

bordoni:
Thanks @stratease I 100% agree with you.

PR #4057, opened 2022-11-20
Fix for call function on null object.

stratease:
[TEC-4566](https://theeventscalendar.atlassian.net/browse/TEC-4566)

Check if global is set before attempting to retrieving value from it.

Artifacts:
Before fix applied when loading event (in certain scenarios) page:
image
After fix applied when loading event page:
image


Camwyn:
@stratease - looks good, fix the tests?

stratease:
@Camwyn was a false negative, just re-ran.

Camwyn (approved):

PR #4058, opened 2022-11-22
Ensure the `Prev` button when using the `Event View` Elementor widget…

EugeneKyale:
Ticket: [FBAR-273](https://theeventscalendar.atlassian.net/browse/FBAR-273)

This PR ensures that the `prev` button when using the `Event View` widget works as expected. 

Artifact:
https://www.loom.com/share/e6a87441906d49ffad97e23fd9157556

### DIFF
--- a/src/resources/js/views/manager.js
+++ b/src/resources/js/views/manager.js
@@ -310,7 +310,7 @@ tribe.events.views.manager = {};
 
 		var $link = $( this );
 		var url = $link.attr( 'href' );
-		var currentUrl = containerData.url;
+		var currentUrl = containerData.prev_url;
 		var nonce = $link.data( 'view-rest-nonce' );
 		var shouldManageUrl = obj.shouldManageUrl( $container );
 		var shortcodeId = $container.data( 'view-shortcode' );


src/resources/js/views/manager.js
```diff
@@ -310,7 +310,7 @@ tribe.events.views.manager = {};
 
 		var $link = $( this );
 		var url = $link.attr( 'href' );
-		var currentUrl = containerData.url;
+		var currentUrl = containerData.prev_url;
```
Camwyn, line 313: This makes sense - but for clarity's sake can we change the var name?

```suggestion
		var prevUrl = containerData.prev_url;
```
(and on line 324 where we consume it)

Camwyn (changes requested):

src/resources/js/views/manager.js
EugeneKyale: @Camwyn 🫡 I've made the update.

Camwyn (approved):

bordoni:
We need to be super careful with these changes.
I would like this to be reverted and tested in Branch.

The URL management of Views needs to be tested in quite a few scenarios before we merge this change safely.

EugeneKyale:
> We need to be super careful with these changes. I would like this to be reverted and tested in Branch.
> 
> The URL management of Views needs to be tested in quite a few scenarios before we merge this change safely.

@bordoni do you mind reverting the merge? I don't the permissions required to update the release branch directly. FYI, I've restored the branch so QA can test in-branch.

PR #4069, opened 2022-11-29
Changing UTC offsets to convert to localized timezone.

stratease:
[TEC-4578](https://theeventscalendar.atlassian.net/browse/TEC-4578)

This PR fixes two separate issues that caused 404's. The core issue is that CT1 table data is not being created, the reasons are completely unrelated.

1) We were creating UTC start/end dates different than legacy in our validators. This would fail to insert any data due to this inconsistency. Now using a consistent approach with how we handle offsets in Legacy and CT1 interpretation. Now DST will be respected when the offset is converted to the localized timezone.
2) There appears to have been some database "corruption" in some clients installations. Basically the foreign key association made in our occurrence table would point to a "missing" table, failing all insertions. Not sure how this happens, but we are removing the foreign key in this version. This will simplify support moving forward, plus is not necessary with our table cleanup.
3) Adding some logging. Both errors were not being logged even with `WP_DEBUG` flagged. Now these failure points will log on failure.

https://www.loom.com/share/eacf772113c644ef8d99b0b8152f085f

### DIFF
--- a/src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Field.php
+++ b/src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Field.php
@@ -49,9 +49,8 @@ abstract class Abstract_Custom_Field implements Field_Schema_Interface {
 	 * immediately before the field creation or update.
 	 *
 	 * @since TBD
-	 *
 	 */
-	protected function before_update() {
+	protected function before_update() :void {
 	}
 
 	/**
--- a/src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Field.php
+++ b/src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Field.php
@@ -24,8 +24,8 @@ abstract class Abstract_Custom_Field implements Field_Schema_Interface {
 	 * {@inheritdoc}
 	 */
 	public function update() {
+		$this->before_update();
 		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
-
 		$results = (array) dbDelta( $this->get_update_sql() );
 		$this->sync_stored_version();
 		$results = $this->after_update( $results );
@@ -43,6 +43,16 @@ abstract class Abstract_Custom_Field implements Field_Schema_Interface {
 	 *                by the `dbDelta` function.
 	 */
 	abstract protected function get_update_sql();
+
+	/**
+	 * Allows extending classes that require it to run some methods
+	 * immediately before the field creation or update.
+	 *
+	 * @since TBD
+	 *
+	 */
+	protected function before_update() {
+	}
 
 	/**
 	 * Allows extending classes that require it to run some methods
--- a/src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Table.php
+++ b/src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Table.php
@@ -78,9 +78,8 @@ abstract class Abstract_Custom_Table implements Table_Schema_Interface {
 	 * immediately before the table creation or update.
 	 *
 	 * @since TBD
-	 *
 	 */
-	protected function before_update() {
+	protected function before_update() :void {
 	}
 
 	/**
--- a/src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Table.php
+++ b/src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Table.php
@@ -163,7 +163,7 @@ abstract class Abstract_Custom_Table implements Table_Schema_Interface {
 	 * @param string $this_field The field of the table that has the foreign key (not the target of the constraint).
 	 * @param string $this_table The table that has the foreign key (not the target of the constraint).
 	 *
-	 * @return stdClass|null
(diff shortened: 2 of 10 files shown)


src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Field.php
```diff
@@ -44,6 +44,16 @@ public function update() {
...
 	abstract protected function get_update_sql();
 
+	/**
+	 * Allows extending classes that require it to run some methods
+	 * immediately before the field creation or update.
+	 *
+	 * @since TBD
+	 *
+	 */
+	protected function before_update() {
```
lucatume, line 54:
```suggestion
	protected function before_update() :void {
```

src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Field.php
```diff
@@ -44,6 +44,16 @@ public function update() {
 	 */
 	abstract protected function get_update_sql();
 
+	/**
+	 * Allows extending classes that require it to run some methods
+	 * immediately before the field creation or update.
+	 *
+	 * @since TBD
+	 *
```
lucatume, line 52:
```suggestion
```

src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Table.php
```diff
@@ -70,6 +73,16 @@ public function update() {
 	 */
 	abstract protected function get_update_sql();
 
+	/**
+	 * Allows extending classes that require it to run some methods
+	 * immediately before the table creation or update.
+	 *
+	 * @since TBD
+	 *
```
lucatume, line 81:
```suggestion
```

src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Table.php
```diff
@@ -70,6 +73,16 @@ public function update() {
...
+	 */
+	protected function before_update() {
```
lucatume, line 83:
```suggestion
	protected function before_update() :void {
```

src/Events/Custom_Tables/V1/Schema_Builder/Abstract_Custom_Table.php
```diff
@@ -129,6 +142,49 @@ protected function has_index( $index, $table_name = null ) {
...
+
+	/**
+	 * Fetches the constraint for a particular field.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $this_field The field of the table that has the foreign key (not the target of the constraint).
+	 * @param string $this_table The table that has the foreign key (not the target of the constraint).
+	 *
+	 * @return stdClass|null
```
lucatume, line 167: Add return value description.

src/Tribe/API.php
```diff
@@ -454,11 +454,25 @@ protected static function prepare_event_date_meta( $event_id, $data ) {
 
 			// Additionally store datetimes in UTC
 			if ( empty( $data['EventStartDateUTC'] ) ) {
-				$data['EventStartDateUTC'] = Tribe__Events__Timezones::to_utc( $data['EventStartDate'], $data['EventTimezone'] );
+				if ( Tribe__Events__Timezones::is_utc_offset( $data['EventTimezone'] ) ) {
```
lucatume, line 457: Collect the result of `Tribe__Events__Timezones::is_utc_offset( $data['EventTimezone'] )` in a variable to avoid running the same logic twice for a small gain.

src/Tribe/API.php
```diff
@@ -454,11 +454,25 @@ protected static function prepare_event_date_meta( $event_id, $data ) {
...
+					// Convert a UTC offset timezone to a localized timezone, e.g. UTC-5 to America/New_York.
+					$data['EventStartDateUTC'] = Tribe__Date_Utils::immutable( $data['EventStartDate'], $data['EventTimezone'] )
+					                                              ->setTimezone( new DateTimeZone( 'UTC' ) )
```
lucatume, line 460: Collect the result of `new DateTimeZone( 'UTC' )` in a variable to avoid building it twice.

src/Tribe/API.php
```diff
@@ -454,11 +454,25 @@ protected static function prepare_event_date_meta( $event_id, $data ) {
...
+					                                              ->format( Tribe__Date_Utils::DBDATETIMEFORMAT );
+				} else {
+					$data['EventStartDateUTC'] = Tribe__Events__Timezones::to_utc( $data['EventStartDate'], $data['EventTimezone'] );
+				}
 			}
 
 			if ( empty( $data['EventEndDateUTC'] ) ) {
-				$data['EventEndDateUTC']   = Tribe__Events__Timezones::to_utc( $data['EventEndDate'], $data['EventTimezone'] );
+				if ( Tribe__Events__Timezones::is_utc_offset( $data['EventTimezone'] ) ) {
```
lucatume, line 468: These changes to the `Tribe__API.php` class will require test coverage; this can be in the form of existing tests requiring update.

lucatume (changes requested):
Smaller doc changes. 
Add/update the failing tests, and add/update the `Tribe__Events__API` missing ones.

lucatume (approved):

lucatume:
The `rest_v1` suite is reported passing in the action, but shows up not passed here. Magic.

PR #4071, opened 2022-12-01
ECP-1319 - prevent serialize notice during migration in PHP 8.1

jesseeproductions:
When on PHP 8.1 if you run the migration you can trigger a notice that stops the migration:

"[TECEventsCustom_TablesV1ModelsEvent implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)]"

This PR fixes that issue and retains backwards compatibility with PHP.

_Ref:_ [ECP-1319](https://theeventscalendar.atlassian.net/browse/ECP-1319)

Artifacts: None


Camwyn (commented):
What's up with the extra readme entry?

jesseeproductions:
> What's up with the extra readme entry?

It showed up when I merged. I removed it.

bordoni (approved):

PR #4077, opened 2022-12-06
Unable to update Events that have Venues assigned when Pro is NOT active. [TEC-4596]

EugeneKyale:
Ticket : [TEC-4596](https://theeventscalendar.atlassian.net/browse/TEC-4596)

**Context ❓** 
When pro is NOT active, events that already had a venue assigned to them could not be updated because the venue website input expected a URL but instead got a string with the venue name based on this [conditional statement](https://github.com/the-events-calendar/the-events-calendar/blob/90f7470fa73a5303e0cd62679d1e3936199d4d59/src/functions/template-tags/venue.php#L266).

Basically, input validation was failing.

**The Fix 👇🏽** 
This update ensures we're fetching the Venue website URL and not the link to the Single Venue page as we'd previously done using `tribe_get_venue_link()`.

**Artifact 🎥**

https://www.loom.com/share/dcb01c5100ef4b3ca8be25e86cb950a3


Camwyn (dismissed):
This looks good.
Sanity check - we don't have the same issue with the Organizer URL?

EugeneKyale:
> This looks good. Sanity check - we don't have the same issue with the Organizer URL?

@Camwyn just checked, the Organizer URL looks good.

Camwyn (approved):

PR #4079, opened 2022-12-07
Fix more PHP 8.1 deprecation issues and PHP 7.4 fatal

lucatume:
Ticket: [ECP-1319](https://theeventscalendar.atlassian.net/browse/ECP-1319)

[Screencast](https://share.cleanshot.com/jSiAGj)

This updates the original fix code to take care of a fatal on PHP 7.4 and remove more deprecated notices.

Note: when the class is updated to use the `__serialize` and `__unserialize` methods, then those will be used in place of the previously implemented `serialize` and `unserialize` ones; this means existing tests are already covering the changes.

### DIFF
--- a/src/Events/Custom_Tables/V1/WP_Query/Monitors/Query_Monitor.php
+++ b/src/Events/Custom_Tables/V1/WP_Query/Monitors/Query_Monitor.php
@@ -208,7 +208,8 @@ trait Query_Monitor {
 	/**
 	 * Return the number of Queries to which at least one modifier is attached.
 	 *
-	 * @since 6.0.0 *
+	 * @since 6.0.0 
+	 *
 	 * @return int The number of modifier instances.
 	 */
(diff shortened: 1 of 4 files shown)


src/Events/Custom_Tables/V1/WP_Query/Monitors/Query_Monitor.php
```diff
@@ -208,10 +208,10 @@ private function drop_modifiers( WP_Query $query ) {
 	/**
 	 * Return the number of Queries to which at least one modifier is attached.
 	 *
-	 * @since 6.0.0
-	 *
+	 * @since 6.0.0 *
```
stratease, line 211:
```suggestion
	 * @since 6.0.0 
	 *
```

stratease (changes requested):

Camwyn (approved):

PR #4081, opened 2022-12-08
Fix/tec 4548 deadlock in migration

stratease:
[TEC-4548](https://theeventscalendar.atlassian.net/browse/TEC-4548)

This adds some preventative measures for Deadlocks. The `INSERT > SELECT` lock events for migration query would do table level locks, and in combination with transactions this would sporadically cause a Deadlock error.

Added some logic to retry the lock + fetch queries to help avoid stuck migrations.

Artifacts:
Debug logs should not have Deadlocks during migration.
image

Existing tests provide coverage.

[TEC-4548]: https://theeventscalendar.atlassian.net/browse/TEC-4548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TEC-4548]: https://theeventscalendar.atlassian.net/browse/TEC-4548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Events/Custom_Tables/V1/Migration/Events.php
```diff
@@ -51,6 +51,9 @@ public function get_ids_to_process( $limit ) {
 		// Batch locking
 		$batch_uid = uniqid( 'tec_ct1_action', true ); // Should be pretty unique.
 
+		// Let's avoid table locks in the following query, this could have Deadlock side effects.
+		$wpdb->query("SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED");
```
lucatume, line 55: This query, as any query, might fail.
I'm not sure what should be done if that is the case, and not collecting and looking at the return value here looks right to me, but I just want to make sure this is a conscious choice.

Also: add a `@see https://dev.mysql.com/doc/refman/5.7/en/set-transaction.html` to know what this is doing.

stratease: Yeah, not sure we want to do anything else here. Don't think we'd want to fail out of this function - the only other option I can think of isn't a great one, but if this fails than to modify the following query to sidestep this potential lock. Also as far as I saw this syntax is valid for many versions of MySQL.

If you have other ideas let me know.

lucatume: No, fine as it is.

src/Events/Custom_Tables/V1/Migration/Events.php
```diff
@@ -83,19 +87,35 @@ public function get_ids_to_process( $limit ) {
...
+		$wpdb->suppress_errors  = true;
 		$wpdb->query( $lock_query );
 		$wpdb->suppress_errors = $suppress_errors_backup;
 
+		// Get our db object so we can inspect. This isn't always an object, so some type checking is needed.
+		$db = is_object( $wpdb->dbh ) ? $wpdb->dbh : null;
+
+		// Deadlock error no.
+		$deadlock_errno = 1213;
+		if ( $db !== null && $db->errno === $deadlock_errno ) {
```
lucatume, line 99: Using the `query` filter as an action, you can mock the `errno` property of the `$wpdb->dbh` object using the `uopz_set_property` function (and add its clean up to the `uopz` dedicated trait). Testing this method, then, would be fairly easy and would cover a function that would be otherwise difficult covering with tests in another way.

stratease: Not as easy as you thought - apparently `mysqli` is somewhat protected, so you can't modify it directly. But got a workaround w/ uopz in place.

lucatume:
The `views_integration` suite tests passed in manual re-run.

lucatume (approved):

lucatume:
The `views_integration` suite tests passed after running them manually from the Actions menu, not reported here.

lucatume (approved):

Camwyn (approved):

PR #4085, opened 2022-12-13
Fix/ecp 1423 relative utc times

stratease:
[ECP-1423](https://theeventscalendar.atlassian.net/browse/ECP-1423)
This does not seem to be related to condense events, but the core relative search mechanism. This will fix relative date searching where it is defaulting to UTC timezone searches.
   
These particular filters would always evaluate to UTC dates, even though they may be targeting the non-UTC date fields.

Making these event filters UTC sensitive over the specific global setting determining which are used.

Adding date search tests for relative date searching, with a focus on validating the UTC period these were failing in.

Artifact: https://www.loom.com/share/ab5f7d7e04ca4026850323aeaba061c2


src/Tribe/Repositories/Event.php
```diff
@@ -263,10 +263,13 @@ public function filter_by_starts_before( $datetime, $timezone = null ) {
 		$date = Tribe__Date_Utils::build_date_object( $datetime, $timezone )
 		                         ->setTimezone( $this->normal_timezone );
 
+		// If this is a UTC date, use our UTC field, else whatever was specified.
+		$key = $date->getTimezone()->getName() === 'UTC' ? '_EventStartDateUTC' : $this->start_meta_key;
```
Camwyn, line 267: We do this 8+ times. In most instances we are setting the timezone just before. Does it make sense to just check `$this->normal_timezone` instead?

Feel like I'm missing something here...

stratease: I guess most of them could be replaced with that... half dozen of one, six of the other. Not sure there's really a big difference, except for a couple places it was made more complicated needlessly.

lucatume (approved):

PR #4092, opened 2022-12-22
refactor(Adjacent_Events) cache closest event results

lucatume:
Ticket: [BTRIA-1570](https://theeventscalendar.atlassian.net/browse/BTRIA-1570) 

Artifacts: tests

This updates the code in the `Tribe__Events__Adjacent_Events::get_closest_event` method to cache the results and avoid running a slow query more than once.
The cached result will either be an `int` or `null`, to avoid issues with cache pre-fetching.
The results are cached for either a week or until an event update happens.


src/Tribe/Adjacent_Events.php
```diff
@@ -189,74 +192,89 @@ public function get_closest_event_where( $where_sql ) {
...
+		// The cached value will be the post ID, or `null`, to avoid pre-fetch issues.
+		$cached = $cache->get( $cache_key, Cache_Listener::TRIGGER_SAVE_POST, false );
+		$event = $cached;
+		if ( ! empty( $cached ) ) {
+			// If not empty, it should be a valid event post ID.
+			$event = get_post( $cached );
+			if ( ! ( $event instanceof WP_Post && $event->post_type === TEC::POSTTYPE ) ) {
+				$event = false;
+			}
+		} elseif ( $cached !== null ) {
```
stratease, line 217:
```
		} elseif ( $cached !== null ) {
```
After the 
```
                if ( ! empty( $cached ) ) {
```
seems odd, it is targeting similar things but I can see a hole in the logic if it is `false` or something else. Is this intentional?

stratease (approved):

src/Tribe/Adjacent_Events.php
lucatume: The expected value should be either a post ID or `null`.
If the `$cached` value is not empty, then it should either be a valid event, or return `false`.
If the `$cached` value is empty, the only valid allowed value is `null`, and that is covered by the second check.

If `$cached` is `0`, then the `$cached !== null` check would fail and the function would return `false`.

Am I missing something?

lucatume: It's a good idea to add a test for this, in any case.

PR #4094, opened 2022-12-27
Fix/rest insertion relative dates

stratease:
This is a fix for REST v1 event insertion + tests that would fail dependent on the current time (and server timezone). The insertion was generating relatives dates based on the server timezone, instead of the one specified (or default setting), thus depending on particular date and timezone it would cross days.

Added more timezone tests and fixed the expected date to be using the timezone specified.

Artifacts: See tests.


stratease:
Shall we make a ticket for this? Not sure how we handle QA for REST endpoints?

tests/restv1/EventUpdateCest.php
```diff
@@ -125,12 +124,14 @@ public function it_should_allow_to_set_the_start_date_using_natural_language( Te
...
 
 		$start = $data[0];
-		$end = $data[1];
+		$end   = $data[1];
 
-		$start_date = date( 'Y-m-d H:i:s', strtotime( $start ) );
-		$end_date   = date( 'Y-m-d H:i:s', strtotime( $end ) );
+		$timezone   = 'America/New_York';
+		$start_obj  = new DateTime( $start, new DateTimeZone( $timezone ) );
+		$end_obj    = new DateTime( $end, new DateTimeZone( $timezone ) );
```
Camwyn, line 131: Why not only generate the TZ once:
```suggestion
		$timezone     = 'America/New_York';
		$timezone_obj = new DateTimeZone( $timezone );
		$start_obj  = new DateTime( $start, $timezone_obj );
		$end_obj    = new DateTime( $end, $timezone_obj );
```
Need the `$timezone` string below, otherwise I'd compact that even further 😉

Camwyn: Rinse and repeat below as needed

Camwyn (changes requested):
This looks good - but I think we can reduce the number of TZ objects we generate significantly. Anything that can make the tests run more efficient/faster 😉

Camwyn (approved):

tests/restv1/EventUpdateCest.php
Camwyn: This is an optimization of tests, so it's not _needed_ for merging.

stratease: Fair point. I'll tackle some of these in my next PR so they don't go by the roadside.

PR #4102, opened 2023-01-12
Improve the Table creation to avoid weird caching problems.

bordoni:
🎫 [TEC-4631]
---

### DIFF
--- a/src/Events/Custom_Tables/V1/Activation.php
+++ b/src/Events/Custom_Tables/V1/Activation.php
@@ -48,41 +48,41 @@ class Activation {
 	 * @since 6.0.0
 	 */
 	public static function init() {
+		$services = tribe();
+
 		// Check if we ran recently.
 		$db_hash = get_transient( static::ACTIVATION_TRANSIENT );
 
-		$schema_builder = tribe( Schema_Builder::class );
+		$schema_builder = $services->make( Schema_Builder::class );
 		$hash           = $schema_builder->get_registered_schemas_version_hash();
 
-		if ( $db_hash == $hash ) {
+		// Sync any schema changes we may have.
+		if ( ! $schema_builder->all_tables_exist( 'tec' ) ) {
+			$schema_builder->up( true );
+		}
+
+		if ( $db_hash === $hash ) {
 			return;
 		}
 
-		set_transient( static::ACTIVATION_TRANSIENT, $hash, DAY_IN_SECONDS );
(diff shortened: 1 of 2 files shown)


src/Events/Custom_Tables/V1/Activation.php
```diff
@@ -48,41 +48,41 @@ public static function activate() {
...
 		$db_hash = get_transient( static::ACTIVATION_TRANSIENT );
 
-		$schema_builder = tribe( Schema_Builder::class );
+		$schema_builder = $services->make( Schema_Builder::class );
 		$hash           = $schema_builder->get_registered_schemas_version_hash();
 
-		if ( $db_hash == $hash ) {
-			return;
+		// Sync any schema changes we may have.
+		if ( ! $schema_builder->all_tables_exist( 'tec' ) ) {
```
stratease, line 60: Just want a sanity check, as I recall we moved this under the transient for performance concerns, maybe it was added after a ticket for performance... @lucatume are you comfy with this check running often?

bordoni: I will run some tests and work on figuring that out a little more. I would love to avoid having to run that method all the time.

lucatume:
I've pushed a set of commits to change the logic and not use the schema hash; using the schema hash would not create tables when they have been removed by other code (e.g. an attempt to stop the migration).
The new logic has less nested ifs, and is more robust to table removal after the tables are supposedly there.
The logic to run the check 1/day is behind a cached value when using real object cache, and behind a transient when not using it. As discusses, I think the case where no real object cache is used **and** transients are not activated should be supported.

src/Tribe/Main.php
```diff
@@ -4058,5 +4063,28 @@ public function register_plugin_autoload_paths( ) {
...
+		 * @return void The method will clear the state of the Custom Tables v1 activation.
+		 */
+		public static function clear_ct1_activation_state(): void {
+			/*
+			 * Value is hard-coded to avoid autoloading the Activation class for the sole purpose of getting the
+			 * transient name.
+			 *
+			 * @see TEC\Events\Custom_Tables\V1\Activation::ACTIVATION_TRANSIENT
+			 */
+			$transient_key = 'tec_custom_tables_v1_initialized';
```
Camwyn, line 4084: Good reason to use a static prop/function instead?

Camwyn (approved):

PR #4110, opened 2023-01-20
[TEC-4663] Add notice to install `Event Tickets`

juanfra:
### 🎫 Ticket

[TEC-4663] 

### 🗒️ Description

- Adding a notice to install `Event Tickets` when TEC is running.

### 🎥 Artifacts 

https://user-images.githubusercontent.com/252415/213775233-f4ad54a3-ff25-40e7-b5ec-e1886bf211a1.mov

### ✔️ Checklist
- [x] I've included a readme.txt Changelog entry. 
- [x] My code is tested. 
- [x] My code has [proper inline documentation](https://the-events-calendar.github.io/products-engineering/docs/code-standards/).

[TEC-4663]: https://theeventscalendar.atlassian.net/browse/TEC-4663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,438 @@
...
+	 * Register update notices.
+	 *
+	 * @since TBD
+	 */
+	public function hook() {
+		if ( ! is_admin() || ! current_user_can( 'install_plugins' ) ) {
+			return;
+		}
+
+		add_action( 'wp_ajax_nopriv_notice_install_event_ticketsr', [ $this, 'ajax_handle_notice_install_event_tickets' ] );
```
Camwyn, line 58: Is the trailing "r" intentional?
`wp_ajax_nopriv_notice_install_event_ticketsr`

Camwyn: Actually, do we need to trigger on `nopriv` Ajax calls (when the user is not logged in)?

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,438 @@
...
+	 * @return void
+	 */
+	public function ajax_handle_notice_install_event_tickets() {
+		if ( ! check_ajax_referer( 'tribe_events_admin_notice_install', 'nonce', false ) ) {
+			$response['message'] = wpautop( __( 'Insecure request.', 'the-events-calendar' ) );
+
+			wp_send_json_error( $response );
+		}
+
+		if ( ! current_user_can( 'install_plugins' ) ) {
```
Camwyn, line 312: Maybe make this a condition of showing the notice in the first place? Do we want to advertise to folks who _can't_ install?

Camwyn (commented):
Maybe we can follow a pattern similar to date-based notices, with an abstract that contains most of this. Then extend it for each of an array of plugins to make it easy to expand?

src/resources/postcss/admin/notice-install-event-tickets.pcss
```diff
@@ -0,0 +1,32 @@
+/* The Events Calendar Utilities */
+@import '../utilities/_all.pcss';
+
+.tribe-notice-event-tickets-install,
+.tribe-notice-event-tickets-activate {
+	border-left-color: var(--tec-color-accent-primary);
+	border-radius: 4px;
```
Camwyn, line 7:
```suggestion
	border-radius: var(--tec-border-radius-default);
```

src/resources/postcss/admin/notice-install-event-tickets.pcss
```diff
@@ -0,0 +1,32 @@
...
+	display: flex;
+	padding: 0;
+
+	h3 {
+		line-height: 1.2;
```
Camwyn, line 12:
```suggestion
		line-height: var(--tec-line-height-0); 
```
That's 1.38 - the smallest we use in our FE common styles;

src/resources/postcss/admin/notice-install-event-tickets.pcss
```diff
@@ -0,0 +1,32 @@
...
+	flex-grow: 0;
+	flex-shrink: 0;
+
+	img {
+		max-width: 36px;
+	}
+}
+
+.tec-admin__notice-install-content {
+	padding: 20px 20px 20px 0;
```
Camwyn, line 31:
```suggestion
	padding: var(--tec-spacer-4) var(--tec-spacer-4) var(--tec-spacer-4) 0;
```

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: Looks like a typo 😬 And yeah, agree on the `nopriv` call. Seems that I added that initially by habit.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: It's already in the hooks method. As it's a public method I feel it'd be more robust to have the check here as well.

Camwyn:
@juanfra no tests? Something like the [Stellar Sale tests](https://github.com/the-events-calendar/tribe-common/blob/master/tests/wpunit/Tribe/Notices/Stellar_SaleTest.php) maybe?

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+	 * @since TBD
+	 *
+	 * @param string $slug The plugin slug.
+	 *
+	 * @return boolean True if active
+	 */
+	public function is_installed( $slug = '' ): bool {
+		$installed_plugins = get_plugins();
+
+		return array_key_exists( $this->get_plugin_path( $slug ), $installed_plugins ) || in_array( $this->get_plugin_path( $slug ), $installed_plugins, true );
```
bordoni, line 141: Checking for the whole path here leads to false negatives in some cases, plugins can be installed in any path. We should look for the `.php` file alone, which is a lot more unlikely to change, the folder can be changed easily.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+	 * Checks if `Event Tickets` is active.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $slug The plugin slug.
+	 *
+	 * @return boolean True if active.
+	 */
+	public function is_active( $slug = '' ): bool {
+		return is_plugin_active( $this->get_plugin_path( $slug ) );
```
bordoni, line 154: This check can lead to false negatives. Plugins can be installed in any path that we choose. We should be checking for `! did_action( 'tribe_tickets_plugin_loaded' )`

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+
+	/**
+	 * Enqueue assets required for the notice.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	private function enqueue_assets() {
+		wp_enqueue_style( 'wp-components' );
```
bordoni, line 228: Make `wp-components` a dependency of our style.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+			return '';
+		}
+
+		$this->enqueue_assets();
+
+		$html = $this->get_template()->template(
+			'notices/install-event-tickets',
+			$this->get_template_data(),
+			false
+		);
```
bordoni, line 208: This is rendering the notice every time.

We should pass a callback to the method `tribe_notice` so it only renders when the conditions are met. Move the conditional to the `tribe_notice` arguments.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+	 * Install notice for `Event Tickets`.
+	 *
+	 * @since TBD
+	 */
+	public function notice_install() {
+		if ( ! $this->should_display_notice_install() ) {
+			return '';
+		}
+
+		$this->enqueue_assets();
```
bordoni, line 202: We don't need to call the assets directly, just let the conditional associated with both assets to load them in the proper order.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+		);
+
+		tribe_notice(
+			'event-tickets-activate',
+			$html,
+			[
+				'dismiss' => true,
+				'type'    => 'warning',
+			]
+		);
```
bordoni, line 263: Same problems as the notice install. Not leveraging `tribe_notice` or `tribe_asset` tools to avoid early rendering and conditionals.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+			$response['message'] = wpautop( __( 'Insecure request.', 'the-events-calendar' ) );
+
+			wp_send_json_error( $response );
+		}
+
+		if ( ! current_user_can( 'install_plugins' ) ) {
+			wp_send_json_error( [ 'message' => wpautop( __( 'Security Error, Need higher Permissions to install plugin.' ), 'the-events-calendar' ) ] );
+		}
+
+		$vars    = $_REQUEST;
```
bordoni, line 314: We should the array of the variables we need, instead of passing all that was received on `$_REQUEST`, could lead to weird security scenarios.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+		}
+
+		if ( $this->is_active( $vars['slug'] ) ) {
+			return true;
+		}
+
+		$activate = activate_plugin( $this->get_plugin_path( $vars['slug'] ), '', false, true );
+
+		return ! is_wp_error( $activate );
+	}
```
bordoni, line 416: These methods should not live in this class and must be tested in some capacity.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -0,0 +1,436 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param array $args Array with arguments to override the defaults.
+	 *
+	 * @return array The template args.
+	 */
+	private function get_template_data( $args = [] ): array {
+		$admin_url    = is_network_admin() ? network_admin_url( 'settings.php' ) : admin_url( 'admin.php' );
+		$redirect_url = add_query_arg( [ 'page' => 'tec-tickets-settings' ], $admin_url );
```
bordoni, line 277: The page's slug should come from the constant, not a literal string.

bordoni (changes requested):
We should add some tests here, one for the Notice and another

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: This is required before ET is installed/active, so we don't have access to that constant at this point, it has to be hardcoded.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: These are reliant on WP core functions that are already tested in WP. Where would you move these? As they are only applicable (at this point) to the functionality related to this notice.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: There's already a nonce check, and the functionality afterward it's calling a template that is not extensible, there's no security risk involved here. Anyways, I can only get the vars we use and send only those to the template.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: There are conditionals before, like many other notices. With those conditionals, the notice won't be rendered. I will adjust the code, though.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: I'm relying on a WP core function, that's used to basically check if a plugin is active. I John Doe is renaming the `event-tickets` (in this case) folder to something else then it should be their problem.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: [See comment](https://github.com/the-events-calendar/the-events-calendar/pull/4110#discussion_r1085156434).

src/Tribe/Admin/Notice/Install_Event_Tickets.php
bordoni: Thanks for the change you made here 🫀

src/Tribe/Admin/Notice/Install_Event_Tickets.php
bordoni: Makes sense.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
bordoni: I would move those into Common, but honestly, it's fine if it lives here now, and we can move the whole class later to Common so we can leverage this for other plugins' behavior of installing other plugins on the go.

src/Tribe/Admin/Provider.php
```diff
@@ -15,6 +15,7 @@ public function register() {
 		$this->container->singleton( 'tec.admin.notice.timezones', Notice\Timezones::class );
 		$this->container->singleton( 'tec.admin.notice.marketing', Notice\Marketing::class );
 		$this->container->singleton( 'tec.admin.notice.update', Notice\Update::class );
+		$this->container->singleton( 'tec.admin.notice.install-event-tickets', Notice\Install_Event_Tickets::class );
```
bordoni, line 18: Lets go with:
```suggestion
		$this->container->singleton( Notice\Install_Event_Tickets::class, Notice\Install_Event_Tickets::class );
```

We should be avoiding the string notation where we can.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
bordoni: You are not wrong, but PUE shouldn't be the standard; it's an older code, honestly. I agree with you that it would be amazing to use WordPress method here, but it's slower and if you have ET active but not in use example; If there is a dependency with TEC failing, it would show the notice in this case and we want to avoid.

bordoni (changes requested):

src/Tribe/Admin/Notice/Install_Event_Tickets.php
juanfra: @bordoni This notice lives in TEC, and is for installing ET. 

The notice will never be shown if there's a dependency mismatch because both plugins will be active. And in that case, the notice is not shown (that's what this check is for). `is_plugin_active` will be the main reason why it won't be shown.

In my opinion, WP core functions are the way to go probably 99% of the time. I don't feel using [this code](https://github.com/WordPress/wordpress-develop/blob/6.1/src/wp-admin/includes/plugin.php#L527-L529) to perform the check will make things significantly slower. This is the code for `is_plugin_active`:

`function is_plugin_active( $plugin ) {
	return in_array( $plugin, (array) get_option( 'active_plugins', array() ), true ) || is_plugin_active_for_network( $plugin );
}`

bordoni (approved):
Looking good, lets find a home release for this and ship it.

borkweb:
This is great, @juanfra! I've extracted the non-notice part of these changes and moved them to a StellarWP library (https://github.com/stellarwp/installer) and made a couple of PRs to put these changes in place:

* https://github.com/the-events-calendar/tribe-common/pull/1872
* https://github.com/the-events-calendar/the-events-calendar/pull/4121

PR #4114, opened 2023-01-24
fix/tec 4536 ct1 tables updates

lucatume:
Ticket: [TEC-4536](https://theeventscalendar.atlassian.net/browse/TEC-4536)

Artifacts: new and existing tests.

Initially, the MySQL type of the date fields in the Occurrences and Events tables was set to `DATETIME`.
While this makes sense, it will trigger warnings on MySQL implementations with stricter settings (`sql_mode="NO_ZERO_IN_DATE,NO_ZERO_DATE"`) in any operation including those fields and treating them like string fields.
Depending on the setting, those warnings could become `wpdb` errors preventing the query from executing completely.
There are a number of plugins out there assuming an operation like `SELECT ... where start_date LIKE '20220-10%` will work and will not trigger exceptions that will benefit from this change.

The existing tests will ensure that the CT1 implementation is fine with the update and is not relying on those fields to be actually `DATETIME` fields; the new tests will cover the regression of the type of those fields.

What will the effect on ECP be? [PR to find it out here](https://github.com/the-events-calendar/events-pro/pull/2179)

[TEC-4536]: https://theeventscalendar.atlassian.net/browse/TEC-4536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Events/Custom_Tables/V1/Tables/Events.php
+++ b/src/Events/Custom_Tables/V1/Tables/Events.php
@@ -27,7 +27,7 @@ class Events extends Abstract_Custom_Table {
 	/**
 	 * @inheritDoc
 	 */
-	const SCHEMA_VERSION = '1.0.0';
+	const SCHEMA_VERSION = '1.0.1';
 
 	/**
 	 * {@inheritdoc}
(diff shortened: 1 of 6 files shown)


src/Events/Custom_Tables/V1/Tables/Events.php
```diff
@@ -58,14 +58,15 @@ protected function get_update_sql() {
 		$table_name = self::table_name(true);
 		$charset_collate = $wpdb->get_charset_collate();
 
+		// VARCHAR(19) to store YYYY-MM-DD HH:MM:SS values as strings and allow partial compare.
 		return "CREATE TABLE `{$table_name}` (
 			`event_id` BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
 			`post_id` BIGINT(20) UNSIGNED NOT NULL,
-			`start_date` DATETIME NOT NULL,
```
stratease, line 64: Not totally against this change. BUT from what I understand of the original issue, wouldn't another fix be to remove the NOT NULL constraint and allow nulls? Hate to move a date to a non-date type if we can avoid it.

src/Events/Custom_Tables/V1/Tables/Occurrences.php
```diff
@@ -60,14 +60,15 @@ protected function get_update_sql() {
 		$table_name      = self::table_name( true );
 		$charset_collate = $wpdb->get_charset_collate();
 
+		// VARCHAR(19) to store YYYY-MM-DD HH:MM:SS values as strings and allow partial compare.
```
stratease, line 63: For partial compares is this because doing date comparisons isn't ideal, i.e. `select start_date from wp_tec_occurrences where DATE(start_date) = '2022-12-01' ` ?

stratease (changes requested):

src/Events/Custom_Tables/V1/Tables/Occurrences.php
lucatume: The `DATE` comparison **is** ideal, and in a world where back-compatibility is not an issue, I would stick with it. The problem is existing TEC and other code that is assuming those fields are strings.

src/Events/Custom_Tables/V1/Tables/Events.php
lucatume: I've ran a test setting the field to `DATETIME` only, it will still throw issues.
The `NO_ZERO_DATE` or `NO_ZERO_IN_DATE` `sql_mode` will condition not only insertions, but selections as well. Just searching a `DATETIME` field under that condition as a string will trigger a warning.

stratease (approved):
Groovy, with a tear.

PR #4116, opened 2023-01-25
Allow filtering the redirection nature of a request

lucatume:
Ticket: [TEC-4511](https://theeventscalendar.atlassian.net/browse/TEC-4511)

This PR is the companion to the [ECP one](https://github.com/the-events-calendar/events-pro/pull/2182) to allow detecting a redirected request only by means of the `tribe_redirected` query var.
The support for the `tribe_redirected` query variable cannot be removed due to its use in the Event Tickets PayPal gateway, but adding a filter allows the request lifecycle to mark the current request as a redirected one without requiring its use.

[TEC-4511]: https://theeventscalendar.atlassian.net/browse/TEC-4511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stratease (commented):
Seems like this would be testable?

stratease (approved):
Groovy!

PR #4117, opened 2023-01-25
fix/tec 4664 support events cat unregister

lucatume:
Ticket: [TEC-4664](https://theeventscalendar.atlassian.net/browse/TEC-4664)

Artifacts: tests.

This updates the code to be robust to the unregistration of the custom Events' category.

I've checked all the places where the result of a `get_terms`, and similar, call is assumed to be a set of terms to check for:
* `WP_Error` response when the taxonomy is not registered
* bad terms, not instances of `WP_Term`

Smaller refactoring to `Tribe__Events__Main` to extract the `register_taxonomy` method; no functional changes.

[TEC-4664]: https://theeventscalendar.atlassian.net/browse/TEC-4664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/functions/template-tags/general.php
+++ b/src/functions/template-tags/general.php
@@ -467,8 +467,16 @@ function tribe_is_past_event( $event = null ) {
  * @category Events
  */
 function tribe_get_event_cat_ids( $post_id = 0 ) {
-	$post_id = Tribe__Events__Main::postIdHelper( $post_id );
-	$terms   = array_filter( (array) get_the_terms( $post_id, Tribe__Events__Main::TAXONOMY ) );
+	$post_id  = Tribe__Events__Main::postIdHelper( $post_id );
+	$terms = get_the_terms( $post_id, Tribe__Events__Main::TAXONOMY );
+
+	if ( $terms instanceof WP_Error ) {
+		return [];
+	}
+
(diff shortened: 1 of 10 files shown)


src/functions/template-tags/general.php
```diff
@@ -682,10 +696,14 @@ function tribe_meta_event_archive_tags( $label = null, $separator = ', ', $echo
 
 		$terms = get_the_terms( get_the_ID(), 'post_tag' );
 
-		if ( is_wp_error( $terms ) ) {
+		if ( empty( $terms ) || is_wp_error( $terms ) ) {
```
Camwyn, line 699: At this point (3rd time's the charm?) I think I'd functionalize this logic.\ 🤔

Camwyn (approved):
This looks good - I left one comment about encapsulating the logic into its own function as a suggestion.

src/functions/template-tags/general.php
lucatume: I'm not sure that would be a good trade-off. It could get bloated over time and introduce side-effects as the `tec_parse_terms_list` function is added more functionality and parameters.

PR #4118, opened 2023-01-26
Replace php-date-formatter submodule with direct commit

borkweb:
This removes the submodule for the php-date-formatter library. We haven't updated it in eons and the submodule is causing issues with deployments for the docs site.

Pull requesting this directly into `master` so we can trickle that down into all the other branches.

Note: I did not fetch any additional code for the library. Instead, I merely removed the `.gitmodule` entry, copied the vendor directory to back it up, deleted the submodule, removed the `.git` directory from the backed up `php-date-formatter/`, moved the backed up `php-date-formatter/` back to `vendor/`, and then direct committed the directory.

Also note: we are not able to use composer for this until we update to composer 2.

---

The reason for this is that the deployment to the docs site repeatedly fails when trying to do a submodule update on that repository specifically. We cannot use composer until we are using composer 2. I don't recommend that we merge the above PR until we are getting ready for B23.android, though it should remain pointed at master so we can keep that branch in sync with master and use that branch when committing submodule updates to the tec-docs repository.


bordoni (approved):

lucatume:
The PR makes sense.
In the future, it might be worth taking a look at what it's actually used of the library and internalize it.

PR #4121, opened 2023-01-28
[TEC-4663] Implement the new stellarwp/installer library for install actions

borkweb:
Since we'll want to use install functionality in more places than one, I've extracted out @juanfra's install/activate logic into a library that can be used in multiple places. That can be seen here: https://github.com/stellarwp/installer

Related:
* https://github.com/the-events-calendar/the-events-calendar/pull/4110
* https://github.com/the-events-calendar/tribe-common/pull/1872

🎥 https://d.pr/v/MqNRTp
:ticket: [TEC-4663]

[TEC-4663]: https://theeventscalendar.atlassian.net/browse/TEC-4663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -3,6 +3,7 @@
  * Main Tribe Events Calendar class.
  */
 
+use TEC\Events\StellarWP\Installer;
 use Tribe\DB_Lock;
 use Tribe\Events\Views\V2;
 use Tribe\Events\Admin\Settings;
@@ -650,6 +651,16 @@ if ( ! class_exists( 'Tribe__Events__Main' ) ) {
 
 			// Load the new third-party integration system.
 			tribe_register_provider( TEC\Events\Integrations\Provider::class );
+
+			Installer\Config::set_hook_prefix( 'tec_events' );
+			$installer = Installer\Installer::get();
+			$installer->register_plugin( 'event-tickets', __( 'Event Tickets', 'the-events-calendar' ), 'event-tickets/event-tickets.php' );
+			add_filter( 'stellarwp/installer/tec_events/button_classes', static function( $classes ) {
+				$classes[] = 'components-button';
+				$classes[] = 'is-primary';
+				$classes[] = 'tec-admin__notice-install-content-button';
+				return $classes;
+			} );
 
 			/**
 			 * Allows other plugins and services to override/change the bound implementations.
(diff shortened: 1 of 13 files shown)


src/Tribe/Main.php
```diff
@@ -651,6 +652,16 @@ public function bind_implementations(  ) {
...
 
+			Installer\Config::set_hook_prefix( 'tec_events' );
+			$installer = Installer\Installer::get();
+			$installer->register_plugin( 'event-tickets', __( 'Event Tickets', 'the-events-calendar' ), 'event-tickets/event-tickets.php' );
+			add_filter( 'stellarwp/installer/tec_events/button_classes', static function( $classes ) {
+				$classes[] = 'components-button';
+				$classes[] = 'is-primary';
+				$classes[] = 'tec-admin__notice-install-content-button';
+				return $classes;
+			} );
```
borkweb, line 663: This doesn't belong here. I'll be moving it.

Camwyn (changes requested):
I think this is because the text changes in the `installer` library weren't versioned?

borkweb:
Nah, just needed to run `slic run integration --debug` to regenerate the snapshots. No updates to installer needed.

src/admin-views/notices/install-event-tickets.php
```diff
@@ -37,17 +33,6 @@
...
-		data-plugin-slug="<?php echo esc_attr( $plugin_slug ); ?>"
-		data-nonce="<?php echo esc_attr( $ajax_nonce ); ?>"
-		data-action="<?php echo esc_attr( $action ); ?>"
-		data-redirect-url="<?php echo esc_attr( $redirect_url ); ?>"
-		data-installing-label="<?php echo esc_attr( $installing_label ); ?>"
-		data-installed-label="<?php echo esc_attr( $installed_label ); ?>"
-		data-activating-label="<?php echo esc_attr( $activating_label ); ?>"
-		data-activated-label="<?php echo esc_attr( $activated_label ); ?>"
-	><?php echo esc_html( $button_label ); ?></button>
+	<?php	Installer::get()->render_plugin_button( 'event-tickets', $action, $button_label, $redirect_url ); ?>
```
juanfra, line 36: We could keep the `plugin_slug` as template var and use that as a parameter instead of using the string `event-tickets` so that we have it living just in one place (given that this template is only for the ET notice)

juanfra (approved):
Looking good. I left one tiny suggestion (I'm ok if we also leave it as is).

PR #4122, opened 2023-01-30
Fix/tec 4653 elementor buttons

stratease:
[TEC-4653](https://theeventscalendar.atlassian.net/browse/TEC-4653)

This fixes situations where a global button style was affecting some of our frontend components that should not be considered buttons, such as "Next" links and our "Calendar Navigation" links.

Artifact:
Some of the "buttons" that were inadvertently affected by global button styles.
image

[TEC-4653]: https://theeventscalendar.atlassian.net/browse/TEC-4653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -7,6 +7,12 @@ on:
         description: 'Zip type: 0 = dev, 1 = production'
         required: false
         default: '0'
+      slack_channel:
+        description: 'Slack channel ID to post to'
+        required: false
+      slack_thread:
+        description: 'Slack thread to post to'
+        required: false
 jobs:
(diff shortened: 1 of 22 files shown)


stratease:
@Camwyn this was the project regarding some `<button>` tags showing up in our templates sometimes (when disabled) instead of as `<a>` tags (when enabled).

Probably worth revisiting with some more markup and styling work in the future.

bordoni (approved):

PR #4129, opened 2023-02-07
Fix fatal regarding rewrite/wpml.

stratease:
Bumped common with some rewrite rule safeguard logic. Adding safeguards around WPML translation logic and rewrite rule filtering.

### DIFF
--- a/src/Tribe/Integrations/WPML/Rewrites.php
+++ b/src/Tribe/Integrations/WPML/Rewrites.php
@@ -58,11 +58,30 @@ class Tribe__Events__Integrations__WPML__Rewrites {
 	 * Filters the rewrite rules array to add support for translated versions of
 	 * venue and organizer slugs in their rules.
 	 *
-	 * @param array $rewrite_rules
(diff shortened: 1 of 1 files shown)


src/Tribe/Integrations/WPML/Rewrites.php
```diff
@@ -58,11 +58,30 @@ public static function instance() {
 	 * Filters the rewrite rules array to add support for translated versions of
 	 * venue and organizer slugs in their rules.
 	 *
-	 * @param array $rewrite_rules
+	 * @since TBD Moving type check down to safeguard this public filter.
```
Camwyn, line 61:
```suggestion
	 * @since 6.0.9 Moving type check down to safeguard this public filter.
```

bordoni (dismissed):

Camwyn (changes requested):
Fix the version # - otherwise looks good

jesseeproductions (approved):

Camwyn (dismissed):

Camwyn (approved):

PR #4133, opened 2023-02-08
Subscribe to calendar dropdown. 

EugeneKyale:
Ticket : [TEC-4388](https://theeventscalendar.atlassian.net/browse/TEC-4388)

Ensure the Subscribe to Calendar Dropdown opens and closes consistently across all themes by adding the event listener to the button instead of the container.

**Screencast 🎥**

https://www.loom.com/share/fe9ab14e2dd04a06a9f1c26379c5ffaa

[TEC-4388]: https://theeventscalendar.atlassian.net/browse/TEC-4388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Camwyn (commented):
Can we also adjust the logic so clicking on the button closes the drop-down? The upward-pointing caret indicates that should work, but it does not.

EugeneKyale:
> Can we also adjust the logic so clicking on the button closes the drop-down? The upward-pointing caret indicates that should work, but it does not.

@Camwyn  Yeah it makes sense to do that. I'll make an update to include that.

Camwyn (approved):

PR #4136, opened 2023-02-09
Fix typo in func name.

stratease:
No ticket, just fixing typo.


Camwyn (changes requested):
The typo fix is good but this mixes code from boilerplate into c3po

Camwyn (approved):

PR #4143, opened 2023-02-15
Add namespace to wp.hooks.addAction to prevent errors.

Camwyn:
In StellarWP Installer, the namespace for this action is the button [data-slug](https://github.com/stellarwp/installer/blob/main/src/assets/js/installer.js#L102) upon inspection of the button, the slug is `event-tickets`

This does not follow the suggested `vendor/plugin/function` pattern from [the docs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-hooks/#the-global-instance) but it seemed prudent to change here and not in the installer.


src/resources/js/admin/notice-install-event-tickets.js
```diff
@@ -53,6 +53,7 @@ tribe.events.admin.noticeInstall = {};
 	obj.ready = function() {
 		wp.hooks.addAction(
 			'stellarwp_installer_tec_error',
+			'event-tickets',
```
borkweb, line 56: I think we want to have this be the namespace of the addAction() call and not the namespace of the activity/description. Maybe follow this format:

```suggestion
			'tec_events',
```

borkweb: Or:

```suggestion
			'tec/events',
```

bordoni (changes requested):
Needs changelog

PR #4155, opened 2023-02-28
Fix WPML Integration meta manipulation

lucatume:
[ECP-1429](https://theeventscalendar.atlassian.net/browse/ECP-1429)

Artifact: [screencast](https://share.cleanshot.com/cmNgCGc0)

Required by [ECP PR](https://github.com/the-events-calendar/events-pro/pull/2212)

The `Tribe__Events__Integrations__WPML__Meta::translate_post_id` mehtod will filter the `get_post_metadata` calls to translate the post IDs that might be found in the meta (i.e. the `_EventOrganizerID` and the `_EventVenueID` ones).
Currently, the filtering will be applied in the `get_metadata_raw` and `metadata_exists` functions and, in both instances, it's for a bail check: one where returning a non-null value would short-circuit the function logic and return the value.  If not found, the value returned should be the original one, `null` in the case of the bail checks, to let the rest of the function logic run its course.

[ECP-1429]: https://theeventscalendar.atlassian.net/browse/ECP-1429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Integrations/WPML/Meta.php
+++ b/src/Tribe/Integrations/WPML/Meta.php
@@ -49,7 +49,9 @@ class Tribe__Events__Integrations__WPML__Meta {
 		$cache_key = 'wpml_meta_translate_post_id_' . $object_id . '-' . $meta_key;
 
 		if ( isset( $cache[ $cache_key ] ) ) {
-			return $cache[ $cache_key ];
+			$cached = $cache[ $cache_key ];
+
+			return $single ? $cached : [ $cached ];
 		}
 
 		$original_value = $value;
@@ -94,7 +96,7 @@ class Tribe__Events__Integrations__WPML__Meta {
 			}
 		}
 
-		$cache[ $cache_key ] = $value;
+		$cache[ $cache_key ] = $value[0];
 
 		return $single ? $value[0] : $value;
 	}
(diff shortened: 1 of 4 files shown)


src/Tribe/Integrations/WPML/Meta.php
```diff
@@ -90,7 +96,7 @@ public function translate_post_id( $value, $object_id, $meta_key, $single ) {
 
 		$cache[ $cache_key ] = $value;
 
-		return $value;
+		return $single ? $value[0] : $value;
```
Camwyn, line 99: We probably want to store the amended value (`$value[0]`) in cache - as we directly return it above (line 52) which could lead to unexpected issues if it's an array.

Camwyn (changes requested):
careful with cache 😉

src/Tribe/Integrations/WPML/Meta.php
lucatume: Fixed.=

lucatume:
I'm debugging the test failure in #4157

lucatume:
I've "fixed" the failing test by skipping it: it passes locally, it passes in CI when running alone ([link](https://github.com/the-events-calendar/the-events-calendar/actions/runs/4306285337/jobs/7509887773)) will `SIGSEV` when running in CI ([link](https://github.com/the-events-calendar/the-events-calendar/actions/runs/4306118295/jobs/7510296677#step:21:28)).
I'm trying to debug the issue in another PR ([link](https://github.com/the-events-calendar/the-events-calendar/pull/4157)).

PR #4156, opened 2023-03-01
[CE-178] Fix allow venue defaults to work with Community Events 

redscar:
This PR fixes an issue where the default venue information wasn't populating when using Community Events.

Screenshot within Admin
image

Screenshot from within Community Events
image

Showing the custom list of countries works as well
image


Camwyn (approved):
I'm assuming this has been checked to make sure it doesn't bleed default venue data where it's not intended

PR #4162, opened 2023-03-03
fix(resources/js) correctly call the addAction method

lucatume:
The `wp.hooks.addAction` method requires a namespace to uniquely
identify the callback (see [doc](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-hooks/#the-global-instance)).
This adds it.

* [before](https://share.cleanshot.com/CM9KlD3Z) 
* [after](https://share.cleanshot.com/hYJtYh37)


lucatume:
Closed in favor of #4164

PR #4163, opened 2023-03-03
Accessibility Improvement.

EugeneKyale:
Ticket : [TEC-3415](https://theeventscalendar.atlassian.net/browse/TEC-3415)

Ensure we only have a single `<main>` element in the single events page for improved accessibility. This applies to sites using both the **Default Events Template** and **Default Page Template** settings.

**Screenshot 📸**

Prior to this update, the single events page had 2 `<main>` elements, one from TEC and the other from the active theme. 

image

[TEC-3415]: https://theeventscalendar.atlassian.net/browse/TEC-3415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Camwyn (dismissed):

Camwyn (approved):
FWIW - I don' think the chosen events template matters much for single event in this instance.

EugeneKyale:
> FWIW - I don' think the chosen events template matters much for single event in this instance.

Copy that, you're right. It applies irregardless of the selected template.

Camwyn (approved):

PR #4170, opened 2023-03-08
Fix ACF integration

lucatume:
[TEC-4621](https://theeventscalendar.atlassian.net/browse/TEC-4621) 

Fix the query side of the ACF integration code to handle both initial PHP state queries for Events, and AJAX-based queries.

> Multiple post-type queries are not supported.

Artifacts:
* tests
* [screencast](https://share.cleanshot.com/HZ5pJZG2) 

[TEC-4621]: https://theeventscalendar.atlassian.net/browse/TEC-4621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Events/Custom_Tables/V1/Integrations/ACF/Controller.php
+++ b/src/Events/Custom_Tables/V1/Integrations/ACF/Controller.php
@@ -111,13 +111,12 @@ class Controller extends Service_Provider {
 	 * @return void Actions and filters are removed.
 	 */
 	public function unregister() {
+		// Unhook the same action twice for each supported field type; see the `register` method.
 		foreach ( self::get_supported_field_types() as $supported_field_type ) {
-			// Detect the start of a specific ACF field type to boot the handling.
 			remove_action( 'acf/render_field/type=' . $supported_field_type, [
 				$this,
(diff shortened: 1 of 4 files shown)


stratease (dismissed):

src/Events/Custom_Tables/V1/Integrations/ACF/Controller.php
```diff
@@ -0,0 +1,208 @@
...
+	/**
+	 * Removes the actions and filters added by the provider.
+	 *
+	 * @since TBD
+	 *
+	 * @return void Actions and filters are removed.
+	 */
+	public function unregister() {
+		foreach ( self::get_supported_field_types() as $supported_field_type ) {
+			// Detect the start of a specific ACF field type to boot the handling.
```
Camwyn, line 115: This inline comment and the one below make sense if you know the context, but less so if you just hit them - as they read like you're adding the actions, not removing them.

Just a thought, not a needed fix.

Camwyn (dismissed):

PR #4176, opened 2023-03-10
Avoid an unbound query for Next and Previous Event date

bordoni:
[TEC-3089](https://theeventscalendar.atlassian.net/browse/TEC-3089)

📹 
1. https://i.bordoni.me/G8YTx91Q
2. https://i.bordoni.me/TB0Nj8vw

[TEC-3089]: https://theeventscalendar.atlassian.net/browse/TEC-3089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Camwyn (approved):

lucatume:
This should not be merged without adequate test coverage.

bordoni:
@lucatume the Snapshots are covering the changes related to the Next and Previous Events.

lucatume:
> @lucatume the Snapshots are covering the changes related to the Next and Previous Events.

The snapshots are covering cold-cache, HTML rendering. I would add tests that hit the updated method multiple times in the same request and assert (e.g. adding a filter on `query`) that no new query actually runs; in Legacy and CT1 context.

And since you're working on that code, I would remove the use of an instance variable (`cached_event_dates`) to store cached values to use the object cache. That will make the methods cache react correctly to WP cache flushes.

lucatume (changes requested):
Add tests to cover the `get_(previous|next)_event_date` methods caching in Legacy and CT1 context.

tests/views_wpunit/Tribe/Events/Views/V2/Day_ViewTest.php
```diff
@@ -0,0 +1,188 @@
...
+			$mock_and_insert( 'events/single/id.template.json', 234234234 ),
+			$mock_and_insert( 'events/single/id.template.json', 2453454355 ),
+			$mock_and_insert( 'events/single/id.template.json', 3094853477 ),
+			$mock_and_insert( 'events/single/id.template.json', 3094855477 ),
+		] );
+
+		/* @var Day_View $day_view */
+		$day_view = View::make( Day_View::class, $this->context );
+
+		$next_event_date = $day_view->get_next_event_date( $now );
```
lucatume, line 115: Since this is about caching, I would pull the same value at least 3 times and check:
1. the value is the same each time
2. hook on the `query` filter to check no queries will run after the first fetch

tests/views_wpunit/Tribe/Events/Views/V2/Day_ViewTest.php
```diff
@@ -0,0 +1,188 @@
...
+	 * @test
+	 */
+	public function should_have_next_date() {
+		$timezone_string = 'Europe/Paris';
+		$timezone        = new \DateTimeZone( $timezone_string );
+		update_option( 'timezone_string', $timezone_string );
+
+		$now = new DateTime( $this->mock_date_value, $timezone );
+
+		$events_data = [
```
lucatume, line 64: Since you're not running a snapshot test that would require the post IDs to be stable, you do not need the mocking feature at all; just use `tribe_events()` to create events happening at the required dates. This will simplify the setup and make the testing code more robust.

tests/views_wpunit/Tribe/Events/Views/V2/Day_ViewTest.php
```diff
@@ -0,0 +1,188 @@
...
+use DateTime;
+
+/**
+ * Class Month_ViewTest
+ *
+ * @since   TBD
+ *
+ * @package Tribe\Events\Views\V2
+ */
+class Day_ViewTest extends \Codeception\TestCase\WPTestCase {
```
lucatume, line 17: The same test should exist in the `ct1_integration` suite to make sure the cold and warm cache values are correct.

tests/views_wpunit/Tribe/Events/Views/V2/Day_ViewTest.php
lucatume: Since there are many caching solutions out there modeling the same data differently, add a test for cache poisoning, where the data is set, but actually not coherent with the format of the expected value. [Example link](https://github.com/the-events-calendar/the-events-calendar/blob/fd5cc4dbc35e7ca3f0a98163e5899dc81f8363ad/tests/integration/Tribe__Events__Adjacent_EventsTest.php#L290)

tests/views_wpunit/Tribe/Events/Views/V2/Month_ViewTest.php
```diff
@@ -0,0 +1,188 @@
...
+use DateTime;
+
+/**
+ * Class Month_ViewTest
+ *
+ * @since   TBD
+ *
+ * @package Tribe\Events\Views\V2
+ */
+class Month_ViewTest extends \Codeception\TestCase\WPTestCase {
```
lucatume, line 17: Same considerations for the Day view apply here.

lucatume (changes requested):

tests/views_wpunit/Tribe/Events/Views/V2/Day_ViewTest.php
bordoni: Oh yeah that makes total sense. I will take that route.

tests/views_wpunit/Tribe/Events/Views/V2/Day_ViewTest.php
bordoni: That is fair, I was mostly using something i knew was stable for View testing, but it makes total sense to change that.

tests/views_wpunit/Tribe/Events/Views/V2/Day_ViewTest.php
stratease: @lucatume don't think cache poisoning is relevant here, these are just memoized values in an array. It's not using the `Tribe__Cache`.

lucatume (approved):

lucatume (approved):
All fine.
I should have fixed the `integration` test in [this PR](https://github.com/the-events-calendar/the-events-calendar)

PR #4177, opened 2023-03-10
Fix for term/taxonomy query returning unintended results.

stratease:
[TEC-4695](https://theeventscalendar.atlassian.net/browse/TEC-4695)

Taxonomy query would break when `get_posts()` was recursed (edge case bug w/ core Wordpress handling of taxonomy). The result would be an additional term where clause added, filtering out the expected results.

Artifact:
The query would contain an additional taxonomy where clause, e.g.
```
WHERE 1=1 
 AND ( 
  wp_term_relationships.term_taxonomy_id IN (182,183,184) 
  AND 
  tt1.term_taxonomy_id IN (184)
)
``` 

[TEC-4695]: https://theeventscalendar.atlassian.net/browse/TEC-4695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Events/Custom_Tables/V1/WP_Query/Modifiers/Events_Only_Modifier.php
```diff
@@ -46,6 +46,26 @@ public function applies_to( WP_Query $query = null ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param Custom_Tables_Query $query A reference to the query object used to get_posts() for Custom Table queries.
+	 *
+	 * @return Custom_Tables_Query The modified query object.
+	 */
+	public function filter_ct_query( Custom_Tables_Query $query ): Custom_Tables_Query {
+		// These query vars conflict, and will result in unintended results. See TEC-4695.
+		if ( ! isset( $query->query['term'] )
```
lucatume, line 63: Skip the `isset( $query->query['term'])` just unset.

lucatume: Should the `$query->query_args` be updated too?

readme.txt
```diff
@@ -241,8 +242,8 @@ Remember to always make a backup of your database and files before updating!
 * Fix - Resolve the issue of the WP customizer not loading correctly when the Colbri Page Builder plugin is active. [TEC-4735]
 * Tweak - Added Filters: `tec_events_linked_posts_my_posts_post_status`, `tec_events_linked_posts_all_posts_post_status` to allow filtering of post status of Linked Posts. [CE-27]
 * Tweak - Ensure we only have a single <main> element in the single events page for improved accessibility. [TEC-3415]
-* Tweak - Added Filters: `tec_events_linked_posts_my_posts_post_status`, `tec_events_linked_posts_all_posts_post_status` to allow filtering of post status of Linked Posts. [CE-27]
```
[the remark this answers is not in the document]
stratease: Yeah, it's a dupe - see line 243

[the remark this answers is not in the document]
EugeneKyale: aah I missed that.

src/Events/Custom_Tables/V1/WP_Query/Modifiers/Events_Only_Modifier.php
stratease: > Skip the `isset( $query->query['term'])` just unset.

As I recall, I don't think the nuanced logic should ignore if that `term` is set. I think that would present another edge case, but I could be wrong... are you sure to remove it?

lucatume: > As I recall, I don't think the nuanced logic should ignore if that `term` is set. I think that would present another edge case, but I could be wrong... are you sure to remove it?

You're still checking if both `term` and `tax_query`: if both set, you can just unset.

lucatume: It's not a bug, though.

lucatume (approved):

src/Events/Custom_Tables/V1/WP_Query/Modifiers/Events_Only_Modifier.php
stratease: Not exactly - it is on different vars. I think that was part of the specific scenario, those vars were somewhat missmatched and were an indicator that someone was not intentionally searching on that "additional" taxonomy. I was trying to rule out intentional search vs the mutated one done by `get_posts()`.

PR #4193, opened 2023-03-29
Update event-status.php

andrasguseo:
Not sure if it's worth adding this to the readme and doing all the other bureaucracy. 🙂


Camwyn (approved):

Camwyn:
Nope, not needed. But it's good to fix this ❤️

PR #4195, opened 2023-03-30
fix(Integrations/WPML) correctly build accented URLs

lucatume:
[TEC-4689](https://theeventscalendar.atlassian.net/browse/TEC-4689) 

Artifacts:
* tests (where possible)
* [screencast](https://share.cleanshot.com/vyMDDCHq) 

When using languages, like the French one, using accented words or
apostrophes, the URL building would break in WPML context.

This fixes that by adding an additional slug flag to the `I18n`
component to closely match the rewrite rules generated when using
permalinks.

### DIFF
--- a/src/Tribe/I18n.php
+++ b/src/Tribe/I18n.php
@@ -57,6 +57,13 @@ class I18n {
 	public const RETURN_BY_LANGUAGE = 8;
 
 	/**
+	 * A flag to require the translations to include the slug version of the translation.
+	 *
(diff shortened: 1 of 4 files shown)


lucatume:
I've restored the `restv1` suite to run with CT1 deactivated; this was always the case. 
In another PR dedicated to it, the same tests as the `restv1` suite, should run with CT1 activated.

jesseeproductions (approved):
Looks good!

PR #4198, opened 2023-03-31
Fix for some collissions.

stratease:
Re: failure in QA on [ECP-1475](https://theeventscalendar.atlassian.net/browse/ECP-1475)

No longer run trash on option updates, let the cron handle it.

[ECP-1475]: https://theeventscalendar.atlassian.net/browse/ECP-1475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Event_Cleaner_Scheduler.php
```diff
@@ -223,10 +223,27 @@ public function move_old_events_to_trash():array {
 			return $results;
 		}
 
+		remove_action( 'save_post_' . Tribe__Events__Main::POSTTYPE, [
```
Camwyn, line 226: I'd _suggest_ moving this (and the re-add) to separate functions so if we find we have to add more remove/add(s) it doesn't clutter this function up so much (it's already the majority of the code in here)

Camwyn (commented):
Question/suggestion

Camwyn (approved):

PR #4200, opened 2023-04-03
Event Tags

EugeneKyale:
Modifies the tag slug for the post_tag taxonomy to include an "events" prefix.

### DIFF
--- a/src/Tribe/Taxonomy/Taxonomy_Provider.php
+++ b/src/Tribe/Taxonomy/Taxonomy_Provider.php
@@ -29,7 +29,6 @@ class Taxonomy_Provider extends \tad_DI52_ServiceProvider {
 		$this->container->singleton( 'events.taxonomy.provider', $this );
 
 		$this->add_filters();
-		$this->add_actions();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/Tribe/Taxonomy/Taxonomy_Provider.php
```diff
@@ -53,4 +54,25 @@ protected function add_filters() {
...
+	public function modify_tag_taxonomy_args( $args, $taxonomy ) {
+		if ( 'post_tag' === $taxonomy ) {
+			$args['rewrite'] = [
+				'slug'       => 'events/tag',
+				'with_front' => false,
+			];
+		}
+
+		return $args;
+	}
```
Camwyn, line 77: This should not go here - the logic does not belong in the Service Provider.
Instead, return 
`$this->container->make( Event_Tag::class )-> modify_tag_taxonomy_args( $args, $taxonomy );`

and put the logic in the `Event_Tag` class.

Camwyn (changes requested):

Camwyn (approved):

PR #4226, opened 2023-04-25
Add classes and styles and a few other tweaks

codingmusician:
Reference: https://github.com/the-events-calendar/the-events-calendar/pull/4196

This PR changes the inline styling: adds classes and moves CSS to a stylesheet.


src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
```diff
@@ -133,47 +133,75 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
+	 *
+	 * @return void
+	 */
+	public function include_event_link_styles( $parent_template ): void {
+		if ( ! $this->should_show_links( $parent_template, $email_class, $args ) ) {
 			return;
 		}
 
-		$use_ticket_email = tribe_get_option( $email_class->get_option_key( 'use-ticket-email' ), false );
+		tribe( Template::class )->template( 'template-parts/header/head/tec-style', $args, true );
```
juanfra, line 165: For consistency, I'd go with `styles`.

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
```diff
@@ -133,47 +133,75 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
+		}
 
-		// Bail early if the email class is not enabled.
-		if ( ! $email_class->is_enabled() ) {
+		$args['event_gcal_link'] = tribe( Google_Calendar::class )->generate_single_url( $args['event']->ID );
+		$args['event_ical_link'] = tribe_get_single_ical_link( $args['event']->ID );
+
+		if ( ! empty( $args['preview'] ) ) {
+			$args['event_gcal_link'] = '#';
+			$args['event_ical_link'] = '#';
```
juanfra, line 145: The fix for gcal is because gcal doesn't always return a link. We can do it like this, to overwrite the ical to `#`, but it's not "necessary" as the gcal is.

src/Events/Integrations/Modules/Ticket_Emails/Email/Ticket.php
```diff
@@ -125,15 +125,57 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
+	 * Includes event link styles in email body for The Events Calendar Tickets.
+	 *
+	 * @since TBD
+	 *
+	 * @param \Tribe__Template $parent_template Event Tickets template object.
+	 *
+	 * @return void
+	 */
+	public function include_event_link_styles( $parent_template ): void {
+		error_log('include_event_styles');
```
juanfra, line 153: Please remove the `error_log`

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
```diff
@@ -133,47 +133,75 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
+	 * Determines whether or not RSVP should show calendar links.
+	 *
+	 * @since TBD
+	 *
+	 * @param \Tribe__Template $parent_template Event Tickets template object.
+	 *
+	 * @return bool
+	 */
+	public function should_show_links( $parent_template, &$email_class, &$args ): bool {
+		$rsvp_class = $email_class = tribe( RSVP_Email::class );
```
juanfra, line 178: Maybe a note on why we are doing the double assignment?

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
```diff
@@ -133,47 +133,75 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
+	/**
+	 * Determines whether or not RSVP should show calendar links.
+	 *
+	 * @since TBD
+	 *
+	 * @param \Tribe__Template $parent_template Event Tickets template object.
+	 *
+	 * @return bool
+	 */
+	public function should_show_links( $parent_template, &$email_class, &$args ): bool {
```
juanfra, line 177: Why are we passing the last two parameters as reference?

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
```diff
@@ -133,47 +133,75 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
-		$use_ticket_email = tribe_get_option( $email_class->get_option_key( 'use-ticket-email' ), false );
+		tribe( Template::class )->template( 'template-parts/header/head/tec-style', $args, true );
+	}
 
+	/**
+	 * Determines whether or not RSVP should show calendar links.
+	 *
+	 * @since TBD
+	 *
+	 * @param \Tribe__Template $parent_template Event Tickets template object.
```
juanfra, line 173: We're missing two parameters in the docblocs.

src/Events/Integrations/Modules/Ticket_Emails/Email/Ticket.php
```diff
@@ -125,15 +125,57 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
 
+		tribe( Template::class )->template( 'template-parts/header/head/tec-style', $args, true );
+	}
+
+	/**
+	 * Determines whether or not Ticket should show calendar links.
+	 *
+	 * @since TBD
+	 *
+	 * @param \Tribe__Template $parent_template Event Tickets template object.
```
juanfra, line 166: Missing parameters in docblocks.

src/Events/Integrations/Modules/Ticket_Emails/Email/Ticket.php
```diff
@@ -125,15 +125,57 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
+	 *
+	 * @return bool
+	 */
+	public function should_show_links( $parent_template, &$email_class, &$args ): bool {
```
juanfra, line 170: Why are we passing the last two by reference?

src/views/v2/emails/template-parts/header/head/tec-style.php
```diff
@@ -0,0 +1,113 @@
+<style>
```
juanfra, line 1:
```suggestion
<style type="text/css">
```

juanfra (dismissed):

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
codingmusician: When I use `styles` it overrides the `styles` file from ET. That's why I named it differently. Or are you suggesting I use `tec-styles`?

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
codingmusician: For preview, I didn't want it to cause problems if users click on the link. By using `#`, it won't take them to another page. Just a better user experience, I feel.

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
codingmusician: I renamed it `tec-styles`

src/Events/Integrations/Modules/Tickets_Emails/Email/RSVP.php
```diff
@@ -133,47 +133,78 @@ public function include_attachments( $attachments, $email_id, $email_class ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param \Tribe__Template    $parent_template Event Tickets template object.
+	 * @param Email_Abstract|null $email_class     References email class.
+	 * @param array|null          $args            References template context arguments.
+	 *
+	 * @return bool
+	 */
+	public function should_show_links( $parent_template, &$email_class, &$args ): bool {
```
bordoni, line 179: Still confused to why are we passing as reference?

bordoni (changes requested):

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
juanfra: I don't see them being modified in any capacity; there's no need to have them as reference.

juanfra: Ref: https://www.php.net/manual/en/language.references.pass.php

codingmusician: `$email_class` and `$args` are null at first, then created within the `should_show_links()` method, so they technically are being modified.

The reason I did it this way is because if the `$email_class` isn't enabled, the the `$args` aren't pulled from the template. I didn't want to create them, if it wasn't needed. Hopefully that makes sense. If there's a better way, let me know.

src/Events/Integrations/Modules/Tickets_Emails/Email/RSVP.php
codingmusician: The `$email_class` and `$args` are needed within the `should_show_links()` method and within the calling method. Rather than creating these variables in both methods, I referenced them. In some cases, the `$args` variable doesn't need to be created, so I didn't want to create them both, then pass them into the method.

bordoni: That is correct an sound logic, the only problem is with the maintenance of that code. If someone makes changes to the reference it could create fatals that are unwanted. Does it work without passing by reference?

src/Events/Integrations/Modules/Ticket_Emails/Email/RSVP.php
codingmusician: Spoke with @bordoni about this. I was overthinking things. LOL! Thanks!

src/Events/Integrations/Modules/Tickets_Emails/Email/RSVP.php
codingmusician: Got it changed. Thank for being a bouncing board!

bordoni (approved):

PR #4227, opened 2023-04-25
Fixed a few css cases I missed.

stratease:
Couple fixes for QA failure - [TEC-4677](https://theeventscalendar.atlassian.net/browse/TEC-4677)


src/resources/postcss/components/full/_events-bar.pcss
```diff
@@ -37,12 +37,36 @@
...
-	.tribe-events-c-events-bar__search-button {
+	.tribe-events-c-events-bar__search-button,
+	button.tribe-events-c-events-bar__search-button {
+		background-color: transparent;
+		background-image: none;
+		border: none;
+		border-radius: 0;
+		box-shadow: none;
 		display: flex;
+		font-size: 1rem;
```
Camwyn, line 48: We should use a var for font-size unless there is a specific reason not to: https://github.com/the-events-calendar/tribe-common-styles/blob/main/variables/_typography.pcss#L22

stratease: Good point. Aside from a bigger strategy conversation around using relative sizing like `em` and `rem` and responsive font, but I think that is for another ticket :)

Camwyn (approved):

PR #4232, opened 2023-04-27
Check for `$wp_query` before `is_tag()`

BrianHenryIE:
Another plugin is running a bland `get_posts()` and The Events Calendar is filtering `parse_query` and calling `is_tag()` resulting in:

```
PHP Notice:  Function is_tag was called <strong>incorrectly</strong>. Conditional query tags do not work before the query is run. Before then, they always return false. Please see <a href="https://wordpress.org/documentation/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 3.1.0.) in /path/to/wp-includes/functions.php on line 5865
```

This PR checks is the `global $wp_query` set before calling `is_tag()`.

### DIFF
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -81,7 +81,8 @@ class Tribe__Events__Query {
 		// Refresh the query post types: they might have been modified.
 		$query_post_types = (array) $query->get( 'post_type' );
 		// Add our post type if we are on the tag archive - we need to check our global wp_query for this.
-		$on_tag_archive_page = is_tag();
+		global $wp_query;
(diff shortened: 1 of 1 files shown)


bordoni:
Related to https://github.com/the-events-calendar/the-events-calendar/pull/4180

@stratease can you own this PR to completion? We need to combine both what existed before and what @BrianHenryIE suggested here.

stratease:
Thanks @BrianHenryIE 

Opened PR https://github.com/the-events-calendar/the-events-calendar/pull/4234 for this. I think we can close this PR now.

PR #4242, opened 2023-05-05
orderby regex match if contains ending space

Matthewnie:
Fix resulting MYSQL error when the orderby string failed to match the regex sent because it conatins an ending space after the ASC|DESC declaration.

Example: If orderby_frag string ends with a space `wp_posts.ID DESC ` this would fail the REGEX test and the result would become `wp_posts.ID DESC DESC` which results in a MYSQL error.

### DIFF
--- a/src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
+++ b/src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
@@ -694,7 +694,7 @@ class Custom_Tables_Query extends WP_Query {
 			}
 
 			// Each `ORDER BY` entry could specify an order (DESC|ASC) or not.
-			if ( preg_match( '~(?<orderby>.*?)\s?(?<order>ASC|DESC)$~i', $orderby_frag, $m ) ) {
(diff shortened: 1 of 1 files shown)


src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
```diff
@@ -694,7 +694,7 @@ public function redirect_posts_orderby( $posts_orderby, $query ) {
 			}
 
 			// Each `ORDER BY` entry could specify an order (DESC|ASC) or not.
-			if ( preg_match( '~(?<orderby>.*?)\s?(?<order>ASC|DESC)$~i', $orderby_frag, $m ) ) {
+			if ( preg_match( '~(?<orderby>.*?)\s?(?<order>ASC|DESC)\s?$~i', $orderby_frag, $m ) ) {
```
therealgilles, line 697: Shouldn't the regexp use `\s*` instead?

therealgilles:
I had pointed out this bug here: https://github.com/the-events-calendar/ql-events/issues/49

therealgilles:
I first found this issue back in December 2022. Why does it take so long for TEC to fix these bugs?

PR #4247, opened 2023-05-10
FF Link in Month View.

EugeneKyale:
WIP 🚧 not ready for review.

Ensure the fast forward link in month view is only displayed when there are upcoming events in the context of the current view.


EugeneKyale:
Fix to be applied in FBAR

PR #4263, opened 2023-05-17
Fix meta noindex on by-day views.

Camwyn:
Based off of https://github.com/the-events-calendar/the-events-calendar/pull/4262

This adds a method for grabbing one event based on the view _date_ restrictions to see if we should add a noindex meta tag.

It's not perfect but it's filterable and better than the current implementation.

Related ECP piece: https://github.com/the-events-calendar/events-pro/pull/2281

[TEC-4717]

[TEC-4717]: https://theeventscalendar.atlassian.net/browse/TEC-4717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Views/V2/Views/Month_View.php
```diff
@@ -785,4 +785,27 @@ protected function get_mobile_messages() {
...
+	 *
+	 * @param Tribe__Repository|false $events     The events repository. False by default.
+	 * @param DateTime                $start_date The start date (object) of the query.
+	 * @param Tribe__Context          $context    The current context.
+	 *
+	 * @return Tribe__Repository|false $events     The events repository results.
+	 */
+	public static function get_noindex_events( $events, $start_date ) {
+		$interval = new \DateInterval( 'P1M' );
+
```
Camwyn, line 802: Potential here to add the view's repository args as well 🤔

src/Tribe/Views/V2/Views/Day_View.php
```diff
@@ -398,4 +398,25 @@ protected function setup_messages( array $events ) {
...
+	 * @since TBD
+	 *
+	 * @param Tribe__Repository|false $events     The events repository. False by default.
+	 * @param DateTime                $start_date The start date (object) of the query.
+	 * @param Tribe__Context          $context    The current context.
+	 *
+	 * @return Tribe__Repository|false $events     The events repository results.
+	 */
+	public static function get_noindex_events( $events, $start_date ) {
+		return tribe_events()
```
stratease, line 414: Not a big performance hit, but might be worth adding cacheability to it?
```
$val = ( new Cache() )->get( $cache_key, Cache_Listener::TRIGGER_SAVE_POST )
if($val) {
  return $val;
}

$cache_me = tribe_events()
// ....
```

src/Tribe/Views/V2/Views/Month_View.php
stratease: Yeah, shouldn't this run after the view has gotten it's events, and just reference those results? Is that possible (the view hook fires in time for the noindex meta tag hooks)? Would avoid the guesswork.

Maybe I'm missing something?

stratease (changes requested):
Some ideas.

src/Tribe/Views/V2/Views/Month_View.php
Camwyn: It can't run after the view query - that doesn't happen until the content starts.
I am doing something to utilize the existing args though, have to test.

src/Tribe/Main.php
```diff
@@ -1509,11 +1581,22 @@ public function issue_noindex() {
...
 		 */
 		public function print_noindex_meta() {
-			echo ' <meta name="robots" content="noindex,follow" />' . "\n";
+			$noindex_meta = ' <meta name="robots" id="tec_noindex" content="noindex,follow" />' . "\n";
+
+			/**
+			 * Filters the noindex meta tag.
+			 *
+			 * @since TBD
+			 *
```
stratease, line 1594: `@param ...?`

src/Tribe/Views/V2/Hooks.php
```diff
@@ -1130,6 +1132,24 @@ public function filter_view_month_today_button_title( $label, $view ) {
 		return $label;
 	}
 
+	/**
+	 * Allow specific views to hook in and add their own calculated events.
+	 *
+	 * @since TBD
+	 *
+	 * @param Tribe__Repository|false $events     The events repository. False by default.
+	 * @param DateTime                $start_date The start date (object) of the query.
```
stratease, line 1141: Looks like some doc formatting TLC here.

stratease (changes requested):

src/Tribe/Main.php
```diff
@@ -1488,18 +1506,72 @@ public function issue_noindex() {
...
+			 */
+			$events = apply_filters( 'tec_events_noindex', null, $start_date, $context );
+
+			$foo = null;
+
+			// If nothing has hooked in ($events is boolean false), we assume a list-style view (no end-date limiter)
+			//  with no params and do a quick query for a single event after the start date.
+			if ( null === $events ) {
+				$cache     = new Tribe__Cache();
```
lucatume, line 1532: Use `tribe_cache()` to get the singleton instance from the Service Locator.
That instance will memorize things over time a new instance will need to recalculate again.

src/Tribe/Main.php
```diff
@@ -1488,18 +1506,72 @@ public function issue_noindex() {
...
+				$trigger   = Tribe__Cache_Listener::TRIGGER_SAVE_POST;
+				$cache_key = $cache->make_key(
+					[
+						'context' => $context,
```
lucatume, line 1536: The `Tribe__Cache::make_key` method will `serialize` what you pass it.
The `Tribe__Context` object is a large one; by serializing it you're getting a very large string that has to moved back and forth. Furthermore, the serialization of the context will eagerly resolve **all its locations**.
Pick the parts of the context you care about for this, build your own hashed key (reduce to a fixed length) and use to store the value.

src/Tribe/Main.php
```diff
@@ -1468,15 +1469,32 @@ public function get_event_label_singular() {
...
 		 *     add_filter( 'tribe_events_add_no_index_meta', '__return_true' );
+		 *
+		 *  Always adding the noindex meta tag for a specific event view is possible with:
+		 *
+		 *     add_filter( "tribe_events_{$view}_add_no_index_meta", '__return_true' );
+		 *
+		 *  Where `$view` above is the view slug, e.g. `month`, `day`, `list`, etc.
 		 */
 		public function issue_noindex() {
+			global $wp_query;
```
lucatume, line 1487: Excellent chance to refactor this out of `Main` into a specialized class.

src/Tribe/Main.php
```diff
@@ -1488,18 +1506,72 @@ public function issue_noindex() {
...
+
+				if ( ! $events ) {
+					$events = tribe_events()->per_page( 1 )->where( 'starts_after', $start_date->format( Tribe__Date_Utils::DBDATEFORMAT ) );
+
+					$cache->set( $cache_key, $events, \Tribe__Cache::NO_EXPIRATION, $trigger );
+				}
+			}
+
+			// No posts = no index.
+			$derp = $events->found();
```
lucatume, line 1553: Remove this.

src/Tribe/Main.php
```diff
@@ -1488,18 +1506,72 @@ public function issue_noindex() {
...
+			 * @since TBD
+			 *
+			 * @param ?Tribe__Repository|null $events     The events repository. False if not hooked in to.
+			 * @param DateTime                $start_date The start date (object) of the query.
+			 * @param Tribe__Context          $context    The current context.
+			 *
+			 */
+			$events = apply_filters( 'tec_events_noindex', null, $start_date, $context );
+
+			$foo = null;
```
lucatume, line 1527: Remove this.

src/Tribe/Main.php
```diff
@@ -1488,18 +1506,72 @@ public function issue_noindex() {
...
+						'view'    => $view,
+						'start'   => $start_date->format( Tribe__Date_Utils::DBDATEFORMAT ),
+					],
+					'tec_noindex_'
+				);
+
+				$events = $cache->get( $cache_key, $trigger );
+
+				if ( ! $events ) {
+					$events = tribe_events()->per_page( 1 )->where( 'starts_after', $start_date->format( Tribe__Date_Utils::DBDATEFORMAT ) );
```
lucatume, line 1546: I think this should be `...->where( 'ends_after', ... )`.
Events that started yesterday and are still running would show in the view.

src/Tribe/Main.php
```diff
@@ -1468,15 +1469,32 @@ public function get_event_label_singular() {
...
+		 *  Always adding the noindex meta tag for a specific event view is possible with:
+		 *
+		 *     add_filter( "tribe_events_{$view}_add_no_index_meta", '__return_true' );
+		 *
+		 *  Where `$view` above is the view slug, e.g. `month`, `day`, `list`, etc.
 		 */
 		public function issue_noindex() {
+			global $wp_query;
+
+			$do_noindex_meta = apply_filters( 'tec_events_add_no_index_meta_tag', true );
```
lucatume, line 1489: Add filter doc-block.

src/Tribe/Main.php
```diff
@@ -1488,18 +1506,72 @@ public function issue_noindex() {
...
+			// If nothing has hooked in ($events is boolean false), we assume a list-style view (no end-date limiter)
+			//  with no params and do a quick query for a single event after the start date.
+			if ( null === $events ) {
+				$cache     = new Tribe__Cache();
+				$trigger   = Tribe__Cache_Listener::TRIGGER_SAVE_POST;
+				$cache_key = $cache->make_key(
+					[
+						'context' => $context,
+						'view'    => $view,
+						'start'   => $start_date->format( Tribe__Date_Utils::DBDATEFORMAT ),
```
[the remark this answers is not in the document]
lucatume: The cache trigger should take care of that.

src/Tribe/Main.php
```diff
@@ -1488,18 +1506,72 @@ public function issue_noindex() {
...
+			 * This *bypasses* the cached query immediately after it.
+			 *
+			 * @since TBD
+			 *
+			 * @param ?Tribe__Repository|null $events     The events repository. False if not hooked in to.
+			 * @param DateTime                $start_date The start date (object) of the query.
+			 * @param Tribe__Context          $context    The current context.
+			 *
+			 */
+			$events = apply_filters( 'tec_events_noindex', null, $start_date, $context );
```
lucatume, line 1525: Since this piece of code just cares about adding a no-index key or not, why not filter a `boolean` value instead of moving around the code a repository?

src/Tribe/Main.php
```diff
@@ -1488,18 +1506,72 @@ public function issue_noindex() {
...
+					],
+					'tec_noindex_'
+				);
+
+				$events = $cache->get( $cache_key, $trigger );
+
+				if ( ! $events ) {
+					$events = tribe_events()->per_page( 1 )->where( 'starts_after', $start_date->format( Tribe__Date_Utils::DBDATEFORMAT ) );
+
+					$cache->set( $cache_key, $events, \Tribe__Cache::NO_EXPIRATION, $trigger );
```
lucatume, line 1548: This cannot be done.
Beside the fact I think this could be just a `boolean` value, objects that are not built-in PHP objects should **never** go to cache.
Most advanced pre-fetching solutions will pre-fetch when only PHP is loaded, not even WordPress is: this would result in a fatal error.

lucatume (changes requested):

src/Tribe/Main.php
Camwyn: A boolean "are there events?"

src/Tribe/Main.php
Camwyn: See `src/Events/SEO/No_Index.php`

src/Events/SEO/No_Index.php
```diff
@@ -0,0 +1,204 @@
...
+  * @since   6.1.0
+  * @package TEC\Events\Telemetry
+  */
+class No_Index {
+	/**
+	 * Runs on the "wp" action. Inspects the main query object and if it relates to an events
+	 * query makes a decision to add a noindex meta tag based on whether events were returned
+	 * in the query results or not.
+	 *
+	 * @since ??
```
lucatume, line 26: This needs filling, if you cannot find the original implementation, then just drop this `@since` tag.

src/Events/SEO/No_Index.php
```diff
@@ -0,0 +1,204 @@
...
+		if ( $add_noindex ) {
+			add_action( 'wp_head', [ $this, 'print_noindex_meta' ] );
+		}
+	}
+
+	/**
+	 * Prints a "noindex,follow" robots tag.
+	 *
+	 * @since TBD
+	 *
```
lucatume, line 142: Remove this line.

src/Events/SEO/No_Index.php
```diff
@@ -0,0 +1,204 @@
...
+	 */
+	public function print_noindex_meta() {
```
lucatume, line 144:
```suggestion
	public function print_noindex_meta() :void {
```

src/Events/SEO/No_Index.php
```diff
@@ -0,0 +1,204 @@
...
+	 *
+	 *     add_filter( 'tribe_events_add_no_index_meta', '__return_true' );
+	 *
+	 *  Always adding the noindex meta tag for a specific event view is possible with:
+	 *
+	 *     add_filter( "tribe_events_{$view}_add_no_index_meta", '__return_true' );
+	 *
+	 *  Where `$view` above is the view slug, e.g. `month`, `day`, `list`, etc.
+	 */
+	public function issue_noindex() {
```
lucatume, line 43:
```suggestion
	public function issue_noindex() :void {
```

src/Events/SEO/No_Index.php
```diff
@@ -0,0 +1,204 @@
...
+						'id'      => true,
+						'name'    => true,
+						'content' => true,
+					],
+				]
+			);
+		}
+
+		/**
+		 * Undocumented function
```
lucatume, line 169:
```suggestion
		 * Returns the end date time object read from the current context.
```

src/Events/SEO/Provider.php
```diff
@@ -0,0 +1,37 @@
...
+ * @package TEC\Events\Legacy\Views\V1
+ */
+class Provider extends Service_Provider {
+	public function register() {
+		$this->container->singleton( static::class, $this );
+
+		$this->hooks();
+	}
+
+	public function hooks() {
```
lucatume, line 30: No need to break out the hooking of one action into a dedicated hooking method, just leave it in `register`.
While I'm there, I suggest to make this a Controller and integrate the functions currently in the `No_Index` class in here.

src/Events/SEO/No_Index.php
```diff
@@ -0,0 +1,204 @@
...
+
+use \Tribe__Date_Utils as Dates;
+
+ /**
+  * Class Provider
+  *
+  * @since   6.1.0
+  * @package TEC\Events\Telemetry
+  */
+class No_Index {
```
lucatume, line 20: I would integrate this code, as suggested in a comment on the Service Provider, in a Controller and reduce the abstractions.
Also: test coverage?

src/Tribe/Views/V2/Views/List_View.php
```diff
@@ -318,4 +318,41 @@ protected function on_page_reset() {
...
+	 * @since TBD
+	 *
+	 * @param Tribe__Repository|false $events     The events repository. False by default.
+	 * @param DateTime                $start_date The start date (object) of the query.
+	 * @param Tribe__Context          $context    The current context.
+	 *
+	 * @return Tribe__Repository|false $events     The events repository results.
+	 */
+	public function get_noindex_events( $events, $start_date, $end_date, $context ) {
+		if ( null === $events ) {
```
lucatume, line 334: Here and elsewhere: refactor for early return.

```suggestion
		if ( null !== $events ) { return; }
		
		// ... do something.
```

lucatume (changes requested):

src/Tribe/Main.php
Camwyn: By passing the repo we allow modifying it - ex: via FBAR rather than just redoing the query and overwriting the value

stratease:
Failing test - thanks for adding the noindex test!

bordoni (approved):

PR #4274, opened 2023-05-25
[ET-1757] Enhance - Make Event title clickable with link to the event.

rafsuntaskin:
[ET-1757]

Screenshot: https://share.cleanshot.com/DwTdnlcg

[ET-1757]: https://theeventscalendar.atlassian.net/browse/ET-1757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/integrations/event-tickets/emails/template-parts/body/event/title.php
+++ b/src/views/integrations/event-tickets/emails/template-parts/body/event/title.php
@@ -28,10 +28,10 @@ if ( empty( $event->title ) ) {
 <tr>
 	<td class="tec-tickets__email-table-content-event-title-container">
 		<h3 class="tec-tickets__email-table-content-event-title">
-			<a href="<?php echo $event->permalink; ?>" target="_blank">
+			<a href="<?php echo esc_url( $event->permalink ); ?>" target="_blank">
 				<?php
 				// phpcs:ignore
-				echo $event->title;
+				echo esc_html( $event->title );
 				?>
 			</a>
 		</h3>
--- a/src/views/integrations/event-tickets/emails/template-parts/body/event/title.php
+++ b/src/views/integrations/event-tickets/emails/template-parts/body/event/title.php
@@ -28,10 +28,12 @@ if ( empty( $event->title ) ) {
 <tr>
 	<td class="tec-tickets__email-table-content-event-title-container">
 		<h3 class="tec-tickets__email-table-content-event-title">
-			<?php
-			// phpcs:ignore
-			echo $event->title;
-			?>
+			<a href="<?php echo $event->permalink; ?>" target="_blank">
+				<?php
+				// phpcs:ignore
+				echo $event->title;
+				?>
+			</a>
 		</h3>
 	</td>
 </tr>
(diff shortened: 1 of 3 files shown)


redscar (approved):

src/views/integrations/event-tickets/emails/template-parts/body/event/title.php
```diff
@@ -28,10 +28,12 @@
 <tr>
 	<td class="tec-tickets__email-table-content-event-title-container">
 		<h3 class="tec-tickets__email-table-content-event-title">
-			<?php
-			// phpcs:ignore
-			echo $event->title;
-			?>
+			<a href="<?php echo $event->permalink; ?>" target="_blank">
```
codingmusician, line 31: Need to escape
```suggestion
			<a href="<?php echo esc_attr( $event->permalink ); ?>" target="_blank">
```

codingmusician: Not sure if `esc_url()` would handle `#` correctly. If it does, the `esc_url()` would be better.

src/views/integrations/event-tickets/emails/template-parts/body/event/title.php
```diff
@@ -28,10 +28,12 @@
...
+				<?php
+				// phpcs:ignore
+				echo $event->title;
```
codingmusician, line 34: Needs escape:
```suggestion
				esc_html_e( $event->title );
```

codingmusician (approved):

src/views/integrations/event-tickets/emails/template-parts/header/head/tec-styles.php
```diff
@@ -170,4 +170,8 @@
 	td.tec-tickets__email-table-content-event-venue-website-container {
 		padding: 0;
 	}
+	
+	h3.tec-tickets__email-table-content-event-title a {
+		text-decoration: none;
```
juanfra, line 175: Should we also set the color so that it's not defaulting to the document default? This could probably be something when viewing the email in some clients.

src/views/integrations/event-tickets/emails/template-parts/body/event/title.php
```diff
@@ -28,10 +28,12 @@
 <tr>
 	<td class="tec-tickets__email-table-content-event-title-container">
 		<h3 class="tec-tickets__email-table-content-event-title">
-			<?php
-			// phpcs:ignore
-			echo $event->title;
-			?>
+			<a href="<?php echo esc_url( $event->permalink ); ?>" target="_blank">
```
juanfra, line 31: if we're using `target="_blank"` we also need to accompany that with `rel="noopener noreferrer"`.

src/views/integrations/event-tickets/emails/template-parts/header/head/tec-styles.php
rafsuntaskin: Yes, but there should be a default text color coming from Event Tickets for all regular texts, right? Do we have something like that in ET?

PR #4277, opened 2023-05-26
Tickets Emails: Add Event Title to Order Emails

codingmusician:
### Issues

- TE-8ADA
- TE-DC94

### Description

The event titles were initially added into ET, but were set to be moved to TEC eventually. That never happened, so this PR handles all of that.

### Abstract

https://www.loom.com/share/8dc76ea87ad2419490a714ba6a2514d7

### DIFF
--- a/src/Events/Integrations/Plugins/Event_Tickets/Emails/Email/Completed_Order.php
+++ b/src/Events/Integrations/Plugins/Event_Tickets/Emails/Email/Completed_Order.php
@@ -2,7 +2,7 @@
 /**
  * Class Completed_Order.
  *
- * @since   TBD
+ * @since TBD
  *
  * @package TEC\Events\Integrations\Plugins\Event_Tickets\Emails
  */
@@ -15,7 +15,7 @@ use TEC\Tickets\Emails\Email\Completed_Order as Completed_Order_Email;
 /**
  * Class Completed_Order.
  *
- * @since   TBD
+ * @since TBD
  *
  * @package TEC\Events\Integrations\Plugins\Event_Tickets
  */
--- a/src/Events/Integrations/Plugins/Event_Tickets/Emails/Email/Completed_Order.php
+++ b/src/Events/Integrations/Plugins/Event_Tickets/Emails/Email/Completed_Order.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Class Completed_Order.
+ *
+ * @since   TBD
+ *
+ * @package TEC\Events\Integrations\Plugins\Event_Tickets\Emails
+ */
+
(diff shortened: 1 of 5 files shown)


src/views/integrations/event-tickets/emails/template-parts/header/head/tec-styles.php
```diff
@@ -174,4 +174,11 @@
 	h3.tec-tickets__email-table-content-event-title a {
 		text-decoration: none;
 	}
+	
+	.tec-tickets__email-table-content-order-event-title,
```
rafsuntaskin, line 178: I think we should not use different selectors for the `Same Event Title` in different emails. As, this will keep making the styles tag long and these would be carried over all emails even if not used. 

cc: @codingmusician @juanfra

src/Events/Integrations/Plugins/Event_Tickets/Emails/Email/Completed_Order.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * Class Completed_Order.
+ *
+ * @since   TBD
```
rafsuntaskin, line 5:
```suggestion
 * @since TBD
```
Let's do a search replace for this change.

src/views/integrations/event-tickets/emails/template-parts/header/head/tec-styles.php
juanfra: Agree.

codingmusician: So what do we do in the case where the styling for the event title is different? The design for the event title in the Completed Order and Purchase Receipt emails is different than the styling for the Event Title in the Ticket and RSVP emails. This is why I was differentiating the class as the "order" event title.

juanfra: Sorry, I misread the initial comment. We _should_ use different class names. My bad, sorry for that!

src/views/integrations/event-tickets/emails/template-parts/body/order/event-title.php
```diff
@@ -0,0 +1,43 @@
...
+	return;
+}
+if ( empty( $event->title ) ) {
+	return;
+}
+
+?>
+<tr>
+	<td class="tec-tickets__email-table-content-order-event-title">
+			<a href="<?php echo esc_url( $event->permalink ); ?>" target="_blank">
```
rafsuntaskin, line 37: let's add the `rel` attribute like Juan suggested here: https://github.com/the-events-calendar/the-events-calendar/pull/4274#discussion_r1206753874

codingmusician:
Closing due to moving post titles into ET for all emails.

PR #4279, opened 2023-05-29
Bug fix – prevent duplicate JOIN statement in query

therealgilles:
Back in December 2022, I had noticed that sometimes the related JOIN statement appeared multiple times in the query, creating a WP database error. I created this workaround, not knowing what the proper fix was. I do not know if this problem has since been fixed some other way.


therealgilles:
This was finally fixed in v6.1.1. It only took 7 months.

PR #4283, opened 2023-05-30
Block Editor Conflict with Map plugins.

EugeneKyale:
Ticket : [TEC-4810](https://theeventscalendar.atlassian.net/browse/TEC-4810)

Avoid the issue where TEC blocks would break when either the `WP Go Maps`, `MapPress Google Maps` and `Leaflet Maps` plugins were active on a site.

In the original code, the `geocoder` variable was declared globally and instantiated only once during module import. This caused the `geocoder` instance to retain its state between different method invocations, resulting in inaccurate geocoding results. By moving the `geocoder` declaration inside the `setCoordinatesState` method, a new `geocoder` instance is created each time the method is called, ensuring accurate and up-to-date geocoding for different addresses.

**Screenshot 📸**

image

[TEC-4810]: https://theeventscalendar.atlassian.net/browse/TEC-4810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EugeneKyale:
@stratease I'm setting you as a reviewer because you're familiar with this portion of the code.

stratease:
> @stratease I'm setting you as a reviewer because you're familiar with this portion of the code.

Yup - looks good. Makes sense considering 3rd party plugins using the same library - although I hope we aren't bleeding state into their modules too.

stratease (dismissed):

EugeneKyale:
@stratease needs a re-approval after updating the base branch.

PR #4290, opened 2023-06-01
Fix for multidaycutoff and all day event updates

stratease:
[TEC-4748](https://theeventscalendar.atlassian.net/browse/TEC-4748)

There was some data consistency issues after 6.0 regarding all day events and `End of day cutoff` setting updates. This would show up with events spanning incorrectly interpreted cut off times.

**This is doing a mass update of All Day events' times** to sync them to the correct interpretation, which should not be effected by the multiday cut off. 

This is also removing the sync/fix All Day end times that happen when the `multidaycutoff` time is changed.

Artifact:
https://share.getcloudapp.com/nOuKJk2v

[TEC-4748]: https://theeventscalendar.atlassian.net/browse/TEC-4748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Views/V2/Views/By_Day_View.php
```diff
@@ -275,40 +274,45 @@ public function get_grid_days( $date = null, $force = false ) {
...
+				$multiday_start = tribe_beginning_of_day( $day->format( Dates::DBDATETIMEFORMAT ) );
+				$multiday_end   = tribe_end_of_day( $day->format( Dates::DBDATETIMEFORMAT ) );
+				$start =  $day->format( 'Y-m-d 00:00:00' ) ;
+				$end   =  $day->format( 'Y-m-d 23:59:59' ) ;
+// @todo Not confident on below changes... site timezone vs event timezone vs timezones assumed on the month view (should be WP or site tz?) and multiday cut off?
+				// Events overlap a day if Event start date <= Day End AND Event end date > Day Start.
 				$results_in_day = array_filter(
 					$day_results,
-					static function ( $event ) use ( $start, $end, $use_site_timezone, $site_timezone, $utc ) {
+					static function ( $event ) use ( $multiday_start, $multiday_end, $start, $end, $use_site_timezone, $site_timezone, $utc ) {
```
stratease, line 293: @bordoni @Camwyn 

For this callback, I was hoping someone could add clarity on it's logic. 

This seems off to me - here we don't use timezones (A):
```
					static function ( $event ) use ( $multiday_start, $multiday_end, $start, $end, $use_site_timezone, $site_timezone, $utc ) {
```
but here we do (B):
```
					static function ( $event ) use ( $multiday_start, $multiday_end, $start, $end, $use_site_timezone, $site_timezone, $utc ) {
```
I don't understand why we do in A but not in B? Seems like a bug...?

bordoni (approved):
Let's not merge this right away, this feels like it deserves a QA in Branch. By Day View changes of this nature always have weird side effects. Do you mind getting QA to do a Full Pass of the affected views so we can be sure we are not breaking in odd scenarios?

src/Tribe/Updater.php
```diff
@@ -50,10 +50,44 @@ public function do_updates() {
...
+		 *
+		 * This is fired immediately before the version change.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $version_option  The version option being used for this plugin. This is necessary to know which plugin is updating.
+		 * @param string $new_version     The new version number.
+		 * @param string $current_version The current version number.
+		 */
+		do_action( 'tec_events_plugin_updater_version_change',
```
jesseeproductions, line 72: What about adding the version option to the action name? 

Or adding a 2nd hook that does that?

So VE would have tec_events_plugin_updater_version_change_events-virtual-schema-version|
ET+ : event-tickets-plus-schema-version

That would make it a it easier to target specific plugins.

src/Tribe/Views/V2/Views/By_Day_View.php
```diff
@@ -715,17 +693,12 @@ protected function build_url_for_date( $date, $canonical, array $passthru_vars =
 	 * @return array<int,\stdClass> A map from each event Post ID to the value object that will represent
 	 *                              the Event ID, start date, end date and timezone.
 	 */
-	protected function prepare_day_results( array $view_event_ids, $use_site_timezone ) {
+	protected function prepare_day_results( array $view_event_ids ) {
```
Camwyn, line 696: deprecate the removed param.
Also: add a `@since` to the docblock for the change.

tests/wpunit/TEC/Events/MigrationsTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+		// We rely on this to validate when to run migrations.
+		$this->assertFalse( $updater->is_version_in_db_less_than( $past_version ) );
+		$this->assertFalse( $updater->is_version_in_db_less_than( $current_version ) );
+		$this->assertTrue( $updater->is_version_in_db_less_than( $future_version ) );
+
+		// Migration should run - but only once.
+		$did_schedule_event = false;
+		$allday_hook        = 'tec_events_migrate_all_day_eod_times'; // Hook for one of our first migrations in this setup.
+		add_filter( 'pre_schedule_event', function ( $pre, $event, $wp_error ) use ( &$did_schedule_event, $allday_hook ) {
```
Camwyn, line 27: Remove the filter after using it - in case cleanup doesn't 😉

Camwyn (changes requested):

src/Tribe/Views/V2/Views/By_Day_View.php
```diff
@@ -191,134 +190,120 @@ public function get_grid_days( $date = null, $force = false ) {
...
 						// * convert it to the current site timezone
 						// * check if the event fits into the day, given shifted start and end of day
-						$event_localized_start_date = Dates::build_date_object( $event->start_date, $utc )
-						                                   ->setTimezone( $site_timezone );
-						$event_localized_end_date   = Dates::build_date_object( $event->end_date, $utc )
-						                                   ->setTimezone( $site_timezone );
-
-						return $event_localized_start_date->format( Dates::DBDATETIMEFORMAT ) <= $end
-						       && $event_localized_end_date->format( Dates::DBDATETIMEFORMAT ) >= $start;
+						$event_localized_start_date = Dates::build_date_object( $event->start_date, $event->timezone ) // @todo but we do here?
```
jesseeproductions, line 289: Not sure on the @todo reference. Is that relevant still or can be removed?

src/Tribe/Views/V2/Views/By_Day_View.php
jesseeproductions, line 696: Does the param have to be removed from the method docblock?

jesseeproductions (commented):
Looks good, with some questions and docblock changes.

src/Events/Custom_Tables/V1/Views/V2/By_Day_View_Compatibility.php
```diff
@@ -39,22 +39,17 @@ public function prepare_day_results( array $ids = [] ) {
...
-		$start_date_prop   = $use_site_timezone ? 'start_date_utc' : 'start_date';
-		$end_date_prop     = $use_site_timezone ? 'end_date_utc' : 'end_date';
-
 		$prepared = [];
 
 		/** @var Occurrence $occurrence */
 		foreach (
-			Occurrence::order_by( $start_date_prop, 'ASC' )
-			          ->find_all( $ids, 'post_id' ) as $occurrence
+			Occurrence::find_all( $ids, 'post_id' ) as $occurrence
```
borkweb, line 46: Why the removal of the order by? Won't this cause the prepared set of occurrences to be collected in a somewhat arbitrary manner?

src/Events/Migrations/Provider.php
```diff
@@ -0,0 +1,101 @@
...
+					ON (pm1.post_id = pm2.post_id AND pm2.meta_key = '_EventAllDay' AND pm2.meta_value = 'yes')
+				INNER JOIN $wpdb->postmeta pm3
+					ON (pm1.post_id = pm3.post_id AND pm3.meta_key = '_EventStartDate')
+				INNER JOIN $wpdb->postmeta pm4
+					ON (pm1.post_id = pm4.post_id AND pm4.meta_key = '_EventDuration')
+				SET pm1.meta_value = DATE_FORMAT( DATE_ADD( pm3.meta_value, INTERVAL pm4.meta_value SECOND ), '%Y-%m-%d %H:%i:%s' )
+				WHERE pm1.meta_key = '_EventEndDate'";
+		$wpdb->query( $fix_start_dates );
+		$wpdb->query( $fix_end_dates );
+	}
```
borkweb, line 99: By moving this away from the `add_action( 'update_option_' . Tribe__Main::OPTIONNAME, [ $this, 'fix_all_day_events' ], 10, 2 );`, we'll lose the ability to react to someone updating their end of day cutoff and having it reflected in their already existing events without triggering a migration.

Instead, should this migration merely trigger the action instead and keep the logic for updating in its current location (or a better location out of Main but tied to the same action?)

src/Tribe/Main.php
```diff
@@ -1585,52 +1586,6 @@ public function safe_redirect_domains() {
...
-					ON (pm1.post_id = pm2.post_id AND pm2.meta_key = '_EventAllDay' AND pm2.meta_value = 'yes')
-				INNER JOIN $wpdb->postmeta pm3
-					ON (pm1.post_id = pm3.post_id AND pm3.meta_key = '_EventStartDate')
-				INNER JOIN $wpdb->postmeta pm4
-					ON (pm1.post_id = pm4.post_id AND pm4.meta_key = '_EventDuration')
-				SET pm1.meta_value = DATE_ADD(pm3.meta_value, INTERVAL pm4.meta_value SECOND )
-				WHERE pm1.meta_key = '_EventEndDate'";
-			$wpdb->query( $fix_start_dates );
-			$wpdb->query( $fix_end_dates );
-		}
```
borkweb, line 1632: We can't outright remove this method since it is public. If we _are_ moving the logic, we'll need to deprecate this method and point to the new location of the logic.

src/Events/Custom_Tables/V1/Views/V2/By_Day_View_Compatibility.php
stratease: @borkweb Good question. The why, is to just help lighten the query load / optimize speed.

From following this logic, the order is completely ignored. We end up [looping on ALL of the items](https://github.com/the-events-calendar/the-events-calendar/blob/master/src/Tribe/Views/V2/Views/By_Day_View.php#L292), to create an array of "which events" show on a particular day. 

Am I missing something else? (definitely see more opportunity to optimize this in the future)

src/Tribe/Main.php
stratease: Moving this to a `Provider` -> `Sync_All_Day_Dates` service.

src/Tribe/Main.php
```diff
@@ -1585,52 +1586,6 @@ public function safe_redirect_domains() {
 			return $domains;
 		}
 
-		/**
-		 * Updates the start/end time on all day events to match the EOD cutoff
-		 *
-		 * @see 'update_option_'.Tribe__Main::OPTIONNAME
-		 */
-		public function fix_all_day_events( $old_value, $new_value ) {
```
stratease, line 1593: This was moved to a service provider.

stratease:
This PR has gone through major changes. Please review again.

src/Events/Updates/Provider.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace TEC\Events\Updates;
+
+use tad_DI52_ServiceProvider as Service_Provider;
+use TEC\Events\Custom_Tables\V1\Provider_Contract;
+use Tribe__Main;
+
+class Provider extends Service_Provider implements Provider_Contract {
```
lucatume, line 9: Doc-block missing.

src/Events/Updates/Provider.php
```diff
@@ -0,0 +1,35 @@
...
+	 * Registers hooks for TEC Updates.
+	 *
+	 * @since TBD
+	 */
+	public function register() {
+		add_action( 'update_option_' . Tribe__Main::OPTIONNAME, [ $this, 'on_cutoff_change_fix_all_day_events' ], 10, 2 );
+	}
+
+	/**
+	 * Removes hooks.
```
lucatume, line 21: Since annotation missing.

src/Events/Updates/Provider.php
lucatume: This is a provider that hooks on one action another class that has one method.
Make this a controller and simplify the code.
It can be broken down if, and when, required. 
Redirect tests on the controller, extend the controller test case (e.g. [this](https://github.com/the-events-calendar/event-tickets/blob/b681bdd577b69dd617a1d05d97c150f0a4dd011b/tests/ft_integration/TEC/Tickets/Flexible_Tickets/BaseTest.php#L18)).

lucatume (changes requested):
Compact the code into a Controller, test the controller.

EugeneKyale (dismissed):

PR #4295, opened 2023-06-05
Fix/freemius migration

Camwyn:
Fix timing issues with modal.

Requires: https://github.com/the-events-calendar/tribe-common/pull/1934

### DIFF
--- a/src/Events/Telemetry/Telemetry.php
+++ b/src/Events/Telemetry/Telemetry.php
@@ -11,6 +11,8 @@ namespace TEC\Events\Telemetry;
 
 use TEC\Common\StellarWP\Telemetry\Config;
 use TEC\Common\StellarWP\Telemetry\Opt_In\Status;
+use TEC\Common\StellarWP\Telemetry\Opt_In\Opt_In_Template;
+
 use TEC\Common\Telemetry\Telemetry as Common_Telemetry;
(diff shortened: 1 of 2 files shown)


src/Events/Telemetry/Telemetry.php
```diff
@@ -228,14 +230,20 @@ public function inject_modal_link() {
...
+		if ( ! $show ) {
+			return;
+		}
 		/**
 		 * Fires to trigger the modal content on admin pages.
 		 *
 		 * @since TBD
 		 */
-		do_action( "stellarwp/telemetry/{$telemetry_slug}/optin" );
+		do_action( 'tec-telemetry-modal', $telemetry_slug );
```
bordoni, line 246: why not `tec_telemetry_modal_slug`?

bordoni (changes requested):

src/Events/Telemetry/Telemetry.php
Camwyn: We trigger `stellarwp/telemetry/{$telemetry_slug}/optin` in common, but rely on the plugins themselves to tell us _where_ they want it - ie in the TEC settings pages

Could change it to `tec_telemetry_modal` though (still no slug - it's meant to be generic

bordoni (approved):

PR #4301, opened 2023-06-14
Use legacy \tad_DI52_ServiceProvider to prevent issues caused by class_exists on this provider in other plugins

borkweb:
This prevents issues on versions of ET < 5.5.2 where ET common is loaded but still checking the existence of classes that require TEC common.


huubl:
Likely related:
https://github.com/mt-support/tec-labs-google-maps-api-key-restrictions/issues/1

Camwyn:
@borkweb is this still needed?

borkweb:
Nope! Deleted

PR #4320, opened 2023-07-06
Fix some deprecation notices and update common

Camwyn:


src/Events/Custom_Tables/V1/Migration/Reports/Event_Report.php
```diff
@@ -526,11 +526,10 @@ public function __get( $prop ) {
 	 * The JSON serializer logic.
 	 *
 	 * @since 6.0.0
-	 *
-	 * @return mixed
 	 */
+	#[\ReturnTypeWillChange]
```
Camwyn, line 530: just adding `: mixed` to the signature breaks some of our tests, so I patched it this way. Prevents deprecation notices on PHP 8.1

stratease (dismissed):

bordoni (approved):

PR #4323, opened 2023-07-10
Narrowing scope of adding our post type.

stratease:
[TEC-4819](https://theeventscalendar.atlassian.net/browse/TEC-4819)

Our post type was being added on the wrong queries.  This would have the side effect of us injecting `tribe_events` when it should only be looking for various ET template post types, thus finding an event post instead of the expected ET template post.

The logic needed to be updated to more narrowly look at the `is_main_query` flag for tag archive queries.

[TEC-4819]: https://theeventscalendar.atlassian.net/browse/TEC-4819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Query.php
```diff
@@ -40,11 +40,10 @@ public static function parse_query( $query ) {
 		}
 
 		// Work out if this is an Event query or not, do not set the flag yet.
-		$is_event_query = (array) $query->get( 'post_type' ) === [ TEC::POSTTYPE ];
-		$any_post_type  = (array) $query->get( 'post_type' ) === [ 'any' ];
-		$is_main_query  = $query->is_main_query();
+		$is_event_query   = (array) $query->get( 'post_type' ) === [ TEC::POSTTYPE ];
```
jesseeproductions, line 43: Is there only formatting changes here? I am having trouble seeing if there is any other changes here.

stratease: Below is the bigger change. Basically all I did was move the block of code into the `if ( $is_main_query ) {` block, instead of being outside below it. You'll notice two closing curly braces at the end of the change vs one.

jesseeproductions (dismissed):
Looks good!

bordoni (approved):

Camwyn (dismissed):

PR #4325, opened 2023-07-12
Fix/tec 4814 invalid nonce

stratease:


stratease:
This approach is a relatively large update - spans ECP and FBAR... maybe others? Still a WIP but would love eyes on it to see if I'm going the direction intended @borkweb ?

src/Tribe/Views/V2/View.php
```diff
@@ -1679,7 +1699,8 @@ static function () {
 			'request_date'         => Dates::build_date_object( $this->context->get( 'event_date', $today ) ),
 			'rest_url'             => $endpoint->get_url(),
 			'rest_method'          => $endpoint->get_method(),
-			'rest_nonce'           => $rest_nonce,
+			'rest_nonce_a'           => $rest_nonce_a,
```
stratease, line 1702: One option here is to instead serialize the nonces on the existing field and not require as much of a lift. But I think having explicitly two nonces is a bit clearer on what we are doing and follow the logic thread, versus a serialized object.

jesseeproductions: I vote for the two nonces showing. I think it makes is clearer too.

src/Tribe/Views/V2/Rest_Endpoint.php
```diff
@@ -20,6 +20,10 @@
  */
 class Rest_Endpoint {
 
+	const NONCE_ACTION = '_view_rest';
+	const NONCE_A_KEY = '_view_rest_nonce_a';
```
jesseeproductions, line 24: Is there a more descriptive name we can give them like main/alt? 
Not sure it makes a difference.

stratease: Good feedback.

src/Tribe/Views/V2/Views/Traits/HTML_Cache.php
```diff
@@ -295,10 +292,14 @@ protected function should_enable_html_cache( $context ) {
 	 * Get the list of fields/input we will do replacement for HTML Cache.
 	 *
 	 * @since 5.0.0
+	 * @since TBD Deprecating for new nonce structure.
+	 *
+	 * @deprecated
```
jesseeproductions, line 297: I think the standard is to add TBD with @deprecated

src/Tribe/Views/V2/Views/Traits/HTML_Cache.php
```diff
@@ -368,28 +379,16 @@ protected function get_view_nonce_json_properties() {
...
-		}
-
-		foreach ( $nonce_attrs as $action => $attr ) {
-			$html = preg_replace( '!(' . preg_quote( $attr, '!' ) . '=")[^"]*(")!', '\1%%NONCE:' . $action . '%%\2', $html );
-		}
-
-		foreach ( $nonce_props as $action => $prop ) {
-			$html = preg_replace( '!("' . preg_quote( $prop, '!' ) . '":")[^"]*(")!', '\1%%NONCE:' . $action . '%%\2', $html );
-		}
+		_deprecated_function( __METHOD__, 'TBD' );
```
jesseeproductions, line 391: What do you think about moving the deprecated methods to the end of the class? I find that helps keep them together and one day get removed.

src/Tribe/Views/V2/Views/Traits/HTML_Cache.php
Camwyn: Yes - so we can add the version on release

src/Tribe/Views/V2/Views/Traits/HTML_Cache.php
Camwyn: +1

src/Tribe/Views/V2/Rest_Endpoint.php
```diff
@@ -47,6 +74,57 @@ class Rest_Endpoint {
...
+	 * @return array<string,string> The field => nonce array.
+	 */
+	public static function get_rest_nonces(): array {
+		$generated_nonces = [];
+		/*
+		 * Some plugins, like WooCommerce, will modify the UID of logged out users; avoid that filtering here.
+		 *
+		 * @see TEC-3579
+		 */
+		$generated_nonces[ static::PRIMARY_NONCE_KEY ] = tribe_without_filters(
```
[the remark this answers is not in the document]
stratease: Good thought. They work in tandum, that's the whole point of them. The idea is in edge case scenarios where a logged in nonce was cached, there would be an unauthenticated nonce to fallback on. We now always check for both, and only one needs to valid.

bordoni:
I feel like this is in a great spot, but I would like to do In Branch QA here too. Just because it's a big change to how navigation works.

jesseeproductions (dismissed):

PR #4335, opened 2023-07-26
Fix/tec 4758 view permalinks not parsed with langs

stratease:
[TEC-4758](https://theeventscalendar.atlassian.net/browse/TEC-4758)

This issue was regarding our rewrite parsing for the view picker url generation. It was failing to parse the translated rewrite properly, and leaving the regex permalinks.

This adds a locale filter that is hit when WPML is active.

There are more than one way to determine current locale, and as we iterate we need to "force filter" both hooks so we iterate on different locales for generating translations.

Artifacts:

Correctly generated URL.

image

Incorrectly generated URL.

image

[TEC-4758]: https://theeventscalendar.atlassian.net/browse/TEC-4758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/plugin_integration.suite.dist.yml
```diff
@@ -0,0 +1,36 @@
...
+        user: %WP_DB_USER%
+        password: %WP_DB_PASSWORD%
+        dump: tests/_data/plugin_integration/dump.sql
+        populate: true # Do import the dump when the suite starts.
+        cleanup: false # Do not clean the db between tests, we'll deal with it.
+        url: '%WP_URL%'
+        originalUrl: 'http://wordpress.test'
+        tablePrefix: '%WP_TABLE_PREFIX%'
+        waitlock: 10
+      WPLoader:
```
lucatume, line 20: You do not need all the configuration parameters when using the `loadOnly: true` mode.
[See here](https://wpbrowser.wptestkit.dev/modules/wploader#wploader-to-only-bootstrap-wordpress).

```suggestion
      WPLoader:
        loadOnly: true
        wpRootFolder: "%WP_ROOT_FOLDER%"
        dbName: "%WP_TEST_DB_NAME%"
        dbHost: "%WP_TEST_DB_HOST%"
        dbUser: "%WP_TEST_DB_USER%"
        dbPassword: "%WP_TEST_DB_PASSWORD%"
        domain: "%WP_DOMAIN%"
```

The plugins are activated in the dump and do not need specification.

tests/plugin_integration/Events/WPML/WPML_Rewrite_Test.php
```diff
@@ -0,0 +1,37 @@
...
+namespace Events\WPML;
+use Spatie\Snapshots\MatchesSnapshots;
+use TEC\Events\Custom_Tables\V1\Provider as CT1_Provider;
+use Tribe\Tests\Traits\With_Uopz;
+use Tribe\Events\Views\V2\Hooks;
+use Tribe__Events__Main as TEC;
+use Tribe__Date_Utils as Dates;
+use Tribe__Events__Venue as Venue;
+
+class WPML_Rewrite_Test extends \Codeception\TestCase\WPTestCase {
```
lucatume, line 11: Do not use the `WPTestCase` when using `WPLoader` in `loadOnly: true` mode as it will try to handle the database as the `WPDb` module is doing.
Look at the `ct1_migration` suite for an example. Extend the base Codeception `Test` class.

tests/plugin_integration.suite.dist.yml
```diff
@@ -0,0 +1,36 @@
+actor: Plugin_integrationTester
```
lucatume, line 1: I would create a suite dedicated to WPML integration only, not to general plugin integration.
It would simplify its maintenance.

lucatume (changes requested):
The fix looks correct, but the test suite and methods require changes.

tests/plugin_integration.suite.dist.yml
stratease: Yeah, I was torn between that and having X suites for each plugin we may do this with in the future. Divi, Elementor, Yoast etc etc.... 
But I'll single this out to WPML, I don't disagree.

lucatume (approved):

PR #4341, opened 2023-08-02
On save of a block event, order venues as entered

borkweb:
If an event was created with the block editor, on save, ensure that the order of the venue IDs in the post meta table match the order of the blocks. I realized that without this, the primary venue display in views would not match the primary venue in event single if the order was changed after adding venues to an event.

:ticket: [ECP-1540]
📷 **Before saving the event**
Screenshot 2023-08-02 053607

📷 **After reordering venues and saving the event**
Screenshot 2023-08-02 053859

[ECP-1540]: https://theeventscalendar.atlassian.net/browse/ECP-1540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Linked_Posts.php
```diff
@@ -902,49 +906,110 @@ public function handle_submission_by_post_type( $event_id, $linked_post_type, $s
...
+	 * @since TBD
+	 *
+	 * @param int    $event_id Event ID.
+	 * @param string $linked_post_type The post type of the linked post.
+	 * @param array  $new_order The new order of the linked posts.
+	 * @param array  $old_order The old order of the linked posts.
+	 *
+	 * @return bool
+	 */
+	public function maybe_reorder_linked_posts_ids( int $event_id, string $linked_post_type, array $new_order = [], array $old_order = [] ): bool {
```
borkweb, line 925: If you are reviewing, be sure to hide whitespace. This method is essentially the same bit of logic that was up above, but I moved it to its own method so that it could be reused. I purposefully did not refactor it to minimize changes. It isn't horrendous, luckily.

src/Tribe/Linked_Posts.php
```diff
@@ -902,49 +906,110 @@ public function handle_submission_by_post_type( $event_id, $linked_post_type, $s
...
+	 * @param string $linked_post_type The post type of the linked post.
+	 * @param array  $original_order The original IDs/order stored in meta.
+	 *
+	 * @return array
+	 */
+	public function maybe_get_new_order_from_blocks( int $event_id, string $linked_post_type, array $original_order = [] ) {
+		// If the post has blocks, we need to update sorting of the post ids to link so it matches block order.
+		if ( ! has_blocks( $event_id ) ) {
+			return $original_order;
+		}
```
bordoni, line 993: Should we also check if someone has blocks editor? Since its possible to have the blocks code in the Content but without the blocks editor active for events?

bordoni (approved):
Besides that one questions we are good.

src/Tribe/Linked_Posts.php
borkweb: A good question, but in this case we are purely talking about data rather - so even if they are editing via the classic editor, if they have blocks in their post content, we need to observe them.

PR #4347, opened 2023-08-04
Ensure we output valid html.

EugeneKyale:
Ticket : [TEC-4812](https://theeventscalendar.atlassian.net/browse/TEC-4812)

Display an empty `<dt>` tag before `<dd>` even when the website title isn't available to ensure we always output valid html.

**Screenshot 📸**

image

[TEC-4812]: https://theeventscalendar.atlassian.net/browse/TEC-4812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/modules/meta/organizer.php
+++ b/src/views/modules/meta/organizer.php
@@ -75,6 +75,17 @@ $website_title = tribe_events_get_organizer_website_title();
 					<dt class="tribe-organizer-url-label">
 						<?php echo esc_html( $website_title ) ?>
 					</dt>
+				<?php else: ?>
+					<dt
+						class="tribe-common-a11y-visual-hide"
+						aria-label="<?php echo sprintf(
+							/* Translators: %1$s is the customizable organizer term, e.g. "Organizer website title" */
+							esc_html_x( '%1$s website title', "The label for the organizer's website title.", 'the-events-calendar' ),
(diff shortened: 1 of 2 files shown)


src/views/modules/meta/organizer.php
```diff
@@ -75,6 +75,17 @@ class="tribe-common-a11y-visual-hide"
...
+				<?php else: ?>
+					<dt
+						class="tribe-common-a11y-visual-hide"
+						aria-label="<?php echo sprintf(
+							/* Translators: %1$s is the customizable organizer term, e.g. "Organizer website title" */
+							esc_html_x( '%1$s website title', "The label for the organizer's website title.", 'the-events-calendar' ),
+							tribe_get_organizer_label_singular()
+						) ; ?>"
+					>
+						<?php // This element is only present to ensure we have a valid HTML, it'll be hidden from browsers but visible to screenreaders for accessibility. ?>
```
Camwyn, line 87: These should probably have actual readable text in them, i.e. `visible to screen readers for accessibility` but we can do a second pass.

Camwyn (approved):

PR #4348, opened 2023-08-04
WIP - testing phpstan workflow.

stratease:


stratease:
I'm no github workflow pro - I was thinking of creating a matrix for this and have it loop on the `phpVersion` for phpstan. Just `sed` replace the value on each item, reusing the same config.

Thoughts on this?

Any areas of optimization?

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install WP CLI
+        run: curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
+      - name: Install WP Core
+        run: php wp-cli.phar core download --path=wpdemo.test --allow-root
+      # ------------------------------------------------------------------------------
+      # Run the PHPStan tests
+      # ------------------------------------------------------------------------------
+      - name: Convert PHPStan Config for Local Paths
+        run: |
+          sed -i 's/..\/..\/..\/wp-includes/wpdemo.test\/wp-includes/g' phpstan.neon
```
lucatume, line 37: This looks a bit contrived: I suggest the creation of a `phpstan.ci.neon` file with the correct path and its use in the workflow.

lucatume (changes requested):

.github/workflows/phpstan.yml
stratease: @lucatume I don't disagree, I considered that. 

The reason I went this route was because my concern was that we would basically be copying an entire file - where all changes would have to be duplicated between a local `neon` versus a CI environment `neon` file, all for one value that needs to be changed. As I understand it there is no "merge" of values, so we can't simply have an extra config that only has that value.

stratease:
> I'm no github workflow pro - I was thinking of converting this to use a `matrix` and have it loop on the `phpVersion` for phpstan. Just `sed` replace the value on each item, reusing the same config.
> 
> Thoughts on this?
> 
> Any areas of optimization?

@lucatume thoughts on this approach?

phpstan.neon
```diff
@@ -0,0 +1,9 @@
+parameters:
+  phpVersion: 70400
+  level: 0
+  paths:
+    - src
+    - the-events-calendar.php
+  scanDirectories:
+    - ../../../wp-includes
```
Camwyn, line 8: Why are we scanning WP?

borkweb: My guess is so that the WP functions/methods/classes aren't shown as invalid. We can get around the need for scanning by including the following in composer.json as a dev dependency:

```
"szepeviktor/phpstan-wordpress": "^1.1",
```

I think doing that would avoid the need to download/install WP via wp-cli

borkweb (changes requested):

phpstan.neon
stratease: Updated to use the PHPStan WP stub library. My only concern is that the stubs stay up to date w/ the latest WP API (deprecations and new prototypes). Rare, but something we need to track over time.

.github/workflows/phpstan.yml
lucatume: Resolving as discussed in 2023-09-11 Dev meeting.

lucatume:
> > I'm no github workflow pro - I was thinking of converting this to use a `matrix` and have it loop on the `phpVersion` for phpstan. Just `sed` replace the value on each item, reusing the same config.
> > Thoughts on this?
> > Any areas of optimization?
> 
> @lucatume thoughts on this approach?

Discussed in 2023-09-11 Dev Meeting.

Camwyn:
> I'm no github workflow pro - I was thinking of converting this to use a `matrix` and have it loop on the `phpVersion` for phpstan. Just `sed` replace the value on each item, reusing the same config.
I'm no phpstan pro but since phpstan will use the composer PHP version if it can find it, can't we just do something like:
`composer config platform.php ${{ matrix.short }}`

Failing that, we _could_ have a phpstan config for each version (which opens the door to version-specific ignores, etc) and swap using the `-c` flag

stratease:
Some approaches in here we might want to use. Using PHP Codesniffer(?) https://github.com/stellarwp/uplink-origin/blob/main/.github/workflows/compatibility.yml

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,47 @@
...
+          submodules: recursive
+      # ------------------------------------------------------------------------------
+      # Set up PHP and run phpstan
+      # ------------------------------------------------------------------------------
+      - name: Configure PHP environment to run phpstan
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.phpVersion }}
+      # ------------------------------------------------------------------------------
+      # Override composer php version
```
Camwyn, line 33: Since we don't set a php version in phpstan, it looks for the composer file.
Setting this overrides the one in composer, so it matches the matrix version we are testing.

composer.json
```diff
@@ -36,15 +36,20 @@
...
+    "phpstan/phpstan": "^1.10",
+    "phpstan/phpstan-deprecation-rules": "^1.1",
+    "phpstan/extension-installer": "^1.3",
+    "szepeviktor/phpstan-wordpress": "*"
   },
   "config": {
     "preferred-install": "dist",
     "platform": {
-      "php": "7.3.33"
+      "php": "7.4"
```
Camwyn, line 48: Our min required PHP version is 7.4. Not sure if there is a good reason to set this lower?

composer.json
```diff
@@ -36,15 +36,20 @@
     "the-events-calendar/coding-standards": "dev-master",
     "the-events-calendar/tec-testing-facilities": "dev-master",
     "vlucas/phpdotenv": "^4.1",
-    "wp-coding-standards/wpcs": "^2.1"
+    "wp-coding-standards/wpcs": "^2.1",
+    "phpstan/phpstan": "^1.10",
+    "phpstan/phpstan-deprecation-rules": "^1.1",
+    "phpstan/extension-installer": "^1.3",
+    "szepeviktor/phpstan-wordpress": "*"
```
Camwyn, line 43: This is in the hopes of never falling behind the current WP version.
See convo here: https://github.com/the-events-calendar/the-events-calendar/pull/4348#discussion_r1313505629

jesseeproductions:
The code looks great. How are we handling the PHP Stan Failures? Are we going to have a followup PR with some of the fixes?

stratease:
> The code looks great. How are we handling the PHP Stan Failures? Are we going to have a followup PR with some of the fixes?

@jesseeproductions they are not required to merge, for now we will iterate and clean them up over time. Eventually they will be required tests to pass.

jesseeproductions (approved):

PR #4354, opened 2023-08-08
Ensure that the EventShowMap(Link) values are always set

borkweb:
This prevents weird cases where the form fields are not passed when checkboxes are unchecked, thus, failing to update the desired state.  Instead, we say:

* If the field is missing when coming from the post editor, consider it false
* If the field is present when coming from the post editor, use the value

:ticket: [ECP-1540]
:movie_camera: https://www.loom.com/share/782199045be94f6bb9d0f13e1ed9f749

[ECP-1540]: https://theeventscalendar.atlassian.net/browse/ECP-1540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/API.php
+++ b/src/Tribe/API.php
@@ -787,6 +787,22 @@ if ( ! class_exists( 'Tribe__Events__API' ) ) {
 				}
 			}
 
+			$is_post_editor = ! empty( $args['action'] ) && $args['action'] === 'editpost';
+
+			if ( $is_post_editor ) {
+				if ( ! empty( $args['EventShowMap'] ) ) {
+					$args['EventShowMap'] = false;
+				} else {
+					$args['EventShowMap'] = true;
+				}
+
+				if ( empty( $args['EventShowMapLink'] ) ) {
+					$args['EventShowMapLink'] = false;
+				} else {
+					$args['EventShowMapLink'] = true;
+				}
+			}
+
 			return $args;
 		}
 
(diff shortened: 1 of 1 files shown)


src/Tribe/API.php
```diff
@@ -787,6 +787,22 @@ public static function sanitize_event_post_create_update_args( $args ) {
...
 			}
 
+			$is_post_editor = ! empty( $args['action'] ) && $args['action'] === 'editpost';
+
+			if ( $is_post_editor ) {
+				if ( ! empty( $args['EventShowMap'] ) ) {
+					$args['EventShowMap'] = false;
+				} else {
+					$args['EventShowMap'] = true;
+				}
```
bordoni, line 797:
```suggestion
				$args['EventShowMap'] = empty( $args['EventShowMap'] );
```

src/Tribe/API.php
```diff
@@ -787,6 +787,22 @@ public static function sanitize_event_post_create_update_args( $args ) {
...
+
+				if ( empty( $args['EventShowMapLink'] ) ) {
+					$args['EventShowMapLink'] = false;
+				} else {
+					$args['EventShowMapLink'] = true;
+				}
```
bordoni, line 803:
```suggestion
				$args['EventShowMapLink'] = ! empty( $args['EventShowMapLink'] );
```

bordoni (changes requested):
Cant we simplify with that?

bordoni (approved):
🍧

PR #4355, opened 2023-08-09
Update Settings tabs sorting. 

EugeneKyale:
Ticket : [TEC-4774](https://theeventscalendar.atlassian.net/browse/TEC-4774)

Fixed a bug where `0` and `1` were displayed on the `general` and `display` tabs for Network admins on multisite installations.

The original snippet using `array_merge` and `array_flip` was causing the keys of the `$first` array to be flipped with its values, resulting in `0` and `1` as keys.

**Screencast 🎥**

https://drive.google.com/file/d/1rKG0NO859t3AmtkuPQpqWg1RWGP0loAO/view?usp=sharing

[TEC-4774]: https://theeventscalendar.atlassian.net/browse/TEC-4774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Admin/Settings.php
```diff
@@ -406,7 +407,17 @@ public function sort_tabs( $tabs, $admin_page ): array {
 
 		// Ensure these are the first tabs.
 		$first   = [ 'general', 'display' ];
-		$tabs   = array_merge( array_flip( $first ), $tabs );
+
+		// Reverse to maintain order when prepending
+		foreach ( array_reverse( $first ) as $sort ) {
```
Camwyn, line 412: Not a major thing when there are only 2 but `array_reverse()` (or any callback in the loop args) will run for _each and every iteration_ - so it's better to do something like:
```suggestion
		$reversed_arr = array_reverse( $first);
		
		foreach ( $reversed_arr as $sort ) {
```

src/Tribe/Admin/Settings.php
```diff
@@ -406,7 +407,17 @@ public function sort_tabs( $tabs, $admin_page ): array {
...
+			if ( ! isset( $tabs[ $sort ] ) ) {
+				continue;
+			}
+			$temp = $tabs[ $sort ];
```
Camwyn, line 416:
```suggestion
			}
			
			$temp = $tabs[ $sort ];
```
whitespace

Camwyn (changes requested):

Camwyn (approved):

PR #4371, opened 2023-08-24
Register Site Editor Templates.

EugeneKyale:
## Ticket
[TEC-4790](https://theeventscalendar.atlassian.net/browse/TEC-4790)

## 🎯Objective
- Introduced registration for both `Events Archive` and `Single Event` templates.
- Implemented logic to toggle the `Activate Block Editor for Events` setting based on the full site editor theme status.
- Modified the `Events → Display → Events Template` setting, ensuring it defaults to Default Events Template when a site editor theme is activated.

## 🎥 Screencast
https://drive.google.com/file/d/1c6QOprxT8bG9pH7kRs6tl46Sl6eEKNtS/view?usp=sharing

## 📸 Screenshot
site_editor_templates

[TEC-4790]: https://theeventscalendar.atlassian.net/browse/TEC-4790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stratease:
Perhaps we need some snapshot tests at least?

src/Events/Editor/Full_Site/Templates.php
```diff
@@ -26,14 +26,23 @@ class Templates {
...
+
 	/**
 	 * Modify the available Templates so that people can edit the template.
 	 *
 	 * @since 5.14.2
 	 *
 	 * @param WP_Block_Template[] $query_result Array of found block templates.
-	 * @param array  $query {
-	 *     Optional. Arguments to retrieve templates.
+	 * @param array               $query        {
```
stratease, line 44:
```suggestion
	 * @param array               $query        
```
I know this wasn't you, but is that curly brace intended? Looks off to me.

src/Events/Editor/Full_Site/Templates.php
```diff
@@ -82,9 +87,71 @@ public function get_template_events_archive() {
 		$template->content        = Template_Utils::inject_theme_attribute_in_content( $template_content );
 		$template->slug           = static::$archive_slug;
 		$template->source         = 'custom';
+		$template->type           = 'wp_template';
+		$template->title          = esc_html__( 'Calendar Views (Event Archive)', 'the-events-calendar' );
```
stratease, line 91:
```suggestion
		$template->title          = esc_html_x( 'Calendar Views (Event Archive)', 'The Full Site editor block navigation title' 'the-events-calendar' );
```
Perhaps something like this? I think we want to use context variations most of the time now right?

src/Events/Editor/Full_Site/Templates.php
```diff
@@ -82,9 +87,71 @@ public function get_template_events_archive() {
...
+
+		$template                 = new WP_Block_Template();
+		$template->id             = 'the-events-calendar//single-event';
 		$template->theme          = 'The Events Calendar';
+		$template->content        = Template_Utils::inject_theme_attribute_in_content( $template_content );
+		$template->slug           = static::$single_slug;
+		$template->source         = 'custom';
 		$template->type           = 'wp_template';
-		$template->title          = esc_html__( 'Events Archive', 'the-events-calendar' );
+		$template->title          = esc_html__( 'Event Single', 'the-events-calendar' );
```
stratease, line 153:
```suggestion
		$template->title          = esc_html_x( 'Event Single', 'Context here...', 'the-events-calendar' );
```
etc

src/Tribe/Editor/Compatibility.php
```diff
@@ -238,4 +239,25 @@ public function filter_is_classic_editor( $is_classic_editor = false ) {
...
+	 * @param string   $old_name  The name of the old theme that was active before switching.
+	 * @param WP_Theme $old_theme WP_Theme instance of the old theme.
+     *
+     * @return void
+     */
+    public function toggle_blocks_editor_setting_on_theme_switch( $old_name, $old_theme = false ) {
+        if ( tec_is_full_site_editor() ) {
+            tribe_update_option( static::$blocks_editor_key, true );
+        } else {
+            tribe_update_option( static::$blocks_editor_key, false );
```
stratease, line 260:
```suggestion
            tribe_update_option( static::$blocks_editor_key, false );
```
Do we really need to disable it? Perhaps it was always set. Just a thought for simplicity / maintenance sake.

stratease (changes requested):

src/Events/Editor/Full_Site/Templates/single-event.html
```diff
@@ -0,0 +1,3 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+<!-- wp:tribe/single-event {} /-->
```
stratease, line 2:
```
<!-- wp:tribe/single-event {} /-->
```
More of a question here.

Is the `tribe/single-event` a reference to our other block? I assume we will not be using the existing blocks for this. 

Also wondering if we should name it like `tec-events/...` or what our naming convention might be (vs `tribe`)?

EugeneKyale:
> Optional. Arguments to retrieve templates.

Yeah I'll a snapshot test for this.

src/Events/Editor/Full_Site/Templates.php
EugeneKyale: I think that's intentional. Additional details provided in the curly braces give a more in-depth description of the structure and intent of this $query parameter.

src/Tribe/Editor/Compatibility.php
EugeneKyale: Ok yeah it makes sense to only update the option on FSE theme activation. I've changed it and updated the docblock to reflect the update.

src/Events/Editor/Full_Site/Templates.php
EugeneKyale: Updates

src/Events/Editor/Full_Site/Templates.php
EugeneKyale: Updated

src/Events/Editor/Full_Site/Templates/single-event.html
bordoni: `tribe/` is the old namespace for blocks, we will need at some point to create a new one and have the old ones map to this new one.

PR #4377, opened 2023-09-12
We need wpnonce to validate user and show user caps.

stratease:
[ECP-1581](https://theeventscalendar.atlassian.net/browse/ECP-1581)

In some situations the user capabilities were failing because the user session was not found. This would result in Event Manager pagination screens to lose the various event actions, and user authenticated logic.

The cause was the `_wpnonce` is used to auth the user.

Artifacts:
f1bc57c1-fbba-4a4a-8617-9f4d22e6d89b
b2cf55e5-f0b7-4f28-afc2-160a3b1e009d

[ECP-1581]: https://theeventscalendar.atlassian.net/browse/ECP-1581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -232,6 +232,7 @@ Remember to always make a backup of your database and files before updating!
 = [TBD] TBD =
 
 * Fix - Resolved "Uncaught ReferenceError: lodash is not defined" error by adding `lodash` as a dependency for the Block Editor Assets. [ECP-1575]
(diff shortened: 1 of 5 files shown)


Camwyn (dismissed):
One tiny whitespace suggestion

redscar (dismissed):

PR #4381, opened 2023-09-18
Fix more undeclared/dynamic properties.

Camwyn:
More PHP compatibility fixes.

[TEC-4847]

[TEC-4847]: https://theeventscalendar.atlassian.net/browse/TEC-4847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -518,6 +518,8 @@ class Tribe__Events__Assets {
 	 * @return bool
 	 */
 	public function is_settings_page() {
+		$admin_helpers = Tribe__Admin__Helpers::instance();
+
 		return $admin_helpers->is_screen( 'settings_page_tribe-settings' );
 	}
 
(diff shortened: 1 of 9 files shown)


src/Tribe/Assets.php
```diff
@@ -518,6 +518,8 @@ public function is_style_option_full() {
 	 * @return bool
 	 */
 	public function is_settings_page() {
+		$admin_helpers = Tribe__Admin__Helpers::instance();
```
Camwyn, line 521: We should not assume that a variable is set like this. It's not a class param, so define it.

src/Tribe/Main.php
```diff
@@ -1012,6 +1014,11 @@ public function init() {
 
 			$this->default_values                             = apply_filters( 'tribe_events_default_value_strategy', new Tribe__Events__Default_Values() );
 
+			/* Deprecated 4.0 */
```
Camwyn, line 1017: I just grouped these and moved them here to make it easier to remove them together.

stratease:
Was this just fixing undefined errors?

stratease (approved):

PR #4383, opened 2023-09-19
Archive Events block showing on template page.

stratease:
[TEC-4792](https://theeventscalendar.atlassian.net/browse/TEC-4792)

Adding support for the Site Editor Archive Events Block Template. You are now able to view, select and save the template and see results for the end users.

Artifact:
https://www.loom.com/share/648bf5b547fe47a39ba65c0e9f561af2?sid=1defc2cc-46c0-44be-9bc3-dfd245986a63

[TEC-4792]: https://theeventscalendar.atlassian.net/browse/TEC-4792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Events/Editor/Full_Site/Templates.php
```diff
@@ -77,14 +78,47 @@ public function add_events_archive( array $query_result, $query, string $templat
 	 * @return WP_Block_Template A reference to the template object for the query.
 	 */
 	public function get_template_events_archive() {
-		$template_content = file_get_contents(
-			Tribe__Events__Main::instance()->plugin_path . '/src/Events/Editor/Full_Site/Templates/archive-events.html'
-		);
 
 		$template                 = new WP_Block_Template();
```
stratease, line 82: Could use some guidance here...

It seems to me, that in our `get_template_events_archive()` we should be looking for a saved template, if none exists create it, then return that. 

Part of my belief for this is because in WP Core, on the `get_block_template`, and quite a few other places, [it is looking for the wp_id](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-templates-controller.php#L693) to hydrate the object.

Also, we (assuming this is correct definition) are defining this a `custom` template. Which with that definition it [bypasses some of the creation](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-templates-controller.php#L353-L360) logic. Also, on the front end the default save operation is a `PUT` (for some reason?) which is not hitting the `create_item()` callback.

Thoughts on that?

One other question - are we intending this `Archive Events` to be a single template? Multiple can't exist right? That is part of the assumption I'm going on.

src/Tribe/Editor.php
```diff
@@ -377,6 +377,7 @@ public function register_blocks() {
 	 * @return bool
 	 */
 	public function is_events_post_type() {
+		return true; // @todo remove
```
stratease, line 380: Temporarily bypassing checks. This is so the existing blocks assets load, and our new block is compiled in the same `app/main.js` file.

I assume we will want to keep that logic here and just adjust these checks to include template pages?

src/Tribe/Editor.php
```diff
@@ -388,6 +389,7 @@ public function is_events_post_type() {
 	 * @return bool
 	 */
 	public function is_edit_screen() {
+		return true; // @todo remove
```
stratease, line 392: Same as above.

src/modules/fs_archive_events.js
```diff
@@ -0,0 +1,37 @@
+const { __ } = wp.i18n;
+const { registerBlockType } = wp.blocks;
+function bob_ross23() {
+	return 'bob_ross23';
+}
+console.log(bob_ross23());
+registerBlockType( 'tec/archive-events', {
```
stratease, line 7: We want to use `tec/` 'theme' namespace here?

src/modules/fs_archive_events.js
```diff
@@ -0,0 +1,37 @@
...
+                    </p>
+                    <p>
+                    <img src={'https://place-hold.it/600x100/333/fff/000?text=@todo&fontsize=22'} />
+                    </p>
+                    <p>
+                    <img src={'https://place-hold.it/600x100/333/fff/000?text=@todo&fontsize=22'} />
+                    </p>
+                </div>
+            );
+        },
```
stratease, line 35: Any definitions I may be missing?

src/modules/index.js
```diff
@@ -2,6 +2,7 @@
  * Internal dependencies
  */
 import * as blocks from './blocks';
+import './fs_archive_events';
```
stratease, line 5: Quick and dirty way to include to primary compiler. I assume we will want to get it loaded via our `./blocks` above? Something else?

stratease (commented):
@EugeneKyale would love your eyes on this. It is very much a rough draft, but I have a number of areas I'd love to get eyes on and see what you think.

src/Events/Editor/Full_Site/Templates.php
EugeneKyale: You're spot on. Using the `wp_id` to hydrate objects is a common pattern in WP Core, and it makes sense for us to align with that. First checking for a saved template is a good idea and should mitigate against unintentionally creating duplicates or overlooking previously set configurations for that template. If none exists, creating it ensures that we have a consistent fallback.

Regarding your question about the Archive Events template, yes, it's designed to be a singular entity. The logic ensures that we don't have multiple instances of the same template, which could lead to inconsistency and confusion. By ensuring only one instance, we can maintain a uniform experience across the board.

src/modules/fs_archive_events.js
EugeneKyale: Yes, we're moving away from `tribe`

src/modules/fs_archive_events.js
EugeneKyale: We probably need a `save` function for frontend output, and the `supports` property for additional block features like custom css classes.

stratease: Could use an example for `save` @EugeneKyale. I'm[ looking at docs](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-edit-save/#save), and don't quite grasp what this will do for us in this v1...? The FE renders fine without it at the moment.

EugeneKyale: Ooh, in that case, block rendering is implemented on the PHP side using the `render_callback` similar to the existing blocks. In this case the save function isn't necessary.

src/Events/Editor/Full_Site/Hooks.php
```diff
@@ -98,6 +130,32 @@ public function filter_include_templates( $query_result, $query, $template_type
...
+	 * @param null|WP_Block_Template $block_template The filtered template.
+	 * @param string                 $id             The block template ID.
+	 * @param string                 $template_type  The template type.
+	 *
+	 * @return null|WP_Block_Template
+	 */
+	public function filter_include_template_by_id( $block_template, $id, $template_type ) {
+		if ( ! is_null( $block_template ) ) {
+			return $block_template;
+		}
```
EugeneKyale, line 147:
```suggestion
		}

```
let's fix the spacing here

src/Events/Editor/Full_Site/Hooks.php
```diff
@@ -98,6 +130,32 @@ public function filter_include_templates( $query_result, $query, $template_type
...
+		if ( $template_type !== 'wp_template' ) {
+			return $block_template;
+		}
+		$archive_template = tribe( Archive_Events::class );
```
EugeneKyale, line 151: Same spacing comment as above.

src/Events/Editor/Full_Site/Templates.php
```diff
@@ -73,26 +77,84 @@ public function add_events_archive( array $query_result, $query, string $templat
...
 	 * @return WP_Block_Template A reference to the template object for the query.
 	 */
 	public function get_template_events_archive() {
-		$template_content = file_get_contents(
-			Tribe__Events__Main::instance()->plugin_path . '/src/Events/Editor/Full_Site/Templates/archive-events.html'
+		$template      = new WP_Block_Template();
+		$archive_block = tribe( Archive_Events::class );
+
+		// Let's see if we have a saved template?
+		$wp_query_args  = array(
```
EugeneKyale, line 89: Use the `[]` shorthand notation when creating arrays.

src/Events/Editor/Full_Site/Templates.php
```diff
@@ -73,26 +77,84 @@ public function add_events_archive( array $query_result, $query, string $templat
...
+			'post_name__in'  => array( $archive_block->slug() ),
+			'post_type'      => 'wp_template',
+			'post_status'    => array( 'auto-draft', 'draft', 'publish', 'trash' ),
+			'posts_per_page' => 1,
+			'no_found_rows'  => true,
+			'tax_query'      => array(
```
EugeneKyale, line 95: Same comment regarding the use of the `[]` shorthand notation.

tests/views_integration/Tribe/Events/Views/Blocks/Archive_EventsTest.php
```diff
@@ -20,29 +22,75 @@ public function tearDown(): void {
 		parent::tearDown();
 	}
 
+	public static function normalize_wp_template( $template ): array {
```
EugeneKyale, line 25: Needs docblock

EugeneKyale (commented):
Overall, the solution looks good. Should be good to go once the spacing issues are addressed.

src/Events/Editor/Full_Site/Templates.php
```diff
@@ -73,26 +77,84 @@ public function add_events_archive( array $query_result, $query, string $templat
...
+			];
+			// Create this template.
+			$id = wp_insert_post( $insert );
+
+			// Setup our "theme" term, for the taxonomy query.
+			$term = get_term_by( 'name', $archive_block->get_namespace(), 'wp_theme', ARRAY_A );
+			if ( ! $term ) {
+				wp_insert_term( $archive_block->get_namespace(), 'wp_theme' );
+			}
+			wp_set_post_terms( $id, $archive_block->get_namespace(), 'wp_theme' );
```
stratease, line 126:
```suggestion
			wp_set_post_terms( $id, $archive_block->get_namespace(), 'wp_theme' );
```
@EugeneKyale just thought of this while working on single. Does it make sense for us to store `tec` as a `wp_theme`? I am not confident in the approach here. Is there a precedence for this you know of?

stratease: I took this approach because WP core attempts to hydrate the template via `_build_block_template_result_from_post()` - and this is how it structures the template post query.

EugeneKyale: @stratease I'm not aware of any precedence and don't have a strong opinion either way. 

A general thought I have is that if `wp_theme` is being used by other themes or plugins, it might be better to use a more specific custom taxonomy to avoid potential conflicts.

stratease: @EugeneKyale it's not a personal choice, it's just based on how the [WP core will look for the block template](https://github.com/WordPress/WordPress/blob/master/wp-includes/block-template-utils.php#L1055-L1068) in a couple places like `get_block_template()` - it is hard coded to use that taxonomy. My concern is maybe I'm using the wrong thing, couldn't find any detailed articles/docs on setting this up, they were pretty high level and didn't touch some of this.

@bordoni do you have thoughts here?

stratease: Also worth noting [how it hydrates the template](https://github.com/WordPress/WordPress/blob/master/wp-includes/block-template-utils.php#L739-L761)

EugeneKyale (approved):

PR #4387, opened 2023-09-27
Feat/tec 4793 single event block template

stratease:
[TEC-4793](https://theeventscalendar.atlassian.net/browse/TEC-4793)

Adding support for the Site Editor Single Event Block Template. You are now able to view, select and save the template and see results for the end users.

Artifact:
https://www.loom.com/share/ddda5e4adb364453a5d79fd1633a2fea?sid=4db96700-be5e-4ffe-b8a9-8f2998464af4

[TEC-4793]: https://theeventscalendar.atlassian.net/browse/TEC-4793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/blocks/single-event/index.js
```diff
@@ -0,0 +1,51 @@
...
+                <p>
+                    <strong>
+                    { __( 'EVENT DATE/TIME', 'the-events-calendar' ) }
+                </strong>
+                </p>
+
+                <FauxLine style={{marginLeft:34, marginRight:'25%'}} />
+                <FauxLine style={{marginRight:'25%'}} />
+                <FauxLine style={{marginRight:'25%'}} />
+                <button type={"button"} class="wp-block-button__link wp-element-button">Subscribe to Event</button>
```
stratease, line 47:
```suggestion
                <button type={"button"} class="wp-block-button__link wp-element-button">Subscribe to Event</button>
```
@EugeneKyale Ignore most of this PR, just wanted thoughts on how we ought to incorporate designs in these blocks. Not sure the future of some of this, but I know on the front-end the button design has the blue border around a white button. But I don't think we want to directly reuse the CSS from that, so for now I was just using the "bare theme" styles...?

Any opinion on which way to go on this for a v1?

EugeneKyale: @stratease I think we should reuse the existing styles and have that as the default styling. This should give users a familiar base before applying any customisations.

stratease: Going to do a style copy facade, not literally bringing in the CSS files.

stratease:
FYI these are forked from https://github.com/the-events-calendar/the-events-calendar/pull/4383 which is still pending CR. Should start there.

src/Events/Editor/Full_Site/Controller.php
```diff
@@ -0,0 +1,399 @@
...
+			if ( $id === $template->id() ) {
+				return $template->get_block_template();
+			}
+		}
+
+		return $block_template;
+	}
+
+	/**
+	 * @param string $template_type The type of templates we are fetching.
```
EugeneKyale, line 320: Let's add a fully qualified docblock.

src/Tribe/Editor.php
```diff
@@ -373,24 +373,29 @@ public function register_blocks() {
...
 	 * @since 5.12.0
+	 * @since TBD Adding support to load on site editor screen.
 	 *
 	 * @return bool
 	 */
 	public function is_edit_screen() {
 		$current_screen = get_current_screen();
 
-		return 'post' === $current_screen->base;
+		return 'post' === $current_screen->base || $current_screen->id === 'site-editor';
```
EugeneKyale, line 398:
```suggestion
		return 'post' === $current_screen->base || 'site-editor' === $current_screen->id;
```

src/Events/Editor/Full_Site/Controller.php
stratease: Did resolve, but something seems off with this CR - it has been updated but still shows unchanged. https://github.com/the-events-calendar/the-events-calendar/pull/4387/commits/80eebbe4095b418ed3f97b8a19da93bd5167e4be

EugeneKyale (approved):

PR #4409, opened 2023-10-24
Fix for nonces being cached on some views.

stratease:
[TEC-4936](https://stellarwp.atlassian.net/browse/TEC-4936)

On the Past Events View, the nonce was being generated twice and one of them would be cached in our HTML cache. This moves the nonce generation up the call stack, to avoid the bug from recursive View HTML generation.

Artifacts:
401 errors on AJAX requests:
```
HTTP 401 when calling /wp-json/tribe/views/v2/html for navigation
```

[TEC-4936]: https://stellarwp.atlassian.net/browse/TEC-4936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -232,12 +232,14 @@ Remember to always make a backup of your database and files before updating!
 = [TBD] TBD =
 
 * Fix - On the Past Events View, the nonce was incorrectly being generated twice, and one of them would be cached in our HTML transient cache. This was causing a 401 nonce errors to occur when the cached nonce expired. The nonce generation was moved outside this HTML generation that would was being cached. [TEC-4936]
+* Fix - Wordpress 6.3 introduce some changes in filters that regressed a prior fix for authentication and our new nonce structure used in view pagination. One symptom of the issue was losing the authenticated user and failing to display user specific capabilities on event views. [ECP-1601]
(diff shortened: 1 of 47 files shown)


EugeneKyale (dismissed):

readme.txt
```diff
@@ -231,6 +231,7 @@ Remember to always make a backup of your database and files before updating!
 
 = [TBD] TBD =
 
+* Fix - On the Past Events View, the nonce was incorrectly being generated twice, and one of them would be cached in our HTML transient cache. This was causing a 401 nonce errors to occur when the cached nonce expired. The nonce generation was moved outside this HTML generation that would was being cached. [TEC-4936]
```
Camwyn, line 234:
```suggestion
* Fix - On the Past Events View, the nonce was incorrectly being generated twice, and one of them would be cached in our HTML transient cache. This was causing a 401 nonce errors to occur when the cached nonce expired. The nonce generation was moved outside the HTML generation that is being cached. [TEC-4936]
```

Camwyn (changes requested):
Small readme tweak

Camwyn (approved):

PR #4431, opened 2023-11-09
Fix/ecp 1535 latest range fail

stratease:
[ECP-1535](https://stellarwp.atlassian.net/browse/ECP-1535)

The "Next" pagination button on week view was not always working. This was due to how we were storing the latest occurrence ID. It was failing to detect when it should store the provisional ID, and it would always store the post ID. 

Test is here (it affected recurring events): https://github.com/the-events-calendar/events-pro/pull/2334

Artifacts:
Screenshot 2023-11-09 140855

[ECP-1535]: https://stellarwp.atlassian.net/browse/ECP-1535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Events/Custom_Tables/V1/Updates/Events.php
+++ b/src/Events/Custom_Tables/V1/Updates/Events.php
@@ -140,10 +140,7 @@ class Events {
 		$latest   = $this->get_latest_occurrence();
 
 		if ( $earliest ) {
-			$earliest_id = $earliest->provisional_id;
-			if ( ! $earliest_id ) {
-				$earliest_id = $earliest->post_id;
-			}
+			$earliest_id = $earliest->provisional_id ? $earliest->provisional_id : $earliest->post_id;
 			tribe_update_option( 'earliest_date', $earliest->start_date_utc );
 			tribe_update_option( 'earliest_date_markers', [ $earliest_id ] );
 		} else {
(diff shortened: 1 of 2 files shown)


src/Events/Custom_Tables/V1/Updates/Events.php
```diff
@@ -139,16 +140,24 @@ public function rebuild_known_range() {
 		$latest   = $this->get_latest_occurrence();
 
 		if ( $earliest ) {
+			$earliest_id = $earliest->provisional_id;
+			if ( ! $earliest_id ) {
+				$earliest_id = $earliest->post_id;
+			}
```
Camwyn, line 146: This _could_ be a ternary:
```suggestion
			$earliest_id = ! empty( $earliest->provisional_id ) ? $earliest->provisional_id : $earliest->post_id;
```

Camwyn: Just not a null coalescing check because (I'm assuming) it's returning a non-null value (like...false?)

Camwyn (changes requested):
Small docblock tweak.

src/Events/Custom_Tables/V1/Updates/Events.php
stratease: It's returning the proper value. This seems to be a bug with null coalescing and magic properties. It returns a positive integer (not `false` or `0` etc) but fails when in a coalescing statement, and seems fine with the traditional ternary check. 

See I checked the return values (it should be a provisional ID): 

image

Camwyn (approved):

PR #4447, opened 2023-12-12
Update local phpcs excludes

Camwyn:


composer.lock
stratease, line 1: This is changed - but no `composer.json` change? Why?

Camwyn: Running composer update will update the lvck file if there are new available versions that fit the version requirements - even if nothing changes in the composer.json file.

bordoni (changes requested):
Remove the lock update since we dont need to update versions unless it's something we meant to update. Do we think we can get this ready still in May?

PR #4452, opened 2024-01-09
Fix/tec 4918 venue block not persisting

stratease:
[TEC-4918](https://stellarwp.atlassian.net/browse/TEC-4918)

There was some leftover single venue state logic that was referencing a single venue. Updated the blocks to reference the `venue` attribute (prop) passed into the components instead of the state.

Artifact:
https://www.loom.com/share/74a97a6901a8412cabc5f52b1e3cb62b?sid=62e8f8f7-e526-4b1c-9f6e-904f18f15269

[TEC-4918]: https://stellarwp.atlassian.net/browse/TEC-4918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/blocks/event-venue/template.js
+++ b/src/modules/blocks/event-venue/template.js
@@ -354,9 +354,11 @@ class EventVenue extends Component {
 	 * Gets the venue ID for the block.
 	 *
 	 * @since 6.2.0
+	 * @since TBD This will now return the value of the `venue` prop.
 	 * @returns {number|null} Venue ID or null.
 	 */
 	getVenueId() {
+		const state = this.props.store.getState();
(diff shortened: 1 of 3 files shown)


src/modules/blocks/event-venue/template.js
```diff
@@ -357,12 +357,7 @@ class EventVenue extends Component {
 	 * @returns {number|null} Venue ID or null.
```
EugeneKyale, line 357: Please update the docblock to include a record of the change.

EugeneKyale (dismissed):

PR #4454, opened 2024-01-09
 [ET-1925] Add information to telemetry for Event Tickets

redscar:
[ET-1925](https://stellarwp.atlassian.net/browse/ET-1925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

This PR focuses on adding stats to the Site Health under 'Event Tickets' that require The Events Calendar. 

It extends this PR https://github.com/the-events-calendar/event-tickets/pull/2877

image

[ET-1925]: https://stellarwp.atlassian.net/browse/ET-1925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


codingmusician:
@redscar We want to start moving away from using the term `Provider` and use `Controller` instead. Also, we can just put all the hooks within the `Controller` and get rid of the `Hooks` file. See the `Tickets_Wallet_Plus` integration for an example.

src/Events/Integrations/Plugins/Event_Tickets/Site_Health/Subsection.php
```diff
@@ -0,0 +1,107 @@
...
+
+	/**
+	 * @inheritDoc
+	 */
+	protected function is_subsection_enabled(): bool {
+		return class_exists(
+			'Tribe__Events__Main',
+			false
+		);
+	}
```
codingmusician, line 30: Is this dependent on TEC? So if TEC isn't enabled, these stats won't show?

codingmusician (commented):
Looks great. I left one question, though.

src/Events/Integrations/Plugins/Event_Tickets/Site_Health/Subsection.php
codingmusician: I just realized that this was for the TEC part of it. If TEC wasn't enabled, then this class will never run. I think we can just return true in this case.

codingmusician (approved):

PR #4465, opened 2024-01-24
Feat/ecp 1620 block template refactor

stratease:
[ECP-1620](https://stellarwp.atlassian.net/browse/ECP-1620) & [ECP-1622](https://stellarwp.atlassian.net/browse/ECP-1622)

Large refactor to get us closer to the[ React specs](https://docs.google.com/document/d/1Nn87aPFTbzDdJlIefaAtZffJy_96JY-dzamRclRuIac/edit#heading=h.8ej7n38ju3g9).

[ECP-1620]: https://stellarwp.atlassian.net/browse/ECP-1620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ECP-1622]: https://stellarwp.atlassian.net/browse/ECP-1622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Main.php
```diff
@@ -683,7 +683,7 @@ public function bind_implementations(  ) {
 			tribe_register_provider( TEC\Events\Blocks\Controller::class );
 
 			// Site Editor
-			tribe_register_provider( TEC\Events\Editor\Full_Site\Controller::class );
```
jesseeproductions, line 686: Do we need to deprecate any of these old clases?

.github/workflows/tests-php.yml
```diff
@@ -18,7 +18,6 @@ jobs:
           - blocks_editor_integration
           - event_status
           - integration
-          - integrations_plugin_wordpress_seo
```
jesseeproductions, line 21: Will this be turned on again in the future?

stratease: It was forked to another action.

src/Tribe/Main.php
stratease: I don't think it's necessary, they are very new. @Camwyn or @bordoni ?

src/Events/Blocks/Controller.php
```diff
@@ -50,25 +50,25 @@ protected function add_actions() {
...
 
 	/**
-	 * Registers the Events Archive template.
+	 * Registers the Events Archive block.
 	 *
 	 * @since 6.2.7
 	 */
-	public function action_register_archive_template() {
-		return $this->container->make( Archive_Block_Template::class )->register();
+	public function register_archive_events_block() {
```
Camwyn, line 62: I know these are "new", but best practice would be to deprecate and return the new function.

src/Tribe/Main.php
Camwyn: I think we should to be honest. Anything public might be in use _somewhere_, and it covers us if we missed something.

Camwyn (changes requested):
I'm with Brian on deprecating.

src/Tribe/Main.php
stratease: I don't want to register both - does the deprecated need to be registered as well?

Camwyn: no

Camwyn (approved):
I only see failing `phpstan` tests so - approved.

PR #4474, opened 2024-02-02
Fix a php compatibility issue

Camwyn:
[TEC-4585]

This fixes the func_get_args notice in PHP 8+
Also cleans up a bit of phpcs issues.

Note the ignore: a lot of these functions aren't built out, so we tell phpcs to ignore the rule in here for now.


stratease (dismissed):
Unfortunately a merge conflict (think we both fixed the same error). Otherwise looks good.

PR #4501, opened 2024-03-27
Make the widgets register and enqueue their own styles.

Camwyn:
Working for all that have existing styles in the template:
Screenshot 2024-03-26 at 9 11 33 PM

Now, as we add styles, we just have to add the param to the widget and the rest happens automatically.

**Now contains the consolidation changes and all the new styling.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "the-events-calendar",
-  "version": "6.3.5",
+  "version": "6.4.0",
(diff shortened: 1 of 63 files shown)


bordoni (dismissed):
I think it's ok for now. But we will need to modify this.

Since these are Elementor Widgets we want to avoid having an instance of the widget initialized just for the assets, I will come back to this next week.

bordoni (approved):

PR #4527, opened 2024-04-05
TEC-5072: Use Default Currency in Event Creation

pattihis:
### 🎫 Ticket

[TEC-5072]

### 🗒️ Description

When creating a new event the three currency fields (symbol, code and position) should be populated from the general settings options.

### ✔️ Checklist

[TEC-5072]: https://stellarwp.atlassian.net/browse/TEC-5072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


readme.txt
```diff
@@ -232,6 +232,9 @@ Remember to always make a backup of your database and files before updating!
 
 == Changelog ==
 
+= [TBD] TBD =
+* Fix - When creating a new event the Currency symbol, code and position fields are populated from the general settings options.
```
stratease, line 236: Add the [ticket id].

src/admin-views/events-meta-box.php
```diff
@@ -223,7 +223,7 @@ class="tribe-field-timezone tribe-dropdown hide-if-js"
 							id='EventCurrencySymbol'
 							name='EventCurrencySymbol'
 							size='2'
-							value='<?php echo isset( $_EventCurrencySymbol ) ? esc_attr( $_EventCurrencySymbol ) : tribe_get_option( 'defaultCurrencySymbol', '$' ); ?>'
+							value='<?php echo isset( $_EventCurrencySymbol ) && ! empty( $_EventCurrencySymbol ) ? esc_attr( $_EventCurrencySymbol ) : tribe_get_option( 'defaultCurrencySymbol', '$' ); ?>'
```
stratease, line 226: Maybe we can trim the long if statements, and resolve into a flag used, something like `$has_currency_position = isset( $_EventCurrencyPosition ) && ! empty( $_EventCurrencyPosition )`?

stratease (changes requested):
Also resolve the code style reviews.

bordoni (dismissed):

readme.txt
bordoni: Lets fix this one.

src/admin-views/events-meta-box.php
bordoni: Since this is super old legacy code that we aim to deprecate completely at some point this year, I would avoid going into any refactor if we can avoid.

readme.txt
Camwyn, line 236:
```suggestion
* Fix - When creating a new event the Currency symbol, code and position fields are populated from the general settings options. [TEC-5072]
```

src/admin-views/events-meta-box.php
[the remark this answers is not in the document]
Camwyn: isset + empty is redundant. empty dosn't throw an error if the variable isn't set.
```suggestion
							value='<?php echo ! empty( $_EventCurrencySymbol ) ? esc_attr( $_EventCurrencySymbol ) : tribe_get_option( 'defaultCurrencySymbol', '$' ); ?>'
```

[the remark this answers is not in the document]
Camwyn: Also, we're not escaping if we get the option? I'd do something like:

```php 
<?php $currency_symbol = $_EventCurrencySymbol ?? tribe_get_option( 'defaultCurrencySymbol', '$' ); ?>
value=<?php echo esc_attr( $currency_symbol ); ?>
```
`null coalesce` simplifies and we ensure the value is always escaped.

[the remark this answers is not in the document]
Camwyn: You can add an ignore with a comment to remove these. It's a legacy variable, we can't easily rename.

src/admin-views/events-meta-box.php
```diff
@@ -234,9 +234,9 @@ class="tribe-dropdown"
 							data-prevent-clear
 						>
 							<?php
-							if ( isset( $_EventCurrencyPosition ) && 'suffix' === $_EventCurrencyPosition ) {
+							if ( isset( $_EventCurrencyPosition ) && ! empty( $_EventCurrencyPosition ) && 'suffix' === $_EventCurrencyPosition ) {
```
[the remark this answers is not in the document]
Camwyn: Similar comment about isset + empty
There are a few of these, I'm not going to tag each one 😉

Camwyn (changes requested):

src/admin-views/events-meta-box.php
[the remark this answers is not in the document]
George-Pat: - Yes but the bug here was exactly because the value was indeed set, but as an empty string `''` .
- I didn't escape the `get_option` because it was already there raw, maybe it is already validated. Ok, will make the change

src/admin-views/events-meta-box.php
```diff
@@ -233,34 +234,25 @@ class='alignleft'
...
-								$suffix = false;
-							}
+							<?php // phpcs:ignore -- Legacy naming.
+							$suffix = $_EventCurrencyPosition ? ( 'suffix' === $_EventCurrencyPosition ? true : false ) : ( tribe_get_option( 'reverseCurrencyPosition' ) ?: false );
 							?>
 							<option value="prefix"> <?php _ex( 'Before cost', 'Currency symbol position', 'the-events-calendar' ) ?> </option>
-							<option value="suffix"<?php if ( $suffix ) {
-								echo ' selected="selected"';
-							} ?>><?php _ex( 'After cost', 'Currency symbol position', 'the-events-calendar' ) ?></option>
+							<option value="suffix" <?php selected( $suffix, true ); ?>><?php _ex( 'After cost', 'Currency symbol position', 'the-events-calendar' ) ?></option>
```
Camwyn, line 241: It's worth pointing out that the `_e()` functions [don't escape](https://developer.wordpress.org/reference/functions/_e/#source). While the chance of using this as an attack vector is small, it's probably best practice to do `echo esc_html( _x() );`

Camwyn (changes requested):

src/admin-views/events-meta-box.php
Camwyn: FWIW WP VIP has dinged me on this kind of thing in the past.

pattihis: @Camwyn My change here was to remove this

```
<?php if ( $suffix ) { 
   echo ' selected="selected"';
} ?>>
```
and change to this

```
<?php selected( $suffix, true ); ?>
```
Because I couldn't stand a three-liner as an html element's attribute.

The `_ex()` was already there. I will escape it, no problem.

src/admin-views/events-meta-box.php
```diff
@@ -233,34 +234,25 @@ class='alignleft'
...
-								$suffix = false;
-							}
+							<?php // phpcs:ignore -- Legacy naming.
+							$suffix = $_EventCurrencyPosition ? ( 'suffix' === $_EventCurrencyPosition ? true : false ) : ( tribe_get_option( 'reverseCurrencyPosition' ) ?: false );
 							?>
-							<option value="prefix"> <?php _ex( 'Before cost', 'Currency symbol position', 'the-events-calendar' ) ?> </option>
-							<option value="suffix"<?php if ( $suffix ) {
-								echo ' selected="selected"';
-							} ?>><?php _ex( 'After cost', 'Currency symbol position', 'the-events-calendar' ) ?></option>
+							<option value="prefix"> <?php _ex( 'Before cost', 'Currency symbol position', 'the-events-calendar' ); ?> </option>
```
Camwyn, line 240:
```suggestion
							<option value="prefix" selected( ! $suffix )> <?php _ex( 'Before cost', 'Currency symbol position', 'the-events-calendar' ); ?> </option>
```

For accessibility purposes, any selected option should be marked as such. Screen readers will read the selected item even before the select is opened.

Camwyn (dismissed):
One small tweak

stratease (dismissed):

PR #4536, opened 2024-04-10
New settings for hide end time.

stratease:
### 🎫 Ticket

[TEC-4371]

### 🗒️ Description

This PR moves **Remove End Time** settings from our **Tweaks** extension into the **TEC > Settings > Display** section. Fixes some compatibility issues with recent updates to the views. Also adds a compatibility layer in case of using an older Tweaks extension.

### 🎥 Artifacts 

image

### ✔️ Checklist

[TEC-4371]: https://stellarwp.atlassian.net/browse/TEC-4371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Events/Views/Modifiers/Visibility_Modifier_Abstract.php
stratease, line 1: @bordoni would love your review on this.

src/Events/Integrations/Plugins/TEC_Tweaks_Extension/Provider.php
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	public static function get_slug(): string {
+		return 'tribe-ext-tec-tweaks';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function load_conditionals(): bool {
+		return class_exists( \Tribe\Extensions\Tec_Tweaks\Main::class );
```
bordoni, line 30:
```suggestion
		return class_exists( \Tribe\Extensions\Tec_Tweaks\Main::class, false);
```

src/Events/Integrations/Plugins/TEC_Tweaks_Extension/Provider.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @inheritDoc
+	 */
+	protected function load(): void {
+		$this->remove_end_time_extension_settings();
+	}
+
+	/**
+	 * This handles removing the end time extension settings from the tweaks extension.
+	 */
+	public function remove_end_time_extension_settings() {
```
bordoni, line 43: Docblock

src/Events/Integrations/Plugins/TEC_Tweaks_Extension/Provider.php
```diff
@@ -0,0 +1,64 @@
...
+		apply_filters(
+			"tribe_get_option_tribe_ext_tec_tweaks_remove_event_end_time",
+			function () {
+				return [];
+			}
+		);
```
bordoni, line 49: Apply filters? And use [`__return_empty_array`](https://developer.wordpress.org/reference/functions/__return_empty_array/)

src/Events/Integrations/Plugins/TEC_Tweaks_Extension/Provider.php
```diff
@@ -0,0 +1,64 @@
...
+		add_filter(
+			'tribe_settings_tab_fields',
+			function ( $fields, $id ) {
```
bordoni, line 52: Static.

src/Events/Integrations/Provider.php
```diff
@@ -33,6 +33,12 @@ public function register() {
 		$this->container->register( Plugins\Colbri_Page_Builder\Provider::class );
 		$this->container->register( Plugins\Event_Tickets\Provider::class );
 		$this->container->register( Plugins\Elementor\Provider::class );
+		add_action(
+			'tribe_plugins_loaded',
+			function () {
+				$this->container->register( Plugins\TEC_Tweaks_Extension\Provider::class );
+			}
+		);
```
bordoni, line 41: Cant we use `$this->container->register_on_action()`

src/Tribe/Views/V2/Hooks.php
```diff
@@ -181,39 +214,73 @@ public function action_set_title_events( $events, $view ) {
...
+						]
+					);
+
+					return $settings;
+				}
+
+				// Is this view flagged to hide the end time?
+				$settings['show_end_time'] = $end_time_modifier->is_visible( $context->get( 'view' ) );
+
+				return $settings;
```
bordoni, line 281: For removing the old plugin anonymous functions are ok, not for mods like this.

src/Tribe/Views/V2/Hooks.php
```diff
@@ -181,39 +214,73 @@ public function action_set_title_events( $events, $view ) {
...
+			'tribe_events_views_v2_bootstrap_pre_get_view_html',
+			function ( $html, $view_slug, $query, $context ) use ( $end_time_modifier ) {
+				// We need further context to determine if we should hide the end time for a particular area.
+				$end_time_modifier->set_context( $context );
+
+				return $html;
+			},
+			10,
+			4
+		);
```
bordoni, line 257: For removing the old plugin anonymous functions are ok, not for mods like this.

src/Tribe/Views/V2/Hooks.php
```diff
@@ -181,39 +214,73 @@ public function action_set_title_events( $events, $view ) {
...
+		$end_time_modifier = new Hide_End_Time_Modifier( $views );
+
+		// Let's setup our context, in either one of two hooks.
+		add_action(
+			'tribe_views_v2_after_setup_loop',
+			function ( $view ) use ( $end_time_modifier ) {
+				// We need further context to determine if we should hide the end time for a particular area.
+				$end_time_modifier->set_context( $view->get_context() );
+			}
+		);
```
bordoni, line 246: For removing the old plugin anonymous functions are ok, not for mods like this.

src/Tribe/Views/V2/Template_Bootstrap.php
```diff
@@ -228,6 +228,13 @@ public function get_view_html() {
 		$should_display_single = apply_filters( 'tribe_events_views_v2_bootstrap_should_display_single', $should_display_single, $view_slug, $query, $context );
 
 		if ( $should_display_single ) {
+			// Set our context to read as a single-event view.
+			add_filter(
+				"tribe_context_pre_view",
+				function () {
+					return 'single-event';
+				}
+			);
```
bordoni, line 237: This doesn't feel like the correct place to do this at all.

src/admin-views/tribe-options-display.php
```diff
@@ -332,6 +332,27 @@ static function( $view ) {
...
+		'options'         => [
+			'recent' => esc_html__( 'Recent past events list', 'tec-labs-tec-tweaks' ),
+			'single-event' => esc_html__( 'Single event page', 'tec-labs-tec-tweaks' ),
+			'day'    => esc_html__( 'Day view', 'tec-labs-tec-tweaks' ),
+			'list'   => esc_html__( 'List view', 'tec-labs-tec-tweaks' ),
+			'month'  => esc_html__( 'Month view tooltip', 'tec-labs-tec-tweaks' ),
+		],
+		'validation_type' => 'options_multi',
+		'can_be_empty'    => true,
+	],
```
bordoni, line 355: Ideally all of this data would come from the actual settings field built by the visibility options abstract.

bordoni (changes requested):

src/Events/Integrations/Provider.php
stratease: Oh, good call. I am blind.

src/Tribe/Views/V2/Template_Bootstrap.php
stratease: @bordoni I agree, but what would you recommend? The flag of whether this is a "single event" is determined here, and without overriding the context the `view` slug still says `list`.

src/admin-views/tribe-options-display.php
stratease: IMHO Ideally in a standalone settings handler, not mixed with the visibility filter. I don't believe it should be [responsible to do multiple things](https://en.wikipedia.org/wiki/Single_responsibility_principle). Especially since we will still need settings to be handled in circumstances that the Visibility Modifier won't be necessary. 

That said, I think we can make this a lot better for sure. Thoughts?

src/Events/Integrations/Plugins/TEC_Tweaks_Extension/Provider.php
stratease: Oops.

src/admin-views/tribe-options-display.php
bordoni: I agree with you that ideally we would have a standalone Settings handler, the problem is that we wont be tackling that project any time soon due to a couple of other priorities that we will need to focus on.

The reason why I had suggested the use of the abstract on ECP was to avoid rework and having to define the options and option name in two places it improves how the reliability of the settings and how these settings are used.

Later when we go about refactoring the Settings framework we can just point those to the new place.

src/Tribe/Views/V2/Template_Bootstrap.php
bordoni: A couple of things:
* Avoid anonymous functions of this nature on production code they can't be removed.
* Move the adding *and removing* of the filter to inside of the `get_v1_single_event_html` which is where we handle all the legacy V1 single problems so the code makes a lot more sense there.
* Potentially use `tribe_context_view` instead of the `pre` filter.

src/admin-views/tribe-options-display.php
stratease: Made some updates. Are you asking me to refactor to put the settings in the View Modifier class?

bordoni: ;)

src/admin-views/tribe-options-display.php
```diff
@@ -332,6 +332,27 @@ static function( $view ) {
 			'11:00' => date_i18n( $site_time_format, strtotime( '11:00 am' ) ),
 		],
 	],
+
+	'remove_event_end_time'             => [
+		'type'            => 'checkbox_list',
+		'label'           => esc_html__( 'Remove event end time', 'tec-labs-tec-tweaks' ),
```
Camwyn: All of the domains are incorrect

src/admin-views/tribe-options-display.php
```diff
@@ -332,6 +332,27 @@ static function( $view ) {
...
+		'tooltip'         => esc_html__(
+			                     'When this box is checked the end time will no longer display for events that end on the same day when viewing the List and Day views, the recent past events list, the tooltip in Month and Week (Pro) views, as well as on the event page itself.',
```
Camwyn: This text references ECP views.

PR #4538, opened 2024-04-12
Updated Docblock for tribe_create_event()

sdokus:
### 🎫 Ticket

No associated ticket (minor tweak)

### 🗒️ Description

Adjusted docblock for tribe_event_create() to redirect users to use the TEC ORM instead since the function is outdated and will be deprecated in the future. 

### ✔️ Checklist
- [ x] Changelog entry in the `readme.txt` file.
- [ NA] Have you added Artifacts?
- [ x] Check the base branch for your PR.

### DIFF
--- a/src/functions/advanced-functions/event.php
+++ b/src/functions/advanced-functions/event.php
@@ -15,7 +15,7 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	/**
 	 * Creates an Event using the legacy method.
 	 *
-	 * Note: This function is outdated and should be replaced with the TEC ORM `create()` method {@see Tribe__Repository::create()}.
+	 * Note: This function is outdated and should be replaced with the [TEC ORM `tribe_events()->create()` method](https://docs.theeventscalendar.com/apis/orm/create/events/).
 	 *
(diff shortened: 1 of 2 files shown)


src/functions/advanced-functions/event.php
```diff
@@ -24,78 +26,79 @@
...
+	 *                                           }
 	 *
 	 * @return int|bool ID of the event that was created. False if insert failed.
+	 * @category Events
+	 *
+	 * @version  TBD
+	 * @since    TBD Updated docblock to indicate future deprecation of function
 	 */
-	function tribe_create_event( $args ) {
+	function tribe_create_event( array $args ): bool|int {
```
bordoni, line 101: We have to be very careful about changes of this nature, for example:
1. Multiple expected return types are not available on our base PHP version 7.4
2. We should avoid adding type hinting on existing method that have public usage, since WP development hasn't fully embraced these yet we can cause Fatals that would be on our software and we would be the ones responsible for the support in some way.

bordoni (changes requested):
With your changes a couple of the docblocks are now broken but also the order is now weird, check those like:
`@category`, `@since` and `@version`

PR #4544, opened 2024-04-16
TEC-4667 Show Notice when REST API is blocked

pattihis:
### 🎫 Ticket

[TEC-4667]

### 🗒️ Description

Added a class names `Rest_Api.php` under `src/Tribe/Admin/Notice` to register an admin notice that would warn the user when our Rest Api endpoints are not accessible.

### ✔️ Checklist

[TEC-4667]: https://stellarwp.atlassian.net/browse/TEC-4667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Events/Admin/Notice/Rest_Api.php
+++ b/src/Events/Admin/Notice/Rest_Api.php
@@ -4,10 +4,10 @@
  *
  * @since
  *
- * @package Tribe\Events\Admin\Notice
+ * @package TEC\Events\Admin\Notice
  */
 
-namespace Tribe\Events\Admin\Notice;
+namespace TEC\Events\Admin\Notice;
 
 use Tribe\Events\Views\V2\Rest_Endpoint as V2;
 
@@ -61,7 +61,7 @@ class Rest_Api {
 	}
 
 	/**
-	 * Checks if we are in an TEC page or in main admin Dashboard.
+	 * Checks if we are in a TEC page or in the main Dashboard.
 	 *
 	 * @since
 	 *
--- a/src/Events/Admin/Notice/Rest_Api.php
+++ b/src/Events/Admin/Notice/Rest_Api.php
@@ -52,9 +52,11 @@ class Rest_Api {
 			$slug,
 			[ $this, 'notice' ],
 			[
-				'type'    => 'error',
-				'dismiss' => 1,
-				'wrap'    => 'p',
+				'type'               => 'error',
+				'dismiss'            => 1,
+				'wrap'               => 'p',
+				'recurring'          => true,
+				'recurring_interval' => 'P7D',
 			],
 			[ $this, 'should_display' ]
 		);
@@ -116,12 +118,17 @@ class Rest_Api {
 			return false;
 		}
 
-		$text = [];
-		// Translators: %s is the "Warning" word in bold.
-		$text[] = sprintf( __( '%s : The Events Calendar REST API endpoints are not accessible! This may be due to a server configuration or another plugin blocking access to the REST API.', 'the-events-calendar' ), '<strong>' . __( 'Warning', 'the-events-calendar' ) . '</strong>' );
-		$text[] = __( 'Please check with your hosting provider or system administrator to ensure that the below is accessible:', 'the-events-calendar' );
-		$text[] = '<p><a href="' . $this->blocked_endpoint . '" target="_blank">' . $this->blocked_endpoint . '</a></p>';
+		$output = sprintf(
+			/* Translators: %1$s and %2$s - opening and closing strong tags, respectively. */
+			__( '%1$sWarning%2$s: The Events Calendar REST API endpoints are not accessible! This may be due to a server configuration or another plugin blocking access to the REST API.', 'the-events-calendar' ),
+			'<strong>',
+			'</strong>'
+		);
+		$output .= '<br />';
+		$output .= __( 'Please check with your hosting provider or system administrator to ensure that the below is accessible:', 'the-events-calendar' );
+		$output .= '<br />';
+		$output .= '<a href="' . $this->blocked_endpoint . '" target="_blank">' . $this->blocked_endpoint . '</a>';
 
-		return implode( '<br />', $text );
+		return $output;
 	}
 }
--- a/src/Events/Admin/Notice/Rest_Api.php
+++ b/src/Events/Admin/Notice/Rest_Api.php
@@ -87,7 +87,7 @@ class Rest_Api {
 	public function is_rest_api_blocked(): bool {
 
 		$v1_api    = new \Tribe__Events__REST__V1__Main();
-		$event_api = get_rest_url( null, $v1_api->get_namespace() . '/events/' . $v1_api->get_version() );
+		$event_api = get_rest_url( null, $v1_api->get_events_route_namespace() );
 		$response  = wp_remote_get( $event_api );
 		if ( is_wp_error( $response ) || wp_remote_retrieve_response_code( $response ) !== 200 ) {
 			$this->blocked_endpoint = $event_api;
--- a/src/Tribe/Admin/Notice/Rest_Api.php
+++ b/src/Tribe/Admin/Notice/Rest_Api.php
@@ -40,8 +40,10 @@ class Rest_Api {
 	 * Constructor.
 	 *
 	 * @since
+	 *
+	 * @return void
 	 */
-	public function hook() {
+	public function hook() : void {
 		$slug = $this->slug;
 
 		tribe_notice(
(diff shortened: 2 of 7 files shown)


src/Tribe/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,122 @@
...
+	/**
+	 * Checks if our endpoints are accessible.
+	 *
+	 * @since
+	 *
+	 * @return boolean
+	 */
+	public function is_rest_api_blocked() {
+
+		$event_api = get_rest_url( null, '/tribe/events/v1/' );
```
bordoni, line 85: These can be filtered and modified on the REST endpoint, we should use the methods there to get the URL we are testing.

src/Tribe/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,122 @@
...
+		$response  = wp_remote_get( $event_api );
+		if ( is_wp_error( $response ) || wp_remote_retrieve_response_code( $response ) !== 200 ) {
+			$this->blocked_endpoint = $event_api;
+			return true;
+		}
+
+		$views_api = get_rest_url( null, 'tribe/views/v2/' );
```
bordoni, line 92: These can be filtered and modified on the REST endpoint, we should use the methods there to get the URL we are testing.

src/Tribe/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	/**
+	 * Checks if we are in an TEC page or in main admin Dashboard.
+	 *
+	 * @since
+	 *
+	 * @return boolean
+	 */
+	public function should_display() {
```
bordoni, line 66:
```suggestion
	public function should_display(): bool {
```

src/Tribe/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	/**
+	 * Checks if our endpoints are accessible.
+	 *
+	 * @since
+	 *
+	 * @return boolean
+	 */
+	public function is_rest_api_blocked() {
```
bordoni, line 83:
```suggestion
	public function is_rest_api_blocked(): bool {
```

src/Tribe/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,122 @@
...
+
+		return false;
+	}
+
+	/**
+	 * HTML for the notice when we have blocked REST API endpoints.
+	 *
+	 * @since
+	 *
+	 * @return string
```
bordoni, line 107:
```suggestion
	 * @return false|string
```

src/Tribe/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @var string
+	 */
+	private $blocked_endpoint = '';
+
+	/**
+	 * Constructor.
+	 *
+	 * @since
+	 */
+	public function hook() {
```
bordoni, line 44:
```suggestion
	public function hook(): void {
```

src/Tribe/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,122 @@
+<?php
+/**
+ * Shows an admin notice when our REST API endpoints are not available.
+ *
+ * @since
+ *
+ * @package Tribe\Events\Admin\Notice
+ */
+
+namespace Tribe\Events\Admin\Notice;
```
bordoni, line 10: Check if these can be created on the `TEC\Events\Admin\Notice` endpoint, it will be in a diff folder, but we should try to add any new classes over there.

bordoni (changes requested):

src/Events/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,127 @@
...
+	 * Checks if our endpoints are accessible.
+	 *
+	 * @since
+	 *
+	 * @return boolean
+	 */
+	public function is_rest_api_blocked(): bool {
+
+		$v1_api    = new \Tribe__Events__REST__V1__Main();
+		$event_api = get_rest_url( null, $v1_api->get_namespace() . '/events/' . $v1_api->get_version() );
```
bordoni, line 90: Lets get this from `Tribe__Events__REST__V1__Main::get_events_route_namespace`, turn that method `public` I see no reason to keep it `protected`, what do you think?

bordoni (changes requested):
One last change. Good stuff so far.

bordoni (approved):

src/Events/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,127 @@
...
+	 * @return false|string
+	 */
+	public function notice() {
+		if ( ! current_user_can( 'activate_plugins' ) ) {
+			return false;
+		}
+
+		$text = [];
+		// Translators: %s is the "Warning" word in bold.
+		$text[] = sprintf( __( '%s : The Events Calendar REST API endpoints are not accessible! This may be due to a server configuration or another plugin blocking access to the REST API.', 'the-events-calendar' ), '<strong>' . __( 'Warning', 'the-events-calendar' ) . '</strong>' );
```
Camwyn, line 121:
```suggestion
		/* Translators: %1$s and %2$s - opening and closing strong tags, respectively. */
				$text[] = sprintf(
			__( '%1$sWarning%2$s: The Events Calendar REST API endpoints are not accessible! This may be due to a server configuration or another plugin blocking access to the REST API.', 'the-events-calendar' ),
			'<strong>',
			'</strong>'
		);
```
Whenever possible, keep strings together in translations for better context. In this case we can just use placeholders for the tags.

src/Events/Admin/Notice/Rest_Api.php
```diff
@@ -0,0 +1,127 @@
...
+		$text[] = __( 'Please check with your hosting provider or system administrator to ensure that the below is accessible:', 'the-events-calendar' );
+		$text[] = '<p><a href="' . $this->blocked_endpoint . '" target="_blank">' . $this->blocked_endpoint . '</a></p>';
```
Camwyn, line 123: Why is this text the only piece wrapped in a paragraph?
Instead of the implode below, wrap them in `<p>`

Camwyn: Maybe 
```
		$text[] = sprintf(
			__( %1$sPlease check with your hosting provider or system administrator to ensure that %2$s is accessible.%3$s', 'the-events-calendar' ),
			'<p>',
			'<a href="' . $this->blocked_endpoint . '" target="_blank">' . $this->blocked_endpoint . '</a>','
			'</p>'
		);
```

And then just 
`return implode( '', $text );`

src/Events/Admin/Notice/Rest_Api.php
Camwyn: Otherwise: the "Warning" gets translated out of context - which could result in a poorer translation when combined.

Camwyn (changes requested):

Camwyn:
@pattihis - can you correct the conflicts as well please?

src/Events/Admin/Notice/Rest_Api.php
pattihis: In my experience, using placeholders for html tags in translation strings is a bad practice because it is prone to user errors. I have encountered un-closed html tags numerous times that broke entire pages, because of a user messing up the placeholders. 

IMO, we shouldn't expect users to know what HTML tags are and translation strings should not be trusted. Thus keeping html out of `__()` is the safest approach. Still, if this is something allowed here, then I will take that shortcut in the future.

As for better context, we have the `_x()` functions whenever needed.

stratease:
Should we remove the hold? I wasn't going to review because of the hold tag.

src/Events/Admin/Notice/Rest_Api.php
pattihis: I would have never use that implode approach for a translation string. This was [copied from here](https://github.com/the-events-calendar/the-events-calendar/blob/master/src/Tribe/Admin/Notice/Timezones.php) and as we discussed in our last meeting, I wasn't aware of which parts of the plugin are legacy code and which to follow as a standard.

As for the actual text, this is an example I came up with and I was waiting for proper specs.

pattihis:
> @pattihis - can you correct the conflicts as well please?

This is not ready to be merged as I am waiting for instructions for the UI/UX and any other features, as discussed in the last team meeting.

PR #4565, opened 2024-04-24
Add note to docblock for tribe_create_event() to indicate deprecation

sdokus:
### 🎫 Ticket

No associated ticket (minor tweak)

### 🗒️ Description

Adjusted docblock for tribe_create_event() to redirect users to use the TEC ORM instead since the function is outdated and will be deprecated in the future.

### ✔️ Checklist


src/functions/advanced-functions/event.php
```diff
@@ -94,6 +94,10 @@
 	 * }
 	 *
 	 * @return int|bool ID of the event that was created. False if insert failed.
+	 * @category Events
+	 *
+	 * @version  TBD
+	 * @since    TBD Updated docblock to indicate future deprecation of function.
```
bordoni, line 100: There is no need for `@version`, but I would move this to the top of the Docblock, right after the free-text "area"
```suggestion
	 * @since    3.0.0
```

bordoni (changes requested):

src/functions/advanced-functions/event.php
```diff
@@ -94,6 +94,9 @@
 	 * }
 	 *
 	 * @return int|bool ID of the event that was created. False if insert failed.
+	 * @category Events
+	 *
+	 * @since    3.0.0
```
bordoni, line 99: Move this `@since` to be the first thing after the Description.

bordoni (changes requested):

sdokus:
I removed the @category tag since it isn't included in the WP Standards, but now I'm second guessing if that will break things or not. Should I re-add it?

bordoni (approved):

bordoni:
> I removed the @category tag since it isn't included in the WP Standards, but now I'm second guessing if that will break things or not. Should I re-add it?

I think it's a good idea to remove.

PR #4576, opened 2024-05-02
Added note to tribe_organizer_create and added @SInCE, and readme.txt…

sdokus:
… note### ### 🎫 Ticket

No associated ticket (minor tweak)

### 🗒️ Description

Adjusted docblock for tribe_create_organizer() to redirect users to use the TEC ORM instead since the function is outdated and will be deprecated in the future.

### ✔️ Checklist
- [x] Changelog entry in the `readme.txt` file.
- [ ] Code is covered by **NEW** `wpunit` or `integration` tests.
- [ ] Code is covered by **EXISTING** `wpunit` or `integration` tests.
- [ ] Are all the **required** tests passing?
- [x] Automated code review comments are addressed.
- [ ] Have you added Artifacts?
- [x] Check the base branch for your PR.
- [ ] Add your PR to the project board for the release.

PR #4582, opened 2024-05-10
DocBlock: `tribe_create_organizer()`

sdokus:
Duplicate of #4576

### DIFF
--- a/src/functions/advanced-functions/organizer.php
+++ b/src/functions/advanced-functions/organizer.php
@@ -27,15 +27,15 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	 * - Website string - URL of the Organizer.
 	 * - Phone string - Phone number for the Organizer.
 	 *
-	 * @link     http://codex.wordpress.org/Function_Reference/wp_insert_post
+	 * @category Organizers
+	 * @since 3.0.0
+	 *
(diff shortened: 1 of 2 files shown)


src/functions/advanced-functions/organizer.php
```diff
@@ -13,8 +13,11 @@
 if ( class_exists( 'Tribe__Events__Main' ) ) {
 
 	/**
-	 * Create an Organizer
+	 * Create an Organizer using the legacy method.
 	 *
+	 * Note: This function is outdated and should be replaced with the [TEC ORM `tribe_organizers()->create()` method](https://docs.theeventscalendar.com/apis/orm/create/organizers/).
```
stratease, line 18: Should we be adding deprecations here...?

stratease (changes requested):

src/functions/advanced-functions/organizer.php
sdokus: @stratease  the deprecation tag is currently removing the function completely from our docs so until that is fixed @bordoni says we should add a note to the DocBlock to at least give people a heads up that it will be deprecated soon

stratease (approved):

stratease (approved):

stratease (approved):

PR #4603, opened 2024-05-29
Fix search by location

dgwatkins:
I had a chat with @andrasguseo over Slack and I created this PR to show a possible fix.


andrasguseo:
For info, here's the discusson:
https://lw.slack.com/archives/C01M100PPPU/p1713263385808199

PR #4608, opened 2024-05-31
Fix/tectria 49 rest password

bordoni:
### 🎫 Ticket

[TECTRIA-49]

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -232,6 +232,10 @@ Remember to always make a backup of your database and files before updating!
 
 == Changelog ==
 
+= [6.5.0.1] 2024-06-03 =
+
+* Fix - Ensure REST API notice only checks on Events admin pages and once every 48 hours regardless of transients rules. [TECTRIA-49]
+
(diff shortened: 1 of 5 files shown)


readme.txt
```diff
@@ -232,6 +232,11 @@ Remember to always make a backup of your database and files before updating!
 
 == Changelog ==
 
+= [6.5.0.1] 2024-06-03 =
+
+* Fix - Ensure REST API notice only checks on Events admin pages and once every 48 hours regardless of transients rules. [TECTRIA-49]
+* Security - Ensure REST API Events that are password protect are not accessible without the password. [TECTRIA-49]
```
Camwyn, line 238:
```suggestion
* Security - Ensure Events that are password protected are not accessible via the REST API  without the password. [TECTRIA-49]
```

Camwyn (changes requested):
This looks good - are we already covering private events as well? If not maybe we should as part of this work?

PR #4618, opened 2024-06-06
Docblock tribe_is_day function and filter

deblynprado:


src/functions/template-tags/day.php
```diff
@@ -1,23 +1,29 @@
...
 	 */
 	function tribe_is_day(): bool {
 		$context  = tribe_context();
 		$is_day = tec_is_view( \Tribe\Events\Views\V2\Views\Day_View::get_view_slug() );
 
 		/**
-		 * Allows filtering of the tribe_is_day boolean value.
+		 * Filters the boolean value indicating if the current view is the Day View.
 		 *
-		 * @since 6.0.7 Added context to parameters.
```
bordoni, line 17: Why did we remove the `@since`?

bordoni (changes requested):

bordoni:
PR is currently missing labels

src/functions/template-tags/day.php
codingmusician: `@since` should have been moved below the description and above the `@param` list.

deblynprado: I should move the `@since` but removed instead, my bad.

bordoni:
Move to `release/T24.babayaga`

PR #4654, opened 2024-07-05
Docblock: `tribe_get_map_link()` and `tribe_get_map_link_html()`

andrasguseo:
* Update dockblocks for `tribe_get_map_link()` and `tribe_get_map_link_html()`
* Add dockblocks for the filters `tribe_events_google_map_link`, `tribe_get_map_link`, and `tribe_get_map_link_html`


src/functions/template-tags/google-map.php
```diff
@@ -66,20 +68,37 @@ function tribe_get_map_link( $post_id = null ) {
...
 
 /**
  * Returns a formed HTML link to Google Maps for the given event.
+ * 
+ * @since 4.6.24
  *
  * @category Events
  *
- * @param string $postId
+ * @param string $postId The event post ID.
```
codingmusician, line 99: I saw that you had change the `$postId` to `int` above. Does this one also need to be updated?
```suggestion
 * @param int $postId The event post ID.
```

codingmusician (approved):

PR #4668, opened 2024-07-12
Docblock: `tribe_embed_google_map()`

andrasguseo:
* Update docblock
* Change a variable name to match current naming convention
* Add docblock to the filter

Readme change is missing on purpose. Waiting for the Changelogger. :)


src/functions/template-tags/google-map.php
```diff
@@ -117,30 +117,35 @@ function tribe_get_embedded_map( $post_id = null, $width = null, $height = null,
 }
 
 /**
- * Google Map Embed Test
- *
- * Check if embed google map is enabled for this event (or venue ).
+ * Check if the "Show Map" option is enabled for the event or venue.
  *
  * @category Events
```
Camwyn, line 122: Should category include Venues?

andrasguseo: Like we discussed on slack, `@category` is deprecated and we're going to leave them as they are for now. :meh:

Camwyn (approved):

pattihis (approved):

PR #4681, opened 2024-07-25
Enhance the REST API disabled notice to avoid false positives

JPry:
### 🎫 Ticket

[TEC-5105]

### 🗒️ Description

This PR changes the logic that determines whether the "REST API blocked" notice is displayed. Most notably, it does 2 things:

* An error due to a timeout isn't counted as the API being blocked.
* An SSL error on a development site isn't counted as the API being blocked.

This PR adds the following new filters:

* `tec_events_rest_api_response_blocked` – This allows 3rd party code to use their own logic to determine if the API is blocked
* `tec_events_rest_api_response_blocked_due_to_timeout` – This filters whether a timeout from the API counts as the API being blocked. Passing `true` to this filter restores the previous behavior.
* `tec_events_site_is_development_mode` – Whether the site is in development mode.

Another notable addition in this API is the `\TEC\Events\Traits\Development_Mode` trait, located in `src/Events/Traits/Development_Mode.php`. This trait provides the `is_site_development_mode()` method, and the result of the check is passed through the `tec_events_site_is_development_mode` filter (mentioned above).

I thought that determining if the site was in development mode could be something that's useful elsewhere, and so I decided that the trait was the best way to implement that addition in a way that would be easy to reuse elsewhere.

### ✔️ Checklist

[TEC-5105]: https://stellarwp.atlassian.net/browse/TEC-5105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JPry:
I see that the PHPStan, npm lint, and some codeception tests are failing, but none of those appear to be failing as a result of the files changed in this PR.

redscar:
While I was reviewing I was thinking, should the trait live in Common so it can be used within any of the plugins?

src/Events/Admin/Notice/Rest_Api.php
```diff
@@ -145,4 +141,94 @@ public function notice() {
...
+	/**
+	 * Checks if the response is blocked.
+	 *
+	 * @since TBD
+	 *
+	 * @param array|WP_Error $response The response from the REST API.
+	 *
+	 * @return bool
+	 */
+	private function is_response_blocked( $response ): bool {
```
redscar, line 154: Would this and  `is_wp_error_response_blocking` be better as protected instead of private?

JPry:
@redscar 

> should the trait live in Common so it can be used within any of the plugins?

Perhaps... but I know that there's been talk of reworking Common, so I don't know if it would make more sense to add it to Common or keep it here.

codingmusician (approved):

src/Events/Admin/Notice/Rest_Api.php
JPry: @redscar If there's a good reason to change it to `protected`, I'm willing to do it. Can you think of a benefit of `protected` for these over `private`? I tend to just prefer `private` as a default, and will loosen restrictions progressively as needed.

src/Events/Admin/Notice/Rest_Api.php
```diff
@@ -96,23 +99,16 @@ public function is_rest_api_blocked( bool $force = false ): bool {
...
+		// Check multiple endpoints to determine if the REST API is blocked.
+		$endpoints = $this->get_routes_to_check();
+		foreach ( $endpoints as $endpoint ) {
+			$response = wp_safe_remote_get( $endpoint );
+			if ( $this->is_response_blocked( $response ) ) {
+				$this->blocked_endpoint = $endpoint;
+				tec_timed_option()->set( $cache_key, $endpoint, $cache_timeout );
 
-			return true;
+				return true;
```
dpanta94, line 110: Do we need to check multiple endpoints to determine this ? If yes:

- The first blocked endpoint bails early and returns true
- Why not the first not-blocked doesn't bail early and return false ?

src/Events/Admin/Notice/Rest_Api.php
dpanta94: I think we could bring this up for discussion, but my belief as well is to go for `protected` by default rather than `private` unless there is a good reason to go for `private`. My logic is that the class if extended should have its method overwrittable unless i have reason not to desire that.

src/Events/Admin/Notice/Rest_Api.php
```diff
@@ -96,23 +99,16 @@ public function is_rest_api_blocked( bool $force = false ): bool {
...
-
-		$views_api = get_rest_url( null, V2::ROOT_NAMESPACE );
-		$response  = wp_remote_get( $views_api );
-		if ( is_wp_error( $response ) || wp_remote_retrieve_response_code( $response ) !== 200 ) {
-			$this->blocked_endpoint = $views_api;
-			tec_timed_option()->set( $cache_key, $views_api, $cache_timeout );
+		// Check multiple endpoints to determine if the REST API is blocked.
+		$endpoints = $this->get_routes_to_check();
+		foreach ( $endpoints as $endpoint ) {
+			$response = wp_safe_remote_get( $endpoint );
```
dpanta94, line 105: Consider the timeout as well please

src/Events/Admin/Notice/Rest_Api.php
```diff
@@ -145,4 +141,94 @@ public function notice() {
...
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	private function get_routes_to_check(): array {
+		$routes  = [];
+		$v1_main = tribe( 'tec.rest-v1.main' );
+
+		// Ensure the what we got from tribe() is the instance we expect.
```
dpanta94, line 225:
```suggestion
		// Ensure that what we got from tribe() is the instance we expect.
```

tests/wpunit/TEC/Tests/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,60 @@
...
+use WP_Error;
+
+/**
+ * Class RESTAPITest
+ *
+ * @since TBD
+ */
+class RESTAPITest extends WPTestCase {
+
+	function test_response_blocked_from_wp_error() {
```
dpanta94, line 22: In general i prefer the method named `it_should_do_that` or `should_do_that` and having `@test` as an annotation

tests/wpunit/TEC/Tests/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,60 @@
...
+		$rest_api = new Rest_Api();
+
+		// Set up a wrapper to test the private method.
+		$is_response_blocking = Closure::bind(
+			function( $response ) {
+				return $this->is_wp_error_response_blocking( $response );
+			},
+			$rest_api,
+			$rest_api
+		);
```
dpanta94, line 46: Please replace with using the `With_Uopz` trait and its method 

`$this->set_fn_return( 'wp_safe_remote_get', $request_return_timeout, true );`

Then you could actually execute the actual method by just faking the return of `wp_safe_remote_get`

tests/wpunit/TEC/Tests/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,60 @@
...
+		$this->assertFalse( $is_response_blocking( $request_return_timeout() ) );
+		add_filter( 'tec_events_rest_api_response_blocked_due_to_timeout', '__return_true' );
+		$this->assertTrue( $is_response_blocking( $request_return_timeout() ) );
+		remove_filter( 'tec_events_rest_api_response_blocked_due_to_timeout', '__return_true' );
+
+		// An SSL error should be blocking, unless we are in development mode.
+		$this->assertTrue( $is_response_blocking( $request_return_ssl_error() ) );
+		add_filter( 'tec_events_site_is_development_mode', '__return_true' );
+		$this->assertFalse( $is_response_blocking( $request_return_ssl_error() ) );
+		remove_filter( 'tec_events_site_is_development_mode', '__return_true' );
```
dpanta94, line 58: How problematic could it be if remove_filter is not called if the tests would fail earlier ? Could the test start bleeding making others fail ? Could explore an `@after` annotated method dedicated to remove the filters so we are sure they are removed even when a test fails

dpanta94 (changes requested):

src/Events/Admin/Notice/Rest_Api.php
JPry: > Do we need to check multiple endpoints to determine this

This is how it was already done, I just combined the endpoints into a single loop instead of duplicating the exact same code for each one.

For your secondary questions, the same answer applies: I just consolidated what was already happening in two separate blocks into a loop with 2 values.

I feel like it's worth checking multiple endpoints that are necessary, and to that end I feel like returning on the first failure makes the most sense.

src/Events/Admin/Notice/Rest_Api.php
JPry: @dpanta94 Could you elaborate?

src/Events/Admin/Notice/Rest_Api.php
JPry: > I think we could bring this up for discussion, but my belief as well is to go for protected by default rather than private unless there is a good reason to go for private. My logic is that the class if extended should have its method overwrittable unless i have reason not to desire that.

That's a good point, but I feel that `private` is the right choice by default, and if/when a class is extended would be the time to move necessary methods from `private` to `protected`. This class isn't extended, and it's also unlikely to be extended, so I think that `private` makes the most sense for where the class is right now.

src/Events/Admin/Notice/Rest_Api.php
dpanta94: Yes! I was talking about the `timeout` offset being passed into `wp_sage_remote_get` through the `$endpoint` variable. By default, if not specified otherwise its `5`. I think we could go to `3`. Not too important, i just wanted to make sure we are considering it.

src/Events/Admin/Notice/Rest_Api.php
dpanta94: Alrighty, sounds good! :D

changelog/fix-TEC-5105-fix-rest-api-disabled-notice
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Update the logic that displays the "REST API blocked" banner to reduce false positives.
```
dpanta94, line 4: Ticket id at the end for the changelog entry please. e.g. `...positives. [TEC-XXX]`

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,80 @@
...
+				'cURL error 60: SSL certificate problem: unable to get local issuer certificate'
+			);
+		};
+
+		$rest_api = new Rest_Api();
+
+		// A timeout should not be blocking, unless we filter it to true.
+		$this->set_fn_return( 'wp_safe_remote_get', $request_return_success, true );
+		$this->assertFalse( $rest_api->is_rest_api_blocked( true ) );
+		$this->unset_uopz_returns();
```
dpanta94, line 64: Shouldn't be here - its being called automatically after every test. If you check the trait's code you will see an `@after` annotation. That means it runs after every test. Similar there are `@before`, `@beforeClass` and `@afterClass`

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,80 @@
...
+
+		$this->set_fn_return( 'wp_safe_remote_get', $request_return_timeout, true );
+		add_filter( 'tec_events_rest_api_response_blocked_due_to_timeout', '__return_true' );
+		$this->assertTrue( $rest_api->is_rest_api_blocked( true ) );
+		$this->unset_uopz_returns();
```
dpanta94, line 69: Similar as above

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,80 @@
...
+		remove_filter( 'tec_events_rest_api_response_blocked_due_to_timeout', '__return_true' );
+
+		// An SSL error should be blocking, unless we are in development mode.
+		$this->set_fn_return( 'wp_safe_remote_get', $request_return_ssl_error, true );
+		$this->assertTrue( $rest_api->is_rest_api_blocked( true ) );
+		add_filter( 'tec_events_site_is_development_mode', '__return_true' );
+		$this->assertFalse( $rest_api->is_rest_api_blocked( true ) );
+		remove_filter( 'tec_events_site_is_development_mode', '__return_true' );
+		$this->unset_uopz_returns();
```
dpanta94, line 78: Similar as above

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,80 @@
...
+
+	/**
+	 * Clean up after this test class finishes.
+	 *
+	 * @return void
+	 */
+	public function wpTearDownAfterClass() {
+		remove_all_filters( 'tec_events_site_is_development_mode' );
+		remove_all_filters( 'tec_events_rest_api_response_blocked_due_to_timeout' );
+	}
```
dpanta94, line 32: I'd prefer if you used `@after` or `@afterClass` instead of the `wpTearDownAfterClass` method. I have discussed with Luca that the annotation versions are better supported in between version of PHPUnit.

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,80 @@
...
+		$this->assertTrue( $rest_api->is_rest_api_blocked( true ) );
+		$this->unset_uopz_returns();
+		remove_filter( 'tec_events_rest_api_response_blocked_due_to_timeout', '__return_true' );
+
+		// An SSL error should be blocking, unless we are in development mode.
+		$this->set_fn_return( 'wp_safe_remote_get', $request_return_ssl_error, true );
+		$this->assertTrue( $rest_api->is_rest_api_blocked( true ) );
+		add_filter( 'tec_events_site_is_development_mode', '__return_true' );
+		$this->assertFalse( $rest_api->is_rest_api_blocked( true ) );
+		remove_filter( 'tec_events_site_is_development_mode', '__return_true' );
```
dpanta94, line 77: Optional: You can skip removing filters here since these are going to be removed after the test in any case.

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
```diff
@@ -0,0 +1,80 @@
...
+		// A timeout should not be blocking, unless we filter it to true.
+		$this->set_fn_return( 'wp_safe_remote_get', $request_return_success, true );
+		$this->assertFalse( $rest_api->is_rest_api_blocked( true ) );
+		$this->unset_uopz_returns();
+
+		$this->set_fn_return( 'wp_safe_remote_get', $request_return_timeout, true );
+		add_filter( 'tec_events_rest_api_response_blocked_due_to_timeout', '__return_true' );
+		$this->assertTrue( $rest_api->is_rest_api_blocked( true ) );
+		$this->unset_uopz_returns();
+		remove_filter( 'tec_events_rest_api_response_blocked_due_to_timeout', '__return_true' );
```
dpanta94, line 70: Same as above :D

dpanta94 (changes requested):

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
JPry: Ah, thanks! I missed that.

What about when I'm overriding the same function multiple times?

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
JPry: Fair point. I intended the ones that run after the class is done to be a fail-safe in case the test failed and these lines weren't run.

tests/wpunit/TEC/Events/Admin/Notice/RESTAPITest.php
dpanta94: If you override - you should call the parent every time. Better to just create your own unique methods and annotate them with `@before` or `@after`

JPry: @dpanta94 I mean the use of `unset_uopz_returns()`. I'm overriding `wp_safe_remote_get()` multiple times in succession. It looks like it just adds to an unindexed array each time.

dpanta94: Ah sorry @JPry, yes your latest overwrite will be the one used

dpanta94 (approved):
Amazing! :rocket:

PR #4682, opened 2024-07-26
Support modified semver versioning - changelogger

dpanta94:
### 🎫 Ticket

Not a ticket.

### 🗒️ Description

Adds a class which is used to parse our modified version of Semver for the changelogger.

The following warning would appear if we would specify a version with 4 digits before.

```bash
Invalid --use-version: Version number “6.0.0.1” is not in a recognized format.
The specified version is not valid. This may cause problems in the future! Continuing anyway.
```

Using the new class for parsing our versions would avoid this warning.

### 🎥 Artifacts 

image

### ✔️ Checklist
- [ ] Changelog entry in the `readme.txt` file.


bordoni:
We should port this code to;

https://github.com/the-events-calendar/actions

Our goal is to have this sync to all repositories. @sdokus

Camwyn (changes requested):
I agree with @bordoni - this looks good but it doesn't belong here, it should get ported.

PR #4689, opened 2024-08-07
Update zip.yml to include NVM install.

redscar:
Update zip.yml to include NVM install.


borkweb (approved):
The updates here look good. 🎈  The unrelated views_integration issue is something that should probably get looked at in the release branch after!

PR #4766, opened 2024-09-30
Fixing broken tests

codingmusician:
### 🗒️ Description
Bug was introduced at some point, which was causing an error when running the `views_integration` tests. This PR fixes the bug that was causing the error and updates the related snapshots.


lucatume (approved):
Nice!

PR #4767, opened 2024-10-09
Docblock: `Tribe__Events__Aggregator__Records`

andrasguseo:
Update docblocks in the `Tribe__Events__Aggregator__Records` class.

### DIFF
--- a/src/Tribe/Aggregator/Records.php
+++ b/src/Tribe/Aggregator/Records.php
@@ -346,8 +346,8 @@ class Tribe__Events__Aggregator__Records {
 	/**
 	 * Returns an appropriate Record object for the given origin.
 	 *
-	 * @param string $origin The record import origin.
-	 * @param int|WP_Post The record post or post ID.
+	 * @param string      $origin The record import origin.
+	 * @param int|WP_Post $post   The record post or post ID.
 	 *
 	 * @return Tribe__Events__Aggregator__Record__Abstract An instance of the correct record class
 	 *                                                     for the origin or an unsupported record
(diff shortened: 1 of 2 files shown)


src/Tribe/Aggregator/Records.php
```diff
@@ -21,27 +25,33 @@ class Tribe__Events__Aggregator__Records {
...
+	 * @since 4.3.0
 	 *
 	 * @var self
 	 */
 	private static $instance;
 
 	/**
+	 * @since 4.3.0
+	 *
 	 * @var string The time, in "Y-m-d H:i:s" format, that's used to query records.
```
codingmusician, line 47:
```suggestion
	 * The time, in "Y-m-d H:i:s" format, that's used to query records.
	 *
	 * @since 4.3.0
	 *
	 * @var string
```

src/Tribe/Aggregator/Records.php
```diff
@@ -76,6 +101,17 @@ public function filter_edit_link( $link, $post, $context ) {
...
 	}
 
+	/**
+	 * Adjusting the "Delete Post" link.
+	 *
+	 * @since 4.3.0
+	 *
+	 * @param string $link    The delete link.
+	 * @param int    $post    The post ID.
+	 * @param bool   $context Whether to bypass the Trash and force deletion. Default false.
```
andrasguseo, line 111: @codingmusician This param should actually be `$force_delete` and not `$context`. It is not used anywhere in the method so it would be safe to change. Should it be changed?

codingmusician: I don't see that it's necessary to change the variable name. You're right. It would make more sense to use something more descriptive like `$force_delete`, but let's leave that for if/when this method is ever updated. Good question!

codingmusician (approved):

PR #4807, opened 2024-11-12
Help Hub - Added additional logic to redirect users who hit the old help page slug.

redscar:
### 🗒️ Description
Redirect users who go to `/edit.php?post_type=tribe_events&page=tec-events-help` to `/edit.php?post_type=tribe_events&page=tec-events-help-hub`. 

This scenario can occur when uploading the plugin and clicking the "Help" menu on the plugin installation confirmation screen.

[Help_Hub_Redirect_old_Slug.webm](https://github.com/user-attachments/assets/efa91ea7-2b11-48bf-9ef5-96877dd3934c)

### DIFF
--- a/src/Tribe/Admin/Settings.php
+++ b/src/Tribe/Admin/Settings.php
@@ -242,7 +242,6 @@ class Settings {
 
 		// Check if the request is for the old help page.
 		if ( Plugin::POSTTYPE === $post_type && 'tec-events-help' === $page ) {
-			// Build the new URL with wp_parse_args for security and consistency.
 			$new_url = add_query_arg(
 				[
 					'post_type' => Plugin::POSTTYPE,
@@ -251,7 +250,6 @@ class Settings {
 				admin_url( 'edit.php' )
 			);
 
-			// Perform a safe redirect to the new Help Hub page.
 			wp_safe_redirect( $new_url );
 			exit;
 		}
--- a/src/Tribe/Admin/Settings.php
+++ b/src/Tribe/Admin/Settings.php
@@ -241,7 +241,7 @@ class Settings {
 		$post_type = tribe_get_request_var( 'post_type' );
 
 		// Check if the request is for the old help page.
-		if ( 'tribe_events' === $post_type && 'tec-events-help' === $page ) {
+		if ( Plugin::POSTTYPE === $post_type && 'tec-events-help' === $page ) {
 			// Build the new URL with wp_parse_args for security and consistency.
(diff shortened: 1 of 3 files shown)


src/Tribe/Admin/Settings.php
```diff
@@ -224,6 +227,34 @@ public function add_admin_pages() {
...
+				[
+					'post_type' => Plugin::POSTTYPE,
+					'page'      => 'tec-events-help-hub',
+				],
+				admin_url( 'edit.php' )
+			);
+
+			wp_safe_redirect( $new_url );
+			exit;
+		}
```
bordoni, line 255: Invert the logic, pull the page slug from the objects you created instead of static strings.

bordoni (changes requested):

src/Tribe/Admin/Settings.php
```diff
@@ -224,6 +234,37 @@ public function add_admin_pages() {
...
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function redirect_to_help_hub(): void {
+		$page      = tribe_get_request_var( 'page' );
+		$post_type = tribe_get_request_var( 'post_type' );
+
+		// Exit if the request is not for the old help page.
+		if ( Plugin::POSTTYPE !== $post_type || 'tec-events-help' !== $page ) {
```
Camwyn, line 251: where is `tec-events-help` defined? can we move that one to a prop as well?

redscar: It was originally hard-coded in the file and was never stored. I moved it into `$original_help_slug`.

src/Tribe/Admin/Settings.php
```diff
@@ -202,6 +216,9 @@ public function add_admin_pages() {
 			]
 		);
 
+		// Redirects users from the outdated Help page to the new Help Hub page if accessed.
```
rafsuntaskin, line 219:
```suggestion
		// Redirects users from the outdated Help page to the new Help Hub page if accessed.
```
I don't have the full context but if the `$original_help_slug` is the actual outdated one then we should not call it original, that may confuse devs in future. Let's call it either classic or old instead.

rafsuntaskin (changes requested):

src/Tribe/Admin/Settings.php
redscar: Good point, I switched to `old_`

rafsuntaskin (approved):

PR #4831, opened 2024-11-21
Include fixes for Export Links to do JIT translation and Aggregator

bordoni:
This is super straightforward change, no translation related methods can run before `after_setup_theme` in WordPress after version 6.7, so for Export links I modified the code so only when we try to use the labels the translations would be called and for Aggregator I delay the registration of the notices.


src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
```diff
@@ -197,11 +197,34 @@ public function is_visible() {
...
+	}
+
+	/**
+	 * Returns the label for the link.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	abstract protected function label(): string;
```
dpanta94, line 211: Isnt it a little risky adding abstacts to already existing classes ?

src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
```diff
@@ -211,27 +234,50 @@ public function get_label( View $view = null ) {
 		 *
 		 * @return string $label The label that will be displayed.
 		 */
-		return apply_filters( "tec_views_v2_subscribe_links_{$slug}_label", $this->label, $this, $view );
+		return (string) apply_filters( "tec_views_v2_single_subscribe_links_{$slug}_label", $value, $this, $view );
```
dpanta94, line 237: Shouldn't we deprecate it instead of directly changing it ?

src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
```diff
@@ -211,27 +234,50 @@ public function get_label( View $view = null ) {
...
+	}
+
+	/**
+	 * Returns the label for the single event view.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	abstract protected function single_label(): string;
```
dpanta94, line 254: Same comment about adding abstract in already existing class.

src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
```diff
@@ -211,27 +234,50 @@ public function get_label( View $view = null ) {
...
-		 * @param Link_Abstract $link_obj The link object the label is for.
-		 * @param View          $view     The current View object.
+		 * @param string        $single_label The label that will be displayed.
+		 * @param Link_Abstract $link_obj     The link object the label is for.
+		 * @param View          $view         The current View object.
 		 *
 		 * @return string $label The label that will be displayed.
 		 */
-		return apply_filters( "tec_views_v2_single_subscribe_links_{$slug}_label", $this->single_label, $this, $view );
+		return (string) apply_filters( "tec_views_v2_single_subscribe_links_{$slug}_single_label", $value, $this, $view );
```
dpanta94, line 280: Deprecating instead of directly changing ?

dpanta94 (changes requested):

dpanta94 (commented):
Also lets make sure tests are passing

src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
bordoni: Why deprecate? It's the same filter

src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
bordoni: It adds a good value, if someone is extending they will know what they need to change it's ok.

src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
dpanta94: Its not, original was `_label` while new is `_single_label`

Camwyn (approved):

Camwyn (changes requested):
Missed the failing test 🤦🏼

dpanta94 (approved):

Camwyn (approved):

PR #4855, opened 2024-12-12
Fix the Venue country not saving

Camwyn:
### 🎫 Ticket

Sheet issue

### 🗒️ Description

We have to get the country name from the list, we were passing the abbreviation (which still should save)


src/Events/Admin/Onboarding/Steps/Venue.php
```diff
@@ -68,7 +68,7 @@ public static function process( $response, $request ): WP_REST_Response {
 		$new_venue['State']         = $venue['state'];
 		$new_venue['Province']      = $venue['state'];
 		$new_venue['Zip']           = $venue['zip'];
-		$new_venue['Country']       = $venue['country'];
+		$new_venue['Country']       = $country;
```
Camwyn, line 71: expand above for context - line 60

dpanta94 (approved):

PR #4869, opened 2024-12-24
Fix/tec 5339 wrong logic

pattihis:
### 🎫 Ticket

[TEC-5339]

### 🗒️ Description

This patch is not needed because `new Tribe__Data( $data, false )` does not return empty.

The actual problems are explained in [ECP-1856](https://stellarwp.atlassian.net/browse/ECP-1856?focusedCommentId=194969) and are related to the [EA Geocoder](https://github.com/the-events-calendar/event-aggregator-site/blob/server/production/wp-content/plugins/event-aggregator/src/Tribe/Geocoder.php)

[TEC-5339]: https://stellarwp.atlassian.net/browse/TEC-5339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ECP-1856]: https://stellarwp.atlassian.net/browse/ECP-1856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Camwyn (approved):

src/Tribe/Venue.php
```diff
@@ -532,8 +532,6 @@ public function create( $data, $post_status = 'publish', $avoid_duplicates = fal
 			$content = $data['Description'] ?? '';
 			$slug    = sanitize_title( $title );
 
-			$data_old = $data;
-
 			$data = new Tribe__Data( $data, false );
```
Camwyn, line 535: default data should be an empty array?

Camwyn: Here we set it as a Tribe__Data object - and the values are in `$data->data`, no?

Camwyn: `save_meta` takes an array for the second param, passing an object _might_ cause issues?

pattihis: It's ok because [Data](https://github.com/the-events-calendar/tribe-common/blob/d60b606481d179dd17d205505df33d10fb2bec47/src/Tribe/Data.php#L7) implements the [ArrayAccess](https://php.net/manual/en/class.arrayaccess.php) interface that allows accessing that data object as an array.

Camwyn: Resolved - Tribe_Data implements ArrayAccess so the save_meta function should not have any issues.

PR #4877, opened 2025-01-12
Register IAN on action

pattihis:
### 🎫 Ticket

[TEC-5336 ]

### 🗒️ Description
During QA for IAN's first release, we discovered that a fatal error was thrown if the user would downgrade from the latest to the previous version of TEC (since the new Notifications class wouldn’t exists in that one).

We patched the fix [with this PR](https://github.com/the-events-calendar/the-events-calendar/pull/4849/files), using a `class_exist()`. 

This PR improves that, by using [register_on_action](https://docs.theeventscalendar.com/reference/classes/tec-common-contracts-container/register_on_action/), and removes the `class_exists()` check.

### ✔️ Checklist

### DIFF
--- a/src/Events/Notifications/Provider.php
+++ b/src/Events/Notifications/Provider.php
@@ -25,20 +25,8 @@ class Provider extends Service_Provider {
 	 * @since 6.4.0
 	 */
 	public function register() {
-		add_action( 'admin_footer', [ $this, 'render_icon' ] );
-	}
+		$this->container->singleton( static::class, $this );
 
-	/**
-	 * Renders the Notification icon.
-	 *
-	 * @since 6.4.0
-	 */
-	public function render_icon() {
-		// TODO: Remove this check when we have a proper way to register it, like register_on_action.
-		if ( ! class_exists( Notifications::class ) ) {
-			return;
-		}
-
-		return $this->container->get( Notifications::class )->render_icon();
+		$this->container->register_on_action( 'tribe_plugins_loaded', Notifications::class );
 	}
 }
--- a/src/Events/Notifications/Provider.php
+++ b/src/Events/Notifications/Provider.php
@@ -25,8 +25,6 @@ class Provider extends Service_Provider {
 	 * @since 6.4.0
 	 */
 	public function register() {
-		$this->container->singleton( static::class, $this );
-
-		$this->container->register_on_action( 'tribe_plugins_loaded', Notifications::class );
+		$this->container->register_on_action( 'tec_common_ian_loaded', Notifications::class );
 	}
 }
--- a/changelog/fix-register_notifications_on_action
+++ b/changelog/fix-register_notifications_on_action
@@ -1,0 +1,4 @@
+Significance: patch
+Type: tweak
+
+Remove class_exists() and use register_on_action() for Notifications [TEC-5336]
--- a/src/Events/Integrations/Provider.php
+++ b/src/Events/Integrations/Provider.php
@@ -6,10 +6,10 @@
  *
  * @package TEC\Events\Integrations
  */
+
 namespace TEC\Events\Integrations;
 
 use TEC\Common\Contracts\Service_Provider;
-
 
 /**
  * Class Provider
--- a/src/Events/Notifications/Notifications.php
+++ b/src/Events/Notifications/Notifications.php
@@ -9,22 +9,42 @@
 
 namespace TEC\Events\Notifications;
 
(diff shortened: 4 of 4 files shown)


stratease (commented):
I assume there's some test somewhere for this already?

src/Events/Notifications/Provider.php
```diff
@@ -25,20 +25,8 @@ class Provider extends Service_Provider {
...
-	 * Renders the Notification icon.
-	 *
-	 * @since 6.4.0
-	 */
-	public function render_icon() {
-		// TODO: Remove this check when we have a proper way to register it, like register_on_action.
-		if ( ! class_exists( Notifications::class ) ) {
-			return;
-		}
+		$this->container->singleton( static::class, $this );
```
dpanta94, line 28: Architecture choices of a provider binding itself i think are confusing. Can we bind this provider 1 (or more) level up ?

src/Events/Notifications/Provider.php
```diff
@@ -25,20 +25,8 @@ class Provider extends Service_Provider {
...
 
-		return $this->container->get( Notifications::class )->render_icon();
+		$this->container->register_on_action( 'tribe_plugins_loaded', Notifications::class );
```
dpanta94, line 30: `register_on_action` calls `register` immediatelly if the action (in this case `tribe_plugins_loaded` ) has already happened or is happening. In this case, i think register is always going to be called immediately.

`register_on_action` should be used by registering on actions that dont happen always. For example,


in ET i have a provider which registers Coupon functionality.
In ETP i have a provider that enriches COupon functionality.
My ETP provider register with register_on_action on the ET's provider registration action.

dpanta94 (changes requested):

src/Events/Notifications/Provider.php
Camwyn: I think of it as a "load right away, unless we haven't done X yet - then wait for that" (or don't load if it doesn't happen, of course)

`tribe_plugins_loaded` _should_ always have happened at this point - so perhaps the hook is a redundant one for this use.

ALSO: `register_on_action` is for _providers_ did we convert `Notifications` to a provider (via Integrations) just to use the on-action method?

Camwyn: I think this is not where we needed to alter things - it should be where we register/bind the Notifications class to ensure it happens before `admin_footer`

pattihis: > did we convert Notifications to a provider (via Integrations) just to use the on-action method?

yes but it also gives us the [load_conditionals()](https://github.com/the-events-calendar/the-events-calendar/pull/4877/files#diff-def87422643db7c06d83ff1115ce16c359b9d15d3f9155527c7fcc7d1650be6aR38) method where we do the  `has_action( 'tribe_common_loaded' )` and replace the previous `class_exist()` check, that was checking for the TCMN. The required `get_slug` method also came in handy.

> register_on_action should be used by registering on actions that dont happen always.

Yes, in theory. Here we are dependent on Common, that it **_should_** always be present. But in a rare scenario, while downgrading to an older version, we had a fatal, hence this and the previous fix.

> tribe_plugins_loaded should always have happened at this point - so perhaps the hook is a redundant one for this use.

then should we change to `tribe_common_loaded` or add a custom action for when IAN is loaded in TCMN?

> I think this is not where we needed to alter things - it should be where we register/bind the Notifications class

there's only two files in TEC, about IAN:
`src/Events/Notifications/Provider.php` and
`src/Events/Notifications/Notifications.php`
Former loads the Latter. Or are you talking about [Main](https://github.com/the-events-calendar/the-events-calendar/blob/7021484082ef021329c2c28c368040bc6405e2f1/src/Tribe/Main.php#L698) ?

dpanta94: > But in a rare scenario, while downgrading to an older version, we had a fatal, hence this and the previous fix.

Then that seems like another issue which should be caught by our Plugin_Register class. I think thats where we should focus on fixing that. If common does not exist (for some reason) which we could check with has_action, no plugin should be loaded.

dpanta94: Because of [this](https://github.com/the-events-calendar/the-events-calendar/blob/7021484082ef021329c2c28c368040bc6405e2f1/src/Tribe/Main.php#L699)

`$this->container->singleton( static::class, $this );`

is not needed.

->register method creates singletons.

dpanta94 (approved):

PR #4878, opened 2025-01-13
[EA-476] Event Aggregator: Replace  deprecated `get_page_by_title` function

codingmusician:
### 🎫 Ticket
[EA-476]

### 🗒️ Description
The WP function `get_page_by_title` has been deprecated as of version 6.2. This PR replaces the function with an equivalent `WP_Query` call.

### 🎥 Artifacts 
https://share.zight.com/7KuwbZgE

### ✔️ Checklist

[EA-476]: https://stellarwp.atlassian.net/browse/EA-476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Aggregator/Record/Abstract.php
+++ b/src/Tribe/Aggregator/Record/Abstract.php
@@ -1812,15 +1812,15 @@ abstract class Tribe__Events__Aggregator__Record__Abstract { //phpcs:ignore TEC.
 							if ( empty( $venue_unique_field ) || ( $lookup_venues_by_title && empty( $venue ) ) ) {
 								$venue_query = new WP_Query(
 									[
-										'post_type'              => Tribe__Events__Venue::POSTTYPE,
-										'title'                  => $event['Venue']['Venue'],
-										'post_status'            => 'all',
-										'posts_per_page'         => 1,
-										'no_found_rows'          => true,
-										'ignore_sticky_posts'    => true,
+										'post_type'      => Tribe__Events__Venue::POSTTYPE,
+										'title'          => $event['Venue']['Venue'],
+										'post_status'    => 'all',
+										'posts_per_page' => 1,
+										'no_found_rows'  => true,
+										'ignore_sticky_posts' => true,
 										'update_post_term_cache' => false,
-										'orderby'                => 'post_date ID',
-										'order'                  => 'ASC',
+										'orderby'        => 'post_date ID',
+										'order'          => 'ASC',
 
 									]
 								);
@@ -1996,15 +1996,15 @@ abstract class Tribe__Events__Aggregator__Record__Abstract { //phpcs:ignore TEC.
 									} else {
 										$organizer_query = new WP_Query(
 											[
-												'post_type'              => Tribe__Events__Organizer::POSTTYPE,
-												'title'                  => $organizer_data['Organizer'],
-												'post_status'            => 'all',
-												'posts_per_page'         => 1,
-												'no_found_rows'          => true,
-												'ignore_sticky_posts'    => true,
+												'post_type'      => Tribe__Events__Organizer::POSTTYPE,
+												'title'          => $organizer_data['Organizer'],
+												'post_status'    => 'all',
+												'posts_per_page' => 1,
+												'no_found_rows'  => true,
+												'ignore_sticky_posts' => true,
 												'update_post_term_cache' => false,
-												'orderby'                => 'post_date ID',
-												'order'                  => 'ASC',
+												'orderby'        => 'post_date ID',
+												'order'          => 'ASC',
 
 											]
 										);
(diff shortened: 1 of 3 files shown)


src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1810,7 +1810,21 @@ public function insert_posts( $items = [] ) {
...
+								$venue_query = new WP_Query(
+									[
+										'post_type'      => Tribe__Events__Venue::POSTTYPE,
+										'title'          => $event['Venue']['Venue'],
+										'post_status'    => 'all',
+										'posts_per_page' => 1,
+										'no_found_rows'  => true,
+										'ignore_sticky_posts' => true,
+										'update_post_term_cache' => false,
+										'orderby'        => 'post_date ID',
```
bordoni, line 1822: Multiple `orderby` should be an array.

bordoni, line 1822: Align the `=>`

bordoni (dismissed):

src/Tribe/Aggregator/Record/Abstract.php
codingmusician: About that... I had them all aligned and `phpcs` threw a fit! This is the spacing it accepted, oddly enough. Have we been having issues with that?

codingmusician: See here: https://github.com/the-events-calendar/the-events-calendar/actions/runs/12755046123/job/35550252359

codingmusician: According to WP, when the sorting order is the same, a string can be used: https://developer.wordpress.org/reference/classes/wp_query/#order-orderby-parameters

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1810,7 +1810,21 @@ public function insert_posts( $items = [] ) {
 							}
 
 							if ( empty( $venue_unique_field ) || ( $lookup_venues_by_title && empty( $venue ) ) ) {
-								$venue = get_page_by_title( $event['Venue']['Venue'], 'OBJECT', Tribe__Events__Venue::POSTTYPE );
+								$venue_query = new WP_Query(
+									[
+										'post_type'      => Tribe__Events__Venue::POSTTYPE,
+										'title'          => $event['Venue']['Venue'],
+										'post_status'    => 'all',
```
dpanta94, line 1817: Is it `all` or `any` ? or both ?

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1810,7 +1810,21 @@ public function insert_posts( $items = [] ) {
...
+										'posts_per_page' => 1,
+										'no_found_rows'  => true,
+										'ignore_sticky_posts' => true,
+										'update_post_term_cache' => false,
```
dpanta94, line 1821: Can we also ignore `update_post_meta_cache` ?

src/Tribe/Aggregator/Record/Abstract.php
```diff
@@ -1810,7 +1810,21 @@ public function insert_posts( $items = [] ) {
...
+										'orderby'        => 'post_date ID',
+										'order'          => 'ASC',
+
+									]
+								);
+								$venue = ! empty( $venue_query->post ) ? $venue_query->post : null;
```
dpanta94, line 1827: Isn't the property we want `posts` ? `post` should be set within the loop and we should not depend being in the loop.

If we are actually ending in the loop because of the updates, this is an issue. I would add a test ensuring that we are not ending up `in_the_loop()`

dpanta94 (changes requested):
Could we have this moved as a methon in TCMN so we can reuse it accross our plugins ?

I see ECP and ETP will also need replaces for example.

src/Tribe/Aggregator/Record/Abstract.php
codingmusician: Looking at the source, when `posts` are found, then the first `post` is set at the start. Otherwise, it remains unset (`null`). We're only looking for one post in this case.

src/Tribe/Aggregator/Record/Abstract.php
codingmusician: Good call. I'll make this change first thing tomorrow.

src/Tribe/Aggregator/Record/Abstract.php
codingmusician: Ah! You're right. It's `any`, not `all`. Great catch! I'll fix it first thing in the AM.

codingmusician:
> Could we have this moved as a methon in TCMN so we can reuse it accross our plugins ?
> 
> I see ECP and ETP will also need replaces for example.

That's a good question. This ticket only referenced one file, but maybe we could do a clean sweep of all the plugins? What are your thoughts on this @bordoni and @Camwyn ?

bordoni:
I am not sure about the common method, I like how explicitly we are doing a query here, but I dont feel strongly about it. What I would like to see here is test.

src/Tribe/Aggregator/Record/Abstract.php
dpanta94: Seems like you are right !

dpanta94 (approved):

rafsuntaskin (approved):

PR #4914, opened 2025-02-03
Fix wrong text domain

dpanta94:
Fixing a couple of wrong text-domain calls.

Adding phpcs rule for catching those in the future during CRs.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -22,6 +22,7 @@
 		</properties>
 	</rule>
 
+	<exclude-pattern>*/common/*</exclude-pattern>
 	<exclude-pattern>*/tests/*</exclude-pattern>
 	<exclude-pattern>*/vendor/*</exclude-pattern>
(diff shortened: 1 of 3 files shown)


phpcs.xml
```diff
@@ -15,6 +15,13 @@
...
 	<rule ref="Generic.Arrays.DisallowLongArraySyntax"/>
 
+	<!-- Enforce our specific plugin's text domain. -->
+	<rule ref="WordPress.WP.I18n">
+		<properties>
+			<property name="text_domain" type="array" value="the-events-calendar" />
+		</properties>
+	</rule>
+
 	<exclude-pattern>*/tests/*</exclude-pattern>
```
Camwyn, line 25:
```suggestion
	<exclude-pattern>*/common/*</exclude-pattern>
	<exclude-pattern>*/tests/*</exclude-pattern>
```

Camwyn (approved):

PR #4921, opened 2025-02-05
[TEC-5376] Created Meta class for Category Colors.

redscar:
### 🎫 Ticket
[TEc-5376]

### 🗒️ Description

- Implemented the Event_Category_Meta class to manage term metadata for the tribe_events_cat taxonomy, allowing for structured and reliable meta handling.
- Introduced chained operations for set(), delete(), and save(), enabling batch updates and deletions before persisting changes.
- Added validation for meta keys and values, ensuring only valid metadata is stored while filtering out invalid inputs.
- Implemented error handling, including exceptions for invalid term IDs and returning WP_Error for invalid metadata operations.
- I also brought back the Controller to a more generic state since this will be the first PR to be merged.

*Changes in Future PR's:*
In my next PR I plan on moving the tests from `integration` to `integration_category_colors` https://github.com/the-events-calendar/the-events-calendar/pull/4940

### ✔️ Checklist


redscar (commented):
We may want to introduce a save method so that instead of setting, saving, setting ,saving we do it all at once.

redscar (commented):
Maybe add an is_valid() that should check the meta, if it's true let everything work, if it's false, then everything else does nothing.

src/Events/Category_Colors/Event_Category_Meta.php
```diff
@@ -75,207 +92,143 @@ public function __construct( int $term_id ) {
...
-		 * @param int    $term_id The term ID where metadata was set.
-		 * @param string $key     The meta key that was updated.
-		 * @param mixed  $value   The new value that was stored.
-		 */
-		do_action( 'tec_events_category_set_meta', $this->term_id, $key, $value );
+		$key   = $this->validate_key( $key );
+		$value = get_term_meta( $this->term_id, $key, true );
 
-		return $this;
+		return ( '' !== $value ) ? $value : null;
```
codingmusician, line 180: I think this would work better (in case it returns false)
```suggestion
		empty( $value ) ? null : $value;
```

src/Events/Category_Colors/Event_Category_Meta.php
```diff
@@ -75,207 +92,143 @@ public function __construct( int $term_id ) {
...
 		/**
-		 * Fires after one or more metadata keys have been deleted.
+		 * Filter the validated meta key before it is used.
 		 *
 		 * @since TBD
 		 *
-		 * @param int      $term_id The term ID where metadata was deleted.
-		 * @param string[] $keys    The meta keys that were deleted.
+		 * @param string $key     The sanitized meta key.
+		 * @param int    $term_id The term ID the meta key belongs to.
```
codingmusician, line 205:
```suggestion
		 * @param int    $term_id The term ID the meta key belongs to.
		 *
		 * @return string The filtered meta key.
```

src/Events/Category_Colors/Event_Category_Meta.php
```diff
@@ -75,207 +92,143 @@ public function __construct( int $term_id ) {
...
-		 * Filter a specific meta key value before returning.
+		 * Filter the meta value before it is saved.
 		 *
 		 * @since TBD
 		 *
-		 * @param mixed  $value   The retrieved metadata value.
-		 * @param string $key     The meta key being retrieved.
-		 * @param int    $term_id The term ID.
+		 * @param mixed $value   The sanitized meta value.
+		 * @param int   $term_id The term ID the meta value belongs to.
```
codingmusician, line 230:
```suggestion
		 * @param int   $term_id The term ID the meta value belongs to.
		 *
		 * @return mixed The filtered meta value.
```

codingmusician (commented):
A few (mostly documentation) suggestions. Also... should we add tests for the new filters?

src/Events/Category_Colors/Event_Category_Meta.php
redscar: I'm not sure if `empty()` is the best way to go. Empty will return true if the value is `0` as well.

redscar: I ended up creating a new test `it_should_properly_handle_empty_values_when_retrieving_meta` to test different scenarios. `empty()` didn't work. However, I ended up refactoring the logic to better handle the different scenarios.

src/Events/Category_Colors/Controller.php
```diff
@@ -1,10 +1,25 @@
...
 namespace TEC\Events\Category_Colors;
 
 use TEC\Common\Contracts\Provider\Controller as Controller_Contract;
-use TEC\Events\Category_Colors\Settings as Category_Colors_Settings;
 
+
+/**
+ * Class Controller
+ *
+ * @since TBD
```
codingmusician, line 21:
```suggestion
 * @since TBD
 *
 * @package TEC\Events\Category_Colors
```

src/Events/Category_Colors/Event_Category_Meta.php
```diff
@@ -0,0 +1,268 @@
...
+	 * @param int $term_id The term ID.
+	 */
+	public function __construct( int $term_id ) {
+		if ( $term_id <= 0 ) {
+			throw new InvalidArgumentException( __( 'Invalid term ID.', 'the-events-calendar' ) );
+		}
+
+		$term = get_term( $term_id, $this->taxonomy );
+
+		if ( ! $term instanceof WP_Term || is_wp_error( $term ) ) {
```
codingmusician, line 93: I think we only need to check this:
```suggestion
		if ( ! $term instanceof WP_Term ) {
```

src/Events/Category_Colors/Event_Category_Meta.php
```diff
@@ -0,0 +1,268 @@
...
+			return $value;
+		}
+
+		// Ensure explicitly stored `false` returns an empty string.
+		if ( false === $value ) {
+			return '';
+		}
+
+		// Ensure empty strings remain empty.
+		return ( '' === $value ) ? '' : $value;
```
codingmusician, line 210: I feel like this can be condensed. Casting `false` and `null` to a string actually results in an empty string, so we can use that to our advantage:
```suggestion
		// Ensure arrays remain unchanged.
		if ( is_array( $value ) ) {
			return $value;
		}

		// Ensure we return a string.
		return (string) $value;
```

src/Events/Category_Colors/Controller.php
```diff
@@ -13,9 +29,7 @@ class Controller extends Controller_Contract {
 	 * @since TBD
 	 */
 	public function do_register(): void {
-		$this->container->singleton( Category_Colors::class );
-		$this->container->singleton( Category_Colors_Settings::class );
-		$this->container->singleton( Quick_Edit::class );
+		$this->container->bind( Event_Category_Meta::class );
```
dpanta94, line 32: This isn't a correct binding - the container wont be able to decide how to build a class which requires an int in its constructor.

You have no way of binding an implementation that requires a int which could change every time in your container.

src/Events/Category_Colors/Controller.php
```diff
@@ -34,102 +48,12 @@ public function unregister(): void {
...
-		add_filter( 'manage_edit-tribe_events_cat_columns', [ $this, 'add_custom_taxonomy_columns' ] );
-
-		add_action( 'manage_tribe_events_cat_custom_column', [ $this, 'populate_custom_taxonomy_color_column' ], 10, 3 );
-		add_action( 'manage_tribe_events_cat_custom_column', [ $this, 'populate_custom_taxonomy_hidden_columns' ], 10, 3 );
-
-		add_action( 'quick_edit_custom_box', [ $this, 'add_custom_quick_edit_field' ], 10, 3 );
-
-		add_action( 'edited_term_taxonomy', [ $this, 'save_quick_edit_custom_fields' ], 10, 2 );
-	}
+	protected function add_filters() {}
```
dpanta94, line 51: remove if we dont need

src/Events/Category_Colors/Controller.php
```diff
@@ -13,9 +29,7 @@ class Controller extends Controller_Contract {
 	 * @since TBD
 	 */
 	public function do_register(): void {
-		$this->container->singleton( Category_Colors::class );
-		$this->container->singleton( Category_Colors_Settings::class );
-		$this->container->singleton( Quick_Edit::class );
```
dpanta94, line 18: What happened to the singletons that were here ?

src/Events/Category_Colors/Event_Category_Meta.php
```diff
@@ -0,0 +1,258 @@
...
+	 * @return $this Returns the instance for chaining.
+	 */
+	public function set( string $key, $value ): self {
+		$key   = $this->validate_key( $key );
+		$value = $this->validate_value( $value );
+
+		// If key or value is invalid, skip but don't break chaining.
+		if ( is_wp_error( $key ) || is_wp_error( $value ) ) {
+			return $this;
+		}
```
dpanta94, line 120: I think is better when fluent apis throw exception on errors.

Somehow you gotta know that something failed.

src/Events/Category_Colors/Event_Category_Meta.php
```diff
@@ -0,0 +1,258 @@
...
+	 * @param string $key The meta key to delete.
+	 *
+	 * @return $this|wp_error  Returns the instance for chaining.
+	 */
+	public function delete( string $key ): self {
+		$key = $this->validate_key( $key );
+
+		if ( is_wp_error( $key ) ) {
+			return $key;
+		}
```
dpanta94, line 141: Here we should not change the return type - which is type hinted to self. Throw exception instead.

dpanta94 (changes requested):

dpanta94 (approved):

PR #4930, opened 2025-02-11
[TEC-5386] CPT and Menu Setup

codingmusician:
### 🎫 Ticket
[TEC-5386]

### 🗒️ Description
Reistering the CPT for Calendar Embeds and adding the "Embed Calendar" submenu.

### 🎥 Artifacts 
image

### ✔️ Checklist

### DIFF
--- a/src/Events/Calendar_Embeds/Calendar_Embeds.php
+++ b/src/Events/Calendar_Embeds/Calendar_Embeds.php
@@ -1,0 +1,147 @@
+<?php
+
+/**
+ * Manages the External Calendar Embeds Feature.
+ *
+ * @since TBD
+ *
+ * @package TEC\Events\Calendar_Embeds
+ */
+
+namespace TEC\Events\Calendar_Embeds;
+
+use Tribe__Events__Main as TEC;
+
+/**
+ * Class Calendar_Embeds
+ *
+ * @since TBD
+ *
+ * @package TEC\Events\Calendar_Embeds
+ */
+class Calendar_Embeds {
+
+	/**
+	 * Calendar Embeds post type slug.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	const POSTTYPE = 'tec_calendar_embed';
+
+	/**
+	 * The page slug
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $slug = 'calendar-embeds';
+
+	/**
+	 * Stores the Registered ID from `add_submenu_page`
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public $ID;
+
+	/**
+	 * Register custom post type for calendar embeds.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register_post_type() {
+		$labels = [
+			'name'               => _x( 'Calendar Embeds', 'post type general name', 'the-events-calendar' ),
+			'singular_name'      => _x( 'Calendar Embed', 'post type singular name', 'the-events-calendar' ),
+			'menu_name'          => _x( 'Calendar Embeds', 'admin menu', 'the-events-calendar' ),
+			'name_admin_bar'     => _x( 'Calendar Embed', 'add new on admin bar', 'the-events-calendar' ),
+			'add_new'            => _x( 'Add New', 'calendar embed', 'the-events-calendar' ),
+			'add_new_item'       => __( 'Add New Calendar Embed', 'the-events-calendar' ),
+			'new_item'           => __( 'New Calendar Embed', 'the-events-calendar' ),
+			'edit_item'          => __( 'Edit Calendar Embed', 'the-events-calendar' ),
+			'view_item'          => __( 'View Calendar Embed', 'the-events-calendar' ),
+			'all_items'          => __( 'Calendar Embeds', 'the-events-calendar' ),
+			'search_items'       => __( 'Search Calendar Embeds', 'the-events-calendar' ),
+			'parent_item_colon'  => __( 'Parent Calendar Embeds:', 'the-events-calendar' ),
+			'not_found'          => __( 'No calendar embeds found.', 'the-events-calendar' ),
+			'not_found_in_trash' => __( 'No calendar embeds found in Trash.', 'the-events-calendar' ),
+		];
+
+		$args = [
+			'labels'             => $labels,
+			'public'             => false,
+			'publicly_queryable' => false,
+			'show_ui'            => true,
+			'show_in_menu'       => false,
+			'show_in_nav_menus'  => true,
+			'query_var'          => true,
+			'rewrite'            => [ 'slug' => 'calendar-embed' ],
+			'capability_type'    => 'post',
+			'has_archive'        => false,
+			'hierarchical'       => false,
+			'supports'           => [ 'title' ],
+			'show_in_rest'       => true,
+		];
+
+		/**
+		 * Filter the arguments for the Calendar Embeds post type.
+		 *
+		 * @since TBD
+		 *
+		 * @param array $args The arguments for the Calendar Embeds post type.
+		 *
+		 * @return array
+		 */
+		$args = apply_filters( 'tec_events_calendar_embeds_post_type_args', $args );
+
+		register_post_type( static::POSTTYPE, $args );
+	}
+
+	/**
+	 * Register the Calendar Embeds menu item.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register_menu_item() {
+		$cpt      = get_post_type_object( TEC::POSTTYPE );
+		$this->ID = add_submenu_page(
+			'edit.php?post_type=' . TEC::POSTTYPE,
+			esc_html( $this->get_page_title() ),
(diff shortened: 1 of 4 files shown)


src/resources/js/events-admin.js
```diff
@@ -778,6 +778,14 @@ jQuery( function( $ ) {
 			.parent()
 			.addClass( 'current' );
 	}
+	if ( $( 'body' ).hasClass( 'post-type-tec_calendar_embed') ) {
```
dpanta94, line 781: Make known to WP the parent file like [this](https://github.com/the-events-calendar/event-tickets/blob/15f56a9a3d97f5cdd8f89835d692b2533981d887/src/Tickets/Commerce/Admin/Singular_Order_Page.php#L58).

Dont try to hack your way into this through JS - can cause other issues.

src/Tribe/Aggregator/Page.php
```diff
@@ -420,6 +420,13 @@ public function register_menu_item() {
 			[ $this, 'render' ]
 		);
 
+		/**
+		 * Fires an Action after the submenu item is created.
+		 *
+		 * @since TBD
+		 */
+		do_action( 'tec_events_after_import_submenu_created' );
```
dpanta94, line 428: Instead of doing that cant we create the page using our admin pages and adjust its placement with the priority param ?

tests/wpunit/TEC/Events/Calendar_Embeds/Calendar_EmbedsTest.php
```diff
@@ -0,0 +1,75 @@
+<?php
```
dpanta94, line 1: Lets create our own test suite instead. I can take care of that. For now leave it in wpunit and ill move it.

tests/wpunit/TEC/Events/Calendar_Embeds/Calendar_EmbedsTest.php
```diff
@@ -0,0 +1,75 @@
...
+	protected function setUp(): void {
+		parent::setUp();
+		$this->calendarEmbeds = new Calendar_Embeds();
+	}
+
+	protected function tearDown() : void {
+		parent::tearDown();
+		$this->calendarEmbeds = null;
+		$this->unset_uopz_returns();
+	}
```
dpanta94, line 26: Both of these methods are not needed.

Also you are binding the Calendar_Embeds to the container as a singleton and then you go ahead and create a new instance, invaliding test cases since they no longer mimic production code.

src/Events/Calendar_Embeds/Controller.php
```diff
@@ -0,0 +1,80 @@
...
+
+		$this->add_actions();
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function unregister(): void {
+		$this->remove_actions();
+	}
```
dpanta94, line 37: I am against the extra code just for the sake of it. Your current code has an unregister which calls the real unregister... Why not just have the add/remove filters in do_register and unregister and avoid having real and fake methods ?

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -0,0 +1,146 @@
...
+	 * Calendar Embeds post type slug.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	const POSTTYPE = 'tec_calendar_embed';
+
+	/**
+	 * The page slug
```
sdokus, line 33:
```suggestion
	 * The page slug.
```

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -0,0 +1,146 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $slug = 'calendar-embeds';
+
+	/**
+	 * Stores the Registered ID from `add_submenu_page`
```
sdokus, line 42:
```suggestion
	 * Stores the Registered ID from `add_submenu_page`.
```

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -0,0 +1,146 @@
...
+			esc_html( $this->get_menu_label() ),
+			$cpt->cap->publish_posts,
+			'edit.php?post_type=' . self::POSTTYPE,
+		);
+
+		return $this->ID;
+	}
+
+	/**
+	 * Gets the Menu label for the Aggregator
```
sdokus, line 126:
```suggestion
	 * Gets the Menu label for the Aggregator.
```

src/Events/Calendar_Embeds/Controller.php
```diff
@@ -0,0 +1,80 @@
...
+	/**
+	 * Add actions for the feature.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function add_actions() {
+		add_action( 'init', [ $this, 'register_post_type' ] );
+		add_action( 'tec_events_after_import_submenu_created', [ $this, 'register_menu_item' ] );
```
dpanta94, line 48: Since this is the main controller dont hook specific actions here which in their callback they just make an instance and they are handled elsewhere. It just adds extra code without a point.

Instead: here register the Calendar_Embeds class (which i would rename to Post_type or something) which should also be a controller and there hook those and handle those.

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -0,0 +1,146 @@
...
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_menu_label() {
+		return __( 'Embed Calendar', 'the-events-calendar' );
+	}
+
+	/**
+	 * Gets the Page title for the Aggregator
```
sdokus, line 137:
```suggestion
	 * Gets the Page title for the Aggregator.
```

dpanta94 (dismissed):

tests/wpunit/TEC/Events/Calendar_Embeds/Calendar_EmbedsTest.php
dpanta94: Expanding because i didnt really explain my point.

When we need to grab the instance of CalendarEmbed we should do it though our container - since this is going to become a controller test case soon we will have test_services available!

the unset_uopz_returns should be called regardless after every test case.

src/Tribe/Aggregator/Page.php
codingmusician: I don't know why I didn't try using the action priority. I was experimenting with the `position` property, which didn't seem to do anything. Using priority 11 worked!

src/Events/Calendar_Embeds/Controller.php
codingmusician: Hmm. That seems to go against the way we use controllers elsewhere. Maybe we can discuss with @bordoni about that. I understand your reasoning, though.

src/Events/Calendar_Embeds/Controller.php
codingmusician: Same here. This is how we use controllers elsewhere, so let's discuss with @bordoni to make sure things are uniform.

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -0,0 +1,168 @@
...
+		 *
+		 * @since TBD
+		 *
+		 * @param array $args The arguments for the Calendar Embeds post type.
+		 *
+		 * @return array
+		 */
+		$args = apply_filters( 'tec_events_calendar_embeds_post_type_args', $args );
+
+		register_post_type( static::POSTTYPE, $args );
```
bordoni, line 102: Registering a Post Type has a return, store that for future reference in a variable; `protected $this->post_type_object;` with a getter; Throw error if the getter is called before the registering.

https://developer.wordpress.org/reference/functions/register_post_type/#return

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -0,0 +1,168 @@
...
+	/**
+	 * Register the Calendar Embeds menu item.
+	 *
+	 * @since TBD
+	 *
+	 * @return int
+	 */
+	public function register_menu_item() {
+		$cpt      = get_post_type_object( TEC::POSTTYPE );
+		$this->ID = add_submenu_page(
```
bordoni, line 114: Dont use ID, what is returned is the Hook Suffix, make this prop protected, and with a public getter. Throw error if the getter is called before the registering.

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -0,0 +1,168 @@
...
+	const POSTTYPE = 'tec_calendar_embed';
+
+	/**
+	 * The page slug.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $slug = 'calendar-embeds';
```
bordoni, line 39: I would avoid this being static or public, expose a `get_url()` method that has access to a protected prop.

bordoni (changes requested):

bordoni (approved):

PR #4941, opened 2025-02-14
Fix linting errors for block modules

codingmusician:
### 🎫 Ticket
N/A

### 🗒️ Description
On a mission to fix all linting tests for the TEC plugin. I'll be doing these in bite-sized chunks so they're not too overwhelming to review.

This PR covers the block modules only.

It also disables linting for tests.


.eslintignore
```diff
@@ -1 +1,2 @@
 src/resources/js/**/*.min.js
+src/modules/**/__tests__/*
```
dpanta94, line 2:
```suggestion
src/modules/**/__tests__/*
tests/**/*.js
```

JPry (approved):
❤️

dpanta94 (approved):

.eslintignore
Camwyn: Honestly, the first (existing) line could likely be `/**/*.min.js` because we don't ever want to lint minimized files, wherever they are.

dpanta94 (approved):

PR #4946, opened 2025-02-17
[TEC-5388] Calendar Embeds: Admin List Table Customizations

codingmusician:
### 🎫 Ticket
[TEC-5388]

### 🗒️ Description
Admin Table List customizations. 

### 🎥 Artifacts
Screen Recording

### ✔️ Checklist

[TEC-5388]: https://stellarwp.atlassian.net/browse/TEC-5388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Events/Calendar_Embeds/Admin/List_Table.php
+++ b/src/Events/Calendar_Embeds/Admin/List_Table.php
@@ -28,7 +28,7 @@ class List_Table {
 	 */
 	public function manage_columns( $columns ): array {
 		$new_columns = [
-			'cb'               => isset( $columns['cb'] ) ? $columns['cb'] : '<input type="checkbox" />',
+			'cb'               => $columns['cb'] ?? '<input type="checkbox" />',
 			'title'            => __( 'Calendar Embeds', 'the-events-calendar' ),
 			'event_categories' => __( 'Categories', 'the-events-calendar' ),
 			'post_tags'        => __( 'Tags', 'the-events-calendar' ),
@@ -49,7 +49,6 @@ class List_Table {
 	 * @return void
 	 */
 	public function manage_column_content( $column_name, $post_id ): void {
-		error_log('testing');
 		switch ( $column_name ) {
 			case 'event_categories':
 				$categories = get_the_terms( $post_id, 'category' );
@@ -71,17 +70,12 @@ class List_Table {
 				break;
 			case 'snippet':
 				add_thickbox();
-				// get permalink
+
 				$permalink = get_permalink( $post_id );
 
-				// create snippet to show iframe
 				$snippet = '<iframe src="' . esc_url( $permalink ) . '" width="800" height="600"></iframe>';
-
-				// create html for modal to display snippet
-				$html = '<div id="snippet_' . $post_id . '" class="hidden"><p>Copy and paste this code to embed the calendar on your website:<br><textarea rows="3" style="width:100%" readonly="readonly">' . esc_html( $snippet ) . '</textarea></p></div>';
-
-				// create thickbox link to show modal
-				$html .= '<a name="Embed Snippet" href="/?TB_inline?width=400&height=300&inlineId=snippet_' . $post_id . '" class="thickbox page-title-action">Get Embed Snippet</a>';
+				$html    = '<div id="snippet_' . $post_id . '" class="hidden"><p>Copy and paste this code to embed the calendar on your website:<br><textarea rows="3" style="width:100%" readonly="readonly">' . esc_html( $snippet ) . '</textarea></p></div>';
+				$html   .= '<a name="Embed Snippet" href="/?TB_inline?width=400&height=300&inlineId=snippet_' . $post_id . '" class="thickbox page-title-action">Get Embed Snippet</a>';
 
 				echo wp_kses_post( $html );
 				break;
--- a/src/Events/Calendar_Embeds/Admin/List_Table.php
+++ b/src/Events/Calendar_Embeds/Admin/List_Table.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * Calendar Embedw List Table
+ *
+ * @since TBD
+ *
+ * @package TEC\Events\Calendar_Embeds\Admin
+ */
+
+namespace TEC\Events\Calendar_Embeds\Admin;
+
+/**
+ * Class List_Table
+ *
+ * @since TBD
+ *
+ * @package TEC\Events\Calendar_Embeds\Admin
+ */
+class List_Table {
+	/**
+	 * Customize columns for the table.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $columns The columns.
+	 *
+	 * @return array
+	 */
+	public function manage_columns( $columns ): array {
+		$new_columns = [
+			'cb'               => isset( $columns['cb'] ) ? $columns['cb'] : '<input type="checkbox" />',
+			'title'            => __( 'Calendar Embeds', 'the-events-calendar' ),
+			'event_categories' => __( 'Categories', 'the-events-calendar' ),
+			'post_tags'        => __( 'Tags', 'the-events-calendar' ),
+			'snippet'          => __( 'Embed Snippet', 'the-events-calendar' ),
+		];
+
+		return $new_columns;
+	}
+
+	/**
+	 * Customize the content of the columns.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $column_name The name of the column.
+	 * @param int    $post_id     The post ID.
+	 *
+	 * @return void
+	 */
+	public function manage_column_content( $column_name, $post_id ): void {
+		error_log('testing');
+		switch ( $column_name ) {
+			case 'event_categories':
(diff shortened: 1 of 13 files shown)


src/Events/Calendar_Embeds/Admin/Page.php
```diff
@@ -0,0 +1,169 @@
...
+
+		Asset::add(
+			'tec-events-calendar-embeds-style',
+			'css/admin/calendar-embeds-page.css'
+		)
+			->enqueue_on( 'admin_enqueue_scripts' )
+			->set_condition( [ __CLASS__, 'is_on_page' ] )
+			->set_dependencies( 'thickbox', 'tribe-common-admin' )
+			->register();
+	}
```
dpanta94, line 168: Set their enqueue_on on the action is being fired by the page's render. So you dont have to worry about the conditional enqueue

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -29,6 +27,24 @@ class Calendar_Embeds {
...
+	public static $meta_key_categories = 'event_categories';
+
+	/**
+	 * The meta key for storing the event tags.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $meta_key_tags = 'event_tags';
```
dpanta94, line 46: If they dont need to change during runtime lets gain the extra speed of a const.

src/Events/Calendar_Embeds/Controller.php
```diff
@@ -116,6 +139,44 @@ public function register_menu_item() {
...
+
+	/**
+	 * Enqueue assets for the admin page.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function enqueue_admin_page_assets(): void {
+		$this->container->make( Admin\Page::class )->register_assets();
```
dpanta94, line 180: Why are we still going with the fake actions that redirect to the real actions ? Having an abstraction layer that only offers more code to maintain... Not optimal

dpanta94 (changes requested):

src/Events/Calendar_Embeds/Admin/Page.php
codingmusician: In this case, we're using the built-in post type page, not rendering it ourselves. That's why I did it this way.

src/Events/Calendar_Embeds/Controller.php
dpanta94: All the hooks are in one place is against my preference. If that is TEC policy though ill oblige :D

I am more of the idea "the hooks are in their appropriate place"

This code is just redirecting to where the hooks should be to begin with.

src/Events/Calendar_Embeds/Admin/Page.php
dpanta94: Thats good! Thank you!

src/Events/Calendar_Embeds/Admin/List_Table.php
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		$this->template = new Tribe__Template();
+		$this->template->set_template_origin( Tribe__Events__Main::instance() );
+		$this->template->set_template_folder( 'src/admin-views/calendar-embeds/' );
+		$this->template->set_template_context_extract( true );
+		$this->template->set_template_folder_lookup( false );
+
+		return $this->template;
+	}
```
dpanta94, line 131: Non blocking comment - as we evolve the functionality we'll probably need to use this same Template in multiple places.


Then we should stop - create a singleton and provide the Template to this class __constructor instead of building anew

codingmusician: I do agree. I don't know if there will be other admin templates, but if there are, we can move it into its own class then.

src/Events/Calendar_Embeds/Controller.php
dpanta94: Of course that isn't strict and code should also include some level of creativity IMO. What i would always vote against is not needed code!

Anyway, not going to force anything to change or continue to be mentioning in every PR.

src/Events/Calendar_Embeds/Admin/List_Table.php
```diff
@@ -0,0 +1,132 @@
...
+	 *
+	 * @param string $column_name The name of the column.
+	 * @param int    $post_id     The post ID.
+	 *
+	 * @return void
+	 */
+	public function manage_column_content( $column_name, $post_id ): void {
+		switch ( $column_name ) {
+			case 'event_categories':
+				// Get events categores from post meta.
```
Camwyn, line 76:
```suggestion
				// Get events categories from post meta.
```
spellcheck on aisle 4 😉

src/Tribe/Assets.php
```diff
@@ -760,4 +761,14 @@ public function customizer_inline_stylesheets( $sheets, $css_template ) {
...
 
+	/**
+	 * Configure the group path for the resources assets folder.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function configure_assets() {
+		TEC\Common\StellarWP\Assets\Config::add_group_path( 'tec-events-resources', Tribe__Events__Main::instance()->plugin_path, 'src/resources/' );
```
dpanta94, line 772: Let's add a use statement in future PR. Approving for now.

dpanta94 (approved):

tests/wpunit/TEC/Events/Calendar_Embeds/Admin/List_TableTest.php
```diff
@@ -0,0 +1,109 @@
...
+		$this->assertEquals( 'All Categories', $category_output );
+
+		// Test before tags are set.
+		ob_start();
+		tribe( List_Table::class )->manage_column_content( 'event_tags', $post_id );
+		$tag_output = ob_get_clean();
+		$this->assertEquals( 'All Tags', $tag_output );
+
+		// Add categories and tags via meta.
+		update_post_meta( $post_id, Calendar_Embeds::META_KEY_CATEGORIES, [
```
Camwyn, line 59: Q: Why not store these as `key => value`? Arrays can handle non-consecutive keys and the ID will always be a unique number.

src/Events/Calendar_Embeds/Admin/Page.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Register the Calendar Embeds menu item.
+	 *
+	 * @since TBD
+	 *
+	 * @return int
+	 */
+	public function register_menu_item() {
+		$cpt               = get_post_type_object( TEC::POSTTYPE );
```
Camwyn, line 41: Since we are registering a CPT in here, can we rename this to `$event_cpt` to avoid potential confusion?

src/Events/Calendar_Embeds/Admin/Page.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Check if the current screen is the Calendar Embeds page.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public static function is_on_page(): bool {
+		$screen = get_current_screen();
```
Camwyn, line 132: FYI: If we need this earlier than `admin_init` (not sure we do here) and you only care about the CPT edit screen you could probably check the URL `post_type` param.

src/admin-views/calendar-embeds/embed-snippet-content.php
```diff
@@ -0,0 +1,44 @@
...
+	<div>
+		<div class="tec-events-calendar-embeds__snippet-modal-text">
+			<?php esc_html_e( 'Copy and paste this code to embed the calendar on your website:', 'the-events-calendar' ); ?>
+		</div>
+		<textarea
+			id="tec_events_calendar_embeds_snippet_code_<?php echo esc_attr( $post_id ); ?>"
+			class="tec-events-calendar-embeds__snippet-modal-textarea"
+			aria-label="<?php esc_attr_e( 'Embed snippet code', 'the-events-calendar' ); ?>"
+			rows="3"
+			readonly><?php echo esc_html( $snippet ); ?></textarea>
```
Camwyn, line 27: Note: this is a textarea, you can (maybe should?) use [esc_textarea](https://developer.wordpress.org/reference/functions/esc_textarea/)

src/resources/postcss/admin/calendar-embeds-page.pcss
```diff
@@ -0,0 +1,67 @@
...
+			&:before {
+				content: '\f335';
+				font: normal var(--tec-font-size-9)/var(--tec-font-size-9) dashicons;
+			}
+		}
+	}
+
+	#TB_ajaxContent {
+		& > div {
+			padding: 0 9px;
```
Camwyn, line 64: change this 9px to `--tec-spacer-1` (8px)

Camwyn (changes requested):

PR #4955, opened 2025-02-24
Improve `get_users()` Performance and removed deprecated args

dilipbheda:
## Description

The `get_users()` function triggers a deprecated notice because it still uses the `who` argument.

Error: Function WP_User_Query was called with an argument that is deprecated since version 5.9.0! who is deprecated. Use capability instead.

## Proposed Solution

We can replace the `who` argument with the `capability` argument.

Support for `capability` was introduced in [WordPress 5.9.0](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-user-query.php#L144). Since The Events Calendar now requires [WordPress 6.5 or higher](https://github.com/the-events-calendar/the-events-calendar/blob/a32495beb5f267bbcd6f3fa3d5f36992cf833f9c/the-events-calendar.php#L6), it is safe to use this feature.

### Other changes for query performance.

Enhance the `get_users()` function by encouraging or defaulting to the use of the `fields` argument where appropriate. This would allow developers to fetch only the necessary data, significantly improving query performance and reducing memory consumption.

```php
// Fetch only user IDs for faster execution
$user_ids = get_users( array(
    'fields' => 'ID',
) );
```

**Potential Benefits:**
- Reduced memory usage  
- Faster query execution  
- Improved scalability for large websites  

### ✔️ Checklist


dilipbheda:
@bordoni You have entered an incorrect name in the changelog. It should be `boundaries props @dilipbheda` instead of `boundaries props @dilipbhed`.

PR #4977, opened 2025-03-12
[TEC-5401] Add one time redirect to First Time Setup page

sdokus:
### 🎫 Ticket

[TEC-5401]

### 🗒️ Description

This PR aims to redirect more first-time users to the First Time Setup page and the Onboarding Wizard. It is meant to only redirect them from other TEC settings to the First Time Setup page one time and if they are a first time user. 

### 🎥 Artifacts 
Before: 

https://github.com/user-attachments/assets/15076ec6-01ec-43f3-a102-49052afcbd22

After:

https://github.com/user-attachments/assets/c9ca78b6-b711-45c7-837a-e6ed38b91134

### ✔️ Checklist

[TEC-5401]: https://stellarwp.atlassian.net/browse/TEC-5401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Admin/Settings.php
```diff
@@ -246,6 +247,75 @@ public function add_admin_pages() {
...
+	/**
+	 * Redirects users to the First Time Setup page when accessing any TEC settings or management page for the first time.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function redirect_tec_pages_to_first_time_setup(): void {
+		// If there is more than one previous version, don't redirect since they're probably already setup.
+		$tec_versions = (array) tribe_get_option( 'previous_ecp_versions', [] );
```
dpanta94, line 259: why ECP and not TEC ?

src/Tribe/Admin/Settings.php
```diff
@@ -221,6 +219,9 @@ public function add_admin_pages() {
 			]
 		);
 
+		// Redirect all TEC admin pages to First Time Setup.
+		$this->redirect_tec_pages_to_first_time_setup();
```
dpanta94, line 223: I would say this is not the appropriate place for this hook. I would choose the `current_screen` action.

src/Tribe/Admin/Settings.php
```diff
@@ -246,6 +247,75 @@ public function add_admin_pages() {
...
+
+		// Only target The Events Calendar-related admin pages.
+		if ( ! in_array( $post_type, [ 'tribe_events', 'tribe_event_series', 'tribe_venue', 'tribe_organizer' ], true ) ) {
+			return;
+		}
+
+		// Prevent infinite redirect: If already on the First Time Setup page, do nothing.
+		if ( 'first-time-setup' === $page ) {
+			return;
+		}
```
dpanta94, line 277: Use `is_tec_page` instead ? I think its appropriate here.

src/Tribe/Admin/Settings.php
```diff
@@ -246,6 +247,75 @@ public function add_admin_pages() {
...
+
+		// Check if the user has already been redirected to the First Time Setup page.
+		$visited_first_time_setup = get_option( 'tec_onboarding_wizard_visited_first_time_setup', false );
+		if ( $visited_first_time_setup ) {
+			// If they already visited, no redirection is needed.
+			return;
+		}
```
dpanta94, line 284: Prefer tec_option instead

src/Tribe/Admin/Settings.php
```diff
@@ -246,6 +247,75 @@ public function add_admin_pages() {
...
+		// Get the current script name and sanitize it.
+		$current_script = isset( $_SERVER['SCRIPT_NAME'] ) ? esc_url_raw( $_SERVER['SCRIPT_NAME'] ) : '';
+
+		if ( in_array( $current_script, $redirect_scripts, true ) || in_array( $page, $redirect_admin_pages, true ) ) {
+			$setup_url = add_query_arg(
+				[
+					'post_type' => 'tribe_events',
+					'page'      => 'first-time-setup',
+				],
+				admin_url( 'edit.php' )
```
andrasguseo, line 308: I believe this should be 'admin_url( 'admin.php' )` and not `edit.php`.

src/Tribe/Admin/Settings.php
```diff
@@ -246,6 +247,75 @@ public function add_admin_pages() {
...
+		];
+
+		// Define admin pages that use the 'page' parameter.
+		$redirect_admin_pages = [
+			'aggregator',         // Import.
+			'tec-events-settings', // Settings.
+		];
+
+		// Get the current script name and sanitize it.
+		$current_script = isset( $_SERVER['SCRIPT_NAME'] ) ? esc_url_raw( $_SERVER['SCRIPT_NAME'] ) : '';
```
dpanta94, line 300: Use the `get_current_screen` instead.

changelog/fix-TEC-5401-add-redirects-for-all-tec-menu-options
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Added one time redirect to the First Time Setup page. [TEC-5401]
```
andrasguseo, line 4: Based on a convo today, there should be a trailing empty line at the end. [link](https://lw.slack.com/archives/C01H7Q57P1C/p1741793702942659)

src/Tribe/Admin/Settings.php
```diff
@@ -246,6 +247,75 @@ public function add_admin_pages() {
...
+				],
+				admin_url( 'edit.php' )
+			);
+
+			// Stop redirecting once we send the user to the First Time Setup page once.
+			update_option( 'tec_onboarding_wizard_visited_first_time_setup', true );
+
+			wp_safe_redirect( $setup_url );
+			exit;
+		}
```
dpanta94, line 316: Use early return approach e.g. `if ( ! in_array() ) { return }`

dpanta94 (changes requested):

src/Tribe/Admin/Settings.php
```diff
@@ -246,6 +247,75 @@ public function add_admin_pages() {
...
 
+	/**
+	 * Redirects users to the First Time Setup page when accessing any TEC settings or management page for the first time.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function redirect_tec_pages_to_first_time_setup(): void {
+		// If there is more than one previous version, don't redirect since they're probably already setup.
```
andrasguseo, line 258: "setup" is a noun. This probably should say "since they're probably already set up."

src/Tribe/Admin/Settings.php
andrasguseo: This is a historical thing. It's named 'ecp' but it is actually the previous TEC versions.

changelog/fix-TEC-5401-add-redirects-for-all-tec-menu-options
Camwyn: Github hides it here - look for the "no new line" symbol. If you use the changelogger the line is there, but won't show here.

src/Tribe/Admin/Settings.php
Camwyn: What Andras said - also I'd use they have instead of they are for this:

```suggestion
		// If there is more than one previous version, don't redirect since they've probably already set up.
```

src/Tribe/Admin/Settings.php
Camwyn: Yes please

src/Tribe/Admin/Settings.php
Camwyn: Same

Camwyn (dismissed):

src/Tribe/Admin/Settings.php
sdokus: I took that from the check in [another part of the Onboarding WIzard ](https://github.com/the-events-calendar/the-events-calendar/blob/e5a59bd72257184c35ff9996d252564f1b75b62a/src/Events/Admin/Onboarding/Landing_Page.php#L555-L560)

src/Tribe/Admin/Settings.php
```diff
@@ -246,6 +247,75 @@ public function add_admin_pages() {
...
+		$post_type = tribe_get_request_var( 'post_type' );
+		$taxonomy  = tribe_get_request_var( 'taxonomy' );
+
+		// Only target The Events Calendar-related admin pages.
+		if ( ! in_array( $post_type, [ 'tribe_events', 'tribe_event_series', 'tribe_venue', 'tribe_organizer' ], true ) ) {
+			return;
+		}
+
+		// Prevent infinite redirect: If already on the First Time Setup page, do nothing.
+		if ( 'first-time-setup' === $page ) {
```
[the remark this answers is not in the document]
sdokus: I didn't know that `is_tec_page` existed so I was trying to reinvent the wheel 😆

src/Tribe/Admin/Settings.php
sdokus: @dpanta94 I chose here because that's where the Help Hub was redirecting - would you mind walking me through why you'd prefer `current_screen`?

src/Tribe/Admin/Settings.php
sdokus: So I'm looking into this more and it seems like we're only adding the 'page' attribute to First Time Setup, Import, Settings, Help, and troubleshooting. Which means we would be missing a redirect on any of these settings:
CleanShot 2025-03-13 at 16 22 53@2x

src/Tribe/Admin/Settings.php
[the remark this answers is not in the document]
sdokus: Oh wait maybe I missed what you were asking about - should the 'first-time-setup' be a variable instead of hard coded?

src/Tribe/Admin/Settings.php
sdokus: It's in `edit.php`: 
CleanShot 2025-03-13 at 16 54 29

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -88,6 +88,53 @@ public function handle_onboarding_page_dismiss(): void {
...
+	 * Redirects users to the Guided Setup page when accessing any TEC settings or management page for the first time.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function redirect_tec_pages_to_guided_setup(): void {
+		// Get some information about the page that we are trying to access.
+		$page      = tribe_get_request_var( 'page' );
+		$post_type = tribe_get_request_var( 'post_type' );
```
andrasguseo, line 101: I would use the `tec_` prefix where available, so we move away from `tribe_` where we can.

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -88,6 +88,53 @@ public function handle_onboarding_page_dismiss(): void {
...
+		}
+
+		// Do not redirect if they are already on the Guided Setup page.
+		if ( 'first-time-setup' === $page ) {
+			return;
+		}
+
+		// Do not redirect if they have been to the Guided Setup page already.
+		$visited_first_time_setup = tribe_get_option( 'tec_onboarding_wizard_visited_guided_setup', false );
+		if ( $visited_first_time_setup ) {
```
andrasguseo, line 121: A simplification, but might be worse for readability. I'm curious about the opinions of @dpanta94 and @Camwyn though.

src/Tribe/Admin/Settings.php
```diff
@@ -221,6 +222,9 @@ public function add_admin_pages() {
 			]
 		);
 
+		// Redirects users from any TEC setting to the Guided Setup page on first visit.
+		do_action( 'tec_redirect_guided_time_setup' );
```
andrasguseo, line 226: A suggestion for adding 'admin_' to the prefix, which feels more in-line with our naming standards.

src/Events/Admin/Onboarding/Controller.php
andrasguseo, line 101:
```suggestion
		$page      = tec_get_request_var( 'page' );
		$post_type = tec_get_request_var( 'post_type' );
```
I would use the tec_ prefix where available, so we move away from tribe_ where we can.

src/Events/Admin/Onboarding/Controller.php
andrasguseo, line 121:
```suggestion
		if ( tribe_get_option( 'tec_onboarding_wizard_visited_guided_setup', false ) ) {
```
A simplification, but might be worse for readability. I'm curious about the opinions of @dpanta94 and @Camwyn though.

src/Tribe/Admin/Settings.php
andrasguseo, line 226:
```suggestion
		do_action( 'tec_admin_redirect_guided_time_setup' );
```
A suggestion for adding 'admin_' to the prefix, which feels more in-line with our naming standards. @Camwyn?

src/Tribe/Admin/Settings.php
```diff
@@ -202,6 +202,7 @@ public function get_tec_events_menu_slug() {
 	 * Adds the menu and pages for The Events Calendar.
 	 *
 	 * @since 5.15.0
+	 * @since TBD Added 'tec_redirect_guided_time_setup' to redirect users to the Guided Setup.
```
andrasguseo, line 205:
```suggestion
	 * @since TBD Added 'tec_admin_redirect_guided_time_setup' to redirect users to the Guided Setup.
```
See comment for line 226.

andrasguseo (approved):
Left some suggestions.

src/Events/Admin/Onboarding/Controller.php
dpanta94: Also @sdokus a tip i would like to see amended!

Consider which condition bails faster!

e.g.
You are getting the page param from the URL.
You are getting the post_type param from the URL
you are quering the database for old TEC versions.
If there were older versions of TEC you bail.
!! If post_type does not contain tribe you bail !!
!! If page is equal to first-time-setup you bail !!

Is that the correct order of operations if you should the bail faster approach ?

for me it would be:

You are getting the page param from the URL.
If page is equal to first-time-setup you bail

You are getting the post_type param from the URL
If post_type does not contain tribe you bail

you are quering the database for old TEC versions.
If there were older versions of TEC you bail.

src/Events/Admin/Onboarding/Controller.php
dpanta94: I prefer it - my approach is to avoid local variables if they can be avoided unless the readability is getting really poor

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -88,6 +88,53 @@ public function handle_onboarding_page_dismiss(): void {
...
+		$setup_url = add_query_arg(
+			[
+				'post_type' => 'tribe_events',
+				'page'      => 'first-time-setup',
+			],
+			admin_url( 'edit.php' )
+		);
+
+		wp_safe_redirect( $setup_url );
+		exit;
```
dpanta94, line 135:
```suggestion
		tribe_exit();
```

We dont use exit or die in our code so that we can test the methods safer.

src/Tribe/Admin/Settings.php
dpanta94: Also needs full docblock.

Actions and filters are documented similar to a function/method would.

/**
 *
 */

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -88,6 +88,53 @@ public function handle_onboarding_page_dismiss(): void {
...
+		$visited_first_time_setup = tribe_get_option( 'tec_onboarding_wizard_visited_guided_setup', false );
+		if ( $visited_first_time_setup ) {
+			return;
+		}
+
+		// If we're still here, redirect to the Guided Setup page.
+		$setup_url = add_query_arg(
+			[
+				'post_type' => 'tribe_events',
+				'page'      => 'first-time-setup',
```
dpanta94, line 129: Replace any occurrence of `first-time-setup` with the variable Stephen pointed to.

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -88,6 +88,53 @@ public function handle_onboarding_page_dismiss(): void {
...
+		$post_type = tribe_get_request_var( 'post_type' );
+
+		// Do not redirect if they have older versions and are probably already set up.
+		$tec_versions = (array) tribe_get_option( 'previous_ecp_versions', [] );
+		if ( count( $tec_versions ) > 1 ) {
+			return;
+		}
+
+		// Do not redirect if the target is not The Events Calendar-related admin pages.
+		if ( ! str_contains( $post_type, 'tribe' ) ) {
```
dpanta94, line 110: I would become more specific to be honest.

i would do something like:

$post_types = apply_filters( 'tec_events_....', [ 'tribe_events', 'tribe_another' ..... ] );

if ( ! in_array() ) {
 return;
}

src/Tribe/Admin/Settings.php
dpanta94: I'll also explain my previous comment about why this place is not ideal for redirects and why i would prefer the `current_screen` hook.

Your custom action is firing along with the add_admin_pages method which is being fired along in the `admin_menu` action.

This is all fine currently.

BUT the method you are adding has one more hidden dependency: html output to not have started (you can't send headers and as a result redirect a user when html output has started)

This is being served by the current `admin_menu` hook. In my mind always lingers a though what if somebody changes the hook for the add_admin_pages method and as a result for my custom action as well ? My redirect could break then.

That's why for admin side redirects i choose for my callback to be directly bound to the `current_screen` hook. This is the last hook before html output starts while WP core has already gathered all the info you might need to evaluate your conditionals. (current screen has been fully set).

src/Events/Admin/Onboarding/Controller.php
dpanta94, line 121:
```suggestion
		if ( (bool) tribe_get_option( 'tec_onboarding_wizard_visited_guided_setup', false ) ) {
```

Also i would cast to bool

src/Tribe/Admin/Settings.php
dpanta94: For FE redirects we should be bound to `template_redirect` hook ;) Just a tip, not relevant here.

src/Tribe/Admin/Settings.php
dpanta94: I explained in another comment - ill resolve this.

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -88,6 +88,59 @@ public function handle_onboarding_page_dismiss(): void {
...
+		if ( (bool) tribe_get_option( 'tec_onboarding_wizard_visited_guided_setup', false ) ) {
+			return;
+		}
+
+		// Do not redirect if they dismissed the Guided Setup page.
+		if ( (bool) Landing_Page::is_dismissed() ) {
+			return;
+		}
+
+		// Do not redirect if the target is not The Events Calendar-related admin pages.
```
dpanta94, line 115: Full doc needed for filters!

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -88,6 +88,59 @@ public function handle_onboarding_page_dismiss(): void {
...
+		$post_types = apply_filters( 'tec_events_admin_post_types', [ 'tribe_events', 'tribe_event_series', 'tribe_venue', 'tribe_organizer' ] );
+
+		$post_type = tec_get_request_var( 'post_type' );
+
+		if ( ! in_array( $post_type, $post_types, true ) ) {
+			return;
+		}
```
dpanta94, line 122: This is lowest cost from quering DB - Should be placed right after the page check.

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -88,6 +88,59 @@ public function handle_onboarding_page_dismiss(): void {
...
+
+		// Do not redirect if they have been to the Guided Setup page already.
+		if ( (bool) tribe_get_option( 'tec_onboarding_wizard_visited_guided_setup', false ) ) {
+			return;
+		}
+
+		// Do not redirect if they dismissed the Guided Setup page.
+		if ( (bool) Landing_Page::is_dismissed() ) {
```
dpanta94, line 111: This is type hinted in the function that it returns bool as a result it CANNOT return anything else. So we don't need to cast it.

check the method [here](https://github.com/the-events-calendar/the-events-calendar/blob/b47be5bd162dd2ac06f44d928b4fcf9d0fbfd40e/src/Events/Admin/Onboarding/Landing_Page.php#L110)

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -75,7 +75,7 @@ public function add_actions(): void {
 		add_action( 'rest_api_init', [ $this, 'register_rest_endpoints' ] );
 		add_action( 'admin_post_' . Landing_Page::DISMISS_ONBOARDING_PAGE_ACTION, [ $this, 'handle_onboarding_page_dismiss' ] );
 		add_action( 'admin_notices', [ $this, 'remove_all_admin_notices_in_onboarding_page' ], -1 * PHP_INT_MAX );
-	}
+		add_action( 'tec_admin_headers_about_to_be_sent', [ $this, 'redirect_tec_pages_to_guided_setup' ] ); }
```
dpanta94, line 78: Make sure the action is actually removed in remove_actions please.

dpanta94 (changes requested):

changelog/fix-TEC-5401-add-redirects-for-all-tec-menu-options
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Add one time redirect to the First Time Setup page. [TEC-5401]
```
dpanta94, line 4: In Andra's PR i saw that the page got renamed to `Setup Guide` - Maybe reflect it here ?

sdokus: Ah good catch!!

dpanta94 (approved):

PR #4986, opened 2025-03-18
Optimize performance for context locations

naxoc:
### 🎫 Ticket
N/A

### 🗒️ Description
The `tribe_context_locations` filter gets called a lot. On my local testing machine it gets called 500+ times pr. page load. 

This PR makes it so that the large array with closures in it will only be created once instead of each filter call. In my unscientific testing this PR makes the "Own Memory" for each page load go from ~25 MB to ~0.0788 MB for the `Tribe\Events\Service_Providers\Context->filter_context_locations` function. This will considerably improve performance.

#### How to test
The only testing is to see that everything works the same as always :) 

##### Before
before
##### After
after

### ✔️ Checklist

### DIFF
--- a/src/Tribe/Service_Providers/Context.php
+++ b/src/Tribe/Service_Providers/Context.php
@@ -41,18 +41,16 @@ class Context extends Service_Provider {
 	}
 
 	/**
-	 * Filters the context locations to add the ones used by The Events Calendar.
+	 * Get the locations array for filtering.
 	 *
-	 * @since 4.9.4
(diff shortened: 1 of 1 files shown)


bordoni:
That is amazing, we have the Common Locations too that might be the correct place to cache this so that it gets cached for all of our plugins. But we will integrate your PR into that for sure. Thank you @naxoc

bordoni:
Resolved by the PR on common, it will be launched on our next version.

PR #4994, opened 2025-03-19
Event tickets notice only on TEC pages

bordoni:


changelog/display-event-tickets-notices-only-on-tec-pages
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Improved user experience by ensuring Event Tickets installation and activation notices only appear on TEC-related admin pages.
```
JPry, line 4: Needs a trailing new line.
```suggestion
Improved user experience by ensuring Event Tickets installation and activation notices only appear on TEC-related admin pages.

```

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -169,7 +169,8 @@ public function should_display_notice_install(): bool {
 
 		return ! $this->is_installed()
 			&& empty( tribe_get_request_var( 'welcome-message-the-events-calendar' ) )
-			&& ! $this->is_install_plugin_page();
+			&& ! $this->is_install_plugin_page()
+			&& $this->is_tec_related_page();
```
JPry, line 173: Minor: we should swap these, so that if `is_tec_related_page()` is `false`, we don't need to also check that it's not `is_install_plugin_page()`.

```suggestion
			&& $this->is_tec_related_page();
			&& ! $this->is_install_plugin_page()
```

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -192,7 +193,67 @@ public function should_display_notice_activate(): bool {
 		 *
 		 * @param bool $should_display True if the notice should display.
 		 */
-		return apply_filters( 'tec_events_admin_notice_event_tickets_should_display', $this->is_installed() && ! $this->is_active() && ! $this->is_install_plugin_page() );
+		return apply_filters(
```
JPry, line 196: For readability, we should put these conditionals into a variable, and then pass that variable to the `apply_filters()` call. With this many conditions strung together, it can be hard to mentally parse.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -192,7 +193,67 @@ public function should_display_notice_activate(): bool {
...
+		// Match any TEC post type screens (events, organizers, venues).
+		$tec_post_types = [
+			\Tribe__Events__Main::POSTTYPE,
+			\Tribe__Events__Organizer::POSTTYPE,
+			\Tribe__Events__Venue::POSTTYPE
+		];
+		
+		if ( in_array( $current_screen->post_type, $tec_post_types, true ) ) {
+			return true;
+		}
```
JPry, line 231: For this kind of comparison, we can (and should) switch away from `in_array()`. This is how I would do it:

```suggestion
		$tec_post_types = [
			\Tribe__Events__Main::POSTTYPE      => true,
			\Tribe__Events__Organizer::POSTTYPE => true,
			\Tribe__Events__Venue::POSTTYPE     => true,
		];
		
		if ( isset( $tec_post_types[ $current_screen->post_type ] ) ) {
			return true;
		}
```

JPry: Also, it would be nice to import the classes at the top of the file, and possibly also alias them at the same time.

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -192,7 +193,67 @@ public function should_display_notice_activate(): bool {
...
+		
+		// Match any screen ID containing 'tribe_events'.
+		if ( false !== strpos( $current_screen->id, \Tribe__Events__Main::POSTTYPE ) ) {
+			return true;
+		}
+		
+		// Match any screen ID containing 'tec-'.
+		if ( false !== strpos( $current_screen->id, 'tec-' ) ) {
```
JPry, line 239: Would this be better with `str_starts_with()`? Would we care if `tec-` is found somewhere else in the string (even if unlikely)?

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -192,7 +193,67 @@ public function should_display_notice_activate(): bool {
...
+			return true;
+		}
+		
+		// Match TEC settings pages.
+		if ( false !== strpos( $current_screen->id, 'tribe-common' ) ) {
+			return true;
+		}
+		
+		// If Admin Helpers class is available, use it for more comprehensive check.
+		if ( class_exists( 'Tribe__Admin__Helpers' ) ) {
```
JPry, line 249:
```suggestion
		if ( class_exists( Tribe__Admin__Helpers::class ) ) {
```

JPry (changes requested):

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -185,14 +188,74 @@ public function should_display_notice_activate(): bool {
...
+		}
+		
+		// Match any TEC post type screens (events, organizers, venues).
+		$tec_post_types = [
+			Events_Main::POSTTYPE,
+			Events_Organizer::POSTTYPE,
+			Events_Venue::POSTTYPE,
+		];
+		
+		if ( in_array( $current_screen->post_type, $tec_post_types, true ) ) {
```
JPry, line 231: @bordoni Did you happen to see [my earlier comment](https://github.com/the-events-calendar/the-events-calendar/pull/4994#discussion_r2003906547) about `in_array()`?

bordoni: Ahh I did I forgot to apply that

src/Tribe/Admin/Notice/Install_Event_Tickets.php
```diff
@@ -185,14 +188,74 @@ public function should_display_notice_activate(): bool {
...
+			return true;
+		}
+		
+		// Match TEC settings pages.
+		if ( false !== strpos( $current_screen->id, 'tribe-common' ) ) {
+			return true;
+		}
+		
+		// If Admin Helpers class is available, use it for more comprehensive check.
+		if ( class_exists( 'Tribe__Admin__Helpers', false ) ) {
```
JPry, line 251: Same suggestion as before, but with an additional question: Why don't we want to allow autoloading? Unless I'm missing something, I don't think this is a place where we would want to avoid it.

```suggestion
		if ( class_exists( Tribe__Admin__Helpers::class, false ) ) {
```

bordoni: Autoloading causes problems when loaded to early

JPry (approved):

PR #5016, opened 2025-03-27
Allow Caching of the Views Requests by switching to GET requests by default.

bordoni:


lucatume (changes requested):
smaller doc block updates

lucatume (approved):
Nice!

PR #5026, opened 2025-03-31
tweak(CT1) introduce query caching in the Model builder.

lucatume:
Ticket: [ECP-1898](https://stellarwp.atlassian.net/browse/ECP-1898)

This introduces query caching in the model builder to cache the results of duplicated queries at the builder level without running them twice.

[Demo](https://share.cleanshot.com/QHLG0szn)

Proper cache invalidation was already directly (`ct1_integration/BuilderTest.php`) and indirectly covered in existing tests.

I had to modify a single test that was insisting on the batched fetching functionality since the whole point of the test is to assert on the number of queries made.

[ECP-1898]: https://stellarwp.atlassian.net/browse/ECP-1898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changelog/fix-ECP-1898-builder-cache
+++ b/changelog/fix-ECP-1898-builder-cache
@@ -1,0 +1,4 @@
+Significance: patch
+Type: performance
+
+Avoid duplicate queries by caching model queries results. [ECP-1898]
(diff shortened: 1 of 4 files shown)


lucatume:
A number of comments from PHPCS come from existing code is prepared and cached elsewhere.

PR #5046, opened 2025-04-03
Enable sharing Widgets across diff Block Editors

bordoni:
### 🗒️ Description

This resolves a problem reported internally in preparation for an internal project, where copying any of our widgets across sites would break.

### 🎥 Artifacts 

https://i.bordoni.me/hc4Yg75F

### DIFF
--- a/src/Tribe/Views/V2/Widgets/Service_Provider.php
+++ b/src/Tribe/Views/V2/Widgets/Service_Provider.php
@@ -105,19 +105,19 @@ class Service_Provider extends Provider_Contract {
 	 * 
 	 * @since TBD
 	 * 
-	 * @param mixed $result The result of the rest request.
-	 * @param WP_REST_Server $server The REST server.
+	 * @param mixed           $result  The result of the rest request.
+	 * @param WP_REST_Server  $server  The REST server.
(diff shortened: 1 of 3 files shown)


src/Tribe/Views/V2/Widgets/Service_Provider.php
```diff
@@ -98,6 +100,59 @@ public function register_widget( $widgets ) {
...
+		if ( ! preg_match( '#^/wp/v2/widget-types/([a-zA-Z0-9_-]+)/encode$#', $route, $matches ) ) {
+			return $result;
+		}
+	
+		// Get the widget type ID from the route.
+		$widget_type_id = $matches[1];
+	
+		global $wp_widget_factory;
+
+		$widget_object = $wp_widget_factory->get_widget_object( $widget_type_id );
```
borkweb, line 133: Move this below the next two conditionals as it is slightly more expensive and not required for either conditional conditional

borkweb: Oh. Actually, I don't think it is being used at all

borkweb (approved):
One small change that seems worthwhile

borkweb (approved):

PR #5059, opened 2025-04-04
Update the CT1 from activating during the Upgrader.

bordoni:
Attempts to resolve the problem of CT1 loading while the upgrade is running.


dpanta94:
Ill add this safe lock by simply registering the controller on tribe_common_loaded instead of checking for actions fired

PR #5082, opened 2025-04-14
Restore our globals for linters

dpanta94:
Restore our globals for js linters


src/Events/Calendar_Embeds/Admin/List_Page.php
```diff
@@ -261,7 +261,7 @@ protected function register_assets(): void {
 			'tec-events-calendar-embeds-style',
 			'css/calendar-embeds/admin/page.css'
 		)
-			->add_to_group_path( 'tec-events-resources' )
+			->add_to_group_path( TEC::class )
```
Camwyn, line 264: Ooh, this is sneaky, I like it - but, can is cause any issues with backwards compatibility?

Camwyn: PS: I did a quick check and didn't see the string `tec-events-resources` anywhere else so I assume not.

Camwyn (approved):

PR #5114, opened 2025-04-28
Fix too strict ECE callbacks

dpanta94:
### 🎫 Ticket

[TEC-5484]

### 🗒️ Description

Making callbacks hooked tp WP-core hooks, in the ECE feature less strict on the expected arguments.
As a result fixing possible fatal errors when the arguments are not of the expected type.

### ✔️ Checklist

[TEC-5484]: https://stellarwp.atlassian.net/browse/TEC-5484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Events/Calendar_Embeds/Admin/Singular_Page.php
+++ b/src/Events/Calendar_Embeds/Admin/Singular_Page.php
@@ -85,12 +85,17 @@ class Singular_Page extends Controller_Contract {
 	 * Modifies the post updated messages for the calendar embed post type.
 	 *
 	 * @since 6.11.0
+	 * @since TBD Made the parameters non-strict.
 	 *
 	 * @param array $messages The post updated messages.
 	 *
 	 * @return array
 	 */
-	public function modify_post_updated_messages( array $messages ): array {
+	public function modify_post_updated_messages( $messages ): array {
+		if ( ! is_array( $messages ) ) {
+			return $messages;
+		}
+
 		if ( ! self::is_on_page() ) {
 			return $messages;
 		}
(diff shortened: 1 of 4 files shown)


redscar (approved):

rafsuntaskin (approved):

src/Events/Calendar_Embeds/Admin/Singular_Page.php
```diff
@@ -85,12 +85,17 @@ public function unregister(): void {
...
 	 * @since 6.11.0
+	 * @since TBD Made the parameters non-strict.
 	 *
 	 * @param array $messages The post updated messages.
 	 *
 	 * @return array
 	 */
-	public function modify_post_updated_messages( array $messages ): array {
+	public function modify_post_updated_messages( $messages ): array {
+		if ( ! is_array( $messages ) ) {
```
JPry, line 95: I didn't look at where the filter for this one is, but is there a special case when some other hook has modified `$messages` and therefore we want to skip our logic entirely? Wouldn't we want to instead do something like `$messages = (array) $messages;`, and then continue our logic?

dpanta94: Yea that is fine. what was happening they would set $messages to null. Giving me the idea that they dont want messages to be set at all.

They can fix their logic to be compatible with what WP expects. I decided to simply bail in this case which is ok

PR #5151, opened 2025-05-16
use `tec_asset` instead of `tribe_asset`

Camwyn:
Addresses issue in QA
https://lw.slack.com/archives/C0714HQSMEW/p1747407258176479


dpanta94 (approved):
Nice have you also checked about any usage of `tribe_assets` that we should update to `tec_assets` ?

Camwyn:
> Nice have you also checked about any usage of `tribe_assets` that we should update to `tec_assets` ?

Yeah - that's why I changed the one in the deprecated Freemius 😉

PR #5163, opened 2025-05-21
Fix EA script bug(s)

Camwyn:
Pull the object from the window, rather than passing it in as a parameter.

### 🎫 Ticket

[TEC-5505]

[TEC-5505]: https://stellarwp.atlassian.net/browse/TEC-5505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/resources/js/aggregator-admin-legacy-settings.js
+++ b/src/resources/js/aggregator-admin-legacy-settings.js
@@ -1,8 +1,8 @@
-var tribe_aggregator = tribe_aggregator || {};
-
-( function ( $, data ) {
+( function ( $ ) {
 	// eslint-disable-line no-unused-vars
 	'use strict';
+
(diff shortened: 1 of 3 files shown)


dpanta94 (changes requested):
To ensure backwards compatibility i would simply expose it to the window to begin with

e.g. instead of

```js
var tribe_aggregator = tribe_aggregator || {};
```

do

```js
window.tribe_aggregator = window.tribe_aggregator || {};
```

bordoni (changes requested):
Changelogs.

Camwyn:
Replaced by https://github.com/the-events-calendar/the-events-calendar/pull/5169

PR #5186, opened 2025-06-03
(Classy) Fix REST date save

JPry:
### 🎫 Ticket

[TICKET_ID]

### 🗒️ Description

This updates Classy so that the date data is properly saved.

Notable changes:

1. Move the Classy meta handling into a new `Meta` class
2. Expand the Meta class functionality to include everything that used to be handled by the `Tribe__Events__Editor__Meta` class
    * All of the fields from the old class have been added
    * The `sanitize_callback` properties have been added, with logic for handling the different types of meta data
    * The handling for custom UTC date generation has been added
    * The handling for `_EventCost` meta has been added
3. The `Supported_Post_Types` trait has been added
4. Tests for Classy updated to match the new behavior, with a new test class added for the `Meta` class

### 🎥 Artifacts 

Proof of Classy event **not** triggering a 404 with these changes:

https://github.com/user-attachments/assets/aa9f0b56-d87c-45dd-9079-832507ed9375

### ✔️ Checklist

### DIFF
--- a/src/Events/Classy/Supported_Post_Types.php
+++ b/src/Events/Classy/Supported_Post_Types.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Classy Supported_Post_Types Trait
+ *
+ * @since TBD
+ */
+
+declare( strict_types=1 );
+
+namespace TEC\Events\Classy;
+
+use Tribe__Events__Main as TEC;
+
+/**
+ * Trait Supported_Post_Types
+ *
+ * @since TBD
+ */
+trait Supported_Post_Types {
+	/**
+	 * Returns the list of post types that this controller supports.
+	 *
+	 * @since TBD
+	 *
+	 * @return array<string> The list of supported post types.
+	 */
+	private function get_supported_post_types(): array {
+		return [
+			TEC::POSTTYPE,
+		];
+	}
+}
--- a/src/Events/Classy/Controller.php
+++ b/src/Events/Classy/Controller.php
@@ -25,6 +25,7 @@ use Tribe__Events__Main as TEC;
 class Controller extends Controller_Contract {
 
 	use Can_Edit_Events;
+	use Supported_Post_Types;
 
 	/**
 	 * The list of event meta keys to be registered.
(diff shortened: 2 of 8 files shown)


src/Events/Classy/Supported_Post_Types.php
```diff
@@ -0,0 +1,32 @@
...
+	 * Returns the list of post types that this controller supports.
+	 *
+	 * @since TBD
+	 *
+	 * @return array<string> The list of supported post types.
+	 */
+	private function get_supported_post_types(): array {
+		return [
+			TEC::POSTTYPE,
+		];
```
bordoni, line 30: Filterable?

bordoni (changes requested):

src/Events/Classy/Supported_Post_Types.php
lucatume: While it looks like this should be filterable, I'm not sure what this would achieve. It's just a way to share the code to get the supported types between two controllers.
The returned values only make sense for the Events post type (look how the return value of this is used in the Meta controller), so the filter would be called something like `tec_events_classy_supported_event_post_types` which I'm not sure how should be explained: "Filters the event post types handled by the Classy controller"?

I think this method will evolve, not in this PR, into code that covers all the TEC-hadled post types (events, venues, organizers, etc.) and their meta types. What could make sense to filter is a map from post types to the meta values that are registered. That would make sense to filter.
This is outside of the scope of this PR, and will be handled the moment we need to deal with that.

lucatume (approved):

src/Events/Classy/Supported_Post_Types.php
JPry: > I think this method will evolve, not in this PR, into code that covers all the TEC-hadled post types (events, venues, organizers, etc.)
> This is outside of the scope of this PR, and will be handled the moment we need to deal with that.

This is the gist of my thought process here as well. We don't want to filter it at this moment, and if/when we find that a filter is necessary, we can add it then.

PR #5197, opened 2025-06-07
(Classy) Use site settings for Event Cost defaults

JPry:
### 🎫 Ticket

[TICKET_ID]

### ✔️ Checklist


src/resources/packages/classy/store/selectors.ts
```diff
@@ -113,9 +114,22 @@ export function areTicketsSupported( state: StoreState ): boolean {
...
 }
+
+/**
+ * Returns the default currency from the settings.
+ *
+ * @since TBD
+ *
+ * @return {Currency} The default currency.
+ */
+export function getDefaultCurrency(): Currency {
```
lucatume: This selector should likely live in the `tec/classy` store, in Common. While the currency data is used here, it will be  used by other components like ET later and it will make sense not have to repeat the work.

src/Events/Classy/Controller.php
```diff
@@ -99,17 +99,26 @@ public function add_supported_post_types( array $supported_post_types ): array {
 	public function filter_data( array $data ): array {
 		$data['settings'] ??= [];
 
-		$time_range_separator                                  = tribe_get_option( 'timeRangeSeparator', ' - ' );
-		$multi_day_cutoff                                      = tribe_get_option( 'multiDayCutoff', '00:00' );
-		[ $multi_day_cutoff_hours, $multi_day_cutoff_minutes ] = array_replace(
+		$additional_settings = [
+			'defaultCurrency'    => [
```
lucatume, line 103: Should the currency information live in Common instead? It would be used by both TEC and ET, and other premium plugins. It's currently used only in TEC. Good for this PR, but it might be moved.

tests/classy_jest/fields/EventCost/EventCost.spec.tsx
```diff
@@ -0,0 +1,190 @@
...
+	'core/editor': {
+		getEditedPostAttribute: (attribute: string) => any;
+	};
+	'tec/classy': {
+		getDefaultCurrency: () => Currency;
+	};
+};
+
+// Mock WordPress data store
+jest.mock('@wordpress/data', () => {
```
lucatume, line 21: This is another case where I would avoid using mocks, where possible, and use real stores and components as much as possible. I've not yet, at the time of this writing, "cracked" the code to make the Classy registry work outside of the Block Editor correctly, which makes integration testing not yet feasible for components like this one. This test is fine for the time being, but should be changed to an integration testing in the future to avoid all this mocking.

tests/classy_jest/jest.config.js
```diff
@@ -22,11 +22,16 @@ module.exports = {
...
+	transformIgnorePatterns: [
+		'/node_modules/(?!client-zip|@wordpress/.*)',
+	],
 	preset: 'ts-jest',
 	moduleFileExtensions: [ 'ts', 'tsx', 'js', 'jsx' ],
 	snapshotSerializers: [ '@emotion/jest/serializer' ],
 	moduleNameMapper: {
-		'@tec/common/(.*)$': '../../../common/src/resources/packages/$1',
+		'@tec/common/(.*)$': '<rootDir>/../../common/src/resources/packages/$1',
+		'@tec/common/classy/(.*)$': '<rootDir>/../../common/src/resources/packages/classy/$1',
```
lucatume, line 35: This one should not be required since Classy is already placed in the right directory.

src/Events/Classy/Controller.php
JPry: Yes, it should. When I moved the CurrencySelector into Common, I missed this.

tests/classy_jest/jest.config.js
JPry: This was an AI addition I meant to clean up before committing, but I missed it when the time came.

PR #5208, opened 2025-06-19
classy/component integration testing

lucatume:
Based on the [Common PR](https://github.com/the-events-calendar/tribe-common/pull/2640).

This PR is a first step toward use of the Classy application outside of the Block Editor context.

Following the refactoring of the Classy code in Common, being able to set up and run a simple test (see the `EventLocation.spec.tsx` file) using the `Provider` of the Classy application means being able to run code relying on the presence anc correct set up of the WordPress default store regsitry **outside** of the Block Editor (i.e. in the tests context for now).

The effort and logic used to make things work in tests brings with itself two benefits:
1. Simpler integration testing, where as little as possible (ideally only API requests) is mocked and components are tested with a fast and realistic usage approach.
2. Preparation and discover for the use of the Classy application outside of the Block Editor context.

### DIFF
--- a/tests/classy_jest/jest.config.js
+++ b/tests/classy_jest/jest.config.js
@@ -1,4 +1,5 @@
 const { defaults: tsjPreset } = require( 'ts-jest/presets' );
+const path = require( 'path' );
 
 module.exports = {
 	verbose: true,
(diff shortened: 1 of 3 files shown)


tests/classy_jest/jest.config.js
```diff
@@ -28,6 +29,8 @@ module.exports = {
 	preset: 'ts-jest',
 	moduleFileExtensions: [ 'ts', 'tsx', 'js', 'jsx' ],
 	snapshotSerializers: [ '@emotion/jest/serializer' ],
+	// Load modules only from TEC, override defaut resolution that could lead Comon loading from its own `nonode_modules`.
```
JPry: Some small typos

```suggestion
	// Load modules only from TEC, override default resolution that could lead Common loading from its own `node_modules`.
```

PR #5222, opened 2025-06-27
Proposal: Split views_integration tests into smaller bites.

Camwyn:
## 🎫 Ticket

No ticket, just hopes and dreams.

#### Includes:

A new bootstrap_base.php file that is used to bootstrap the test environment.
New test Helpers for each of the new suites.
Modified Views Integration Test Suite to use the new suites.
Added a new README.md file to the tests directory.
New test suites for:
- Unit Views
- Views Blocks
- Views Components
- Views Core
- Views Data
- Views Modules
- Views Partials
- Views SEO
See the new VIEWS_INTEGRATION_REFACTORING_SUMMARY.md file for more details.

#### Includes:
- Added a new reusable workflow for basic setup.
- Added a new reusable workflow for PHP change detector.
- Added a new reusable workflow for SLIC test runner.
- Modified tests to use the new reusable workflows.
- Added a new  skip flags parser workflow to parse the PR description for skip flags and output a comment on the PR.
- Added a WORKFLOWS.md file to document the new workflows and their purpose.

PR #5225, opened 2025-06-30
WIP: git workflows

Camwyn:
### 🗒️ Description

In a [previous PR](https://github.com/the-events-calendar/the-events-calendar/pull/5222) attempting to split up the views_integration tests, I made some changes to the workflow process.

I realized those should get pulled out into their own changeset. This is that and a bit more.

After converting what we have to leverage reusable parts, I then refactored the release tasks into reusable parts as well, allowing me to create one task orchestrating them all.

The next piece was to begin taking manual tasks that are currently not handled in our actions and adding them, thus a new reusable step equivalent to `tut analyze` - runs after the changelog is created and adds the changelog entries before processing. As a standalone, I've preserved the varied output formats, so they can still be  used to generate lists, etc for use elsewhere.

PS: Once this is working, my intention is to move as much of it as possible to`the-events-calendar/actions/.github/actions/`


.github/workflows/phpstan.yml
```diff
@@ -6,7 +6,19 @@ on:
       - '*.php'
       - 'src/**.php'
 jobs:
+  conditional:
+    # MODIFIED: Using reusable php-change-detector workflow
```
lucatume: Left-over.
```suggestion
```

.github/workflows/lint.yml
```diff
@@ -6,35 +6,22 @@ on:
...
       - 'src/resources/postcss/**.pcss'
+
 jobs:
   lint:
     runs-on: ubuntu-latest
     if: "!contains(github.event.pull_request.body, '[skip-lint]')"
     steps:
-      - name: Checkout the repository
-        uses: actions/checkout@v4
+      # MODIFIED: Using reusable basic-setup workflow
```
lucatume:
```suggestion
```

.github/workflows/phpcs.yml
```diff
@@ -1,44 +1,20 @@
 name: 'PHPCS'
 on: [pull_request]
+
 jobs:
   conditional:
-    if: "!contains(github.event.pull_request.body, '[skip-phpcs]')"
-    runs-on: ubuntu-latest
+    # MODIFIED: Using reusable php-change-detector workflow
```
lucatume:
```suggestion
```

.github/workflows/phpstan.yml
```diff
@@ -21,33 +33,33 @@ jobs:
...
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 1000
-          submodules: recursive
-      # ------------------------------------------------------------------------------
-      # Set up PHP and run phpstan
-      # ------------------------------------------------------------------------------
-      - name: Configure PHP environment to run phpstan
-        uses: shivammathur/setup-php@v2
+      # MODIFIED: Using reusable basic-setup workflow
```
lucatume:
```suggestion
```

lucatume (changes requested):
It's a bit difficult to parse this after all the work in #5222.

This seems a more gradual approach that would be merged before #5222, correct?

Camwyn:
> It's a bit difficult to parse this after all the work in #5222.
> 
> This seems a more gradual approach that would be merged before #5222, correct?

Correct - I realized this piece was more digestible and more gradual.

I've closed #5222 as (from our conversations) I think the splitting is not the best path forward.

.github/workflows/main-release.yml
```diff
@@ -0,0 +1,187 @@
...
+        type: choice
+        options:
+          - amend
+          - generate
+          - amend-version
+      target-branch:
+        description: "Target branch for the workflow"
+        required: false
+        type: string
+        default: "main"
```
dpanta94, line 29: Probably the default should be master until we switch over

.github/workflows/release-process-changelog.yml
```diff
@@ -23,120 +23,81 @@ on:
...
+          action-type: ${{ inputs.action-type }}
+          target-branch: ${{ inputs.target-branch || github.head_ref || github.ref_name }}
 
+      # Create PR only when run individually
       - name: Create new branch
         id: versions
+        if: github.event_name == 'workflow_dispatch'
         run: |
-          changeBranch="task/process-changelog/${{ steps.format_date.outputs.RELEASE_DATE }}/${{ steps.figure_out_version.outputs.RELEASE_VERSION }}/${{ steps.vars.outputs.sha_short }}"
+          changeBranch="task/process-changelog/$(date +%Y-%m-%d)/${{ steps.process-changelog.outputs.changelog-content || 'unknown' }}/$(git rev-parse --short ${{ github.sha }})"
```
dpanta94: This is problematic, the changelog-content should not be part of the branch name

.github/workflows/release-process-changelog.yml
```diff
@@ -23,120 +23,81 @@ on:
...
+        if: github.event_name == 'workflow_dispatch'
         uses: peter-evans/create-pull-request@v6
         id: cpr
         with:
           token: ${{ secrets.GHA_BOT_TOKEN_MANAGER }}
-          base: ${{ github.ref }}
-          branch: "task/process-changelog/${{ steps.format_date.outputs.RELEASE_DATE }}/${{ steps.figure_out_version.outputs.RELEASE_VERSION }}/${{ steps.vars.outputs.sha_short }}"
-          title: "[BOT] Process changelog for '${{ github.head_ref || github.ref_name }}'"
+          base: ${{ inputs.target-branch || github.ref }}
+          branch: "task/process-changelog/$(date +%Y-%m-%d)/${{ steps.process-changelog.outputs.changelog-content || 'unknown' }}/$(git rev-parse --short ${{ github.sha }})"
```
dpanta94: Similar :D

.github/workflows/reusable/basic-setup.yml
```diff
@@ -0,0 +1,83 @@
+name: Basic Setup
+on:
+  workflow_call:
+    inputs:
+      fetch-depth:
+        description: 'Git fetch depth'
+        required: false
+        type: number
+        default: 1000
```
dpanta94: I would set the default to `1`. Just because it's faster/cheaper. If an action needs something else aka `0` to pull everything or whatever better if we overwrite it there IMO

.github/workflows/reusable/release-process/analyze-changes.yml
```diff
@@ -0,0 +1,272 @@
...
+          fi
+          if [ ${#ACTIONS_REMOVED[@]} -gt 0 ]; then
+            CHANGELOG_ENTRIES+=("* Tweak - Removed actions: \`$(IFS='`, `'; echo "${ACTIONS_REMOVED[*]}")\`")
+          fi
+          if [ ${#VIEWS_CHANGED[@]} -gt 0 ]; then
+            CHANGELOG_ENTRIES+=("* Tweak - Changed views: \`$(IFS='`, `'; echo "${VIEWS_CHANGED[*]}")\`")
+          fi
+
+          # Create changelog entry
+          if [ ${#CHANGELOG_ENTRIES[@]} -gt 0 ]; then
```
dpanta94: You can reuse the action we have for creating changelog entries instead.

.github/workflows/reusable/release-process/analyze-changes.yml
```diff
@@ -0,0 +1,272 @@
...
+            # Create changelog directory if it doesn't exist
+            mkdir -p changelog
+
+            # Generate filename
+            TIMESTAMP=$(date +%Y%m%d-%H%M%S)
+            SHA_SHORT=$(git rev-parse --short ${{ github.sha }})
+            FILENAME="task-analyze-changes-$TIMESTAMP-$SHA_SHORT"
```
dpanta94: For example here your filename will probably fail on the second loop since its possible still the same second. The reusable action has a way of always producing unique filenames for each changelog

.github/workflows/reusable/release-process/replace-tbd-entries.yml
```diff
@@ -0,0 +1,167 @@
...
+
+          # Define file extensions to search
+          FILE_EXTENSIONS=(
+            "*.php"
+            "*.js"
+            "*.jsx"
+            "*.ts"
+            "*.tsx"
+            "*.css"
+            "*.scss"
```
dpanta94:
```suggestion
            "*.pcss"
```

Or it could be an addition instead

.github/workflows/reusable/release-process/replace-tbd-entries.yml
```diff
@@ -0,0 +1,167 @@
...
+            "*.md"
+            "*.txt"
+          )
+
+          # Define directories to exclude
+          EXCLUDE_DIRS=(
+            "*/vendor/*"
+            "*/node_modules/*"
+            "*/.git/*"
+            "*/tests/*"
```
dpanta94:
```suggestion
            "*/tests/*"
            "*/common/*"
```

TEC and ET should not affect TCMN

.github/workflows/templates/test-suite.yml
```diff
@@ -0,0 +1,56 @@
+name: Test Suite Template
```
dpanta94: What is that template ?

dpanta94 (changes requested):
In general i like what is being done here! Left some comments though

.github/workflows/main-release.yml
Camwyn: Yeah, this PR is into and based off `main` though.

.github/workflows/templates/test-suite.yml
Camwyn: A unused (as yet) thing I did not delete.

.github/workflows/reusable/basic-setup.yml
Camwyn: Agreed - and we are overriding it on a per-job basis already.

.github/workflows/templates/test-suite.yml
Camwyn: Deleted.

PR #5235, opened 2025-07-01
Enhanced Datepicker Accessibility Implementation

pattihis:
### 🎫 Ticket

[TEC-5177]

### 🗒️ Description

This PR implements comprehensive keyboard accessibility for calendar datepicker component, making it fully  WCAG compliant and enabling seamless navigation for users with disabilities.

This is a follow-up to: https://github.com/the-events-calendar/the-events-calendar/pull/5216 

####  Complete Keyboard Navigation

  - Arrow key navigation for all datepicker views (days, months, years, decades)
  - Tab/Shift+Tab support with proper focus management
  - Enter/Space activation for all interactive elements
  - Escape key functionality to close datepicker and return focus

#### ARIA Compliance

  - Added proper role="button" attributes to navigation controls
  - Implemented descriptive aria-label attributes for screen readers
  - Applied tabindex="0" to make elements keyboard focusable

#### Roving Tabindex Pattern

  - Implements WCAG-recommended grid navigation pattern
  - Only one element in each section is tabbable at a time
  - Arrow keys move focus within sections seamlessly

#### Core Function: `enhanceDatepickerA11yForAll()`

  - Location: /src/resources/js/views/datepicker.js (lines 758-1140)
  - Purpose: Dynamically enhances any visible datepicker with accessibility features
  - Scope: Supports all Bootstrap datepicker views (days, months, years, decades)

#### Enhancements

  - MutationObserver integration to detect DOM changes
  - Event-driven enhancement on month changes and view switches
  - Automatic initialization for dynamically loaded datepickers

#### For Keyboard Users

  - Complete datepicker navigation without mouse dependency
  - Intuitive arrow key movement matching visual layout
  - Clear focus indicators and logical tab order

#### For Screen Reader Users

  - Descriptive labels for all interactive elements
  - Proper semantic roles and ARIA attributes
  - Logical navigation announcements

####  For All Users

  - Escape key support for quick datepicker dismissal
  - Auto-close on focus loss for better UX
  - Maintained visual design with zero visual impact

#### Keyboard Testing

  - Tab navigation through all datepicker elements
  - Arrow key navigation in all views (days/months/years/decades)
  - Enter/Space activation of dates and controls
  - Escape key closing functionality

####  Screen Reader Testing

  - NVDA/JAWS announcement verification
  - Proper role and label reading
  - Navigation instruction clarity

---
Note: This implementation maintains 100% backward compatibility and requires no changes to existing HTML structure or CSS styles.

### DIFF
--- a/src/resources/js/views/datepicker.js
+++ b/src/resources/js/views/datepicker.js
@@ -90,6 +90,12 @@ tribe.events.views.datepicker = {};
 	 */
 	obj.keyCode = {
 		ENTER: 13,
+		ESCAPE: 27,
+		SPACE: 32,
+		LEFT: 37,
+		UP: 38,
+		RIGHT: 39,
+		DOWN: 40,
 	};
 
 	/**
@@ -649,6 +655,12 @@ tribe.events.views.datepicker = {};
 			.bootstrapDatepicker( obj.options )
 			.on( changeEvent, { container: $container }, changeHandler )
 			.on( 'show', { datepickerButton: $datepickerButton }, obj.handleShow )
+			.on( 'show', function() {
+				// Enhance accessibility when datepicker is shown
+				setTimeout(function() {
+					enhanceDatepickerA11yForAll();
+				}, 100);
+			})
 			.on(
 				'hide',
 				{ datepickerButton: $datepickerButton, input: $input, observer: obj.observer },
@@ -666,6 +678,20 @@ tribe.events.views.datepicker = {};
 
 		// deinit datepicker and event handlers before success
 		$container.on( 'beforeAjaxSuccess.tribeEvents', { container: $container, viewSlug }, obj.deinit );
+
+		// Hook into month change events to re-apply accessibility enhancements
+		$container.on('handleMutationMonthChange.tribeEvents', function() {
+			setTimeout(function() {
+				enhanceDatepickerA11yForAll();
+			}, 50);
+		});
+
+		// Also hook into datepicker's own changeMonth event as backup
+		$input.on('changeMonth', function() {
+			setTimeout(function() {
+				enhanceDatepickerA11yForAll();
+			}, 50);
+		});
 
 		$container.trigger( 'afterDatepickerInit.tribeEvents', [ index, $container, data ] );
 	};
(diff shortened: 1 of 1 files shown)


src/resources/js/views/datepicker.js
```diff
@@ -649,6 +655,12 @@ tribe.events.views.datepicker = {};
 			.bootstrapDatepicker( obj.options )
 			.on( changeEvent, { container: $container }, changeHandler )
 			.on( 'show', { datepickerButton: $datepickerButton }, obj.handleShow )
+			.on( 'show', function() {
+				// Enhance accessibility when datepicker is shown
+				setTimeout(function() {
+					enhanceDatepickerA11yForAll();
+				}, 100);
```
JPry: Proper spacing is needed.

Also, it's preferable to use an arrow function instead of the `function` keyword.

```suggestion
				setTimeout( function() {
					enhanceDatepickerA11yForAll();
				}, 100 );
```

src/resources/js/views/datepicker.js
```diff
@@ -667,6 +679,20 @@ tribe.events.views.datepicker = {};
 		// deinit datepicker and event handlers before success
 		$container.on( 'beforeAjaxSuccess.tribeEvents', { container: $container, viewSlug }, obj.deinit );
 
+		// Hook into month change events to re-apply accessibility enhancements
+		$container.on('handleMutationMonthChange.tribeEvents', function() {
+			setTimeout(function() {
+				enhanceDatepickerA11yForAll();
+			}, 50);
+		});
```
JPry: Same as the above.

```suggestion
		$container.on( 'handleMutationMonthChange.tribeEvents', () => {
			setTimeout( () => { enhanceDatepickerA11yForAll() }, 50 );
		} );
```

src/resources/js/views/datepicker.js
```diff
@@ -667,6 +679,20 @@ tribe.events.views.datepicker = {};
...
+
+		// Also hook into datepicker's own changeMonth event as backup
+		$input.on('changeMonth', function() {
```
JPry: Spacing and arrow functions here too.

src/resources/js/views/datepicker.js
```diff
@@ -722,48 +748,363 @@ tribe.events.views.datepicker = {};
...
-				.attr('role', 'button')
-				.attr('tabindex', '0')
-				.attr('aria-label', 'Previous month')
-				.off('keydown.a11y')
-				.on('keydown.a11y', function(e) {
-					if (e.key === 'Enter' || e.key === ' ' || e.keyCode === 13 || e.keyCode === 32) {
-						$(this).trigger('click');
-						e.preventDefault();
+		// Target all currently visible datepicker views (days, months, years, decades)
+		var $datepickers = $('.datepicker:visible').find('.datepicker-days:visible, .datepicker-months:visible, .datepicker-years:visible, .datepicker-decades:visible');
```
JPry: For both this selector and the one a couple lines down, it might be more readable to have an array of selectors and then `.join()` them into a single string.

src/resources/js/views/datepicker.js
```diff
@@ -722,48 +748,363 @@ tribe.events.views.datepicker = {};
...
+
+		// Clean up any existing accessibility event handlers first
+		$datepickers.find('th.prev, th.next, th.datepicker-switch, td.day, span.month, span.year, span.decade').off('.a11y');
```
JPry: For this entire function, please use correct spacing.

src/resources/js/views/datepicker.js
```diff
@@ -722,48 +748,363 @@ tribe.events.views.datepicker = {};
...
+				var $control = $(this);
+				var isNext = $control.hasClass('next');
+				var isPrev = $control.hasClass('prev');
+				var isSwitch = $control.hasClass('datepicker-switch');
+				$control
+					.attr('role', 'button')
+					.attr('tabindex', '0')
+					.off('keydown.a11y focus.a11y click.a11y')
+					.on('focus.a11y', function() {
+						// Header control received focus
```
JPry: Do we need to do something here? If not, then we should skip adding the listener.

src/resources/js/views/datepicker.js
```diff
@@ -722,48 +748,363 @@ tribe.events.views.datepicker = {};
...
+					})
+					.on('click.a11y', function() {
+						if (isPrev || isNext) {
+							setTimeout(function() {
+								enhanceDatepickerA11yForAll();
+							}, 100);
+						}
+					})
+					.on('keydown.a11y', function(e) {
```
JPry: We're deep in the nesting structure by this point. For readability, I think the event handler function should be separate from the rest of this nested structure as its own function.

src/resources/js/views/datepicker.js
```diff
@@ -722,48 +748,363 @@ tribe.events.views.datepicker = {};
...
+							if ($firstCell && $firstCell.length) {
+								$firstCell.focus();
+							}
+							return false;
+						}
+					});
+
+				// Set appropriate aria-labels
+				if (isPrev) {
+					$control.attr('aria-label', 'Previous month');
```
JPry: This and the other strings below need to be translated.

src/resources/js/views/datepicker.js
```diff
@@ -722,48 +748,363 @@ tribe.events.views.datepicker = {};
...
+
+			/**
+			 * Cell Navigation Accessibility Enhancement
+			 */
+			// Enhance selectable cells for better keyboard navigation
+			// Different views have different selectable elements
+			var $selectableCells;
+			var cellType = '';
+
+			if ($datepickerView.hasClass('datepicker-days')) {
```
JPry: If possible, I think it would be better to use the native [`.classList`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) property to get all the classes, and then handle as needed. That would be nicer than this big if/elseif block.

src/resources/js/views/datepicker.js
```diff
@@ -722,48 +748,363 @@ tribe.events.views.datepicker = {};
...
+			} else if ($datepickerView.hasClass('datepicker-decades')) {
+				$selectableCells = $datepickerView.find('span.decade');
+				cellType = 'decade';
+			} else {
+				$selectableCells = $();
+				cellType = 'unknown';
+			}
+
+
+			$selectableCells.each(function(index) {
```
JPry: This is also a big enough function to justify putting it elsewhere and referencing it here.

PR #5258, opened 2025-07-15
Moves bidnings from Main to dedicated cotnroller

dpanta94:
Moves provider registrations from Main to dedicated up to date Controller.

Adds codeowners to make sure we no longer continue to add registrations in src/Tribe/Main.php

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,7 @@
 * @the-events-calendar/engineers
 
-src/Tribe/Main.php @dpanta94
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+* @the-events-calendar/engineers
+
+src/Tribe/Main.php @dpanta94
```
lucatume: Are there names missing from here?

PR #5264, opened 2025-07-16
TEC-5537 - avoid TIMESTAMPDIFF where possible

lucatume:
Ticket: [TEC-5537](https://stellarwp.atlassian.net/browse/TEC-5537)

In this PR:
1. Update the repository function to fetch overlapping event to behave differently depending on the requested seconds overlap and avoid use of the `TIMESTAMPDIFF` function.
2. Update the `By_Day_View` logic to use an overlap of `1` (thus using the new logic) over an overlap of `2`. I'm not sure about the original intention to use an overlap of `2`, but the same, non inclusive matching effect can be obtained using the overlap value of `1`.

This updates the repository method fetching overlapping events, to use
logic that will not require casting of the start / end dates and the use
of the `TIMESTAMPDIFF` function.

The method changed is the one serving this repository API:

```php
tribe_events()->where('date_overlaps', $start, $end);
```

By default the overlap interval is 1 second, i.e. a non-inclusive check,
but a different overlap can be specified:

```php
// Overlap of 23 seconds.
tribe_events()->where('date_overlaps', $start, $end, $timezone, 23);
```

If the specified overlap is more than `1`, then the function must use
the previous logic based on `TIMESTAMPDIFF`, else it will use a
string-based logic leveraging the fact that the start/end date/times are
saved in the format `Y-m-d H:i:s` that allows for string comparison
while making sense in numbers as well.

Artifacts:
* existing tests keep running
* new dedicated `wpunit` and `ct1_integration` tests

Note: comparing `Y-m-d H:i:s` format strings in MySQL
The string to string comparison is fine to establish greater/lesser (inclusive or not), but cannot be used for seconds comparison:
```
mysql> select '2025-01-01 10:00:23' - '2025-01-01 10:00:00';
+-----------------------------------------------+
| '2025-01-01 10:00:23' - '2025-01-01 10:00:00' |
+-----------------------------------------------+
|                                             0 |
+-----------------------------------------------+
1 row in set, 2 warnings (0.00 sec)

mysql> select if('2025-01-01 10:00:23' > '2025-01-01 10:00:00', 'yes', 'no');
+----------------------------------------------------------------+
| if('2025-01-01 10:00:23' > '2025-01-01 10:00:00', 'yes', 'no') |
+----------------------------------------------------------------+
| yes                                                            |
+----------------------------------------------------------------+
1 row in set (0.03 sec)

mysql> select if('2025-01-01 10:00:23' >= '2025-01-01 10:00:00', 'yes', 'no');
+-----------------------------------------------------------------+
| if('2025-01-01 10:00:23' >= '2025-01-01 10:00:00', 'yes', 'no') |
+-----------------------------------------------------------------+
| yes                                                             |
+-----------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> select if('2025-01-01 10:00:23' < '2025-01-01 10:00:00', 'yes', 'no');
+----------------------------------------------------------------+
| if('2025-01-01 10:00:23' < '2025-01-01 10:00:00', 'yes', 'no') |
+----------------------------------------------------------------+
| no                                                             |
+----------------------------------------------------------------+
1 row in set (0.00 sec)
```

The first query returns a difference of `0`, where the expected difference we're looking for is in seconds and comparable to an integer value.
This is the reason the logic is forking to support (the probably little used) case where the requested difference in seconds is more than `1`. 

[TEC-5537]: https://stellarwp.atlassian.net/browse/TEC-5537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### [6.14.1] 2025-07-16
+
+* Fix - Ensure all featured images on calendar views have dimensions set to reduce layout shifts. [TEC-4715]
+* Fix - Ensure views selector links have proper aria-current attributes.
+ Tweak - Changed views: `v2/components/events-bar/views/list/item`, `v2/day/event/featured-image`, `v2/latest-past/event/featured-image`, `v2/list/event/featured-image`, `v2/month/calendar-body/day/calendar-events/calendar-event/featured-image`, `v2/month/mobile-events/mobile-day/mobile-event/featured-image`
+* Language - 0 new strings added, 1 updated, 0 fuzzied, and 0 obsoleted.
+
 ### [6.14.0] 2025-07-03
 
 * Feature - Category Colors functionality has been merged into The Events Calendar. [TEC-5364]
--- a/changelog/tweak-TEC-5179-featured-alt
+++ b/changelog/tweak-TEC-5179-featured-alt
@@ -1,0 +1,4 @@
+Significance: minor
+Type: accessibility
+
+Updated day views with proper aria label and titles for event linked featured images. [TEC-5179]
--- a/lang/the-events-calendar.pot
+++ b/lang/the-events-calendar.pot
@@ -2,14 +2,14 @@
 # This file is distributed under the GPLv2 or later.
 msgid ""
 msgstr ""
-"Project-Id-Version: The Events Calendar 6.14.0\n"
+"Project-Id-Version: The Events Calendar 6.14.1\n"
 "Report-Msgid-Bugs-To: https://evnt.is/191x\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"POT-Creation-Date: 2025-07-03T16:19:05+00:00\n"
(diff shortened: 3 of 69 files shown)


src/Tribe/Repositories/Event.php
```diff
@@ -459,37 +459,79 @@ public function filter_by_date_overlaps( $start_datetime, $end_datetime, $timezo
...
+		if ( $min_sec_overlap > 1 ) {
+			/*
+			 * If the minimum overlap requested is more than one second, then we have to use a not performant logic.
+			 * The interpolated variables come from the code.
+			 *
+			 * @phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+			 */
+			$alt_where = $wpdb->prepare(
+				"(
+				TIMESTAMPDIFF ( SECOND, {$join_start_key}.meta_value, %s ) >= %d
```
Camwyn, line 494: Why not pass `$join_start_key` (and `$join_end_key`) in as a param as well - quotes?

lucatume: If you mean using the `%i` placeholder, that cannot be used for aliases of tables (I've tried).

Camwyn (approved):

bordoni (approved):

PR #5271, opened 2025-07-22
Fix - Add URL validation for Website URLs of Events, Venues and Organizer.

rafsuntaskin:
### 🎫 Ticket
Issue - 14

### 🗒️ Description
* Added URL validation for Event Details Website input.
* Organizer Website URL.
* Venue Website URL.
* Invalid URLs are highlighted and skipped from saving.

### 🎥 Artifacts 
 📷 screenshot(s):
 https://share.cleanshot.com/vvVHNmsX
 https://share.cleanshot.com/K1WtzJ4r

### DIFF
--- a/src/resources/packages/classy/fields/EventDetails/EventDetails.tsx
+++ b/src/resources/packages/classy/fields/EventDetails/EventDetails.tsx
@@ -26,6 +26,7 @@ export default function EventDetails( props: FieldProps ) {
 
 	const [ description, setDescription ] = useState< string >( postContent || '' );
 	const [ eventUrlValue, setEventUrlValue ] = useState< string >( eventUrlMeta );
+	const [ isUrlInvalid, setIsUrlInvalid ] = useState< boolean >( false );
 
 	useEffect( () => {
 		setDescription( postContent );
@@ -44,9 +45,11 @@ export default function EventDetails( props: FieldProps ) {
 		const urlValue = nextValue ?? '';
 
 		if ( ! isValidUrl( urlValue ) ) {
+            setIsUrlInvalid( true );
 		    return;
 		}
 
+        setIsUrlInvalid( false );
         setEventUrlValue( urlValue );
         editPost( { meta: { [ METADATA_EVENT_URL ]: nextValue } } );
 	};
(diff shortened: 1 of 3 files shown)


src/resources/packages/classy/fields/EventDetails/EventDetails.tsx
```diff
@@ -25,6 +26,7 @@ export default function EventDetails( props: FieldProps ) {
 
 	const [ description, setDescription ] = useState< string >( postContent || '' );
 	const [ eventUrlValue, setEventUrlValue ] = useState< string >( eventUrlMeta );
+	const [ isUrlInvalid, setIsUrlInvalid ] = useState< boolean >( false );
```
lucatume: The starting value should be based on `eventUrlMeta`, right?

lucatume: Can you change the const name to `isUrlValid`? Easier to use in functions following human preference for positive logic.

src/resources/packages/classy/fields/EventLocation/VenueUpsert.tsx
```diff
@@ -69,6 +70,7 @@ export default function VenueUpsert( props: {
 
 	// At a minimum, a Venue requires a name.
 	const [ confirmEnabled, setConfirmEnabled ] = useState( currentValues.name !== '' );
+	const [ isUrlInvalid, setIsUrlInvalid ] = useState< boolean >( false );
```
lucatume: Same question about initial value here, should it be based on the venue current values?
And same observation about making the const a positive one, `isValidUrl`.

src/resources/packages/classy/fields/EventOrganizer/OrganizerUpsert.tsx
```diff
@@ -29,6 +30,23 @@ export default function OrganizerUpsert( props: {
 
 	// At a minimum an Organizers requires a name.
 	const [ confirmEnabled, setConfirmEnabled ] = useState( currentValues.name !== '' );
+	const [ isUrlInvalid, setIsUrlInvalid ] = useState< boolean >( false );
```
lucatume: Same observations done for the Venue and Event.

lucatume (changes requested):

src/resources/packages/classy/fields/EventDetails/EventDetails.tsx
rafsuntaskin: @lucatume starting value is not tied to saved value intentionally, this is to avoid showing red signs where previously saved values might be invalid. It will show the warning when the user changes the value and won't save without a valid value.

lucatume (approved):

PR #5288, opened 2025-08-07
Fix multiple venue support in REST API and ORM

Camwyn:
These commits fix the issue where multiple venues were being filtered to only the first venue when passed through the REST API or not being assigned at all.

**Requires:** https://github.com/the-events-calendar/tribe-common/pull/2694

### Primary Fix:
- Modified With_Transform_Organizers_And_Venues trait to check the
  venue linked post type's 'allow_multiple' setting before deciding
  whether to use all venues or just the first one
- When Events Pro is active and multiple venues are allowed, the full
  venues array is now preserved instead of being truncated

### Repository Improvements:
- Fixed missing return statement in filter_by_cost_less_than method
- Removed debug error_log statements from Event repository
- Fixed inconsistent return statements in cost filter methods

### Endpoint Updates:
- Removed redundant transform_input_params calls from Event and Events
  endpoints since the transformation is now handled properly in the trait

### Linked Posts Enhancement:
- Added get_type_args method to Linked_Posts class for easier access
  to linked post type configuration
- Now uses the Linked Post "allow_multiple" to determine if more than one linked post will get attached.

## Testing Results:
✅ Multiple venues work correctly when Events Pro is active
✅ Single venue fallback when Events Pro is inactive
✅ Event creation, updating, and venue management all functional
✅ Backwards compatibility maintained

Fixes the core issue where Postman API calls with multiple venues
were only saving the first venue in filter_meta_input.

----

Updated common submodule to include fix for Update_Entity_Response trait
that was causing TypeError when updating venues and organizers via REST API.

The issue was in the response logic which tried to call array_keys() on
a repository object instead of using the entity ID directly.

Changes:
- Updated common submodule
- Venue and organizer updates now return proper HTTP 200 responses
- Error handling for invalid IDs still works correctly (403/404)

This resolves the production-blocking 500 error while maintaining
proper error handling for edge cases.

**See:** https://github.com/the-events-calendar/tribe-common/pull/2694

----

When Events Pro is inactive, multiple venues should fallback to storing
only the first venue as a single integer value rather than a serialized
array. This ensures proper venue retrieval and display.

Changes:
- Modified venue type check from allow_multiple negation to empty() check
  to handle cases where allow_multiple is an empty string
- Store single venue ID as integer instead of array when multiple venues
  are not allowed
- Fixes issue where venue ID 1 was incorrectly assigned instead of the
  first venue from the input array

Tested scenarios:
- ECP Active: Multiple venues [413, 414] stored and retrieved correctly
- ECP Inactive: Multiple venues [413, 414] -> First venue 413 stored as
  integer and retrieved correctly

[skip-changelog][skip-lint]

### DIFF
--- a/src/Events/REST/TEC/V1/Endpoints/Event.php
+++ b/src/Events/REST/TEC/V1/Endpoints/Event.php
@@ -306,16 +306,7 @@ class Event extends Post_Entity_Endpoint implements RUD_Endpoint {
 		return $schema;
 	}
 
-	/**
-	 * Returns the arguments for the delete request.
-	 *
-	 * @since TBD
-	 *
-	 * @return QueryArgumentCollection
-	 */
-	public function delete_args(): QueryArgumentCollection {
-		return new QueryArgumentCollection();
-	}
+
 
 	/**
 	 * Returns the OpenAPI schema for deleting an event.
@@ -331,7 +322,7 @@ class Event extends Post_Entity_Endpoint implements RUD_Endpoint {
 			$this->get_operation_id( 'delete' ),
 			$this->get_tags(),
 			$this->get_path_parameters(),
-			null,
+			$this->delete_args(),
 			null,
 			true
 		);
--- a/src/Events/REST/TEC/V1/Endpoints/Event.php
+++ b/src/Events/REST/TEC/V1/Endpoints/Event.php
@@ -68,9 +68,6 @@ class Event extends Post_Entity_Endpoint implements RUD_Endpoint {
 	 * @return WP_REST_Response The response object.
 	 */
 	public function update( array $params = [] ): WP_REST_Response {
-		// Transform venues array to use first element (TEC default behavior).
-		$params = $this->transform_input_params( $params );
-
 		$id = $params['id'] ?? null;
 		if ( ! $id ) {
 			return new WP_REST_Response(
--- a/src/Events/REST/TEC/V1/Endpoints/Events.php
+++ b/src/Events/REST/TEC/V1/Endpoints/Events.php
@@ -72,9 +72,6 @@ class Events extends Post_Entity_Endpoint implements Readable_Endpoint, Creatabl
 	 * @return WP_REST_Response The response object.
 	 */
 	public function create( array $params = [] ): WP_REST_Response {
-		// Transform venues array to use first element (TEC default behavior).
-		$params = $this->transform_input_params( $params );
-
 		// Call the parent create method.
 		$entity = $this->get_orm()->set_args( $params )->create();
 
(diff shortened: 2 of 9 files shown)


src/Tribe/Repositories/Event.php
```diff
@@ -1061,11 +1066,9 @@ public function filter_by_cost_currency_symbol( $symbol ) {
 	 *                       e.g. "USD" for U.S. dollars, or a currency symbol, e.g. "$".
 	 *                       In the latter case results will include any event with the matching currency symbol,
 	 *                       this might lead to ambiguous results.
-	 *
-	 * @return array An array of query arguments that will be added to the main query.
 	 */
 	public function filter_by_cost_between( $low, $high, $symbol = null ) {
-		return $this->by( 'cost', [ $low, $high ], 'BETWEEN', $symbol );
+		$this->by( 'cost', [ $low, $high ], 'BETWEEN', $symbol );
```
bordoni: Dont make this change.

src/Tribe/Repositories/Event.php
```diff
@@ -1080,11 +1083,9 @@ public function filter_by_cost_between( $low, $high, $symbol = null ) {
 	 *                           e.g. "USD" for U.S. dollars, or a currency symbol, e.g. "$".
 	 *                           In the latter case results will include any event with the matching currency symbol,
 	 *                           this might lead to ambiguous results.
-	 *
-	 * @return array An array of query arguments that will be added to the main query.
 	 */
 	public function filter_by_cost_greater_than( $value, $symbol = null ) {
-		return $this->by( 'cost', $value, '>', $symbol );
+		$this->by( 'cost', $value, '>', $symbol );
```
bordoni: Dont make this change.

bordoni (changes requested):

src/Events/REST/TEC/V1/Endpoints/Event.php
```diff
@@ -68,9 +68,6 @@ class Event extends Post_Entity_Endpoint implements RUD_Endpoint {
 	 * @return WP_REST_Response The response object.
 	 */
 	public function update( array $params = [] ): WP_REST_Response {
-		// Transform venues array to use first element (TEC default behavior).
```
dpanta94, line 71: Everything that is here already happens in the update method of the trait. So it makes the overwriting here not adding any value.

src/Events/REST/TEC/V1/Endpoints/Events.php
```diff
@@ -72,9 +72,6 @@ class Events extends Post_Entity_Endpoint implements Readable_Endpoint, Creatabl
 	 * @return WP_REST_Response The response object.
 	 */
 	public function create( array $params = [] ): WP_REST_Response {
-		// Transform venues array to use first element (TEC default behavior).
-		$params = $this->transform_input_params( $params );
```
dpanta94, line 76: Similar as above

dpanta94 (changes requested):

src/Events/REST/TEC/V1/Endpoints/Event.php
Camwyn: I had created and now removed that function in the Trait. We're not overwriting anything.

PR #5290, opened 2025-08-11
Fix transform_entity filter chain in REST API

Camwyn:
- Call parent::transform_entity() first in With_Transform_Organizers_And_Venues trait
- This ensures occurrence ID transformation filters are applied before venue/organizer transformations
- Fixes issue where occurrence IDs were returned instead of real post IDs
- Critical for proper database relationships between events, tickets, and other entities

Previously, the trait was overriding transform_entity() without calling the parent method that applies the tec_rest_v1_tribe_events_transform_entity filter, breaking the occurrence ID to post ID transformation.

See: https://github.com/the-events-calendar/events-pro/pull/2822


dpanta94 (changes requested):
I think we should call the filters before returning not at the start of the method. Also there are tests breaking

PR #5296, opened 2025-08-15
TEC-5571 PHP 8.3 deprecation notices

andrasguseo:
### 🎫 Ticket

[TEC-5571]

### 🗒️ Description

This PR addresses some PHP 8.3 compatibility warnings based on the respective Jira ticket.
Note: fixing the test files is irrelevant for production code, so those have not been touched.

### 🎥 Artifacts 
N/A

### ✔️ Checklist


src/views/blocks/event-links.php
```diff
@@ -3,17 +3,18 @@
...
- * @link http://evnt.is/1aiy
+ * @link https://evnt.is/1aiy
  *
  * @version 4.7
  *
  * @var Tribe__Events__Editor__Template $this
  */
 
-// don't show on password protected posts
+// Don't show on password protected posts.
```
Camwyn: Is this in the wrong place? Should be above line 20

Camwyn (changes requested):
One quibble on a docblock.

src/views/blocks/event-links.php
andrasguseo: Totally valid. Moved.

src/views/blocks/event-links.php
```diff
@@ -25,7 +26,7 @@
 $has_outlook_365  = $this->attr( 'hasOutlook365' );
 $has_outlook_live = $this->attr( 'hasOutlookLive' );
 
-// don't show on password protected posts
+// Don't show on password protected posts.
```
dpanta94: Twice the same conditional ?

src/Events/Custom_Tables/V1/Models/Builder.php
```diff
@@ -290,38 +290,42 @@ public function set_batch_size( $size = 100 ) {
...
+	 * @since TBD Create working copies to preserve original parameter values for debug_backtrace().
 	 *
 	 * @param array<string>            $unique_by A list of columns that are marked as UNIQUE on the database.
 	 * @param array<string,mixed>|null $data      The data to be inserted or updated into the table.
 	 *
 	 * @return false|int The rows affected flag or false on failure.
 	 */
 	public function upsert( array $unique_by, array $data = null ) {
-		if ( empty( $unique_by ) ) {
+		$working_unique_by = $unique_by;
```
dpanta94, line 303: I dont understand why we are renaming the variables. Could you explain ?

andrasguseo: Since PHP 7.0, functions inspecting arguments, like `func_get_args()` and `debug_backtrace()`, no longer report the original value as passed to a parameter, but will instead provide the current value.
(We call `debug_backtrace` on line 393.)

With this change we make sure that the original parameter stays the same and is used by `debug_backtrace()`, and the current values stored in `$working_...` also get passed as separate parameters.

dpanta94 (dismissed):

Camwyn (approved):

PR #5297, opened 2025-08-16
Add Request Query_Vars controller for query variable management

bordoni:
## Summary
- Introduces a new Request controller system for managing custom query variables in The Events Calendar
- Provides centralized registration and management of query variables
- Ensures proper integration with WordPress query system

## Changes
- **New Query_Vars class**: Handles registration of custom query variables (`tec_post_parent`, `tec_should_update_meta`, etc.)
- **New Request Controller**: Manages the lifecycle and registration of query variable handlers
- **Integration**: Registered Request Controller in main Events Controller provider stack
- **Tests**: Added comprehensive integration tests covering all query variable functionality

## Test Plan
- [x] Run integration tests: `codecept run tests/integration/Events/Request/Query_Vars_Test.php`
- [ ] Verify query variables are properly registered in WordPress
- [ ] Test that custom query variables work in WP_Query operations
- [ ] Ensure no conflicts with existing query variable handling

## Benefits
This implementation provides a clean, extensible architecture for managing custom query variables that The Events Calendar needs for various operations. It follows the established controller pattern used throughout the codebase and includes full test coverage.


src/Events/Request/Query_Vars.php
```diff
@@ -0,0 +1,140 @@
...
+			}
+
+			$normalized = $this->normalize_ical( $value );
+
+			if ( null !== $normalized ) {
+				$vars['ical'] = $normalized;
+			} else {
+				unset( $vars['ical'] );
+			}
+		}
```
Camwyn, line 71: This, and the casting to `null` seems overkill - we _never_ pass this as false so we should just unset it if it's not true

src/Events/Request/Query_Vars.php
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	/**
+	 * Normalizes the `ical` value to either `1` or `null` (unset).
+	 *
+	 * @since TBD
+	 *
+	 * @param mixed $value The raw value to normalize.
+	 *
+	 * @return int|null `1` when truthy, `null` when not.
```
Camwyn, line 130:
```suggestion
	 * @return ?int`1` when truthy, `null` when not.
```

src/Events/Request/Query_Vars.php
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	protected function normalize_ical( $value ) {
```
Camwyn, line 132:
```suggestion
	protected function normalize_ical( $value ): ?int {
```

src/Events/Request/Query_Vars.php
```diff
@@ -0,0 +1,140 @@
...
+		// Support presence-only query var (?ical) as truthy.
+		if ( '' === $value || null === $value ) {
+			return 1;
+		}
+
+		return tribe_is_truthy( $value ) ? 1 : null;
```
Camwyn, line 138: Optionally:
```suggestion
		return (int) tribe_is_truthy( $value ) || null;
```

src/Events/Request/Query_Vars.php
Camwyn: Also - I am not psyched about "normalizing" an `int` into a null...

Normalization should not change the expected **`type`** - just ensure we have a valid value.

Camwyn (changes requested):

src/Events/Request/Controller.php
```diff
@@ -0,0 +1,53 @@
...
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	protected function do_register(): void {
+		$this->container->singleton( Query_Vars::class, Query_Vars::class, [ 'register' ] );
+
+		// Initialize the singleton and register the hooks.
+		$this->container->make( Query_Vars::class );
+	}
```
dpanta94, line 37: Why isn't Query_Vars code here ?

Also why Query var is register as a singleton which on init it should call the register method ? Meaning a provider/controller !

bordoni:
@Camwyn lets make sure this gets released soon.

PR #5397, opened 2025-09-30
Cast return as string.

redscar:
While testing, Victor received a fatal 
```
PHP Fatal error:  Uncaught TypeError: Return value of TEC\Events\Integrations\Plugins\Elementor\Template\Controller::filter_events_template_setting_option() must be of the type string, null returned in /.../wp-content/plugins/the-events-calendar/src/Events/Integrations/Plugins/Elementor/Template/Controller.php:124
Stack trace:
#0 /.../wp-includes/class-wp-hook.php(326): TEC\Events\Integrations\Plugins\Elementor\Template\Controller->filter_events_template_setting_option(NULL)
#1 /.../wp-includes/plugin.php(205): WP_Hook->apply_filters(NULL, Array)
#2 /.../wp-content/plugins/the-events-calendar/common/src/functions/template-tags/general.php(57): apply_filters('tribe_get_optio...', NULL, 'tribeEventsTemp...', 'default')
#3 /.../wp-content/plugins/the-events-calendar/src/Tribe/Views/V2 in /.../wp-content/plugins/the-events-calendar/src/Events/Integrations/Plugins/Elementor/Template/Controller.php on line 124
```

We were unable to replicate it. However, this fix should fix the issue if it was to occur again.

[skip-changelog]

### DIFF
--- a/src/Events/Integrations/Plugins/Elementor/Template/Controller.php
+++ b/src/Events/Integrations/Plugins/Elementor/Template/Controller.php
@@ -154,7 +154,7 @@ class Controller extends Controller_Contract {
 	 * @return mixed results of option query.
 	 */
 	public function filter_tribe_get_single_option( $option, $default_value, $option_name ) {
-		// ONly this option.
+		// Only this option.
 		if ( 'tribeEventsTemplate' !== $option_name ) {
 			return $option;
 		}
--- a/src/Events/Integrations/Plugins/Elementor/Template/Controller.php
+++ b/src/Events/Integrations/Plugins/Elementor/Template/Controller.php
@@ -113,6 +113,7 @@ class Controller extends Controller_Contract {
 	 * Force the correct Single Event template object for Elementor theme.
 	 *
 	 * @since 6.4.0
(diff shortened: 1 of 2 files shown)


src/Events/Integrations/Plugins/Elementor/Template/Controller.php
Camwyn:
```suggestion
		// Only this option.
```

Camwyn (approved):

src/Events/Integrations/Plugins/Elementor/Template/Controller.php
```diff
@@ -121,12 +122,12 @@ public function remove_filters(): void {
...
 		// Only for events.
 		if ( ! is_singular( TEC::POSTTYPE ) ) {
-			return $option;
+			return (string) $option;
 		}
 
 		// Only for single events.
 		if ( ! tribe( Template_Bootstrap::class )->is_single_event() ) {
-			return $option;
+			return (string) $option;
```
rafsuntaskin, line 130: We should not cast here.

PR #5407, opened 2025-10-01
TEC-5583 Change optin default from false to null

sdokus:
### 🎫 Ticket

[TEC-5583]

### 🗒️ Description

This PR goes along with https://github.com/the-events-calendar/tribe-common/pull/2787 to make sure that the opt-in setting is not prematurely set before a user is actually prompted to make a decision about opting into Telementry or not. This PR is just a small tweak to not set the default value to `false` instead of `null` when grabbing the option and processing the initial Onboarding tab to account for if a user skips the Onboarding Wizard (in which case we would still want to show the Telementry modal). 

This PR is preferred over https://github.com/the-events-calendar/the-events-calendar/pull/5404 because it addresses the root of the issue instead of a workaround. 

### ✔️ Checklist

[TEC-5583]: https://stellarwp.atlassian.net/browse/TEC-5583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Events/Admin/Onboarding/Steps/Optin.php
```diff
@@ -35,14 +35,15 @@ class Optin extends Abstract_Step {
 	 * Process the optin data.
 	 *
 	 * @since 6.8.4
+	 * @since TBD Change default opt-in status to null.
```
Camwyn:
```suggestion
	 * @since TBD Change default opt-in status to null to prevent false positives in checks when the value is not set.
```

Camwyn (changes requested):
Docblock change.

dpanta94 (approved):

src/Events/Admin/Onboarding/Steps/Optin.php
```diff
@@ -35,14 +35,15 @@ class Optin extends Abstract_Step {
...
 	 * @param WP_REST_Request  $request  The request object.
 	 *
 	 * @return WP_REST_Response
 	 */
 	public function process( $response, $request ): WP_REST_Response {
-		$current_optin = tribe_get_option( 'opt-in-status', false );
+		$current_optin = tribe_get_option( 'opt-in-status', null );
 		$optin         = $request->get_param( 'optin' );
 
 		if ( $current_optin === $optin ) {
```
rafsuntaskin, line 49: Question: In the case of both `current_optin` and `optin` as `null` we are returning here (Line 49). Is this expected? We are not saving anything but the message will say saved as requested.

Camwyn: `optin` comes from the wizard and will always be a bool. We can cast it here to be sure though.

(edit) Done.

src/Events/Admin/Onboarding/Steps/Optin.php
```diff
@@ -35,14 +35,15 @@ class Optin extends Abstract_Step {
...
 	 *
 	 * @param WP_REST_Response $response The response object.
 	 * @param WP_REST_Request  $request  The request object.
 	 *
 	 * @return WP_REST_Response
 	 */
 	public function process( $response, $request ): WP_REST_Response {
-		$current_optin = tribe_get_option( 'opt-in-status', false );
+		$current_optin = tribe_get_option( 'opt-in-status', null );
 		$optin         = $request->get_param( 'optin' );
```
Camwyn:
```suggestion
		$optin         = (bool) $request->get_param( 'optin' );
```

PR #5408, opened 2025-10-02
[TEC-5531] - Fixed incorrect heading usage to improve A11y.

rafsuntaskin:
### 🎫 Ticket

[TEC-5531]

### 🗒️ Description
* Event Date time block now uses `p` instead of `h2` tag to improve a11y.
* With the new `p` tag, we don't have the extra margin like we had before; this looks better, so we didn't force that old margin with `p`. All other styles are kept same.

### 🎥 Artifacts 
📷 screenshot(s): 
Before: https://share.cleanshot.com/9kHl4cGL
After: https://share.cleanshot.com/XqVTsDsQ

After css class style update: https://share.cleanshot.com/rNBm28m1

### ✔️ Checklist

[TEC-5531]: https://stellarwp.atlassian.net/browse/TEC-5531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/resources/postcss/views/skeleton/_single-event.pcss
+++ b/src/resources/postcss/views/skeleton/_single-event.pcss
@@ -82,15 +82,17 @@
 }
 
 /* Datetime block */
-.tribe-events-schedule p {
-	margin-bottom: var(--tec-spacer-1);
-	margin-right: var(--tec-spacer-1);
-	order: 0;
-	width: 100%;
+.tribe-events-schedule {
+	.tribe-events-schedule__datetime {
+		margin-bottom: var(--tec-spacer-1);
+		margin-right: var(--tec-spacer-1);
+		order: 0;
+		width: 100%;
 
-	@media (--viewport-medium) {
-		margin-bottom: 0;
-		width: auto;
+		@media (--viewport-medium) {
+			margin-bottom: 0;
+			width: auto;
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


src/views/blocks/event-datetime.php
```diff
@@ -75,7 +75,7 @@
 $classes = isset( $attributes['className'] ) ? array_merge( $default_classes, [ $attributes['className'] ] ) : $default_classes;
 ?>
 <div <?php tec_classes( $classes ); ?>>
-	<h2 class="tribe-events-schedule__datetime">
+	<p class="tribe-events-schedule__datetime">
```
Camwyn, line 78: Can we target the class instead of the tag?
(In the CSS)

Camwyn (changes requested):

src/resources/postcss/views/skeleton/_single-event.pcss
```diff
@@ -562,11 +564,6 @@
  * ----------------------------------------------------------------------------- */
 
 .tribe-theme-twentytwenty {
-
-	.tribe-events-schedule h2 {
-		margin-left: 0;
-	}
```
rafsuntaskin, line 568: This was a h2 reset, not needed any more.

PR #5637, opened 2026-05-27
Merge/rsvp merge project

larodiel:
## Summary

The main purpose of this PR is to bring the latest changes from `main` into the `bucket/rsvp-merge` integration branch (via the `merge/rsvp-merge-project` sub-branch), so the RSVP merge work stays current with `main`.

- Merges latest `main` into the `bucket/rsvp-merge` integration bucket via the `merge/rsvp-merge-project` sub-branch
- Common submodule moved to the `merge/rsvp-merge-project` branch at `2feb50f259`
- Bumps outer `package-lock.json` version to match main

## Changes we needed to make to get the tests passing

Pulling `main` in surfaced a few places where this branch and the newer shared `tribe-common` library were out of sync, which broke CI. The fixes below are small and targeted — each one maps to a specific failing check, with no unrelated changes. The only behavior changes are the two bug fixes called out (QR decode and the event `dates` field).

- **QR code link decoding** — `src/Events/QR/Routes.php`
  A QR link is stored as `id:type:salt`. The security salt can itself contain a `:`, which sometimes split the value into too many pieces and made decoding fail with "Invalid QR code data format." We now split into exactly three parts so a salt with a colon stays intact. _(fixes the `integration` suite)_

- **Event REST API `dates` field** — `src/Tribe/Models/Post_Types/Event.php`
  The newer `tribe-common` changed the rule for which event fields get attached to REST API responses: a field is now only included if it's marked `true` (or given an alias name). The event's `dates` field was marked with an object instead of `true`, so it was being silently dropped from API responses. Marking it `true` (like every other field) restores it. _(fixes the `rest_tec_v1_integration` suite)_

- **Action Scheduler loader** — `src/Events/Custom_Tables/V1/Migration/Provider.php`
  The newer `tribe-common` relocates the bundled Action Scheduler into a "prefixed" vendor folder. A migration helper was still pointing at the old location and erroring out. We updated it to look in the new location, with a fallback to the old path. _(fixes the `ct1_migration` suite)_

- **Coding-standards (phpcs) cleanups** — `src/Tribe/REST/V1/Endpoints/Single_Event.php`, `src/Tribe/Views/V2/Url.php`
  Small lint fixes on code already in this PR: corrected the order of a docblock `@param`, switched `parse_url()` to `wp_parse_url()`, and reformatted a multi-line function call. No behavior change. _(fixes the `phpcs` check)_


src/Events/Custom_Tables/V1/Migration/Provider.php
```diff
@@ -302,8 +302,25 @@ public function add_phase_callback( $upgrade_fields ) {
...
+			$common_vendor . '/vendor-prefixed/woocommerce/action-scheduler/action-scheduler.php',
+			$common_vendor . '/woocommerce/action-scheduler/action-scheduler.php',
+		];
+
+		foreach ( $candidates as $action_scheduler_file ) {
+			if ( file_exists( $action_scheduler_file ) ) {
+				require_once $action_scheduler_file;
+				return;
+			}
+		}
```
dpanta94, line 323: I think it should always be loaded by Shepherd in common. So maybe we simoply depracate this method ?

dpanta94 (changes requested):

PR #5645, opened 2026-06-01
Fix bucket/rsvp-merge CI failures after common bump

sdokus:
[skip-changelog]

## Summary

Three targeted fixes for CI breakages that surfaced on `bucket/rsvp-merge` once the newer shared `tribe-common` library was pulled in. Each maps to a specific failing suite; no unrelated changes.

## Fixes

- **QR code decode** — `src/Events/QR/Routes.php`
  A QR link is stored as `id:type:salt`. The security salt can itself contain a `:`, which over-split the value and made decoding fail with "Invalid QR code data format." We now limit the split to exactly three parts so a salt with a colon stays intact. _Fixes the `integration` suite (3 errors)._

- **Event REST API `dates` field** — `src/Tribe/Models/Post_Types/Event.php`
  The newer `tribe-common` only attaches an event property to REST responses when it's marked `true` (or given an alias). The `dates` field was marked with an object instead of `true`, so it was silently dropped. Marking it `true` like every other field restores it. _Fixes the `rest_tec_v1_integration` suite._

- **Action Scheduler late-load** — `src/Events/Custom_Tables/V1/Migration/Provider.php`
  The already-deprecated `load_action_scheduler_late()` still `require_once`'d Action Scheduler from the old `common/vendor/woocommerce/...` path, which no longer exists after Strauss prefixing — causing a fatal. Per Dimi's review feedback on #5637, Action Scheduler is now **always loaded by Shepherd in Common**, so the method is simply deprecated with no load. _Fixes the `ct1_migration` suite._


src/Tribe/Models/Post_Types/Event.php
```diff
@@ -414,16 +414,7 @@ public static function get_properties_to_add(): array {
...
-					'start'         => true,
-					'start_utc'     => true,
-					'start_site'    => true,
-					'start_display' => true,
-					'end'           => true,
-					'end_utc'       => true,
-					'end_site'      => true,
-					'end_display'   => true,
-				],
+				'dates'                  => true,
```
dpanta94, line 417: Im curious for this. Why doesn't the old format work anymore ?

sdokus: The old nested-object value was actually being ignored the whole time. Before the alias change, `add_properties_to_model()` used `array_intersect_key( $original_post, $properties_to_add )`, which matches on the map's *keys* only and never looks at the values — so any value next to `dates` worked, including the `(object)`. Luca's alias commit (`9dfac3d00`, "feat: support field aliases in the TEC REST API") changed the loop to branch on the *value*: it now only accepts a string (alias) or strict `=== true`, and silently skips anything else — so the object value drops the field. Marking it `true` just puts it back on the supported path; the nested structure was never read by any version of the consumer, so nothing is lost.

dpanta94 (approved):

PR #5660, opened 2026-06-22
Update WP version min to 6.8

redscar:
Bump WP version min to 6.8
[skip-changelog][skip-phpcs]

### DIFF
--- a/.github/workflows/tests-php.yml
+++ b/.github/workflows/tests-php.yml
@@ -137,7 +137,7 @@ jobs:
         run: |
           ${SLIC_BIN} up wordpress
           ${SLIC_BIN} wp core version
-          ${SLIC_BIN} wp core update --force --version=6.9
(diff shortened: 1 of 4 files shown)


.github/workflows/tests-php.yml
```diff
@@ -137,7 +137,7 @@ jobs:
         run: |
           ${SLIC_BIN} up wordpress
           ${SLIC_BIN} wp core version
-          ${SLIC_BIN} wp core update --force --version=6.8
+          ${SLIC_BIN} wp core update --force --version=6.9
```
sdokus:
```suggestion
          ${SLIC_BIN} wp core update --force --version=6.8
```

This is probably why that test is failing, we already bumped the version here, even if we didn't update the readme.txt file

sdokus (approved):

PR #5665, opened 2026-06-24
[SMTNC-1646] PHP Compatibility improvements

dpanta94:
### 🎫 Ticket

SMTNC-1646

### 🗒️ Description

Improves php compatibility for php version from 7.4 up to 8.4.

Breaking contract by adding explicitly nullable `?` is not a breaking change.

PHPCompatibility and PHPCompatibilityWP are included via the `sh` script because the latest stable version of PHPCompatibilityWP does not cover 8+ compatibility concerns.

### 🎥 Artifacts 
Tests continue passing are part of the artifact.

Test ensuring that the cotract being broken is not considered a breaking change.

CleanShot 2026-06-25 at 00 33 40

### ✔️ Checklist

### DIFF
--- a/.github/workflows/php-compatibility.yml
+++ b/.github/workflows/php-compatibility.yml
@@ -1,0 +1,39 @@
+name: 'PHP Compatibility'
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: php-compat-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  php-compat:
+    # Allow opting out on a specific PR with [skip-php-compat] in the description.
+    if: "!contains(github.event.pull_request.body, '[skip-php-compat]')"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.3'
+          coverage: none
+          tools: composer
+
+      # Cache the downloaded phpcs phar + PHPCompatibility/PHPCSUtils standards.
+      # Key is bound to bin/php-compat.sh, so bumping the pinned versions there
+      # automatically invalidates the cache.
+      - name: Cache PHP compatibility toolchain
+        uses: actions/cache@v4
+        with:
+          path: dev/php-compat
+          key: php-compat-${{ runner.os }}-${{ hashFiles('bin/php-compat.sh') }}
+
+      # Runs `bash bin/php-compat.sh`, which scans the paths in phpcompat.xml.dist
+      # against testVersion 8.0-8.4 and exits non-zero on any finding (hard gate).
+      - name: Run PHP 8 compatibility scan
+        run: composer compat
--- a/.github/workflows/php-compatibility.yml
+++ b/.github/workflows/php-compatibility.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
--- a/src/Events/Custom_Tables/V1/Migration/CSV_Report/File_Download.php
+++ b/src/Events/Custom_Tables/V1/Migration/CSV_Report/File_Download.php
@@ -105,7 +105,7 @@ class File_Download {
 		header( "Cache-Control: no-cache, must-revalidate" );
 		header( "Expires: Sat, 26 Jul 1997 05:00:00 GMT" );
 
-		fputcsv( $output, self::CSV_COLUMNS, $delimiter );
+		fputcsv( $output, self::CSV_COLUMNS, $delimiter, '"', "\\");
 		foreach ( $reports as $report ) {
 			$has_error = (bool) $report->error;
 			if ( $has_error ) {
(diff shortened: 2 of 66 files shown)


src/Tribe/Importer/File_Reader.php
```diff
@@ -17,7 +17,6 @@ class Tribe__Events__Importer__File_Reader {
 	 * @param string $file_path The full path to the file.
 	 */
 	public function __construct( $file_path ) {
-		ini_set( 'auto_detect_line_endings', true );
```
defunctl, line 20: Flagging this as you must now manually normalize Mac CSV line endings that use `\r` only, e.g. `str_replace(["\r\n", "\r"], "\n", …)` like this setting used to do automatically.

src/Events/Custom_Tables/V1/Migration/CSV_Report/File_Download.php
```diff
@@ -105,7 +105,7 @@ public function download_csv( $should_exit = true ) {
 		header( "Cache-Control: no-cache, must-revalidate" );
 		header( "Expires: Sat, 26 Jul 1997 05:00:00 GMT" );
 
-		fputcsv( $output, self::CSV_COLUMNS, $delimiter );
+		fputcsv( $output, self::CSV_COLUMNS, $delimiter, '"', '\\' ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fputcsv
```
defunctl, line 108: I actually think the `\\` should be left out?

https://www.php.net/manual/en/function.fputcsv.php

Image

src/Events/Custom_Tables/V1/Migration/CSV_Report/File_Download.php
```diff
@@ -121,11 +121,11 @@ public function download_csv( $should_exit = true ) {
 				$has_error ? "Yes" : "No"
 			];
 
-			fputcsv( $output, $item, $delimiter );
+			fputcsv( $output, $item, $delimiter, '"', '\\' ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fputcsv
```
defunctl, line 124: Same as above

src/Events/Custom_Tables/V1/WP_Query/Modifiers/Events_Admin_List_Modifier.php
```diff
@@ -24,7 +24,7 @@ class Events_Admin_List_Modifier extends Base_Modifier {
 	/**
 	 * {@inheritDoc}
 	 */
-	public function applies_to( WP_Query $query = null ) {
+	public function applies_to( ?WP_Query $query = null ) {
```
defunctl, line 29: This needs a `null` guard check for `$query`, if it comes in as `null` this would fatal.

src/Tribe/Views/V2/View.php
```diff
@@ -1331,12 +1331,13 @@ protected function filter_template_vars( array $template_vars ) {
 	 * @since 4.9.3
 	 * @since 6.0.5 Now will merge a "global" repository arg filter, which will be applied elsewhere as well as this
 	 *                primary repository query.
+	 * @since TBD Made $context explicty nullable.
```
defunctl:
```suggestion
	 * @since TBD Made $context explicity nullable.
```

This typo appears elsewhere as well

src/Events/Custom_Tables/V1/WP_Query/Provider.php
```diff
@@ -27,7 +27,7 @@
  *
  * @package TEC\Events\Custom_Tables\V1\WP_Query
  */
-class Provider extends Service_Provider implements Serializable, Provider_Contract {
+class Provider extends Service_Provider implements Provider_Contract {
```
defunctl, line 29: By removing this interface, I think you should be deprecating the [serialize/unserialize methods](https://github.com/the-events-calendar/the-events-calendar/blob/a46bf6b962036fe5cb6027d672b337e668ece017/src/Events/Custom_Tables/V1/WP_Query/Provider.php#L96-L120) in addition to making them call the underlying modern/magic [__serialize/__unserialize methods](https://github.com/the-events-calendar/the-events-calendar/blob/a46bf6b962036fe5cb6027d672b337e668ece017/src/Events/Custom_Tables/V1/WP_Query/Provider.php#L122-L146)

defunctl (changes requested):
A few things need addressing here and I think a bunch of classes/methods are missing the `@TBD` docblock.

src/Tribe/Views/V2/View.php
dpanta94: I found it nowhere else.

src/Tribe/Importer/File_Reader.php
dpanta94: Not doing that was because my findings were that the `\r` file ending in MAC has been deprecated since 2002 and you would actually need to use Excel's 2008 version to have that issue. I feel like a 18 year old problem shouldn't concern us here.

src/Events/Custom_Tables/V1/Migration/CSV_Report/File_Download.php
dpanta94: Well, i'd like us to think it through. You are correct that providing an empty string is the better way producing RFC 4180 CSVs. Thats great.

But it would also change the bytes for fields containing backslashes or embedded quotes. Could that be a bad experience maybe ? Should we worry about it ? i feel like no, but i'd also like you to think through.

Providing `'\\'` ensures that we continue doing what we have been doing already. That was my choice here.

.github/workflows/php-compatibility.yml
```diff
@@ -0,0 +1,39 @@
...
+  cancel-in-progress: true
+
+jobs:
+  php-compat:
+    # Allow opting out on a specific PR with [skip-php-compat] in the description.
+    if: "!contains(github.event.pull_request.body, '[skip-php-compat]')"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
```
estevao90:
```suggestion
        uses: actions/checkout@v6
```

PR #5697, opened 2026-07-16
[SMTNC-267] Event times render in UTC instead of site timezone

WayneRocha:
### 🎫 Ticket

[SMTNC-267](https://linear.app/nexcess/issue/SMTNC-267/woocommerce-enhanced-templates-emails-render-event-times-in-utc)

### Commom part

https://github.com/the-events-calendar/tribe-common/pull/2948

### 🗒️ Description

When talking about timezone in this context we have...
- **WP Timezone** - That we set on admin panel and is native from WP.
- **Server-timezone** - Normally configured by the host, in the php.ini or by `date_default_timezone_set()`

Default PHP functions "guess the timezone" based on server-timezone and in our case TEC expects it to be UTC, but when it's not it shifts the timezone twice, which give us a incorrect result.

*Solution here is to not rely on `strtotime` and use UTC timezone.*

### 🧪 Testing Instructions

In a clean install you probably will not be able to reproduce it. So create a 1line plugin with `date_default_timezone_set("America/Chicago");`. That way you can mimic this behavior.

To undertand the timezone example. try this:

```bash
lando wp --path=wordpress eval 'echo strip_tags( tribe_events_event_schedule_details( EVENT_ID ) ) . "\n";'
# → July 22 @ 11:30 am - 1:00 pm  (PHP tz = UTC)

lando wp --path=wordpress eval 'date_default_timezone_set("America/Chicago"); echo strip_tags( tribe_events_event_schedule_details( EVENT_ID ) ) . "\n";'
# → July 22 @ 4:30 pm - 6:00 pm   (PHP tz != UTC → BUG)
```

---

Helper plugin:

```php
<?php
/**
 * Plugin Name: SMTNC-267 Repro - TZ override
 * Description: Simulates a plugin/snippet that sets PHP default timezone to the site timezone (like the customer's workaround).
 */
date_default_timezone_set( 'America/Chicago' );
```

### 🎥 Artifacts 

https://www.loom.com/share/af125ff0bba4415b8493beb6b9c6c151

### ✔️ Checklist

### DIFF
--- a/changelog/smtnc-267-event-times-render-in-utc
+++ b/changelog/smtnc-267-event-times-render-in-utc
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Resolved an issue where event times rendered in UTC on WooCommerce order pages and emails when PHP's default timezone was changed by third-party code.
--- a/src/Tribe/Timezones.php
+++ b/src/Tribe/Timezones.php
@@ -135,6 +135,12 @@ class Tribe__Events__Timezones extends Tribe__Timezones {
 
 		$timestamps = tribe_get_var( $cache_var_name, [] );
 
+		if ( null === $timezone ) {
+			$timezone = self::mode();
+		}
+
+		// Build the cache key after resolving the timezone mode so `null` and the
+		// resolved mode string share the same cache entry.
 		$cache_key = "{$event_id}:{$type}:{$timezone}";
 
 		if ( isset( $timestamps[ $cache_key ] ) ) {
@@ -144,10 +150,6 @@ class Tribe__Events__Timezones extends Tribe__Timezones {
 		$event    = get_post( Tribe__Events__Main::postIdHelper( $event_id ) );
 		$event_tz = get_post_meta( $event->ID, '_EventTimezone', true );
 		$site_tz  = self::wp_timezone_string();
-
-		if ( null === $timezone ) {
-			$timezone = self::mode();
-		}
 
 		// Should we use the event specific timezone or the site-wide timezone?
 		$use_event_tz = self::EVENT_TIMEZONE === $timezone;
@@ -161,7 +163,11 @@ class Tribe__Events__Timezones extends Tribe__Timezones {
 		if ( $use_event_tz || ( $use_site_tz && $site_zone_is_event_zone ) ) {
 			$datetime = get_post_meta( $event->ID, "_Event{$type}Date", true );
 
-			return $timestamps[ $cache_key ] = strtotime( $datetime );
+			$timestamps[ $cache_key ] = self::wall_clock_timestamp( $datetime );
+
+			tribe_set_var( $cache_var_name, $timestamps );
+
+			return $timestamps[ $cache_key ];
 		}
 
 		// Otherwise lets load the event's UTC time and convert it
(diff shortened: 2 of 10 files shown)


src/Tribe/Timezones.php
```diff
@@ -161,7 +163,9 @@ protected static function get_event_timestamp( $event_id, $type = 'Start', $time
 		if ( $use_event_tz || ( $use_site_tz && $site_zone_is_event_zone ) ) {
 			$datetime = get_post_meta( $event->ID, "_Event{$type}Date", true );
 
-			return $timestamps[ $cache_key ] = strtotime( $datetime );
+			$timestamps[ $cache_key ] = self::wall_clock_timestamp( $datetime );
+
+			return $timestamps[ $cache_key ];
```
coutinhocouto, line 168: This branch populates $timestamps[ $cache_key ] but never calls tribe_set_var( $cache_var_name, $timestamps ) before returning, so the event‑timezone‑mode result isn't cached for the rest of the request. Pre‑existing, but worth adding the tribe_set_var() call here now that these lines are being touched.

src/functions/template-tags/general.php
```diff
@@ -1319,11 +1319,14 @@ function tribe_events_event_schedule_details( $event = null, $before = '', $afte
 				$inner .= ( $html ? '</span>' : '' ) . $time_range_separator;
 				$inner .= $html ? '<span class="tribe-event-date-end">' : '';
 
-				$end_date_full           = tribe_get_end_date( $event, true, Tribe__Date_Utils::DBDATETIMEFORMAT );
-				$end_date_full_timestamp = strtotime( $end_date_full );
+					$end_date_full = tribe_get_end_date( $event, true, Tribe__Date_Utils::DBDATETIMEFORMAT );
+
+				// Use UTC so the wall-clock timestamps remain independent of PHP's default timezone.
+				$end_date_full_timestamp    = Tribe__Date_Utils::build_date_object( $end_date_full, 'UTC' )->getTimestamp();
+				$beginning_of_day_timestamp = Tribe__Date_Utils::build_date_object( tribe_beginning_of_day( $end_date_full ), 'UTC' )->getTimestamp();
```
coutinhocouto: For consistency with Tribe__Events__Timezones::wall_clock_timestamp() and the new tribe_format_date() guard, consider matching their build_date_object( …, 'UTC', false ) + strtotime() fallback here instead of relying on the default "now" fallback — otherwise a malformed end date silently renders as today rather than degrading to the legacy behavior.

coutinhocouto (changes requested):
Just two minor comments, then is good to go for me :)

src/Tribe/Timezones.php
WayneRocha: @coutinhocouto You can add it if you can manage to make [tests](https://github.com/the-events-calendar/the-events-calendar/actions/runs/29513555658/job/87672912656?pr=5697) run. With this, data will leak and tests will fail. I removed it intentionally.

src/functions/template-tags/general.php
WayneRocha: 0feb70133d25cfcaf72e8ceea4b6cbfe342ba00e

coutinhocouto (approved):

changelog/smtnc-267-event-times-render-in-utc
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Resolved an issue where event times rendered in UTC on WooCommerce order pages and emails when PHP's default timezone was changed by third-party code.
```
sdokus, line 4: Can we combine these two changelogs? Or are these two separate issues being combined in one PR/ticket? 🤔

WayneRocha: Yes, when digging it, I had to solve this minor issue along with the main problem @sdokus

sdokus (approved):
Just a small heads up, let's make sure that we get the merge order correct - https://github.com/the-events-calendar/tribe-common/pull/2948 should be merged first and then we should double check that the common changes here aren't moving the SHA pointer away from the `release/M26.kadabra` branch in tribe-common

WayneRocha:
Oh, thanks for the tip @sdokus !

PR #5608, opened 2026-04-07
Promoter - Fix TEC REST endpoints to give better errors

sdokus:
## Summary

Cherry-pick of #5540 targeting `release/M26.feraligatr`.

- Fixes incorrect HTTP status codes for the `/wp-json/tribe/events/v1/events/{id}` GET endpoint (non-existent events now return 404 instead of 400)
- Improves error messages for trashed, draft, private, and password-protected events
- Adds manual validation for event ID parameters in archive endpoints

### 🎫 Ticket

[[TEC-5737](https://stellarwp.atlassian.net/browse/TEC-5737)]

## Test plan

- [x] Verify non-existent events return 404 with descriptive message
- [x] Verify trashed events return 403 with "The event is trashed"
- [x] Verify draft/private/pending events return 403 with "not accessible"
- [x] Verify password-protected events return 403 with password message
- [x] Verify published events still return 200 normally

### Artifacts
CleanShot 2026-04-08 at 10 50 22@2x
CleanShot 2026-04-08 at 10 56 31@2x
CleanShot 2026-04-08 at 10 57 41@2x

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[TEC-5737]: https://stellarwp.atlassian.net/browse/TEC-5737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5698, opened 2026-07-16
Promoter - Fix TEC REST endpoints to give better errors

sdokus:
## Summary

Fixes [SMTNC-249](https://linear.app/nexcess/issue/SMTNC-249/rest-api-giving-wrong-responses): the REST API single-event endpoint currently returns a generic `rest_invalid_param` / 400 for deleted, trashed, draft, pending, private, and password-protected events, which breaks Promoter's error handling and causes unwarranted soft-blocking.

This is a rebase of the previously-approved, unmerged work from #5540 / #5608 (originally targeting release branches that have since shipped), plus two small follow-ups:
- The trashed-event error now uses `event-no-longer-available` / "This event is no longer available" to match the SMTNC-249 spec exactly (the original branches used `event-is-trashed`).
- Changelog entry no longer references a specific ticket number, per current changelog procedure.

`GET /events/{id}` now returns:
- 404 `rest-event-not-found` for permanently deleted/non-existent events
- 403 `event-no-longer-available` for trashed events
- 403 `event-not-accessible` for draft/pending/private events (unchanged)
- 403 `event-password-protected` for password-protected events (unchanged)

Archive endpoints (`Archive_Event`, `Archive_Organizer`, `Archive_Venue`, `Term_Archive_Base`) now validate event ID params and return proper 400s for malformed/non-existent IDs instead of surfacing errors downstream.

`is_event_id()` behavior and the public `tribe_events_validator_is_event_id` filter signature are unchanged (no breaking change); a new `is_event_id_format()` method handles format-only validation for the REST `id` param's `validate_callback`.

## Artifact
https://www.loom.com/share/94716b0f53e544bda7a1d5cf803223d0

### DIFF
--- a/changelog/fix-rest-api-validation
+++ b/changelog/fix-rest-api-validation
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Improve REST API error handling by providing specific error messages for trashed events and ensuring invalid or non-existent event ID parameters in archive endpoints return proper 400 errors.
--- a/changelog/fix-rest-api-validation
+++ b/changelog/fix-rest-api-validation
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Improve REST API error handling by providing specific error messages for trashed events and ensuring invalid or non-existent event ID parameters in archive endpoints return proper 400 errors.
+Improve REST API error handling by providing specific error messages for trashed events and ensuring invalid or non-existent event ID parameters in archive endpoints return proper 400 errors. [TEC-5737]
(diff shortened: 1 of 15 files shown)


redscar (approved):
So excited to see this being implemented. This should fix those pesky Promoter issues. 

And since I know that Dimi will say we should switch to the newer API (and he's right)- as much as I would love to it was a bit task to even try to.

sdokus:
> _And since I know that Dimi will say we should switch to the newer API (and he's right)- as much as I would love to it was a bit task to even try to._

I agree, that's the more technically correct approach, but this will at least hold us over until we can get to that 🤞

PR #5701, opened 2026-07-17
[SMTNC-1797] fix: _load_textdomain_just_in_time notice bug

larodiel:
### 🎫 Ticket

[SMTNC-1797](https://linear.app/nexcess/issue/SMTNC-1797/global-load-textdomain-notice-flooding-logs)

### 🗒️ Description

We are having a bug with `_load_textdomain_just_in_time` being triggered when some plugins are installed with TEC.
This fix address to prevent the user log to be flooded by these notice messages.

### 🎥 Artifacts 
https://www.loom.com/share/0c3a7f49e2724b2182b52f02e0ddf404

### ✔️ Checklist

### DIFF
--- a/changelog/fix-smtnc-1797-global-load_textdomain-notice-flooding-logs
+++ b/changelog/fix-smtnc-1797-global-load_textdomain-notice-flooding-logs
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Resolved an issue where the _load_textdomain_just_in_time notice was triggered because the cron schedule label was translated before the init hook fired.
--- a/changelog/fix-smtnc-1797-global-load_textdomain-notice-flooding-logs
+++ b/changelog/fix-smtnc-1797-global-load_textdomain-notice-flooding-logs
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Resolved an issue where the `_load_textdomain_just_in_time` notice could be triggered on the site.
+Resolved an issue where the `_load_textdomain_just_in_time` notice could be triggered on the site by deferring registration of the Event Aggregator `cron_schedules` filter until the `init` hook fires.
--- a/changelog/fix-smtnc-1797-global-load_textdomain-notice-flooding-logs
+++ b/changelog/fix-smtnc-1797-global-load_textdomain-notice-flooding-logs
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Resolved an issue where the _load_textdomain_just_in_time notice was triggered because the cron schedule label was translated before the init hook fired.
+Resolved an issue where the `_load_textdomain_just_in_time` notice could be triggered on the site.
--- a/src/Tribe/Aggregator/Cron.php
+++ b/src/Tribe/Aggregator/Cron.php
@@ -206,10 +206,14 @@ class Tribe__Events__Aggregator__Cron {
 		// Ensure schedules is an array.
 		$schedules = is_array( $schedules ) ? $schedules : [];
 
+		$display = ( did_action( 'init' ) || doing_action( 'init' ) )
+			? esc_html_x( 'Every 15 minutes', 'aggregator schedule frequency', 'the-events-calendar' )
+			: 'Every 15 minutes'; // fallback, avoids translating before init
+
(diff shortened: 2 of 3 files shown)


changelog/fix-smtnc-1797-global-load_textdomain-notice-flooding-logs
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Resolved an issue where the _load_textdomain_just_in_time notice was triggered because the cron schedule label was translated before the init hook fired.
```
estevao90: I'd be more generic here, but not sure about how we deal with it in TEC.


```suggestion
Resolved an issue where the `_load_textdomain_just_in_time` notice could be triggered on the site.
```

src/Tribe/Aggregator/Cron.php
```diff
@@ -206,10 +206,14 @@ public function filter_add_cron_schedules( $schedules = [] ) {
 		// Ensure schedules is an array.
 		$schedules = is_array( $schedules ) ? $schedules : [];
 
+		$display = ( did_action( 'init' ) || doing_action( 'init' ) )
+			? esc_html_x( 'Every 15 minutes', 'aggregator schedule frequency', 'the-events-calendar' )
```
estevao90: I think you should investigate whether we can make this filter called after `init`. Not offering a translation in some cases feels wrong unless we don't have any other option.

estevao90 (commented):
It looks good to me, but tests are not passing.

tests/wpunit/Tribe/Events/Aggregator/CronTest.php
```diff
@@ -576,6 +598,72 @@ public function purge_expired_records_max_allowed_packet_provider(): array {
...
+	 */
+	public function should_defer_cron_schedules_filter_registration_until_init() {
+		$cron = $this->make_real_instance();
+
+		$this->assertFalse(
+			has_filter( 'cron_schedules', [ $cron, 'filter_add_cron_schedules' ] ),
+			'The cron_schedules filter should not be registered before init has fired.'
+		);
+
+		do_action( 'init' );
```
d4mation: Something to be aware of with tests like this:

If you write `do_action( 'init' )` within a test, _anything_ you have attached to `init` will run. While that's how it works in a production environment, when you consider that WordPress has already been initialized by `wp-browser` it can get weird and you could end up running code multiple times doing this.

You may want to find a way to mock the action being ran instead to avoid triggering unexpected code.

larodiel: Ty fo the tip @d4mation it helped to fix the tests that were not passing

tests/wpunit/Tribe/Events/Aggregator/CronTest.php
```diff
@@ -576,6 +598,91 @@ public function purge_expired_records_max_allowed_packet_provider(): array {
...
+	public function should_defer_cron_schedules_filter_registration_until_init() {
+		$cron = $this->make_real_instance();
+
+		$this->assertFalse(
+			has_filter( 'cron_schedules', [ $cron, 'filter_add_cron_schedules' ] ),
+			'The cron_schedules filter should not be registered before init has fired.'
+		);
+
+		// Simulate `init` firing for this hook, without triggering the global `init` action.
+		$cron->register_cron_schedules_filter();
```
d4mation, line 631: This unfortunately doesn't result in us properly testing that it is attached at `init`, which for this specific fix is actually pretty important. A `has_action()` assertion for that would probably be a good idea to help ensure it doesn't drift in a future update. You could also make sure the error message on a failed test states _why_ it has to be hooked at `init`.

Testing to see if `_doing_it_wrong()` was called (with a message matching the one from [`_load_textdomain_just_in_time()`](https://developer.wordpress.org/reference/functions/_load_textdomain_just_in_time/)) may also be a good idea, but that could get a bit messy with how to _accurately_ test for it you'd need to ensure that `did_action( 'init' )` returns `false` until you fake run the `init` hook and I don't know if it is worth it here.

d4mation (approved):

estevao90 (approved):

nikolaystrikhar (approved):

PR #5703, opened 2026-07-21
Redirect to the guided setup wizard on single-site activation

dave-green-uk:
## Summary

When The Events Calendar is activated on its own, it now redirects to the **Guided Setup** wizard (`first-time-setup`), the way sister plugins like LearnDash behave on a new install. Bulk plugin activations are deliberately left alone.

https://www.loom.com/share/20921ceb62324738ac56bf6adc4c4a46

Previously the activation-redirect path was dead: `Tribe__Events__Main::activate()` still set the `_tribe_events_activation_redirect` transient, but its only consumer (`activation_page()` → `Tribe__Admin__Activation_Page`) was deprecated in 6.8.2 and is never called. The modern redirect only fired when a user *navigated* to a TEC admin page. This reconnects a consumer to that existing transient — no new transient, no change to the activation callback beyond removing a dead guard.

🎫 SMTNC-1826

## Behavior

| Scenario | Result |
|---|---|
| Activate TEC on its own (fresh install) | → redirect to Guided Setup |
| Bulk-activate TEC with other plugins | stays on the Plugins screen |
| Already visited / dismissed the setup screen | no redirect |
| Upgraded install (more than one recorded version) | no redirect |
| User lacks the page capability | no redirect |

### On bulk activation
The `activate-multi` guard belongs in the **consumer**, not the activation hook: WordPress only appends `activate-multi` to the Plugins-screen URL on the request *after* activation, so the flag isn't set while the activation hook runs. The consumer checks it on that next load, where it's actually present.

## What's in here (7 commits)

- **refactor:** extract `should_redirect_to_guided_setup()` + `get_guided_setup_url()` so the on-visit and on-activation paths share one guard. No behavior change.
- **feat:** `maybe_redirect_to_guided_setup_on_activation()` + hook wiring + integration tests.
- **fix:** honor a dismissed setup screen in the redirect. `Landing_Page` opts out of `Abstract_Admin_Page::$is_dismissible`, so `is_dismissed()` always returned false; the guard now reads the dismiss option directly — **without** hiding the Setup Guide menu item (deliberate: dismiss stops the redirect only).
- **tweak:** drop the no-op `activate-multi` guard from `Main::activate()` (see above).
- **test:** on-visit characterization coverage + a cross-test option-cache isolation fix (see Testing).
- **tweak:** changelog wording.
- **tweak:** read the `activate-multi` flag through `tec_get_request_var()` instead of a raw `$_GET` + `phpcs:ignore` (review feedback from @bordoni). Same presence check, one less local ignore.

## Testing

Integration suite via `slic`: **18 tests, 38 assertions — green, deterministic across runs.**

- Activation redirect: fresh single activation redirects; bulk / no-transient / already-visited / dismissed / upgraded-install do not. Bulk guard mutation-tested (breaking it fails the test).
- On-visit redirect characterized: slug gate, post-type gate, visited guard, force-filter bypass, happy path.
- **Isolation fix:** adding the on-visit test surfaced a real cross-test leak — the tribe options sit in both the WP object cache and a tribe var, neither rolled back with the DB, so `previous_ecp_versions` bled between test files (the activation suite was passing only by test ordering). Fixed with `wp_cache_flush()` + `tribe_unset_var()` in both files' setup.

## Notes / follow-ups (not in this PR)

- `Landing_Page::is_dismissed()` is inert because the page doesn't override `$is_dismissible` (defaults false) — so the base-class menu gate never treats it as dismissed. Left as-is by design here (we did not want dismiss to remove the menu item). Worth a separate decision.
- A StellarWP phpcs sniff (`StellarWP.Whitespace.DocCommentSpacing`) crashes on a pre-existing `@see` docblock in `Main.php` — unrelated, but it breaks full-file phpcs runs on that file.

Changelog entries: `feat` (activation redirect) + `fix` (dismiss).

Ref: SMTNC-1826

### DIFF
--- a/src/Events/Admin/Onboarding/Controller.php
+++ b/src/Events/Admin/Onboarding/Controller.php
@@ -213,8 +213,7 @@ class Controller extends Controller_Contract {
 
 		// Do not hijack a bulk plugin activation. WordPress adds `activate-multi` to the Plugins
 		// screen URL after several plugins are activated together, so the user stays in context.
-		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		if ( isset( $_GET['activate-multi'] ) ) {
+		if ( null !== tec_get_request_var( 'activate-multi' ) ) {
 			return;
 		}
 
--- a/src/Events/Admin/Onboarding/Controller.php
+++ b/src/Events/Admin/Onboarding/Controller.php
@@ -88,6 +88,7 @@ class Controller extends Controller_Contract {
 	 *
 	 * @since 6.8.4
 	 * @since 6.11.0 Changed the priority of `admin_menu` to reposition menu item.
+	 * @since TBD Redirect to the Guided Setup page on a fresh activation.
 	 */
 	public function add_actions(): void {
 		add_action( 'admin_menu', [ $this, 'landing_page' ], 20 );
@@ -96,6 +97,7 @@ class Controller extends Controller_Contract {
 		add_action( 'admin_post_' . Landing_Page::DISMISS_PAGE_ACTION, [ $this, 'handle_onboarding_page_dismiss' ] );
 		add_action( 'admin_notices', [ $this, 'remove_all_admin_notices_in_onboarding_page' ], -1 * PHP_INT_MAX );
 		add_action( 'tec_admin_headers_about_to_be_sent', [ $this, 'redirect_tec_pages_to_guided_setup' ] );
+		add_action( 'tec_admin_headers_about_to_be_sent', [ $this, 'maybe_redirect_to_guided_setup_on_activation' ] );
 	}
 
 	/**
@@ -118,6 +120,7 @@ class Controller extends Controller_Contract {
 	 *
 	 * @since 6.8.4
 	 * @since 6.11.0 Changed the priority of `admin_menu`.
+	 * @since TBD Removed the activation redirect to the Guided Setup page.
 	 */
 	public function remove_actions(): void {
 		remove_action( 'admin_menu', [ $this, 'landing_page' ], 20 );
@@ -126,6 +129,7 @@ class Controller extends Controller_Contract {
 		remove_action( 'admin_post_' . Landing_Page::DISMISS_PAGE_ACTION, [ $this, 'handle_onboarding_page_dismiss' ] );
 		remove_action( 'admin_notices', [ $this, 'remove_all_admin_notices_in_onboarding_page' ], -1 * PHP_INT_MAX );
 		remove_action( 'tec_admin_headers_about_to_be_sent', [ $this, 'redirect_tec_pages_to_guided_setup' ] );
+		remove_action( 'tec_admin_headers_about_to_be_sent', [ $this, 'maybe_redirect_to_guided_setup_on_activation' ] );
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


src/Events/Admin/Onboarding/Controller.php
```diff
@@ -189,36 +181,120 @@ public function redirect_tec_pages_to_guided_setup(): void {
...
+
+		// Only ever act on it once, regardless of what happens below.
+		delete_transient( '_tribe_events_activation_redirect' );
+
+		// Do not hijack a bulk plugin activation. WordPress adds `activate-multi` to the Plugins
+		// screen URL after several plugins are activated together, so the user stays in context.
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( isset( $_GET['activate-multi'] ) ) {
+			return;
+		}
```
bordoni: Why ignore? Should we be using `tec_get_request_var`

bordoni (dismissed):

src/Events/Admin/Onboarding/Controller.php
dave-green-uk: @bordoni resolved in https://github.com/the-events-calendar/the-events-calendar/pull/5703/commits/c37448a3f8b8f5ee20d4cbe588cfe7bda336beaf

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -189,36 +181,119 @@ public function redirect_tec_pages_to_guided_setup(): void {
...
+		if ( (bool) tribe_get_option( 'tec_onboarding_wizard_visited_guided_setup', false ) ) {
+			return false;
+		}
+
+		// Do not redirect if they dismissed the Guided Setup page. Read the option directly:
+		// the page opts out of Abstract_Admin_Page::$is_dismissible, so is_dismissed() always
+		// returns false, and dismissing should stop the redirect without hiding the menu item.
+		if ( (bool) tribe_get_option( Landing_Page::DISMISS_PAGE_OPTION, false ) ) {
+			return false;
+		}
```
dpanta94, line 277: Prefer `Landing_Page::is_dismissed()` unless there is a reason not to prefer it.

src/Events/Admin/Onboarding/Controller.php
```diff
@@ -189,36 +181,119 @@ public function redirect_tec_pages_to_guided_setup(): void {
...
+	 * @return void
+	 */
+	public function maybe_redirect_to_guided_setup_on_activation(): void {
+		// Set on activation in Tribe__Events__Main::activate(). Absent means this is a normal page load.
+		if ( ! get_transient( '_tribe_events_activation_redirect' ) ) {
+			return;
+		}
+
+		// Only ever act on it once, regardless of what happens below.
+		delete_transient( '_tribe_events_activation_redirect' );
```
dpanta94: I think we should delete the transient later ? Why delete it for example if later we find out that the user doesn't have the capability to see the page ?

i would remove it after we check that the current user has the cap to see the page.

dpanta94 (changes requested):

src/Events/Admin/Onboarding/Controller.php
dave-green-uk: @dpanta94 `is_dismissed()` doesn't work on this page. It's inherited from `Abstract_Admin_Page` and returns false immediately unless $is_dismissible is true — `Landing_Page` never sets that, so it's always false here.

That's the bug this PR fixes: the old code called `is_dismissed()`, so dismissing the Setup Guide never actually stopped the redirect.

We can't just set `$is_dismissible = true`, because `admin_page()` also bails on `is_dismissed()` — dismissing would then delete the Setup Guide menu item too. So we're reading the option directly to keep the fix contained. 

Happy to wire up the dismissal API properly in a follow-up if you'd prefer?

src/Events/Admin/Onboarding/Controller.php
dave-green-uk: Good shout! Resolved in https://github.com/the-events-calendar/the-events-calendar/pull/5703/commits/fd6d10a25ec6800ee5407fb49cb7a3304ba02f28

dpanta94 (approved):

dave-green-uk:
@bordoni Could you leave an approval if you're happy with this resolution? https://github.com/the-events-calendar/the-events-calendar/pull/5703#discussion_r3637597176

PR #5707, opened 2026-07-22
[SMTNC-231] recurring events show duplicated incorrect entries when ECP is deactive

larodiel:
### 🎫 Ticket

[SMTNC-231](https://linear.app/nexcess/issue/SMTNC-231/tec-recurring-events-show-duplicatedincorrect-entries-when-pro-is)

### 🗒️ Description

We had a bug where, when the user created an event and then deactivated it, the events kept repeating, and it shouldn't have happened. The right behavior was to show only the original event.

To avoid checking if the premium plugin class exists, the solution was to check if the filter `tec_events_custom_tables_v1_occurrence_select_fields` exists, and using a static method, we can check if the functionality exists or not.

We have a test suite to it on ECP as well https://github.com/the-events-calendar/events-pro/pull/3067

### 🎥 Artifacts 
https://www.loom.com/share/986bdd2d99e94d5b99793341e158918f

### ✔️ Checklist

### DIFF
--- a/src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
+++ b/src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
@@ -128,6 +128,25 @@ class Custom_Tables_Query extends WP_Query {
 	}
 
 	/**
+	 * Determines whether an active add-on has wired up genuine support for rendering multiple
+	 * Occurrences of the same Event (e.g. each date of a recurring series) as distinct entries.
+	 *
+	 * This is determined by checking whether anything has hooked the
+	 * `tec_events_custom_tables_v1_occurrence_select_fields` filter: the extension point an add-on
+	 * must use to redirect the query `SELECT` to per-Occurrence identifiers. Declaring support
+	 * without implementing that redirection, and the post and meta hydration that depends on it,
+	 * has no functional effect: every row would still resolve back to the same Event post and the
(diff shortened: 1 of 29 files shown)


src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
```diff
@@ -150,8 +169,9 @@ public function get_posts() {
 		// While not ideal, this is the only way to intervene on `GROUP BY` in the `get_posts()` method.
 		add_filter( 'posts_groupby', [ $this, 'group_posts_by_occurrence_id' ], 10, 2 );
 		add_filter( 'posts_where', [ $this, 'filter_by_date' ], 10, 2 );
-		add_filter( 'posts_where', [ $this, 'filter_where' ], 10, 2 );
+		add_filter( 'posts_where', [ $this, 'filter_where' ], 20, 2 );
```
sdokus, line 172: Just curious for my own learning, why are we changing the priority here?

sdokus (approved):

src/Events/Custom_Tables/V1/WP_Query/Custom_Tables_Query.php
larodiel: @sdokus as it's a "where" filter, I want to make sure that it will run after the other hooks, it makes sense when you think on it as a SQL query structure.

PR #5708, opened 2026-07-22
Add license activation to guided setup and the Setup Guide

dave-green-uk:
Relates to [SMTNC-1834](https://linear.app/nexcess/issue/SMTNC-1834/add-activation-ui-to-tec-pro-onboarding) · Parent: [SMTNC-1833](https://linear.app/nexcess/issue/SMTNC-1833/add-activation-ui-to-first-step-of-all-product-onboarding)

## Summary

Onboarding had no route to activating a license — a user who had bought TEC but not activated it here had to find the portal themselves, then find their way back.

This adds that route in both places onboarding appears:

- **Guided setup wizard** — an **Activate license** button beside "Set up my calendar".
- **Setup Guide page** — a **License activated** step at the head of the first-time setup checklist.

Both carry a return address, so the portal brings the user back where they left off.

**Neither appears unless the site runs an active premium plugin.** The calendar is free, and a license only unlocks the premium plugins built on top of it — so a site running none of them is never shown licensing UI implying otherwise.

## Artifact

(new audio)
https://www.loom.com/share/f86c82af864541fb9600ddcb8ce7bf03

Screenshot 2026-07-23 162127

Screenshot 2026-07-23 161405

Screenshot 2026-07-23 161527

## ⚠️ Inert until the Harbor bump

TEC bundles Harbor 1.4.0, whose global functions do not yet include the activation-URL API (`lw_harbor_get_activation_url()` / `lw_harbor_get_product_activation_url()`) — those arrive in [stellarwp/harbor#181](https://github.com/stellarwp/harbor/pull/181). Following review there, the onboarding calls those stable global functions rather than resolving Harbor's `Activation_Url` class from the container: the functions always resolve to the loaded, highest-version Harbor copy, so a consumer is never tied to whichever version it happens to bundle.

Both entry points gate on `can_build_activation_url()` — now a `function_exists()` check on those global functions — and return `''` when they are missing. The React component renders `null`; the checklist step is skipped *and* left out of its "steps completed" tally, so the count stays correct. On today's `main` neither appears — no fatal, no broken link.

Safe to merge whenever; starts working on its own once Harbor lands. It does mean **neither is visible on a clean checkout** — see Testing.

## What changed

| File | Change |
| --- | --- |
| `Admin/Onboarding/License_Data.php` | **New** — every licensing question the onboarding UI asks, including whether a premium plugin is active at all |
| `Admin/Onboarding/Landing_Page.php` | The new checklist step, gated on an active premium plugin; delegates all licensing to the above |
| `Admin/Onboarding/Controller.php` | Registers the new service |
| `wizard/components/buttons/activate.tsx` | New — secondary `Button`, renders `null` on an empty URL |
| `wizard/components/tabs/welcome/tab.tsx` | Wraps both buttons in a flex group |
| `wizard/index.css` | The button group |
| `…/Landing_Page_License_Step_Test.php` | New — 22 cases |

### Why there's a new service

The licensing logic first landed as six methods on `Landing_Page`, which left a class whose job is *rendering an admin page* knowing about Harbor's license repository, product collections, entitlement validity and portal URL formats. Those change for entirely different reasons than the page layout does.

The pattern to follow was already in the same namespace: **`Data`** is a container-resolved class holding exactly this kind of "fetch the state the onboarding UI needs" logic, and `Landing_Page` already delegates its organizer, venue, view and wizard-settings lookups to it. Licensing was the only such lookup that ended up on the page instead.

`License_Data` now sits beside `Data`, and **`Landing_Page` carries no Harbor references at all**. The service takes the return address as a parameter rather than reading `Landing_Page::$slug`, so it knows nothing about the page consuming it and the dependency runs one way.

It is registered in `Controller.php` as a `singleton()`, alongside `Data` and `Landing_Page`.

## Review notes

**1. The checklist step behaves differently from the wizard button — on purpose.** The button disappears once the license is live; the step stays and marks itself complete. A checklist that shrinks as you finish it is worse than one that fills in.

That is why `License_Data` exposes two URL methods rather than one. A single one had conflated two questions:

| Method | Question | Empty when… |
| --- | --- | --- |
| `build_activation_url()` | Can a URL be built at all? | Harbor is too old |
| `get_activation_url()` | Has the user still got something to do? | …that, **or** already activated |

The wizard wants the second, so its button disappears. The checklist wants the first, so it can render the step as **done** rather than hide it.

The step's link follows the state:

| State | Link | Destination | New tab |
| --- | --- | --- | --- |
| Not activated | Activate license | scoped activation URL, returns here | **No** |
| Activated | Manage license | in-WP Software Manager (`options-general.php?page=lw-software-manager`) | **No** |
| Harbor too old | — | step not rendered, not counted | — |

**Neither link opens a new tab, and only the activation link is external.** Before activation the link leaves WordPress for the portal, so it navigates in place — the activation URL carries a return address, and a round trip to a tab the user has wandered away from would strand them looking at a stale step — and carries the `tec-admin-page__link--external` affordance. Once activated the management link stays in wp-admin, so it navigates in place too (the browser back button is the way back) and drops the external affordance.

Management points at the in-WP **Software Manager** page Harbor registers (`options-general.php?page=lw-software-manager`, via `lw_harbor_get_license_page_url()`) — where an activated user manages their products without leaving the site, matching the LearnDash onboarding. There's a test pinning it.

**2. Step order and `$condition` are deliberately independent.** The step leads the list, but `$condition` only feeds the tally — `count()` and `array_filter()` don't care about order — so the existing `$condition[0..4]` references are untouched and the new entry is simply appended. **Please keep these independent** rather than reordering the array to match the markup; that would reintroduce the positional coupling. Tally goes `5` → `6` when the step shows.

**3. `is_activated()` vs `get_licensed_entry()` are not redundant.** The first filters on activation, the second doesn't. That second one is what lets the URL be scoped: the tier is known as soon as the key covers the product, well before activation, so `sku=the-events-calendar%3Aelite` can be set on exactly the installs this targets and the portal pre-selects the right subscription. No covering key, no tier to name — the user lands on their subscription list instead.

<details>
<summary>CSS note — why the margin override matters</summary>

`.tec-events-onboarding__tab-content` is a grid, so the buttons stacked. They're now a flex row with a 20px gap (`--tec-spacer-4`), which needed:

```css
.tec-events-onboarding__button-group > .tec-events-onboarding__button,
… > .tec-events-onboarding__button.is-primary,   /* + is-secondary, is-tertiary */
{
	margin-left: 0;
	margin-right: 0;
}
```

The existing rule centres a lone button with `margin: 0 auto`. Auto margins absorb all free space once the element is a flex item, spreading the buttons apart and leaving `gap` nothing to distribute. Zeroing them hands centring back to `justify-content`. Please keep this if the group is refactored. `flex-wrap: wrap` restores stacking on narrow viewports.

The checklist step needed no new CSS — it reuses `step-list__item` and `tec-admin-page__onboarding-step--completed`.
</details>

**4. The premium plugin gate.** The calendar is free on wp.org, and a license only unlocks the premium plugins built on top of it — so offering activation UI to a site running none of them would imply a license is part of setting the calendar up. `License_Data::has_active_premium_plugin()` gates both surfaces: the wizard button via `get_activation_url()`, the checklist step via an empty `$license_url`.

It delegates to Common's own `Tribe__Dependency::has_active_premium_plugin()` rather than keeping a second list of premium plugins here. Deliberately **not** the `lw_harbor/premium_plugin_exists` filter: that one is cross-brand by design, so an unrelated brand's premium plugin would have switched the calendar's onboarding on.

## Testing

Verified against a real install with Harbor's API manually injected, using live license data on a site holding an `elite` key that covers the calendar without being activated here:

```text
the-events-calendar    activated_entry=no    -> shown        entry=yes   sku=the-events-calendar%3Aelite
learndash              activated_entry=yes   -> HIDDEN
kadence                activated_entry=no    -> shown
not-a-real-product                                           entry=no    sku=(none)
```

Layout and labels confirmed in the browser.

The 22 new tests run in CI (`test (integration)`).

They cover the checklist step in every state, plus `get_activation_url()`'s gating directly. A stand-in `License_Data` is bound into the container, overriding only the four methods that read the outside world — so the real gating logic runs against those answers rather than being stubbed out with them. Two cases bind a stand-in that **throws** if the next question is asked, which pins the guard *ordering* rather than just its result: the premium gate short-circuits before Harbor is asked anything, and the Harbor check short-circuits before licensing state is read.

**Checks:** `php -l` clean; **phpcs clean on the added lines** — warnings are back to the pre-change baseline of 13, all on pre-existing sibling arrays; eslint clean on `activate.tsx`; stylelint unchanged at 38 pre-existing errors in `index.css`, none in the new block.

<details>
<summary>How to see this before the Harbor bump</summary>

Update `common`'s bundled Harbor to the stellarwp/harbor#181 branch (or cherry-pick the activation-URL API into `common/vendor/vendor-prefixed/stellarwp/harbor/`) so the `lw_harbor_get_activation_url()` / `lw_harbor_get_product_activation_url()` global functions register. Because the onboarding now calls those functions — defined in `global-functions.php`, loaded via Composer's `files` autoload — rather than resolving `Activation_Url` directly, the `setClassMapAuthoritative(true)` constraint that keeps new *classes* from autoloading does not apply to them.
</details>

## Before merge

- `@since TBD` on the new PHP members needs a real version.
- Depends on stellarwp/harbor#181 and a Harbor bump in `common`.

*Unrelated:* in Firefox the wizard button opens the portal in a new window. Nothing in this branch, TEC, common, Harbor, or WP core's `Button` sets a `target`, and Chrome is fine. Not caused by this PR — noted so it isn't filed twice.

https://claude.ai/code/session_011uxoDruAaZXzRy3hk8mr4W

### DIFF
--- a/changelog/feat-smtnc-1833-onboarding-activate-license-button
+++ b/changelog/feat-smtnc-1833-onboarding-activate-license-button
@@ -1,4 +1,4 @@
 Significance: minor
-Type: feat
+Type: tweak
 
-Added an "Activate license" button to the first step of the guided setup, and a matching "License activated" step to the Setup Guide checklist. Both send the user to the Liquid Web portal to activate, and bring them back where they left off once they are done. The wizard button is hidden after the calendar is licensed and activated on the site. The checklist step stays on show, marked as complete, and offers a "Manage license" link to the portal instead.
+Improved the unified licensing page experience.
--- a/changelog/feat-smtnc-1833-onboarding-activate-license-button
+++ b/changelog/feat-smtnc-1833-onboarding-activate-license-button
@@ -1,4 +1,4 @@
 Significance: minor
 Type: feat
 
-Added an "Activate license" button to the first step of the guided setup, and a matching "License activated" step to the Setup Guide checklist. Both send the user to the Liquid Web portal and return them to where they left off. The wizard button is hidden once the calendar is licensed and activated on the site; the checklist step stays on show and is marked as complete.
+Added an "Activate license" button to the first step of the guided setup, and a matching "License activated" step to the Setup Guide checklist. Both send the user to the Liquid Web portal to activate, and bring them back where they left off once they are done. The wizard button is hidden after the calendar is licensed and activated on the site. The checklist step stays on show, marked as complete, and offers a "Manage license" link to the portal instead.
--- a/changelog/feat-smtnc-1833-onboarding-activate-license-button
+++ b/changelog/feat-smtnc-1833-onboarding-activate-license-button
@@ -1,4 +1,4 @@
 Significance: minor
-Type: tweak
+Type: feat
 
 Improved the unified licensing page experience.
--- a/changelog/feat-smtnc-1833-onboarding-activate-license-button
+++ b/changelog/feat-smtnc-1833-onboarding-activate-license-button
@@ -1,4 +1,4 @@
 Significance: minor
 Type: feat
 
-Added an "Activate license" button to the first step of the guided setup, sending the user to the Liquid Web portal and returning them to where they left off. The button is hidden when the calendar is already licensed and activated on the site.
+Added an "Activate license" button to the first step of the guided setup, and a matching "License activated" step to the Setup Guide checklist. Both send the user to the Liquid Web portal and return them to where they left off. The wizard button is hidden once the calendar is licensed and activated on the site; the checklist step stays on show and is marked as complete.
--- a/changelog/feat-smtnc-1833-onboarding-activate-license-button
+++ b/changelog/feat-smtnc-1833-onboarding-activate-license-button
@@ -1,0 +1,4 @@
+Significance: minor
+Type: feat
+
+Added an "Activate license" button to the first step of the guided setup, sending the user to the Liquid Web portal and returning them to where they left off. The button is hidden when the calendar is already licensed and activated on the site.
--- a/src/Events/Admin/Onboarding/Controller.php
+++ b/src/Events/Admin/Onboarding/Controller.php
@@ -16,6 +16,7 @@ use TEC\Events\Admin\Onboarding\Steps\Venue;
 use TEC\Events\Admin\Onboarding\Steps\Tickets;
 use TEC\Events\Admin\Onboarding\Data;
 use TEC\Events\Admin\Onboarding\Landing_Page;
+use TEC\Events\Admin\Onboarding\License_Data;
 use TEC\Common\StellarWP\Assets\Config;
 use Tribe__Events__Main;
 
@@ -56,6 +57,7 @@ class Controller extends Controller_Contract {
 
 		$this->container->singleton( Landing_Page::class );
 		$this->container->singleton( Data::class );
+		$this->container->singleton( License_Data::class );
 	}
 
 	/**
--- a/src/Events/Admin/Onboarding/Landing_Page.php
+++ b/src/Events/Admin/Onboarding/Landing_Page.php
@@ -14,6 +14,10 @@ use TEC\Common\StellarWP\Installer\Installer;
 use TEC\Common\Admin\Abstract_Admin_Page;
 use TEC\Common\Admin\Traits\Is_Events_Page;
 use TEC\Common\Asset;
+use TEC\Common\LiquidWeb\Harbor\Licensing\Product_Collection;
+use TEC\Common\LiquidWeb\Harbor\Licensing\Repositories\License_Repository;
+use TEC\Common\LiquidWeb\Harbor\Licensing\Results\Product_Entry;
+use TEC\Common\LiquidWeb\Harbor\Portal\Activation_Url;
 use Tribe__Events__Main as TEC;
 
 /**
@@ -43,6 +47,15 @@ class Landing_Page extends Abstract_Admin_Page {
 	 * @var string
 	 */
 	const DISMISS_PAGE_OPTION = 'tec_events_onboarding_page_dismissed';
+
+	/**
+	 * The product slug this plugin is licensed under in the Liquid Web catalog.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	const LICENSE_PRODUCT_SLUG = 'the-events-calendar';
 
 	/**
 	 * The slug for the admin menu.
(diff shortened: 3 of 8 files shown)


changelog/feat-smtnc-1833-onboarding-activate-license-button
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: feat
+
+Added an "Activate license" button to the first step of the guided setup, and a matching "License activated" step to the Setup Guide checklist. Both send the user to the Liquid Web portal to activate, and bring them back where they left off once they are done. The wizard button is hidden after the calendar is licensed and activated on the site. The checklist step stays on show, marked as complete, and offers a "Manage license" link to the portal instead.
```
WayneRocha: I looked into [changelogs we already have](https://github.com/the-events-calendar/the-events-calendar/blob/smtnc-1833/onboarding-activate-license-button/changelog.md) and I don't see lines so long.
Idk the character limit for that, but if it's there some way you can make this more short or maybe divide into more files would be good

WayneRocha (approved):
LGTM @dave-green-uk. only flagging a detail for changelog

changelog/feat-smtnc-1833-onboarding-activate-license-button
dave-green-uk: Done!

dave-green-uk:
@WayneRocha Added a couple of new commits if you could re-review?

changelog/feat-smtnc-1833-onboarding-activate-license-button
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: tweak
```
WayneRocha:
```suggestion
Type: feat
```

WayneRocha (approved):
tweak is small adjust. You really did a good amount of work here, so it can be `feat`. Just a minor adjust, but looks good for me.
I have some understand about TEC, but not much about harbor, so I would you  to request the other reviewers to also take a look @dave-green-uk, not just me. They can see some context I don't have.

dave-green-uk:
Leaving a note before I go on leave tomorrow. This PR is effectively no longer required in light of the [Harbor dev guidelines](https://docs.google.com/document/d/1jWmNRrsfjKgviWzBo-l52p2L3X-oVvfI0LU5d8FcqUQ/edit?tab=t.0#heading=h.i4p5fzeiaozd), and I'll pick this up when I'm back.

We need to implement the same UI, but do this in https://github.com/the-events-calendar/events-pro

PR #5728, opened 2026-08-17
[SMTNC-1979] posts_per_page in List view creates an infinite crawler trap

WayneRocha:
### 🎫 Ticket

[SMTNC-1979](https://linear.app/nexcess/issue/SMTNC-1979/posts-per_page-in-list-view-creates-an-infinite-crawler-trap)

### 🗒️ Description

Resolved an issue where the List view exposed an internal pagination look-ahead value in generated navigation and canonical URLs, creating an unbounded, self-generating crawl space. `posts_per_page` is now scrubbed from generated URLs, and the number of events fetched per page is capped at 200 so unauthenticated requests cannot drive unbounded repository queries.

### ✔️ Checklist
- [x] Are all the **required** tests passing? (slic unavailable locally — not run)

### DIFF
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -1361,7 +1361,7 @@ class View implements View_Interface {
 	 *
 	 * @return int The bounded number of events per page.
 	 */
-	protected function get_events_per_page( Context $context = null ) {
+	protected function get_events_per_page( ?Context $context = null ) {
 		$context ??= $this->get_context();
 		$per_page = absint( $context->get( 'events_per_page', 12 ) );
 
--- a/src/Tribe/Views/V2/View.php
+++ b/src/Tribe/Views/V2/View.php
@@ -48,6 +48,16 @@ class View implements View_Interface {
 	use Json_Ld_Data;
 
 	/**
+	 * The maximum number of events a View will fetch per page.
+	 *
+	 * The `posts_per_page` request variable can drive the number of events per page through the Context,
+	 * so the value is capped to keep unauthenticated requests from forcing unbounded queries.
+	 *
+	 * @since TBD
+	 */
+	const MAX_EVENTS_PER_PAGE = 200;
+
+	/**
 	 * An instance of the DI container.
 	 *
 	 * @var Container
@@ -1275,6 +1285,12 @@ class View implements View_Interface {
 		// By default let's use the locations set in the Context to map the arguments to query args.
 		$query_args = tribe_context()->map_to_read( $args, Context::REQUEST_VAR );
 
+		/*
+		 * `posts_per_page` is an internal repository argument carrying the pagination look-ahead value; it must
+		 * never be reflected into a generated View URL or an unbounded, self-generating crawl space results.
+		 */
+		unset( $query_args['posts_per_page'] );
+
 		global $wp;
 
 		return array_intersect_key( $query_args, array_combine( $wp->public_query_vars, $wp->public_query_vars ) );
@@ -1334,6 +1350,25 @@ class View implements View_Interface {
 	}
 
 	/**
+	 * Returns the number of events a View should show per page, bounded to a safe maximum.
+	 *
+	 * The value can be driven by the `posts_per_page` request variable through the Context, so it is
+	 * capped here to keep unauthenticated requests from forcing unbounded repository queries.
+	 *
+	 * @since TBD
+	 *
+	 * @param Context|null $context The context to read the value from, or `null` to use the View Context.
+	 *
+	 * @return int The bounded number of events per page.
+	 */
+	protected function get_events_per_page( Context $context = null ) {
+		$context ??= $this->get_context();
+		$per_page = absint( $context->get( 'events_per_page', 12 ) );
+
+		return min( $per_page, static::MAX_EVENTS_PER_PAGE );
+	}
+
+	/**
 	 * Sets up the View repository arguments from the View context or a provided Context object.
 	 *
 	 * @since 4.9.3
@@ -1358,7 +1393,7 @@ class View implements View_Interface {
 		 * @since 5.0.0
 		*/
 		$args = [
-			'posts_per_page'       => (int) $context_arr['events_per_page'] + 1,
+			'posts_per_page'       => $this->get_events_per_page( $context ) + 1,
 			'paged'                => max( Arr::get_first_set( array_filter( $context_arr ), [
 				'paged',
 				'page',
@@ -1427,7 +1462,7 @@ class View implements View_Interface {
 			1
 		);
 
-		return ( $current_page - 1 ) * (int) $context->get( 'events_per_page' );
+		return ( $current_page - 1 ) * $this->get_events_per_page();
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


WayneRocha:
@coderabbitai review

src/Tribe/Views/V2/View.php
```diff
@@ -47,6 +47,16 @@ class View implements View_Interface {
...
 
+	/**
+	 * The maximum number of events a View will fetch per page.
+	 *
+	 * The `posts_per_page` request variable can drive the number of events per page through the Context,
+	 * so the value is capped to keep unauthenticated requests from forcing unbounded queries.
+	 *
+	 * @since TBD
+	 */
+	const MAX_EVENTS_PER_PAGE = 200;
```
coutinhocouto, line 58: Can we make this filterable rather than a hard constant? Something like `tec_events_views_v2_max_events_per_page`.

Two reasons. A site that has legitimately set "Events per page" above 200 gets silently truncated with nothing to turn off. And more importantly, `Hooks.php::filter_pre_get_posts_events_per_page()` still sets the main query's `posts_per_page` from `tribe_get_option( 'posts_per_page', … )` uncapped — that filter exists specifically to keep the main query in step with the View ("so pagination (found_posts, max_num_pages) matches the View and AJAX", per its docblock), and capping one side without the other puts them back out of sync on exactly those sites.

A filter gives us one value to apply in both places.

src/Tribe/Views/V2/View.php
```diff
@@ -1592,7 +1627,7 @@ public function set_has_next_event( $value ) {
 	 * @return mixed                   Weather the array of events has a next page.
 	 */
 	public function has_next_event( array $events, $overwrite_flag = true ) {
-		$has_next_events = count( $events ) > (int) $this->get_context()->get( 'events_per_page', 12 );
+		$has_next_events = count( $events ) > $this->get_events_per_page();
```
coutinhocouto, line 1630: This one's now capped, but `Hooks::action_set_title_events()` still trims with the raw context value:

    $cnt = $view->get_context()->get( 'events_per_page' );
    $events = array_slice( $events, 0, (int) $cnt );

With `?posts_per_page=250` the repository returns at most 201 events, so that slice is a no-op and the look-ahead event ends up in the title's post list. Before the cap the two numbers always agreed, so it's fallout from this change rather than something pre-existing. Worth routing that call site through `get_events_per_page()` too.

src/Tribe/Views/V2/View.php
```diff
@@ -1333,6 +1349,25 @@ protected function filter_template_vars( array $template_vars ) {
...
+	 * The value can be driven by the `posts_per_page` request variable through the Context, so it is
+	 * capped here to keep unauthenticated requests from forcing unbounded repository queries.
+	 *
+	 * @since TBD
+	 *
+	 * @param Context|null $context The context to read the value from, or `null` to use the View Context.
+	 *
+	 * @return int The bounded number of events per page.
+	 */
+	protected function get_events_per_page( ?Context $context = null ) {
```
coutinhocouto, line 1364: Could take a native `: int` here. Brand-new method, so the BC rule about not adding types to already-released methods doesn't apply, and this file already declares return types in a dozen places (622, 778, 791, 852, 2436, 2467, 2535, 2542, 2586, 3195).

tests/views_integration/Tribe/Events/Views/V2/ViewTest.php
```diff
@@ -253,6 +253,43 @@ public function should_correctly_produce_a_view_prev_and_next_canonical_urls() {
...
+	 * It should not expose the posts_per_page look-ahead value in the View URL.
+	 *
+	 * @test
+	 */
+	public function should_not_expose_posts_per_page_in_the_view_url() {
+		add_filter( 'tribe_events_views', static function () {
+			return [ 'test' => Test_View::class ];
+		} );
+
+		global $wp;
```
coutinhocouto, line 266: This mutates the global `$wp->public_query_vars` and never restores it, and the `tribe_events_views` filter two lines up isn't removed either. There's no `tearDown()` on the class, so both leak into any test that runs after this one in the same process.

`View_RegisterTest.php` already handles this exact global — backs it up in `wpSetUpBeforeClass()`, restores it in `tearDownAfterClass()`. Worth copying that shape.

tests/views_integration/Tribe/Events/Views/V2/ViewTest.php
```diff
@@ -253,6 +253,43 @@ public function should_correctly_produce_a_view_prev_and_next_canonical_urls() {
...
+	 *
+	 * @test
+	 */
+	public function should_cap_the_events_per_page_repository_argument() {
+		add_filter( 'tribe_events_views', static function () {
+			return [ 'test' => Test_View::class ];
+		} );
+
+		$view = View::make( 'test' );
+		$view->set_context( tribe_context()->alter( [ 'events_per_page' => 3517 ] ) );
```
coutinhocouto, line 286: `3517` is the figure straight out of the ticket's crawler log, and bare like this it doesn't say why it was picked. `View::MAX_EVENTS_PER_PAGE * 2` would keep the test meaningful if the cap ever moves — as written it quietly stops testing anything if someone raises the constant past 3517.

coutinhocouto (changes requested):
**`Event_Query_Controller.php:128` is still unbounded.** It builds the SSR initial-state ORM args from `max( 1, (int) $context->get( 'events_per_page', 12 ) )`, which maps straight from the public `posts_per_page` request var. The admin option only overrides it when `event_display` is `list`, `month` or `day` (lines 137-145) — any other display value lets the request value drive the query, so `?posts_per_page=100000` can still reach an expensive unauthenticated query. The ticket's third ask was to bound the value wherever it's read from a request, and this is the call site still open.

src/Tribe/Views/V2/View.php
WayneRocha: Addressed in 9b48c7c9f27890cf32bb7c73d84cb4a707efa561

src/Tribe/Views/V2/View.php
WayneRocha: Addressed in 9b48c7c9f27890cf32bb7c73d84cb4a707efa561

src/Tribe/Views/V2/View.php
WayneRocha: Addressed in bc3b7074dfbf493f3b67104d62e57cb79c71c759

tests/views_integration/Tribe/Events/Views/V2/ViewTest.php
WayneRocha: Addressed in 1e38e4dd6793673f187b83638ce8d036df7f1a9a

tests/views_integration/Tribe/Events/Views/V2/ViewTest.php
WayneRocha: Addressed in 1e38e4dd6793673f187b83638ce8d036df7f1a9a

WayneRocha (commented):
Addressed in 29373079c7c7701a145230ba26afb560845dd06b

WayneRocha:
Conflicts resolved by Rebase

PR #5745, opened 2026-08-28
 Improve error reporting

sdokus:
### 🎫 Ticket

[SMTNC-2262](https://linear.app/nexcess/issue/SMTNC-2262/event-aggregator-fails-to-import-from-eventbrite-despite-successful)

### 🗒️ Description

This PR is to improve the specificity of the error handling for EA. Before, there were four different scenarios that would result in the error message “There was an error fetching the results from your import: There may be an issue with the Event Aggregator server. Please try your import again later.”

After this PR, specific error messages will show or if the error is a `403` it will show the message 'Event Aggregator server has blocked your request. Please try your import again later or contact support to know why.'

I don't think that these changes will necessarily straight up fix the linear ticket linked, but it at least will give us more information about why a user is being blocked from importing events from eventbrite. 

### ✔️ Checklist

### DIFF
--- a/changelog/fix-smtnc-2262-ea-post-error-handling.yaml
+++ b/changelog/fix-smtnc-2262-ea-post-error-handling.yaml
@@ -1,0 +1,4 @@
+significance: patch
+type: tweak
+entry: Improve specificity of error handling when syncing event with Event Aggregator.
+timestamp: 2026-08-28T16:05:44.919Z
--- a/changelog/fix-smtnc-2262-ea-post-error-handling.yaml
+++ b/changelog/fix-smtnc-2262-ea-post-error-handling.yaml
@@ -1,4 +1,4 @@
 significance: patch
 type: tweak
-entry: Improve specificity of error handling when syncing event with Event Aggregator.
+entry: Improved specificity of error handling when syncing events with Event Aggregator.
 timestamp: 2026-08-28T16:05:44.919Z
--- a/src/Tribe/Aggregator/Service.php
+++ b/src/Tribe/Aggregator/Service.php
@@ -288,10 +288,12 @@ class Tribe__Events__Aggregator__Service {
 	/**
 	 * Performs a POST request against the Event Aggregator service
 	 *
-	 * @param string $endpoint   Endpoint for the Event Aggregator service
-	 * @param array  $data       Parameters to send to the endpoint
+	 * @since TBD Moved the transport-error check ahead of the response-code read and added distinct handling for a blocked (403) response.
 	 *
-	 * @return stdClass|WP_Error
+	 * @param string $endpoint Endpoint for the Event Aggregator service.
+	 * @param array  $data     Parameters to send to the endpoint.
+	 *
+	 * @return stdClass|WP_Error The decoded response body, or a WP_Error describing the failure.
 	 */
 	public function post( $endpoint, $data = [] ) {
 		$url = $this->build_url( $endpoint );
@@ -314,9 +316,11 @@ class Tribe__Events__Aggregator__Service {
 
 		$response = $this->requests->post( esc_url_raw( $url ), $args );
 
-		// A transport-level failure (DNS, TLS, refused connection, timeout) never carries a
-		// response code, so this has to be checked before `$code` is read: a `WP_Error` would
-		// otherwise yield `0` and get reported as a generic server-side problem.
+		/*
+		 * A transport-level failure (DNS, TLS, refused connection, timeout) never carries a
+		 * response code, so this has to be checked before `$code` is read: a `WP_Error` would
+		 * otherwise yield `0` and get reported as a generic server-side problem.
+		 */
 		if ( is_wp_error( $response ) ) {
 			tribe( 'logger' )->log_debug(
 				'Request failed during the creation: ' . $response->get_error_message(),
@@ -329,6 +333,8 @@ class Tribe__Events__Aggregator__Service {
 		$code = (int) wp_remote_retrieve_response_code( $response );
 
 		if ( 403 === $code ) {
+			tribe( 'logger' )->log_debug( 'Request denied with a 403 - during the creation.', 'EA Service' );
+
 			return new WP_Error(
 				'core:aggregator:request-denied',
 				esc_html__( 'Event Aggregator server has blocked your request. Please try your import again later or contact support to know why.', 'the-events-calendar' )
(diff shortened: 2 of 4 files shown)


sdokus:
@coderabbitai review

src/Tribe/Aggregator/Service.php
```diff
@@ -314,8 +314,28 @@ public function post( $endpoint, $data = [] ) {
 
 		$response = $this->requests->post( esc_url_raw( $url ), $args );
 
-		// we know it is not a 404 or 403 at this point.
+		// A transport-level failure (DNS, TLS, refused connection, timeout) never carries a
+		// response code, so this has to be checked before `$code` is read: a `WP_Error` would
+		// otherwise yield `0` and get reported as a generic server-side problem.
```
coutinhocouto:
```suggestion
        /**
		* A transport-level failure (DNS, TLS, refused connection, timeout) never carries a
		* response code, so this has to be checked before `$code` is read: a `WP_Error` would
		* otherwise yield `0` and get reported as a generic server-side problem.
		* /
```

tests/muintegration/Tribe/Events/Aggregator/ServiceTest.php
```diff
@@ -303,13 +303,95 @@ public function it_should_apply_subsite_limit_if_subsite_has_local_key() {
...
 	 */
 	private function make_instance() {
 		return new Service( $this->requests->reveal() );
 	}
 
+	/**
+	 * @param int $code The HTTP status code the mocked response should carry.
+	 *
+	 * @return array
+	 */
```
coutinhocouto, line 379:
```suggestion
	/**
	 * @since TBD
	 *
	 * @param int $code The HTTP status code the mocked response should carry.
	 *
	 * @return array{response: array{code: string}, headers: array{content-type: string}, body: string} A mocked `wp_remote_post()` response carrying the given status code.
	 */
```

tests/muintegration/Tribe/Events/Aggregator/ServiceTest.php
```diff
@@ -303,13 +303,95 @@ public function it_should_apply_subsite_limit_if_subsite_has_local_key() {
...
 	}
 
+	/**
+	 * A transport-level failure has no response code, so `post()` has to check for it before
+	 * reading one. Regression: the code check used to run first, turning every cURL failure
+	 * into `core:aggregator:bad-response` and discarding the real reason.
+	 *
+	 * @test
+	 * it should return the transport error when the post request fails at the transport level
+	 */
```
coutinhocouto, line 309: This paragraph describes the change rather than the test, and every other test in the file is just @test plus a one-liner. The why is already at the call site.

```suggestion
	/**
	 * @test
	 * it should return the transport error when the post request fails at the transport level
	 */
```

tests/muintegration/Tribe/Events/Aggregator/ServiceTest.php
```diff
@@ -303,13 +303,95 @@ public function it_should_apply_subsite_limit_if_subsite_has_local_key() {
...
+		$this->assertEquals( 'core:aggregator:request-denied', $response->get_error_code() );
+	}
+
+	/**
+	 * The generic message is still correct for a genuine server-side fault; it just must not be
+	 * the catch-all for causes the client can tell apart.
+	 *
+	 * @test
+	 * it should report a bad response when the post request returns a server error
+	 */
```
coutinhocouto, line 352: Same thing here.

```suggestion
	/**
	 * @test
	 * it should report a bad response when the post request returns a server error
	 */
```

changelog/fix-smtnc-2262-ea-post-error-handling.yaml
```diff
@@ -0,0 +1,4 @@
+significance: patch
+type: tweak
+entry: Improve specificity of error handling when syncing event with Event Aggregator.
```
coutinhocouto:
```suggestion
entry: Improved specificity of error handling when syncing events with Event Aggregator.
```

src/Tribe/Aggregator/Service.php
coutinhocouto, line 297: Docblock still reads like nothing changed, but you've reordered the error handling and added a branch. Needs a `@since` TBD line saying what moved, and the @return has never had a description.

```suggestion
	/**
	 * Performs a POST request against the Event Aggregator service
	 *
	 * @since TBD Moved the transport-error check ahead of the response-code read and added distinct handling for a blocked (403) response.
	 *
	 * @param string $endpoint Endpoint for the Event Aggregator service.
	 * @param array  $data     Parameters to send to the endpoint.
	 *
	 * @return stdClass|WP_Error The decoded response body, or a WP_Error describing the failure.
	 */
```

src/Tribe/Aggregator/Service.php
```diff
@@ -314,8 +314,28 @@ public function post( $endpoint, $data = [] ) {
...
+		}
+
 		$code = (int) wp_remote_retrieve_response_code( $response );
+
+		if ( 403 === $code ) {
+			return new WP_Error(
+				'core:aggregator:request-denied',
+				esc_html__( 'Event Aggregator server has blocked your request. Please try your import again later or contact support to know why.', 'the-events-calendar' )
+			);
+		}
```
coutinhocouto, line 342: The other two branches here both log, this one doesn't. It matters more than usual because of the New.php issue above: if the message gets swallowed, the log is all that's left. And a 403 from the license or domain check is exactly the theory you raised in the ticket comments.

```suggestion
		if ( 403 === $code ) {
			tribe( 'logger' )->log_debug( 'Request denied with a 403 - during the creation.', 'EA Service' );

			return new WP_Error(
				'core:aggregator:request-denied',
				esc_html__( 'Event Aggregator server has blocked your request. Please try your import again later or contact support to know why.', 'the-events-calendar' )
			);
		}
```

coutinhocouto (changes requested):
> [!NOTE]  
> @sdokus a question poped up when reviewing this for `src/Tribe/Aggregator/Tabs/New.php:486`:
>
> This is where the new messages get lost. ajax_create_import() only passes the error code to get_service_message(), and neither core:aggregator:request-denied nor core:aggregator:bad-response is registered in $service_messages or in Errors.php. So both fall through to get_unknow_message() and the user sees "Unknown service message". The 403 text you added never makes it to the screen on this path. It already behaved this way for bad-response, so it's not something you broke, but it does mean the extra specificity stops here. 
>
> So I would suggest using something like: 'message' => $service->get_service_message( $result->get_error_code(), $result->get_error_data() ?? [], $result->get_error_message() ),

coutinhocouto (approved):
@sdokus thanks for the changes, all look good to me now

bordoni (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/the-events-calendar/the-events-calendar/pull/5745)</sub>

PR #5746, opened 2026-08-31
feat: free Occurrence infrastructure and one-by-one event dates

bordoni:
### 🎫 Ticket

[TBD]

### 🗒️ Description

Free sites get events with multiple explicit dates, without Pro. Each date gets its own URL and its own row on the calendar views, and the event gets an archive of its dates at `/event/{slug}/all/`. Everything sits behind `TEC\Events\Recurrence\Controller` (kill-switch constant, env, a default-off option, then the `tec_events_recurrence_enabled` filter), so this is dark-shipped and nothing changes for anyone until we flip it on.

Today an event in free is one date, full stop. The plumbing that could do better (the CT1 tables, the occurrence ORM, the query redirection) already lives in free with no engine on top of it, and this PR adds that engine for explicit date lists. Rule-based recurrence (RRULE) and Series stay in Pro, and php-rrule does not move.

It also fixes a real migration bug. A legacy event whose `_EventRecurrence` holds only a list of dates fails migration today with an "install Events Calendar PRO" error, and since migration is fail-fast, that one event aborts the whole site's migration. That shape is exactly what free's own `Date_Rules` writes. The new `tec-date-rules-strategy` migrates those events to one Occurrence per date, so only genuinely rule-based events still need Pro to migrate.

What free now owns: the provisional post ID system and its query/meta hydration, the four recurrence columns (`rset`, `has_recurrence`, `sequence`, `is_rdate`) with the same schema versions and option keys Pro registers (the `Schema_Builder` dedupes the double registration), per-occurrence links and the date URL rewrite rules, a dates-only generator on the existing `tec_events_custom_tables_v1_occurrences_generator` seam, the `Dates_Service` authoring API that writes the legacy `_EventRecurrence` date shape, the Event Dates authoring UI in both editors, the recurring markers on the Views v2 cards, the `/all/` archive plus the dateless URL redirect, the occurrence template tags, the date-list migration strategy, and the guard rails around Pro-authored rule-based events: the freeze of their dates, the lock setting and the conversion to individual dates.

Compatibility has no release train in either direction. The controller refuses to register when a Pro older than `7.9.0-dev` is active, so an older Pro keeps sole ownership and runs exactly as released. A newer Pro force-enables the gate and defers its own copies to these classes; that side is the companion PR, the-events-calendar/events-pro#3078. The handshake Pro reads is the static `Controller::provides_occurrences()`, the whole gate as one read that constructs nothing at `plugins_loaded::1`; `is_active()` delegates to it. `_EventRecurrence` stays the canonical authored format and `rset` stays derived from it, so there are no stored data changes and either plugin can be rolled back on its own. When an rset contains rules and no rule engine is registered, the existing Occurrences are frozen instead of deleted.

On a rule-locked event (Pro data, Pro absent) the Event Dates section in both editors lists every scheduled date as a chip: each one links to its occurrence, carries a tooltip with the full range and its status, and pairs with a button opening the occurrence edit screen in a new tab; past dates collapse behind a toggle. Building that surfaced a timezone bug, fixed here: the Occurrences table stores event-local wall-clock values, and the section was formatting them as UTC epochs shifted by the site offset.

Saving an occurrence edit screen now moves that occurrence only, in both editors. Before this, it silently corrupted the parent: every meta write against a provisional post ID is retargeted to the real post, so the posted dates landed on the event's own date, its first occurrence got pruned, and the occurrence on screen never changed. The date meta posted against a provisional ID is now buffered before that retargeting and applied as a move of that one occurrence. On a dates-only event the matching authored date is replaced (the event's own date when the occurrence is the first one) and the set regenerates. On a rule-locked event the move is refused: its occurrences follow the Pro rules, which only Pro or the conversion below may change, so the refusal is recorded with the `Freeze_Guard` and the user gets the same warning a refused date write leaves (an untouched save, which re-posts the occurrence's own dates, stays silent). The occurrence row moves in place before anything regenerates, so its ID, and the edit screen, stay put.

Rule-locked events are now protected from every write path, not only from the Event Dates section. A `Freeze_Guard` on the post meta filters swallows changes to the date and `_EventRecurrence` meta of such an event, whether they come from a classic save, a REST save, the ORM or an import; the title, content, venue and the rest still save, a classic save that tried to change dates leaves a warning notice, and both editors render the Start/End, all-day and timezone controls disabled (the classic editor posts hidden mirrors so an untouched save stays byte-identical). The occurrence edit screens of such an event get the same treatment: disabled date controls, an occurrence-specific notice linking back to the recurring event, and, with the setting off, the same conversion controls, nonced for the parent event and landing on its screen. Whether the lock can be lifted is a setting: Events > Settings > General > Editing gains "Keep Events Calendar Pro recurrence rules locked", on by default, shown only on sites that have used Pro (`Pro_History` reads the traces Pro leaves: schema version options, its settings, its license, its Series table or posts, rule-based rows) and linking to Pro unless upsells are hidden. With the setting off, the Event Dates section and the block editor panel explain what a conversion does and, after an "I understand" checkbox, post to `admin-post.php`: `Rules_Conversion` turns every frozen occurrence into an authored date, keeping occurrence IDs and permalinks, drops the rules and Pro's block meta, removes the event's rows from `tec_series_relationships` and trashes an autogenerated Series left empty, then redirects back with a notice the block editor receives through its config. Conversion is irreversible for the rules and capped at 1,000 dates (`tec_events_recurrence_conversion_max_dates`); programmatic writers hitting a locked event are no-op'd for the frozen keys, with `tec_events_recurrence_freeze_meta_write` as the escape hatch.

The numbers: 78 commits, +22,593 / −457 across 185 files. 52 new production PHP files, 6 new view templates, 33 changelog entries written to the readme house style. 45 new PHP test files across 4 suites (wpunit, ct1_integration, ct1_migration, views_integration) adding 258 tests, plus 2 jest suites, 12 of them snapshot files. Writing those tests surfaced two production bugs that are fixed here as well (a third, the occurrence save corrupting the parent, came from the owner testing the chips): a stale occurrence match cache churned occurrence IDs when the same request saved twice, and the controller re-bound resolved singletons and orphaned their hooks.

All CI checks pass. Full local runs: ct1_integration green (the Recurrence directory alone is 167 tests and 883 assertions), ct1_migration green including the 17 new migration tests, the full wpunit suite green, views_integration partials green, jest green at 257 tests. I checked the free RSET serializer output byte for byte against what Pro derives for the same dates, and verified the TEC/Pro version combos live on Lando.

### 🧪 Human QA test plan

Free only (Pro deactivated), pretty permalinks on, both editors available. Testers can work from the interactive checklist at https://claude.ai/code/artifact/a8060127-ddab-42ed-addc-6c7e0435f2fe (progress saves in your browser); it also carries the cross-version matrix for the TEC/Pro combos, which this list only touches in step 9.

1. Create an event and add 2 or 3 extra dates in the classic editor. Save twice. The date list must be identical after both saves. Repeat in the block editor and check both editors show the same rows, including an All Day row.
2. Open a dated URL (`/event/{slug}/{YYYY-MM-DD}/`): it renders that single occurrence. Open the dateless URL (`/event/{slug}/`): it redirects to the next upcoming date. Move every date into the past and the same URL redirects to `/all/`.
3. Open `/event/{slug}/all/`: upcoming dates listed one card per date, each card linking its own dated URL, with a working link to past dates and pagination that does not 404.
4. On the list, month, and day views the event shows the recurring marker, each card links its own date, and the marker's "see all" link lands on `/all/`.
5. Open a single occurrence for editing through the edit button of a scheduled date chip: the editor shows the single-occurrence notice with a working link back to the parent event, and saving with the dates untouched changes nothing.
6. Open a rule-locked event (Pro-authored rules, Pro deactivated) in each editor: every scheduled date shows as a chip, hovering one shows the weekday range and status, the chip opens the dated occurrence URL and its edit button opens the occurrence edit screen, both in new tabs, and the past dates toggle works. Set the site timezone to something other than UTC and the chip times still match the event's own times.
7. From a chip's edit button, change the occurrence's date and time and save, in each editor. The screen reloads on the same occurrence showing the new date, the parent event's own date and every other occurrence are unchanged, the parent's chips show the moved date, and the classic editor shows the success notice. Repeat on a dates-only event (an extra date, then the event's own first date). On a rule-locked event the occurrence's date controls are disabled instead (see step 8).
8. Still on a rule-locked event with Pro deactivated, and the "Keep Events Calendar Pro recurrence rules locked" setting at its default (checked; it only shows under Events > Settings > General > Editing on sites that have used Pro, with a link to Pro in its description): in each editor the Start/End, all-day and timezone controls are disabled and the Event Dates section explains the lock with a link to the setting. Change the title and save: the title saves, the dates do not move. In the classic editor, force a date change (re-enable the input in dev tools) and save: the dates stay, and a warning notice says they were left unchanged. Open one of its occurrences from a chip's edit button: in each editor the date controls are disabled too, the Event Dates section says "This is one date of an event…" with the settings link and an "Edit the recurring event." link and shows no chips; saving the title keeps the occurrence's date, and a forced date change leaves the occurrence row, the parent and the rules unchanged with the same warning notice.
9. Uncheck the setting. The Event Dates section (and the block editor panel) now shows the warning, the "I understand" checkbox and the "Convert to individual dates" button, disabled until the checkbox is ticked; an occurrence screen of the event shows the same, plus "Converting sends you to the recurring event.", and converting from there lands on the parent event. Convert: the screen reloads on the same event with a success notice, the Event Dates rows are editable, the occurrence IDs and dated URLs are unchanged (`SELECT occurrence_id, start_date FROM wp_tec_occurrences WHERE post_id = <id>` before and after), the event's rows in `wp_tec_series_relationships` are gone and its autogenerated Series is trashed. Reactivate Pro: the event reads as a list of dates, no fatal.
10. On a legacy pre-CT1 site with one date-list recurring event and one rule-based one, run the migration preview. The date-list event previews under the new strategy, the rule-based one reports it needs Events Calendar Pro, and the run is not aborted by it. Run the migration for real and the date-list event ends up with one occurrence per date.
11. Activate a Pro older than `7.9.0-dev`: the feature goes fully inert (no Event Dates section, no markers, no `/all/`) and Pro behaves as released. Swap in the Pro branch from events-pro#3078: Pro owns the rendering with nothing doubled. Deactivate Pro while a rule-based event exists: its occurrences stay in place, nothing is deleted.
12. The event feed (`/event/{slug}/feed/`) renders the event once and does not redirect.

### 🎥 Artifacts
| | | |
|---|---|---|
| screenshot-01 | screenshot-02 | screenshot-03 |
| screenshot-04 | screenshot-05 | screenshot-06 |

### ✔️ Checklist


src/Tribe/Editor.php
```diff
@@ -179,9 +180,9 @@ public function flag_post_from_classic_editor() {
 
 		$has_flag_classic_editor = metadata_exists( 'post', $post, $this->key_flag_classic_editor );
 
-		// If we already have a flag we bail
+		// Flagged, yet the content still has no blocks: a previous conversion was interrupted, retry it.
 		if ( $has_flag_classic_editor ) {
-			return false;
+			return (bool) $this->update_post_content_to_blocks( $post );
```
coutinhocouto, line 207: `flag_post_from_classic_editor()` is hooked on `load-post.php` at priority 0, unconditionally, nowhere near the recurrence gate. So this branch is live for every site the moment the plugin updates.

It also flips a read-only guard into a write on a GET. Any flagged event that still has no blocks now runs `update_post_content_to_blocks()`, and that calls `wp_update_post()` with no attempt counter and no re-flagging on failure. If the conversion can't produce blocks (a site filtering `tribe_blocks_editor_update_classic_content` to an empty string, or `wp_update_post()` failing for any reason) then every open of that post's edit screen rewrites the content and spawns another revision. Forever.

src/Events/Custom_Tables/V1/Models/Provisional_Post_Meta.php
```diff
@@ -0,0 +1,156 @@
...
+
+		$provisional = tribe( Provisional_Post::class );
+
+		if ( ! $provisional->is_provisional_post_id( $object_id ) ) {
+			return $meta_value;
+		}
+
+		$post = get_post( $object_id );
+		// Maybe already hydrated? Use `get_object_vars` as `isset` will trigger the `WP_Post::__get` method.
+		$occurrence_id = get_object_vars( $post )['_tec_occurrence_id'] ?? null;
```
coutinhocouto, line 129: `get_post()` returns `null` here for a provisional ID that no longer maps to a live Occurrence, and `is_provisional_post_id()` only checks `is_numeric() && > base`. A deleted occurrence, a stale bookmark, or a hand-typed `?p=<base+n>` all get through. Then `get_object_vars( null )` is a `TypeError` on PHP 8 and the request dies, and this filter is on `get_post_metadata` globally whenever the engine is active.

I know this is inherited from the Pro copy rather than newly written, but the port is the one chance to close it.

```suggestion
		$post = get_post( $object_id );

		if ( ! $post instanceof \WP_Post ) {
			return $meta_value;
		}

		// Maybe already hydrated? Use `get_object_vars` as `isset` will trigger the `WP_Post::__get` method.
		$occurrence_id = get_object_vars( $post )['_tec_occurrence_id'] ?? null;
```

src/Events/Recurrence/Occurrences_List.php
```diff
@@ -0,0 +1,326 @@
...
+		$post_id = Occurrence::normalize_id( $post_id );
+
+		$builder       = Occurrence::where( 'post_id', '=', $post_id );
+		$count_builder = Occurrence::where( 'post_id', '=', $post_id );
+
+		if ( 'upcoming' === $view ) {
+			$now = current_time( 'mysql' );
+			$builder->where( 'end_date', '>=', $now );
+			$count_builder->where( 'end_date', '>=', $now );
+		}
```
coutinhocouto: `current_time( 'mysql' )` is the site's local wall clock, but `end_date` holds the event's local wall clock. The comments in this same file at lines 147, 169 and 257 say so explicitly.

So for any event in a timezone other than the site's, this compares two strings from different frames and occurrences near the boundary land in the wrong bucket. `$data['total']` and the page count go wrong with them, since `$count_builder` gets the same clause.

`get_scheduled_dates()` right below already does this properly, and the query two lines down orders on `start_date_utc`, so the UTC columns are right there.

```suggestion
		if ( 'upcoming' === $view ) {
			$now = current_time( 'mysql', true );
			$builder->where( 'end_date_utc', '>=', $now );
			$count_builder->where( 'end_date_utc', '>=', $now );
		}
```

src/Events/Custom_Tables/V1/Schema_Builder/Schema_Builder.php
```diff
@@ -115,7 +117,57 @@ public function get_registered_field_schemas() {
...
+
+			if ( isset( $deduped[ $key ] ) ) {
+				continue;
+			}
+
+			$deduped[ $key ] = $schema;
+		}
+
+		return array_values( $deduped );
 	}
```
coutinhocouto, line 180: The dedupe identity is the schema version *option* and never `SCHEMA_VERSION`, and the winner is whichever registered first on a hook order this code doesn't control.

Right now free and Pro both declare `1.0.1`, so it's invisible. But the transition window the docblock describes is exactly the moment they stop matching. The first time one side bumps `SCHEMA_VERSION` or changes `fields()`, the newer definition can be dropped from `get_registered_field_schemas()` entirely, its `dbDelta` never runs, the column never gets created, and `sync_stored_version()` on the surviving handler still writes to the shared option. The dropped handler's `is_schema_current()` then reports "current" against a value
some other class wrote.

Fold the declared version into the key and keep the highest, or hard-fail when two schemas collide on a key with different versions. "First wins" is the one thing that isn't safe here.

src/Events/Recurrence/Dates_Service.php
```diff
@@ -0,0 +1,174 @@
...
+			$timezone    = new DateTimeZone( (string) $event->timezone );
+			$event_start = new DateTimeImmutable( (string) $event->start_date, $timezone );
+			$event_end   = new DateTimeImmutable( (string) $event->end_date, $timezone );
+
+			$periods = [];
+			foreach ( $dates as $date ) {
+				if ( ! is_array( $date ) || ! isset( $date['start'], $date['end'] ) ) {
+					// A malformed entry would silently author an Occurrence at the current time.
+					return false;
+				}
```
coutinhocouto, line 67: `isset()` is true for `''`, and `new DateTimeImmutable( '', $tz )` doesn't throw on an empty string, it quietly returns now. So `[ 'start' => '', 'end' => '2026-01-01 10:00:00' ]` walks straight past this guard and authors an occurrence at the current time, which is the exact thing the comment on the next line says it prevents.

Both callers in this repo pre-filter, so it only bites someone using this as the documented authoring API. Worth closing anyway since that's what it's advertised as.

```suggestion
			foreach ( $dates as $date ) {
				if ( ! is_array( $date ) || ! isset( $date['start'], $date['end'] ) ) {
					// A malformed entry would silently author an Occurrence at the current time.
					return false;
				}

				foreach ( [ 'start', 'end' ] as $bound ) {
					if ( ! $date[ $bound ] instanceof DateTimeImmutable && '' === trim( (string) $date[ $bound ] ) ) {
						// An empty string passes `isset()` and resolves to the current time.
						return false;
					}
				}
```

src/Events/Custom_Tables/V1/Models/Provisional_Post.php
```diff
@@ -0,0 +1,379 @@
...
+		$fetched    = $occurrence;
+
+		if ( $fetched instanceof Occurrence ) {
+			$cache[ $cache_key ] = $occurrence;
+
+			return $fetched;
+		}
+
+		return $meta_value;
+	}
```
coutinhocouto: I think this is dead code, so I recommend removing it.

tests/_support/Traits/With_Recurrence_Engine.php
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * @after
+	 */
+	public function deactivate_recurrence_engine(): void {
+		remove_all_filters( 'tec_events_recurrence_enabled' );
+		// Symmetric cleanup: no engine hook or extended model state leaks into other tests.
+		tribe( Controller::class )->unregister();
+		tribe()->setVar( Controller::class . '_registered', false );
+		Model::reset_extensions();
+	}
```
coutinhocouto, line 75: `activate_recurrence_engine()` sets `ct1_fully_activated` to `true` and the paired `@after` never puts it back. It resets the controller and `Model::extensions`, which is what makes the omission easy to miss.

I recommend two edits:

```php
	public function activate_recurrence_engine(): void {
		add_filter( 'tec_events_recurrence_enabled', '__return_true' );
		$this->previous_ct1_activation = tribe()->getVar( 'ct1_fully_activated', false );
		tribe()->setVar( 'ct1_fully_activated', true );
```

and restore it in the cleanup, alongside a `private $previous_ct1_activation = false;` on the trait:

```php
		tribe( Controller::class )->unregister();
		tribe()->setVar( Controller::class . '_registered', false );
		tribe()->setVar( 'ct1_fully_activated', $this->previous_ct1_activation );
		Model::reset_extensions();
```

tests/_support/Traits/With_Recurrence_Engine.php
```diff
@@ -0,0 +1,89 @@
...
+					'start' => '2026-11-12 09:00:00',
+					'end'   => '2026-11-12 10:00:00',
+				],
+			];
+		}
+
+		$this->assertTrue( tribe( Dates_Service::class )->set_dates( $post->ID, $dates ) );
+
+		return $post;
+	}
```
coutinhocouto, line 118: The shared fixture hard-codes `2026-11-05` and `2026-11-12`, which is about nine weeks out.

Once those pass, `Occurrences_List's` `end_date >= now` filter drops the rows and a few things go red on a branch nobody touched:

`All_Occurrences_ProviderTest.php:222` asserts `eventDate=2026-11-05` is "the next   upcoming occurrence", with a comment saying "The event date is in the future" - the `should_render_the_scheduled_dates_metabox` snapshot bakes in "November 5, 2026"

`Admin_Views_SnapshotTest.php:134` already uses 2050 and the comment there names this exact hazard, so the pattern is right in the same PR. The shared fixture should follow it.

tests/ct1_integration/TEC/Events/Recurrence/Admin_Views_SnapshotTest.php
```diff
@@ -0,0 +1,178 @@
...
+
+		foreach ( \TEC\Events\Custom_Tables\V1\Models\Occurrence::where( 'post_id', '=', $event_id )->all() as $occurrence ) {
+			$ids[] = tribe( \TEC\Events\Custom_Tables\V1\Events\Provisional\ID_Generator::class )->current() + (int) $occurrence->occurrence_id;
+		}
+
+		// IDs appear as `post=<id>`, `value="<id>"` or path fragments: match whole numbers only.
+		foreach ( $ids as $i => $id ) {
+			$placeholder = 0 === $i ? '{{EVENT_ID}}' : '{{OCCURRENCE_ID_' . $i . '}}';
+			// The boundary classes exclude date and time adjacency: `2050-01-10`, `10:00am`.
+			$html = preg_replace( '/(?<![\w.\-:])' . preg_quote( (string) $id, '/' ) . '(?![\w.\-:])/', $placeholder, $html );
```
coutinhocouto, line 142: When `event_id` is `0` this regex matches every standalone `0` in the markup, not just the ID-shaped ones. The committed snapshot has

    viewBox="{{EVENT_ID}} {{EVENT_ID}} 20 20"
    padding: {{EVENT_ID}};
    margin: 12px {{EVENT_ID}} 16px;
    var nextIndex = {{EVENT_ID}};

Those regions can't fail any more. Change the icon viewBox or that padding and the snapshot still matches.

Skip normalization entirely when the id is `0`, or exclude a bare zero from the match.

src/Events/Custom_Tables/V1/WP_Query/Custom_Query_Filters.php
```diff
@@ -0,0 +1,352 @@
+<?php
```
coutinhocouto, line 1: `Custom_Query_Filters` (352 lines) and `Replace_Results` (192) don't have a test naming either class, and between them they rewrite `WP_Query` results for every event listing on the site. `Provisional_Post`, `Provisional_Post_Cache`, `Provisional_Post_Meta`, `ID_Generator`, `Event_Sequence` and `Rset_Formatter` are in the same shape, exercised incidentally through other tests but never directly.

158 new tests is genuinely a lot, and the `src/Events/Recurrence/` half is well covered. It's the ported CT1 half that's bare, and that's where the null-pointer I flagged in `Provisional_Post_Meta` lives. A direct test on that class would have caught it.

src/modules/blocks/event-datetime/dashboard/event-dates/template.js
```diff
@@ -0,0 +1,394 @@
...
+			stash.current = rows;
+			updateRows( [] );
+		}
+
+		setIsOpen( checked );
+	};
+
+	const removeRow = ( index ) => {
+		const nextRows = rows.filter( ( unused, i ) => i !== index );
+		updateRows( nextRows );
```
coutinhocouto, line 414: `removeRow()` shrinks `rows` but leaves `editing` alone, and `editing` is keyed by array index (`${ index }:${ field }`) while `rows.map()` keys by index too.

Delete a row above one with uncommitted keystrokes in its time picker, and the stale `editing[key]` now resolves against a different row. `renderTimePicker` shows the leftover text on the wrong date's field.

```suggestion
	const removeRow = ( index ) => {
		const nextRows = rows.filter( ( unused, i ) => i !== index );
		updateRows( nextRows );
		setEditing( {} );
```

changelog/fix-dateless-recurring-single-url
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Resolved an issue where the dateless URL of a multi-date event rendered the single event template once per occurrence; it now redirects to the next upcoming occurrence, or to the occurrences archive when all dates are past.
```
coutinhocouto: Opens correctly in past tense and then switches to "it now redirects". Same "now" in `changelog/tweak-recurrence-dates-chips:4`, `changelog/feat-all-occurrences-link:4`, `changelog/feat-recurrence-field-schemas:4` and `changelog/feat-date-rules-migration-strategy:4`.

changelog/feat-all-occurrences-link
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: feat
+
+Added the tribe_all_occurrences_link() template tag; the recurring event markers in List, Month and Day views now link to the event's occurrences archive.
```
coutinhocouto: All changelogs are wrong, they must be in .yaml format `(stellarwp/chageloger)`, please re-create all of them

src/Events/Recurrence/Controller.php
```diff
@@ -0,0 +1,236 @@
...
+		 * use this filter to force-enable the feature at `PHP_INT_MAX` priority: with
+		 * such a Pro active this filter cannot disable the feature, and the supported
+		 * kill-switch is the `TEC_EVENTS_RECURRENCE_DISABLED` constant.
+		 *
+		 * @since TBD
+		 *
+		 * @param bool $active Whether the feature is active or not; defaults to the
+		 *                     value of the `tec_events_recurrence_active` option.
+		 */
+		return (bool) apply_filters( 'tec_events_recurrence_enabled', $active );
```
coutinhocouto: Raw cast on a filter return: `tribe_is_truthy()` rather than `(bool)`.

Same pattern at `ID_Generator.php:136` and `:154`, `Occurrences_List.php:98` and `:251`, and `All_Occurrences_Provider.php:267`. Worth noting the PR already gets this right at `Date_Rules.php:167` and `Admin_Provider.php:269`, so it's just drift.

src/Events/Recurrence/Controller.php
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	/**
+	 * Removes the filters and actions hooks added by the controller.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function unregister(): void {
```
coutinhocouto, line 184: `unregister()` clears `tec_events_recurrence_fully_activated` and every sub-provider, but not `Controller::class . '_registered'`. So a later `register()` returns at the base class's `is_registered()` guard and `do_register()` never runs again.

`do_register()`'s own comment promises the opposite ("a register/unregister/register cycle re-attaches their hooks"), and `With_Recurrence_Engine::activate_recurrence_engine()` setting that var by hand is the giveaway that it doesn't hold.

Nothing in production unregisters this controller today, so either reset the var here or reword the comment to say the guarantee is per sub-provider, not per controller.

src/Events/Custom_Tables/V1/Links/Links.php
```diff
@@ -0,0 +1,277 @@
...
+										->where( 'end_date_utc', '>', $current_date )
+										->order_by( 'start_date' )
+										->first();
+
+			if ( $upcoming instanceof Occurrence ) {
+				return $upcoming;
+			}
+
+			// If the most recent event has passed, open the one closest to the end date of the event, meaning
+			// the closest to the current date.
```
coutinhocouto:
```suggestion
            /**
			* If the most recent event has passed, open the one closest to the end date of the event, meaning
			* the closest to the current date.
			* 
			*/
```

src/Events/Recurrence/Controller.php
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	/**
+	 * Registers the feature sub-controllers and the capability handshake state.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	protected function do_register(): void {
```
coutinhocouto, line 226: `protected function do_register()` sits between two public methods. Order is public → protected → private.

src/Events/Recurrence/All_Occurrences_Provider.php
```diff
@@ -0,0 +1,488 @@
...
+		if ( '' !== (string) $query->get( 'eventDate' ) ) {
+			return false;
+		}
+
+		// The Occurrences archive URLs match the dateless conditions too: never touch them.
+		if ( 'all' === $query->get( 'eventDisplay' ) || ! empty( $query->query_vars['tribe_recurrence_list'] ) ) {
+			return false;
+		}
+
+		if ( $check_surfaces && ( $query->is_feed() || $query->is_embed() || $query->is_404() ) ) {
```
coutinhocouto: Feed, embed and 404 are excluded but preview isn't.

`get_preview_post_link()` on an already-published post reuses the pretty permalink and just appends `?preview=true`, so `name` is set exactly as on a normal front-end hit. Click Preview Changes on a multi-date event and this matches, `redirect_dateless_request()` fires on `template_redirect`, and the author gets sent to the next occurrence's live URL with `preview=true` dropped. They see the published page and never their draft.

```suggestion
		if ( $check_surfaces && ( $query->is_feed() || $query->is_embed() || $query->is_404() || $query->is_preview() ) ) {
```

coutinhocouto (changes requested):

PR #5747, opened 2026-08-31
Prevent fatal when a snippet returns a non-array from tec_views_v2_subscribe_links

sdokus:
### 🎫 Ticket

[SMTNC-1042](https://linear.app/nexcess/issue/SMTNC-1042/snippet-for-hiding-all-subscribe-links-causes-a-fatal-error)

### 🗒️ Description

The snippet in [our KB article](https://theeventscalendar.com/knowledgebase/remove-ical-and-google-calendar-links-from-single-event-views/) hides the subscribe links with `add_filter( 'tec_views_v2_subscribe_links', '__return_false', 100 )`, which replaces the list of links with `false`. Calendar views only ever check that value with `empty()`, so it appears to work. Single events hand it to `count()`, and PHP 8 fatals:

```
Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, bool given
in .../src/Tribe/Views/V2/iCalendar/Single_Events.php:85
```

`Single_Events::single_event_links()` already got a guard in 6.10.1, but that only covered the one caller the report named. This moves the check to where all of them route through:

- `iCalendar_Handler::get_subscribe_links()` now returns the array its docblock has always promised, so no consumer — templates, the single event links, the block — can receive a scalar.
- `Link_Abstract::filter_tec_views_v2_subscribe_links()` bails on a non-array instead of writing an offset into it. A snippet hooked below priority 10 reaches the link classes before they append, which is the same bug one step earlier in the chain.

The snippet in the KB article now does what it says on all views, so no workaround (`__return_empty_array`) is needed.

### Artifacts
https://www.loom.com/share/7320407c316d4e32b22fa3a403687690

### ✔️ Checklist
- [x] Ran `npm run changelog` to add changelog file(s).

### DIFF
--- a/src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
+++ b/src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
@@ -119,6 +119,11 @@ abstract class Link_Abstract implements Link_Interface, JsonSerializable {
 			return $subscribe_links;
 		}
 
+		// An earlier filter may have replaced the list with a non-array; don't append to it.
+		if ( ! is_array( $subscribe_links ) ) {
+			return $subscribe_links;
+		}
+
 		$subscribe_links[ self::get_slug() ] = $this;
 
 		return $subscribe_links;
--- a/src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
+++ b/src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
@@ -112,6 +112,8 @@ abstract class Link_Abstract implements Link_Interface, JsonSerializable {
 
 	/**
 	 * {@inheritDoc}
+	 *
+	 * @since TBD Return the value untouched when an earlier callback replaced it with a non-array.
 	 */
 	public function filter_tec_views_v2_subscribe_links( $subscribe_links ) {
 		// Bail early if we're not supposed to show this link.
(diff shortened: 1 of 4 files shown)


larodiel (approved):

tests/views_integration/Tribe/Events/Views/V2/iCalendar/Subscribe_LinksTest.php
```diff
@@ -0,0 +1,47 @@
...
+ * from `tec_views_v2_subscribe_links`.
+ *
+ * @see https://theeventscalendar.com/knowledgebase/remove-ical-and-google-calendar-links-from-single-event-views/
+ */
+class Subscribe_LinksTest extends WPTestCase {
+
+	/**
+	 * @return iCalendar_Handler
+	 */
+	protected function handler() {
```
coutinhocouto: `protected function handler()` is declared above all three `public function test_*` methods. 

The correct order is public → protected → private

tests/views_integration/Tribe/Events/Views/V2/iCalendar/Subscribe_LinksTest.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Regression coverage for snippets that hide the subscribe links by returning a non-array
+ * from `tec_views_v2_subscribe_links`.
+ *
+ * @see https://theeventscalendar.com/knowledgebase/remove-ical-and-google-calendar-links-from-single-event-views/
+ */
+class Subscribe_LinksTest extends WPTestCase {
+
+	/**
+	 * @return iCalendar_Handler
```
coutinhocouto:
```suggestion
	 * @since TBD
	 *
	 * @return iCalendar_Handler The subscribe links handler.
```

src/Tribe/Views/V2/iCalendar/Links/Link_Abstract.php
coutinhocouto, line 117:
```suggestion
	/**
	* {@inheritDoc}
	*
	* @since TBD Return the value untouched when an earlier callback replaced it with a non-array.
	*/
```

coutinhocouto (changes requested):

coutinhocouto (approved):
@sdokus thanks for the changes, now it's okay for me, great work 😄

PR #5752, opened 2026-09-02
[SOFT-3323] fix: RSVP V2 Import issue

larodiel:
### 🎫 Ticket

[SOFT-3323](https://linear.app/nexcess/issue/SOFT-3323/rsvp-import)

### 🗒️ Description

**Fix** (bug fix - CSV import data loss)
Fixed a bug in `Tribe__Events__Aggregator__Record__CSV::prep_import_data()` where a rejected column mapping (missing a required field) was permanently saved as the site-wide default. This caused every subsequent import preview, for any content type (events, venues, organizers, RSVP tickets), to pre-select "Do Not Import" for the required column and fail again until manually corrected. The previous valid mapping is now restored whenever validation fails.

Complement to: https://github.com/the-events-calendar/event-tickets/pull/4440

### 🎥 Artifacts 
https://www.loom.com/share/1eb893f46e2e44ddb4753c04ce8e51bd

### ✔️ Checklist

### DIFF
--- a/src/Tribe/Aggregator/Record/CSV.php
+++ b/src/Tribe/Aggregator/Record/CSV.php
@@ -201,11 +201,8 @@ class Tribe__Events__Aggregator__Record__CSV extends Tribe__Events__Aggregator__
 
 		try {
 			$importer = $this->get_importer();
-		} catch ( RuntimeException $e ) {
-			// Only roll back if nothing else has saved a newer mapping since our write above.
-			if ( get_option( $option_key ) === $data['column_map'] ) {
-				update_option( $option_key, $previous_map );
-			}
+		} catch ( InvalidArgumentException $e ) {
+			update_option( $option_key, $previous_map );
 			return tribe_error( 'core:aggregator:missing-csv-file' );
 		}
 
@@ -221,10 +218,7 @@ class Tribe__Events__Aggregator__Record__CSV extends Tribe__Events__Aggregator__
 		$missing         = array_diff( $required_fields, $data['column_map'] );
 
 		if ( ! empty( $missing ) ) {
-			// Only roll back if nothing else has saved a newer mapping since our write above.
-			if ( get_option( $option_key ) === $data['column_map'] ) {
-				update_option( $option_key, $previous_map );
-			}
+			update_option( $option_key, $previous_map );
 
 			$mapper = new Tribe__Events__Importer__Column_Mapper( $content_type );
 
--- a/src/Tribe/Aggregator/Record/CSV.php
+++ b/src/Tribe/Aggregator/Record/CSV.php
@@ -196,7 +196,7 @@ class Tribe__Events__Aggregator__Record__CSV extends Tribe__Events__Aggregator__
 		// permanently overwrite it - get_importer() needs the option set to build the
 		// importer with *this* submission's map, but we must not let a rejected map stick
 		// around as the default for the next preview if we end up returning an error.
-		$previous_map = get_option( $option_key, array() );
+		$previous_map = get_option( $option_key, [] );
 		update_option( $option_key, $data['column_map'] );
 
 		try {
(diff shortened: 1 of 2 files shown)


larodiel:
@coderabbitai review

common
sdokus, line 1: Are these common changes intentional? 🤔

larodiel: no, it was a leak 🫠

larodiel: @sdokus removed! :)

src/Tribe/Aggregator/Record/CSV.php
```diff
@@ -191,11 +191,21 @@ public function prep_import_data( $data = array() ) {
...
+		// Remember the last-known-good mapping so an invalid submission below doesn't
+		// permanently overwrite it - get_importer() needs the option set to build the
+		// importer with *this* submission's map, but we must not let a rejected map stick
+		// around as the default for the next preview if we end up returning an error.
+		$previous_map = get_option( $option_key, [] );
+		update_option( $option_key, $data['column_map'] );
 
 		try {
 			$importer = $this->get_importer();
 		} catch ( RuntimeException $e ) {
```
sdokus: The rollback misses the exception type this PR is actually about.

 `File_Importer::get_importer()` throws `InvalidArgumentException` — not `RuntimeException` — when no importer is registered for a content type (`src/Tribe/Importer/File_Importer.php:109`).

 For `rsvp_tickets`, the importer comes from the `tribe_events_import_rsvp_tickets_importer` filter, which lives in Event Tickets. On a site where ET is inactive, older, or hasn't hooked it, line 200 has already written the map, the throw sails past this catch, and the rejected mapping stays saved — the exact data loss being fixed here, plus an uncaught fatal.

 Same hole for anything thrown by `maybe_set_default_category()`, `maybe_set_default_post_status()`, or `get_required_fields()` further down.

 A `finally` (or writing the option only after validation passes) covers all of them and lets you delete the duplicated rollback block at 224–226.

src/Tribe/Aggregator/Record/CSV.php
```diff
@@ -191,11 +191,21 @@ public function prep_import_data( $data = array() ) {
...
+			// Only roll back if nothing else has saved a newer mapping since our write above.
+			if ( get_option( $option_key ) === $data['column_map'] ) {
+				update_option( $option_key, $previous_map );
```
sdokus: This guard can't do what the comment says, and it can skip the rollback.

`update_option()` at line 200 populates the in-request options cache, so this `get_option()` always reads back our own write. A concurrent request's DB write is invisible to it — the "nothing else has saved a newer mapping" check has no way to observe that case.

Its only live effect is in the failure direction: a third-party `option_tribe_events_import_column_mapping_*` filter that normalizes or reorders the map breaks the strict `===`, the rollback silently doesn't happen, and we're back to the bug this PR fixes.

Since the guard can't detect the race it's written for, dropping it (and rolling back unconditionally) is both simpler and strictly safer. Same comment applies to 224–226.

src/Tribe/Aggregator/Record/CSV.php
```diff
@@ -211,6 +221,11 @@ public function prep_import_data( $data = array() ) {
 		$missing         = array_diff( $required_fields, $data['column_map'] );
 
 		if ( ! empty( $missing ) ) {
+			// Only roll back if nothing else has saved a newer mapping since our write above.
+			if ( get_option( $option_key ) === $data['column_map'] ) {
+				update_option( $option_key, $previous_map );
+			}
```
sdokus: See comment above

src/Tribe/Aggregator/Record/CSV.php
```diff
@@ -228,8 +243,6 @@ public function prep_import_data( $data = array() ) {
 			);
 		}
 
-		update_option( 'tribe_events_import_column_mapping_' . $content_type, $data['column_map'] );
-
```
dpanta94, line 232: Why don't you only save here instead of saving before and then having to potentially roll back the value ?

dpanta94 (changes requested):

dpanta94 (approved):

PR #3016, opened 2020-01-22
VR > Tax breadcrumbs

juanfra:
🎫 https://moderntribe.atlassian.net/browse/TEC-3127
🎥 https://www.loom.com/share/353e44a77e474a9e96f76bcd573e459d

This one came back from smoketest, it's not related to the original fix.

PR #5770, opened 2026-09-16
[SMTNC-2752] Prevent event rendering failures from TEC blocks

dpanta94:
### 🎫 Ticket

[SMTNC-2752](https://linear.app/nexcess/issue/SMTNC-2752)

### 📋 Plan

OpenSpec waived by requester; its CI check remains failing.

### 🗒️ Description

Keep event pages renderable when descriptions contain Single Event or Events Archive blocks. Register `events.editor.template` independently of the editor setting and prevent recursive full-view rendering inside event descriptions, while preserving site-template and ordinary-page rendering. Include common’s separate export filtering and description-preservation fix, pinned to `ff15ea882`; merge [common #3016](https://github.com/the-events-calendar/tribe-common/pull/3016) first.

Reproduce:
1. With Twenty Twenty and the event block editor disabled, publish an event with `<!-- wp:tec/single-event /-->` between two paragraphs.
2. Open its page and `?ical=1` export: before, missing-template fatal; after, both work and retain the paragraphs without a nested calendar.

Expanded QA: [ticket comment](https://linear.app/nexcess/issue/SMTNC-2752#comment-043dc914).

### 🎥 Artifacts

Slic results:
- Rendering regression before: `14 tests, 11 errors, 1 failure`.
- After: views `940 tests, 1575 assertions, 1 skipped`; integration `232 tests, 628 assertions, 6 skipped`; no errors/failures.
- Existing iCal: `49 tests, 169 assertions`; editor/template checks pass.

Tests cover both content renderers, nested blocks, template/page controls, and exports without the rendering service. Stabilizes an existing list snapshot’s cached organizer classification. Targeted PHPStan passes; PHPCS findings are pre-existing. Browser QA remains unperformed.

### ✔️ Checklist

- [ ] OpenSpec plan (waived).
- [x] Changelog added.
- [x] Regression coverage added.
- [ ] All required checks pass (new CI pending; OpenSpec fails).
- [ ] Automated review comments addressed.
- [x] Test evidence added.
- [x] Base: `release/M26.octillery`.
- [x] Ticket includes expanded QA instructions.
- [ ] Ready for human review.

### 🤖 AI Disclosure

- [ ] Human ownership/review of the latest changes confirmed.

**AI usage:** Drafted with GPT-6 (Codex); agent-reviewed and tested, with human review of the latest changes pending.


src/Events/Blocks/Single_Event/Block.php
```diff
@@ -43,12 +43,18 @@ public function default_attributes(): array {
...
 	 * @since 6.3.3
+	 * @since TBD Prevents full event views from rendering inside event descriptions.
 	 *
 	 * @param array $attributes The block attributes.
 	 *
 	 * @return string The block HTML.
 	 */
 	public function render( $attributes = [] ): string {
+		// Rendering the event view inside its own description would recurse.
+		if ( doing_filter( 'the_content' ) && get_post_type() === \Tribe__Events__Main::POSTTYPE ) {
```
coutinhocouto: The recursion this guards against is specific to `tec/single-event` re-rendering itself. Views V2 list and grid templates never call `the_content()` or `get_the_excerpt()` per event card, so `tec/archive-events` inside an event description doesn't recurse.

Copying the guard here means someone who embeds "other upcoming events" in an event's own description now gets nothing, with no error anywhere to explain it. `Rendering_Test::content_cases()` asserts that blanking as intended behaviour, so if it's a deliberate call it's worth a sentence saying why. If it's symmetry with the sibling class, the archive block can probably keep rendering.

coutinhocouto: `get_post_type()` with no argument reads the ambient global `$post`, which isn't necessarily the post whose content is mid-filter. Anything that calls `apply_filters( 'the_content', $other_post->post_content )` without `setup_postdata()` while the global post happens to be an event gets the wrong answer, and it's wrong in both directions. A legitimate block on that other post gets blanked, or the guard misses and the single-event block renders a full event view inside its own description.

The second case matters more now than it would have before this PR. With `events.editor.template` always bound, a miss is unbounded recursion rather than a clean fatal.

WordPress already passes `$attributes, $content, $block` to a render callback, and `render( $attributes = [] )` is discarding the last two. Widening the signature and reading the block's own context would tie this to the post actually being rendered instead of to whatever is sitting in the global.

tests/views_integration/Tribe/Events/Views/V2/Views/List_ViewTest.php
```diff
@@ -81,6 +81,10 @@ public function test_render_with_upcoming_events() {
 			'events/single/2.json'
 		] );
 
+		// Remapped IDs may have belonged to organizers in earlier tests; these fixtures have none.
+		tribe_unset_var( 'Tribe__Events__Main::isOrganizer' );
+		$this->assertFalse( tribe_has_organizer( 7 ) );
```
coutinhocouto: Worth saying first: the snapshot edit is correct. The `"organizer"` object you removed had the event's own title as its `name`, which is the signature of `Tribe__Events__Main::isOrganizer()`'s per-ID static cache being stale-true for post ID 7. The old snapshot was recording polluted output.

My concern is where the fix landed. `Rendering_Test` documents this exact leak model in its own `setUp()` ("Rendering caches post types by ID, and later fixtures can reuse those IDs") but only resets the cache for its own run, and `List_ViewTest` gets a `tribe_unset_var()` plus a rewritten snapshot. That fixes the one test that happened to run next.

Resetting `isOrganizer`/`isVenue` in the shared `TestCase` that every view snapshot test extends would stop the next ordering change from doing this again to a different victim.

coutinhocouto (changes requested):

src/Tribe/Editor/Provider.php
```diff
@@ -19,6 +20,9 @@ public function register() {
 
 		tribe( 'events.editor' )->hook();
 
+		// Registered blocks still need to render when the event editor is disabled.
+		$this->container->singleton( 'events.editor.template', 'Tribe__Events__Editor__Template' );
```
larodiel, line 24: Shouldn't it go after the check on the line below?

dpanta94: No that's the point. To register the implementation regardless of whether the setting is turned on or off.

src/Events/Blocks/Single_Event/Block.php
dpanta94: I checked this by removing just the Archive block’s guard. Both blocks call `Template_Bootstrap::get_view_html()`, and on an event request that selects the single-event view, even for tec/archive-events. So the Archive block can end up rendering the event’s description again.

Removing the guard triggered the recursion checks under both classic and block themes. I think we should keep protection for both blocks, and make that reason clearer in the code.

coutinhocouto (approved):
Seems okay for me now, thanks @dpanta94

src/Events/Blocks/Single_Event/Block.php
```diff
@@ -43,12 +45,18 @@ public function default_attributes(): array {
...
 	 *
 	 * @since 6.3.3
+	 * @since TBD Prevents full event views from rendering inside event descriptions.
 	 *
 	 * @param array $attributes The block attributes.
 	 *
 	 * @return string The block HTML.
 	 */
 	public function render( $attributes = [] ): string {
+		// Both blocks render the single-event view on event requests, regardless of the global post.
```
sdokus:
```suggestion
		// On a single event request the view already contains this event's content, so rendering the block again would recurse.
```
Without the context of this PR, "both blocks" loses meaning. A similar change would be nice on `Archive_Events/Block.php:55`.

PR #5772, opened 2026-09-16
[SMTNC-2761] Accept whatever a foreign filter delivers

dpanta94:
### 🎫 Ticket

[SMTNC-2761](https://linear.app/nexcess/issue/SMTNC-2761/loosen-core-filter-callback-type-requirements)

### 📋 Plan

None. OpenSpec was explicitly waived for this change by the author.

### 🗒️ Description

Bug fix. Seventeen callbacks attached to filters this plugin does not own — WordPress core, plus Avada, WPML and WP Rocket — declared a scalar or array parameter type, or declared a return type while handing the incoming value straight back. WordPress passes a filter's value through every callback in the chain, so any of them could be handed a type it rejected and fatal the request before our code ran. Each now takes the value untyped and sorts out the type inside.

The reported symptom was `Harbor\PUE::filter_pre_http_request()` dying on the null URL `wp_remote_get( null, ... )` produces; that one is fixed in tribe-common#3018, and this is the same bug class everywhere else it exists. The worst of these is `the_content` with a declared `string $content` — a null from any plugin ahead of us takes down the page.

Left alone deliberately: declarations for arguments core constructs itself (`WP_Query`, `WP_Post`, `WP_Rewrite`, `WP_REST_Request`), which no filter can substitute, and every action callback, whose arguments are not filterable.

Includes the common bump for tribe-common#3018.

### 🎥 Artifacts 

New `Core_Filter_Callback_Types_Test` fires each patched callback with a wrong-typed value that filter can really deliver. Red before the fix, green after:

```
Tests: 17, Assertions: 16, Errors: 1, Failures: 16     <- before
OK (17 tests, 17 assertions)                            <- after
```

Also run: `Tribe__Events__Admin_ListTest` (OK 2/2), `integration_category_colors` (OK 267/867), `views_wpunit HooksTest` (OK 16/20). `embed_calendar_integration` reports 4 failures, all pre-existing path assertions that hardcode `plugins/the-events-calendar/` and break in a differently-named worktree — unrelated to this change.

No browser QA: the ticket states QA is not needed.

### ✔️ Checklist
- [ ] There is an OpenSpec plan for this work in `tec-plans`, and it still describes what this PR actually does. — **waived by the author for this change**
- [x] Changelog file(s) added.
- [x] Code is covered by tests.
- [x] All the **required** tests are passing. — the suites listed under Artifacts were run, not the full matrix
- [x] Artifacts added.
- [x] Check the base branch for your PR. — `release/M26.octillery`
- [x] The linked ticket includes all the information for a QA engineer to verify the work. — ticket states QA is not needed
- [x] Pull Request is ready to be reviewed by a Human.

### 🤖 AI Disclosure

Was AI used for this PR [**Y**/n]: **Y**

Which model(s): Claude Opus 5

In what way was AI used: To sweep the codebase for the bug class, write the regression tests, apply the fix and draft this PR.

- [x] I own the work being done in this PR and I have personally reviewed it before asking another human to.


src/Tribe/Integrations/WPML/Language_Switcher.php
```diff
@@ -31,9 +31,11 @@ public static function instance() {
 	 *
 	 * @return array The languages with maybe updated URLs
 	 */
-	public function filter_icl_ls_languages( array $languages = [] ) {
+	public function filter_icl_ls_languages( $languages = [] ) {
 		global $wp_query;
 
+		$languages = (array) $languages;
```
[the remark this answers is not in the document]
dpanta94: Is this ok now ?

src/Tribe/Integrations/WPML/Language_Switcher.php
```diff
@@ -31,9 +31,13 @@ public static function instance() {
 	 *
 	 * @return array The languages with maybe updated URLs
 	 */
-	public function filter_icl_ls_languages( array $languages = [] ) {
+	public function filter_icl_ls_languages( $languages = [] ) {
 		global $wp_query;
 
+		if ( ! is_array( $languages ) ) {
+			return $languages;
+		}
```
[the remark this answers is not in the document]
dpanta94: I think i prefer to simply return what bad data i was given to not affect 3rd party potentially.

larodiel (approved):

src/Events/Calendar_Embeds/Calendar_Embeds.php
```diff
@@ -97,13 +97,14 @@ public function redirect_to_embed(): void {
...
 	 * @since 6.11.2.1 Made the parameters non-strict.
+	 * @since TBD Dropped the return type; a count or WP_Error leaves untouched.
 	 *
 	 * @param array  $terms      The terms.
 	 * @param ?array $taxonomies The taxonomies.
 	 *
 	 * @return array
 	 */
-	public function modify_term_count_on_term_list_table( $terms, $taxonomies = null ): array {
+	public function modify_term_count_on_term_list_table( $terms, $taxonomies = null ) {
```
sdokus, line 107: The new `@since` says "a count or WP_Error leaves untouched", but `$terms = (array) $terms;` still runs once taxonomies are set. A `WP_Error` gets cast to an array of its own properties and a scalar becomes `[ 5 ]`, and that's what we hand back. 

Same call you made on the WPML one, return the bad value as-is:
```
if ( ! is_array( $terms ) ) { 
	return $terms;
}
```
Then the docblock is actually true.

src/Tribe/Views/V2/Template/Title.php
```diff
@@ -305,16 +305,19 @@ public static function build_post_range_title( Context $context, $event_date, ar
...
 	 * @since 4.9.10
+	 * @since TBD Made the parameter non-strict.
 	 *
 	 * @param array $title The document title parts.
 	 *
 	 * @return array The filtered document title parts.
 	 */
-	public function filter_document_title_parts( array $title = [] ) {
+	public function filter_document_title_parts( $title = [] ) {
+		$title = (array) $title;
```
coutinhocouto, line 315:
```suggestion
		$title = is_array( $title ) ? $title : [];
```

src/Events/Category_Colors/Admin/Controller.php
coutinhocouto, line 155:
```suggestion
	 * @since 6.14.0
	 * @since TBD Wrapped the forwarded value; `manage_edit-tax_columns` can deliver a non-array.
	 *
```

tests/wpunit/TEC/Events/Hooks/Core_Filter_Callback_Types_Test.php
```diff
@@ -0,0 +1,208 @@
...
+ * returned. A scalar type declaration on such a callback turns another plugin's sloppy
+ * value into an uncaught TypeError that takes down the whole request, before our code
+ * gets a chance to ignore it.
+ *
+ * Reported as SMTNC-2761: SG Optimizer's purge queue calls wp_remote_get( null, ... ),
+ * WP_Http::request() forwards that null to `pre_http_request`, and
+ * Harbor\PUE::filter_pre_http_request()'s `string $url` fatals on it.
+ *
+ * @see https://linear.app/nexcess/issue/SMTNC-2761
+ */
```
coutinhocouto:
```suggestion
 * Reported from the field: SG Optimizer's purge queue calls wp_remote_get( null, ... ),
 * WP_Http::request() forwards that null to `pre_http_request`, and
 * Harbor\PUE::filter_pre_http_request()'s `string $url` fatals on it.
 */
```

coutinhocouto (changes requested):

src/Tribe/Views/V2/Template/Title.php
dpanta94: I think this is fine as is.

src/Events/Category_Colors/Admin/Controller.php
dpanta94: Nah the functionality is not changing in any way to require a new since tag

coutinhocouto (approved):

